PR #415, opened 2015-10-21
Terraform Prototype

ajordens:
- clouddriver only


tomaslin:
killing this with prejudice

PR #444, opened 2015-10-30
[almost] build an aws ami

ewiseblatt:
This doesnt quite work because I dont know how to properly set up S3
permissions so the packer instance can get at it. I assume that's easy
enough for someone else to fix.

To use this:
   (assuming ~/.aws/credentials exists)
   S3_RELEASE_PATH=s3://my-release-bucket
   mkdir build
   cd build

   ../spinnaker/dev/refresh_source.sh --github_user="upstream" --pull_origin
   ../spinnaker/dev/build_release.sh --release_path=$S3_RELEASE_PATH
   ../spinnaker/dev/build_aws_ami.sh --release_path=$S3_RELEASE_PATH

This will result in an error, something like:
    amazon-ebs: RuntimeError: The path "s3://ewiseblatt-20151029" does not seem to exist within S3. aws s3 ls returned "Unable to locate credentials. You can configure credentials by running "aws configure"."

### DIFF
--- a/dev/build_aws_ami.packer
+++ b/dev/build_aws_ami.packer
@@ -1,0 +1,37 @@
+{
+ "variables": {
+   "release_path": null,
+   "installer_path": null,
+   "aws_access_key": "",
+   "aws_secret_key": "",
+   "region": "us-east-1",
+   "source_ami": "ami-d05e75b8",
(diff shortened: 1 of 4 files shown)


ewiseblatt:
no. I'll close it.

PR #476, opened 2015-11-09
Add Contributing to Spinnaker submenu under Community top navigation …

rbuskens:
…menu,

including guidelines for contributing and first drafts of Individual and
Corporate CLAs.

@aglover: Please review. Note:
1. The CLA text itself is adapted from Apache2.
2. I have left warning signs and placeholders in the text where we should find a way to add electronic signing. (If we can't figure it out, then we'll need to setup manual signing in the meantime.) Screenshot of navigation menu with "Contributing to Spinnaker" menu item below:

image

### DIFF
--- a/_data/navigation.yml
+++ b/_data/navigation.yml
@@ -17,6 +17,7 @@
 -
   title: Community
   children:
+    - contributing_guidelines
     - collab_dev
     - code_of_conduct
 -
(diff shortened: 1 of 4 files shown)


duglin:
What was the result of this? Looking thru the web sites/doc I can't find anything that talks about DCOs or CLAs. What do people need to do when they submit a PR? Is there a URL to a doc I'm missing?

PR #557, opened 2015-11-16
Add CF support to prop files

gregturn:
This patch will add Cloud Foundry property settings to the default configuration.

It also updates echo with slack settings.

### DIFF
--- a/config/clouddriver.yml
+++ b/config/clouddriver.yml
@@ -41,6 +41,18 @@ google:
       project: ${providers.google.primaryCredentials.project}
       jsonPath: ${providers.google.primaryCredentials.jsonPath}
 
+cf:
+  enabled: ${providers.cf.enabled:false}
+
+  accounts:
+    - name: ${providers.cf.primaryCredentials.name}
+      api: ${providers.cf.primaryCredentials.api}
+      console: ${providers.cf.primaryCredentials.console}
+      org: ${providers.cf.defaultOrg}
+      space: ${providers.cf.defaultSpace}
+      username: ${cf.account.name:}
+      password: ${cf.account.password:}
+
 credentials:
-  primaryAccountTypes: ${providers.aws.primaryCredentials.name}, ${providers.google.primaryCredentials.name}
-  challengeDestructiveActionsEnvironments: ${providers.aws.primaryCredentials.name}, ${providers.google.primaryCredentials.name}
+  primaryAccountTypes: ${providers.aws.primaryCredentials.name}, ${providers.google.primaryCredentials.name}, ${providers.cf.primaryCredentials.name}
+  challengeDestructiveActionsEnvironments: ${providers.aws.primaryCredentials.name}, ${providers.google.primaryCredentials.name}, ${providers.cf.primaryCredentials.name}
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -29,6 +29,22 @@ providers:
       project: ${SPINNAKER_GOOGLE_PROJECT_ID}
       jsonPath: ${SPINNAKER_GOOGLE_PROJECT_CREDENTIALS_PATH:}
 
+  cf:
+    # If you want to deploy some services to Cloud Foundry (cf),
+    # set enabled and provide primary credentials for deploying.
+    # Enabling cf is independent of other providers.
+    enabled: false
+    defaultOrg: spinnaker-cf-org
+    defaultSpace: spinnaker-cf-space
+    primaryCredentials:
+      name: my-cf-account
+      api: my-cf-api-uri
+      console: my-cf-console-base-url
+    # Either uncomment and plugin credentials here, or supply as environment variables for maximum security
+    # account:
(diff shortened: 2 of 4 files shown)


config/clouddriver.yml
```diff
@@ -40,3 +40,20 @@ google:
     - name: ${providers.google.primaryCredentials.name}
       project: ${providers.google.primaryCredentials.project}
       jsonPath: ${providers.google.primaryCredentials.jsonPath}
+
+cf:
+  enabled: ${providers.cf.enabled:false}
+
+  # TODO: Migrate this out of clouddriver and into either orca or igor
+  jenkins:
```
tomaslin: Did you address @ewiseblatt comments on the other PR about having these settings here? Are they still needed ( jenkins )

config/clouddriver.yml
```diff
@@ -40,3 +40,20 @@ google:
...
+    username: ${services.jenkins.defaultMaster.username}
+    password: ${services.jenkins.defaultMaster.password}
+
+  accounts:
+    - name: ${providers.cf.primaryCredentials.name}
+      api: ${providers.cf.primaryCredentials.api}
+      console: ${providers.cf.primaryCredentials.console}
+      org: ${providers.cf.defaultOrg}
+      space: ${providers.cf.defaultSpace}
+      username: ${cf.account.name}
```
duftler: Where are these defined? I don't see `cf.account.name` or `cf.account.password`. Maybe I'm just looking in the wrong files.

tomaslin: environment variables.

duftler: So won't this fail if those env variables are not defined, and there are no defaults here?

config/clouddriver.yml
gregturn: I haven't fixed the core issue yet, but it's on my radar.

config/clouddriver.yml
gregturn: I should stick some NA defaults.

duftler: Can't you just add an empty default like this one? https://github.com/spinnaker/spinnaker/blob/master/config/default-spinnaker-local.yml#L30

gregturn: Ahh. Didn't know you could do that. Perfect.

ewiseblatt: is it possible to make environment variables look like environment variables for clarity (UPPER_CASE)?

tomaslin: wouldn't that actually stop you from adding them by hand in another yml file though?

config/default-spinnaker-local.yml
```diff
@@ -103,5 +116,8 @@ services:
         account: # twilio account id
         token: # twilio auth token
         from: # phone number by which sms messages are sent
+      slack:
+        enabled: false
+        token: # the part after http://hooks.slack.com/services/
```
ewiseblatt: Is this an independent change?
slack isnt tied to CF, is it?
If it is the other way around, the slack change can get in there first anyway.

config/echo.yml
```diff
@@ -14,6 +14,10 @@ orca:
 
 endpoints.health.sensitive: false
 
+slack:
+  enabled: ${services.echo.slack.enabled:false}
```
ewiseblatt: service.notifications

config/default-spinnaker-local.yml
tomaslin: I would say slack should be a different PR

config/clouddriver.yml
ewiseblatt: Is this an independent change?
slack isnt tied to CF, is it?
If it is the other way around, the slack change can get in there first anyway.

config/clouddriver.yml
ewiseblatt: @tomaslin, you are right. I stand corrected.

gregturn: FYI I'm not a fan of declaring UPPER_CASE env variables, when every single property you're seeing support overrides via the same method.

gregturn:
I pulled slack stuff and will file a separate PR.

config/clouddriver.yml
ewiseblatt: I'm not a fan of environ variables at all but recognize they are common and
accepted practice.
I'm just noting the convention that is in use to keep things more
readable/predictable in the "global" namespace.
Especially since the possible sources of these values are sharded to the
wind.
Once you drop into a provider, you can (and should be able to) follow the
providers conventions or idiosyncrasies.

gregturn:
See #610 for slack updates.

anotherchrisberry:
What's the story on this PR? Is it still valid or should we close it?

gregturn:
@anotherchrisberry I need to update and revise. The outstanding issues I mentioned earlier should now be resolved.

gregturn:
Okay, I've updated and polished this PR. Please review.

anotherchrisberry:
LGTM - @duftler @ewiseblatt @tomaslin any objections to merging?

PR #563, opened 2015-11-16
Update the docs for https git clone

mstantoncook:


cfieber:
@mstantoncook we all run with 2FA on our github accounts, which means contributing via https involves generating an API token

If you could reword this as either `git@` or `https://` that would work

PR #564, opened 2015-11-16
Update README.adoc

mstantoncook:
Need to install caskroom first.


tomaslin:
brew install brew-cask seems to do "Installing brew-cask from caskroom/homebrew-cask" for me

cfieber:
thanks @mstantoncook 
that is the recommended path from http://caskroom.io/

PR #566, opened 2015-11-17
Add install option for dependencies only -- no spinnaker services

ewiseblatt:
@dpeterka, @cfieber

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -44,7 +44,9 @@ function print_usage() {
   cat <<EOF
 usage: $0 [--cloud_provider <aws|google|none|both>]
     [--aws_region <region>] [--google_region <region>]
-    [--quiet]
+    [--quiet] [--dependencies_only]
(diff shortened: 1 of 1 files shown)


ewiseblatt:
I'll let you merge so you can have all the others ready at the same time.

dpeterka:
Quiet install usually means no output, can we change the option name to something like "dependencies-only"?

ewiseblatt:
@dpeterka PTAL

duftler:
Not as part of this PR, but it would be nice to have a flag that just installs using the default settings derived from the metadata server (i.e. provider and region/zone), with no prompting. The way it is now, the only way to avoid prompts for user input is to pre-configure zero providers (and that flag is combined with the no-output behavior).

dpeterka:
I have no iddues here

duftler:
First I thought you said you have "no ideas here", then I realized it was "no issues here"...

duftler:
LGTM.

PR #575, opened 2015-11-17
Copy all files to .spinnaker

gregturn:
Need to include steps to copy ALL config files into $HOME/.spinnaker.


ewiseblatt:
It shoudlnt. Will discuss this offline with you then update docs.

PR #579, opened 2015-11-18
Refactored install_development.sh to use production install

ewiseblatt:
@duftler Rewrote the install_development.sh and updated the dev/bootstrap_dev.sh script.


duftler:
LGTM.

PR #580, opened 2015-11-18
Updated create_google_dev_vm (and run_dev)

ewiseblatt:
@duftler (needs the previous 01_.. PR first)
Uses the refactored install_development.sh rather than the old python script.


duftler:
LGTM.

PR #583, opened 2015-11-18
Configure AWS AMI only on boot

dpeterka:


packer/init.aws.region
```diff
@@ -84,6 +96,7 @@ echo '    HostName' $public_hostname
 echo '    IdentityFile /path/to/private/key'
 echo '    LocalForward 8081 127.0.0.1:9000'
 echo '    LocalForward 8084 127.0.0.1:8084'
+echo '    LocalForward 8084 127.0.0.1:8087'
```
duftler: Looks like a typo: s/8084/8087

PR #593, opened 2015-11-19
Auto-generate spinnaker-local.yml on run_dev first's run.

ewiseblatt:
@duftler

### DIFF
--- a/dev/dev_runner.py
+++ b/dev/dev_runner.py
@@ -23,10 +23,69 @@ import subprocess
 import sys
 import time
 
-from spinnaker.fetch import fetch
 from spinnaker.configurator import InstallationParameters
+from spinnaker.fetch import AWS_METADATA_URL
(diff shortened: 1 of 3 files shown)


dev/dev_runner.py
```diff
@@ -23,11 +23,70 @@
...
+                                           'providers.aws.defaultRegion')
+
+  return content
+
+
+def populate_google_yml(content):
+  credentials = {'project': '', 'jsonPath': ''}
+  google_dict = {'enabled': False,
+                 'defaultRegion': 'us-central1',
+                 'defaultZone': 'us-central1-b',}
```
duftler: I think `us-central1-f` is a better choice for a baked-in default.

duftler:
LGTM.
I did not test this, but can do so tomorrow.

ewiseblatt:
ok. I updated the README as well

PR #606, opened 2015-11-20
Always check if log dirs exist before starting services via spinnaker…

duftler:
… umbrella service.

Stop/start spinnaker instead of restart since restart seems not happy about a service that is not already running.
@skorten or @dpeterka please review.


dpeterka:
LGTM

PR #610, opened 2015-11-20
Add slack property settings

gregturn:


config/default-spinnaker-local.yml
```diff
@@ -103,5 +103,8 @@ services:
         account: # twilio account id
         token: # twilio auth token
         from: # phone number by which sms messages are sent
+      slack:
```
tomaslin: should add a line here linking to slack api

PR #611, opened 2015-11-21
Updated script to create google development environment

ewiseblatt:
Added checks for gcloud and copy over configuration to avoid the oauth flow
on the other end.
Default to "upstream" rather than the gituser so by default you dont need forked repos.
Refactored copying files to copy lists of files.
Copy .gradle/gradle.properties

### DIFF
--- a/dev/bootstrap_dev.sh
+++ b/dev/bootstrap_dev.sh
@@ -93,6 +93,7 @@ function git_clone() {
   else
     git -C $git_project remote add upstream \
         https://github.com/$upstream_user/${git_project}.git
+    git -C $git_project remote set-url --push upstream disabled
   fi
 }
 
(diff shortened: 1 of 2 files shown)


ewiseblatt:
@duftler

dev/bootstrap_dev.sh
```diff
@@ -260,8 +264,14 @@ EOF
...
   print_invoke_instructions
 else
-  exec bash -l
   print_source_instructions
 fi
 
 print_spinnaker_reference
+
+if [[ $(basename "/bogus/$0") == "bootstrap_dev.sh" ]]; then
+  echo "Restart you shell to pick up path changes."
```
duftler: s/you/your

dev/create_google_dev_vm.py
```diff
@@ -65,6 +67,22 @@ def get_project(options):
...
+
+    The default zone is the current zone if on GCE or a random zone.
+    """
+    if not options.zone:
+      result = fetch(os.path.join(GOOGLE_INSTANCE_METADATA_URL, 'zone'),
+                                  google=True)
+      if result.ok():
+        options.zone = os.path.basename(result.content)
+      else:
+        options.zone = 'us-central1-b'
```
duftler: Please use `us-central1-f` here instead.

duftler:

ewiseblatt:
@duftler, I made additional changes and accidentally squashed them already. Sorry.
The gist of the extra changes is "maybe" copy methods that report unused useful options to streamline setup. In bootstrap, accept "none" as a github user (so you dont need a github id) to indicate skipping git setup.
PTAL
and "none" to not pull source

dev/create_google_dev_vm.py
```diff
@@ -65,6 +67,22 @@ def get_project(options):
     return options.project
 
 
+def get_zone(options):
+    """Determine the default availability zone.
+
+    The default zone is the current zone if on GCE or a random zone.
```
duftler: s/a random/an arbitrary

dev/bootstrap_dev.sh
```diff
@@ -182,12 +190,13 @@ if prompt_YN "Y" "Install (or update) Google Cloud Platform SDK?"; then
...
    echo "*** BEGIN installing gcloud..."
    curl https://sdk.cloud.google.com | bash
-   if ! $(gcloud auth list 2>&1 | grep "No credential"); then
+   if [[ ! -f $HOME/.config/gcloud/credentials ]]; then
       echo "Running gcloud authentication..."
       gcloud auth login
    else
       echo "*** Using existing gcloud authentication:"
       gcloud auth list
+      echo "To add anotheraccount, run gcloud auth login"
```
duftler: s/anotheraccount/another account

duftler:
What's the use case for running bootstrap and not checking out the source? I get that it will still do all the configuration, but why would someone do that if not actually working with the source?

ewiseblatt:
A few:
  - You are only working on a single repository or two and dont need the scripting at all
  - My own case where I'm working in my spinnaker/ fork but upstream/ for the rest.
  - The bootstrap assumes our preferences for https two-factor authentication, but cameron prefers ssh
    In fact, the current the github documentation invokes refresh_source.sh with --use_ssh.

I'd add back the ssh option in bootstrap_vm but need to cut scope somewhere for now.

duftler:
Thanks for the detailed explanation.
LGTM.

PR #615, opened 2015-11-23
Preserve original credentials of spinnaker-local.yml when transforming it.

ewiseblatt:
@duftler 
Just a recap of the past few commits:

I used to explicitly set the mode of files after copying them over within create_dev_vm.
I took that out to simplify a bit since gcloud already preserves files. However, there is a temp file created within create_dev to transform the jsonPath used in the new instance. That temp file was not being protected. In this PR, I copy the original file modes for the temp file.

When all is said and done, this can still result in new failures due to the check of sensitive file permissions within run_dev. However, these are legit in the sense that they would only occur if the source machine files were not properly protected.


duftler:
LGTM.

PR #618, opened 2015-11-23
update settings.js, removing preferredZones config

anotherchrisberry:
`primaryAccounts` and `challengeDestructiveActions` should be coming from Clouddriver now (https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.groovy#L73), as should `preferredZones` (https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentials.java#L132)

`whatsNew` is a Netflix-only component, so should not need to be in settings.js.

@duftler or @ewiseblatt can you review/sanity-check this?

### DIFF
--- a/config/settings.js
+++ b/config/settings.js
@@ -34,24 +34,14 @@ window.spinnakerSettings = {
         account: `${googlePrimaryAccount}`,
         region: `${googleDefaultRegion}`,
         zone: `${googleDefaultZone}`,
-      },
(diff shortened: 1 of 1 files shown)


duftler:
LGTM.

PR #624, opened 2015-11-25
Updated InstallSpinnaker.sh to include --local-install option

skorten:


skorten:
I think I've addressed all of your comments.  The one thing I will note is that apt repos for the dependency software are always added.  For instance, if you do --local-install apt repos for redis and cassandra are still added while the java 8 and spinnaker repos are not.

InstallSpinnaker.sh
```diff
@@ -245,6 +262,92 @@ function set_defaults_from_environ() {
       echo "No providers are enabled by default."
   fi
 }
+
+function add_apt_repositories() {
+  # Redis
+  # https://launchpad.net/~chris-lea/+archive/ubuntu/redis-server
+  add-apt-repository -y ppa:chris-lea/redis-server
```
ewiseblatt: Arent you still adding these repositories?
You cant do that in DOWNLOAD=true mode.

InstallSpinnaker.sh
```diff
@@ -245,6 +262,92 @@ function set_defaults_from_environ() {
...
+
+after installing java 8 run the following to install cassandra:
+
+sudo apt-get install -y --force-yes cassandra=2.1.11 cassandra-tools=2.1.11
+sudo apt-mark hold cassandra cassandra-tools
+
+EOF
+  else
+    apt-get install -y --force-yes cassandra=2.1.11 cassandra-tools=2.1.11
+    apt-mark hold cassandra cassandra-tools
```
ewiseblatt: when the apt-get install fails, this infinite loops

PR #627, opened 2015-11-25
Treat files in /opt/spinnaker/config as dpkg configuration files so t…

duftler:
…hey survive removal/upgrade.
@dpeterka please review.


dpeterka:
+1

PR #633, opened 2015-11-27
replace let statements with var statements

anotherchrisberry:
Safari and (possibly) Firefox throw exceptions on `let` statements in strict mode. This changeset attempts to get around this issue by foregoing some of the ES6 sugar we've introduced (`let`, string templates).

I honestly have no idea if this will resolve the issues we've been hearing about from Safari users. It's been over a week since I tried to perform a clean install of Spinnaker via the documentation, and I was never able to get it to work. Hoping @ewiseblatt or @tomaslin or @duftler can take a look and see if these changes do the trick.

Also updating the scripts to remove some very Netflix-specific settings (feature flags) and obsolete settings (preferred zones, primary account, challenge destructive actions) that should be coming from Clouddriver's /credentials endpoint.

### DIFF
--- a/pylib/spinnaker/configurator.py
+++ b/pylib/spinnaker/configurator.py
@@ -192,17 +192,17 @@ class Configurator(object):
     original_block = source[offset:end]
     # Remove all the explicit declarations in this block
     # Leaving us with just comments
-    block = re.sub('\n\s*let\s+\w+\s*=(.+)\n', '\n', original_block)
+    block = re.sub('\n\s*var\s+\w+\s*=(.+)\n', '\n', original_block)
     settings = [source[:offset]]
 
-    # Now iterate over the comments looking for let specifications
+    # Now iterate over the comments looking for var specifications
     offset = 0
-    for match in re.finditer('//\s*let\s+(\w+)\s*=\s*(.+?);?\n', block) or []:
+    for match in re.finditer('//\s*var\s+(\w+)\s*=\s*(.+?);?\n', block) or []:
       settings.append(block[offset:match.end()])
       offset = match.end()
       name = match.group(1)
       value = self.bindings.replace(match.group(2))
-      settings.append('let {name} = {value!r};\n'.format(
+      settings.append('var {name} = {value!r};\n'.format(
          name=name, value=value))
 
     settings.append(block[offset:])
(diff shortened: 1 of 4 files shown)


tomaslin:
The configurator would need to updated to replace lets with vars. https://github.com/spinnaker/spinnaker/blob/master/pylib/spinnaker/configurator.py#L195 ( and lines 200 and 205 )

tomaslin:
instructions on how to test deck reconfig is here - http://spinnaker.io/documentation/troubleshooting.html#i-changed-my-configuration-how-do-i-get-spinnaker-to-pick-up-the-modified-configuration but basically create a build directory on the same level as your spinnaker direction, checkout deck and then run `../spinnaker/runtime/reconfigure_spinnaker.sh`

ewiseblatt:
This looks right (with the configurator.py change) but I havent tried it
yet. Will try tomorrow.

ewiseblatt:
Actually spinnaker/ can be anywhere (including in build).
The constraint is that when you run the script, you need to be in the build
directory (parent of deck) so it knows where deck/ is.

Unfortunately reconfigure_spinnaker.sh is going to rewrite the same
deck/settings.js that is under git version control, but that is because
there isnt a way to inject a settings.js file (as we discussed a while
back).

pylib/spinnaker/configurator.py
```diff
@@ -192,17 +192,17 @@ def process_deck_settings(self, source):
...
+    # Now iterate over the comments looking for var specifications
     offset = 0
-    for match in re.finditer('//\s*let\s+(\w+)\s*=\s*(.+?);?\n', block) or []:
+    for match in re.finditer('//\s*var\s+(\w+)\s*=\s*(.+?);?\n', block) or []:
       settings.append(block[offset:match.end()])
       offset = match.end()
       name = match.group(1)
       value = self.bindings.replace(match.group(2))
-      settings.append('let {name} = {value!r};\n'.format(
+      settings.append('var {name} = {value!r};\n'.format(
```
anotherchrisberry: updated per comments from Tomás/Eric (also updated base settings.js in deck - PR pending here: https://github.com/spinnaker/deck/pull/1759)

ewiseblatt:
The other occurrences of 'let' in the test need to be 'var' as well for the test to pass.
Seems ok after that.

anotherchrisberry:
@ewiseblatt thanks for catching that - I've updated the `let` statements in `configurator_test.py` to `var`

PR #636, opened 2015-11-30
Allow Configurator unittest to be run from any directory.

ewiseblatt:
@lwander

### DIFF
--- a/unittest/configurator_test.py
+++ b/unittest/configurator_test.py
@@ -23,18 +23,30 @@ from spinnaker.configurator import InstallationParameters
 from spinnaker.yaml_util import YamlBindings
 
 
+
 class ConfiguratorTest(unittest.TestCase):
     def test_configuration(self):
-        configurator = Configurator()
-        self.assertEquals(os.path.join(os.environ['HOME'], '.spinnaker'),
-                          configurator.user_config_dir)
+        temp_homedir = tempfile.mkdtemp()
+        home_spinnaker_dir = os.path.join(temp_homedir, '.spinnaker')
+        os.mkdir(home_spinnaker_dir)
+        with open(os.path.join(home_spinnaker_dir,
+                               'spinnaker-local.yml'), 'w') as f:
+          f.write('empty:\n')
+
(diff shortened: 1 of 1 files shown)


unittest/configurator_test.py
```diff
@@ -23,18 +23,30 @@
...
+                               'spinnaker-local.yml'), 'w') as f:
+          f.write('empty:\n')
+
+        # We arent using HOME in a test, but want to verify that it is
+        # being used.
+        old_home = os.environ['HOME']
+        os.environ['HOME'] = temp_homedir
+        try:
+            configurator = Configurator()
+        finally:
```
lwander: Does this catch signals that could kill the test before restoring `$HOME`?

unittest/configurator_test.py
```diff
@@ -99,5 +114,14 @@ def test_update_deck_settings(self):
...
+
+  # The configurator requires we be run in the build directory
+  # containing deck. So let's pretend this was the case.
+  temp_indir = tempfile.mkdtemp()
+  os.environ['PWD'] = temp_indir
+  os.mkdir(os.path.join(temp_indir, 'deck'))
+  try:
+      unittest.TextTestRunner(verbosity=2).run(suite)
+  finally:
+      shutil.rmtree(temp_indir)
```
lwander: Is it OK to remove `$PWD`?

unittest/configurator_test.py
ewiseblatt: Why is that a concern here?

lwander: Canceling the test with `^C` could leave you with a broken home directory.

ewiseblatt: Subprocess environments shouldnt leak up to the parent's environment.

lwander: Ah, right

anotherchrisberry:
@ewiseblatt @lwander is this PR still active or can it be closed?

ewiseblatt:
submitted. thansk for the headds up.

PR #638, opened 2015-12-01
Adds apt-mark hold/unhold around doing apt-get updates

skim1420:
@spinnaker/google

### DIFF
--- a/install/first_google_boot.sh
+++ b/install/first_google_boot.sh
@@ -222,8 +222,10 @@ else
 fi
 
 # apply outstanding updates since time of image creation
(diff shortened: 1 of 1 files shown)


install/first_google_boot.sh
```diff
@@ -222,8 +222,10 @@ else
 fi
 
 # apply outstanding updates since time of image creation
```
ewiseblatt: Update comment here to say
Apply outstanding OS updates since time of image creation
but keep spinnaker version itself intact.

ewiseblatt: Do we want to pin redis while we are at it?

ewiseblatt: redis-server

skim1420:
Leaving out pinning redis for now, to handle in separate PR if desired.

PR #645, opened 2015-12-04
How to build local images and run them in compose

tomaslin:


tomaslin:
@duftler PTAL

experimental/docker-compose/README.md
```diff
@@ -37,7 +37,45 @@ Under Kitematic, click on the application 'deck' -> settings -> ports. Click on
...
+
+## Using locally built images
+
+Sometimes it is useful to use the docker compose configuration to test changes that span multiple services locally.
+
+We provide a development configuration that will build a service from source instead of downloading an image from the remote repository. 
+
+The setup is similar to the local configuration outlined at the root of this project.
+
+* Create a directory called build that sits at the same level as the spinnaker directory ( ../../../build/ from the docker-compose directory ).
```
duftler: Maybe say "is a sibling to the spinnaker directory" instead of "sits at the same level as"? Could probably drop the ../../../build stuff then.

experimental/docker-compose/README.md
```diff
@@ -37,7 +37,45 @@ Under Kitematic, click on the application 'deck' -> settings -> ports. Click on
...
+
+* Call `../spinnaker/dev/refresh_source.sh --pull_origin --use_ssh --github_user default` to download Spinnaker source files. 
+
+* Navigate back to the docker-compose directory.
```
duftler: Could use back-ticks around `docker-compose`.

experimental/docker-compose/README.md
```diff
@@ -37,7 +37,45 @@ Under Kitematic, click on the application 'deck' -> settings -> ports. Click on
 
 <img src="https://cloud.githubusercontent.com/assets/74310/11158618/4bba7122-8a0e-11e5-83b6-8ff2297562b2.png"/>
 
-Alternatively, you can just enter ```DOCKER_IP=`docker-machine ip default` && open http://$DOCKER_IP:9000```
+Alternatively, you can just enter ```open http://`docker-machine ip default`:9000```
+
+## Using locally built images
+
```
duftler: Don't we need steps (1) and (2) and half of (3) (just the `DOCKER_IP=...` bit) from "Local Installation" above?

duftler: The first 2 of the 3 `docker-compose` sample commands below don't include `DOCKER_IP=...`. So we could add it to those 2, but you'd probably still want to include steps (1) and (2) from above.

duftler: After reading through this several times, I think it just needs to be made a bit clearer if the user is expected to complete the "Local Installation" section first or just some of it or something else.

duftler:

experimental/docker-compose/README.md
tomaslin: you only need docker ip when you are running but not when you are calling it to build an image.

PR #651, opened 2015-12-11
fix linking error on post install during upgrade

layered-esell:
When upgrading via apt-get this error pops up:

    Setting up spinnaker (0.15.0-3) ...
    ln: failed to create symbolic link ‘/opt/spinnaker/scripts/bin’: File exists
    dpkg: error processing package spinnaker (--configure):
     subprocess installed post-installation script returned error exit status 1
    Errors were encountered while processing:
     spinnaker
    E: Sub-process /usr/bin/dpkg returned an error code (1)

This PR just adds some logic to only create that link if it doesn't already exist.


duftler:
Thanks for the contribution!
LGTM.

PR #658, opened 2015-12-15
Mark /etc/default/spinnaker as a configuration file so it is not sile…

duftler:
…ntly overwritten on an upgrade.
@dpeterka please review.


dpeterka:
+1, no issues.

PR #661, opened 2015-12-16
use images in quay.io

tomaslin:
Deck just seems unable to consistently build on docker hub automated builds.


tomaslin:
@ttomsu ptal

PR #663, opened 2015-12-16
Replace `app` with `component` in pod query.

duftler:
@ttomsu please review.


ttomsu:
d'oh! Find/Replace all is hard.

PR #672, opened 2015-12-17
Specify explicit machine type for container nodes.

duftler:
Need enough cpu capacity when running everything on one node.
@ttomsu please review.

### DIFF
--- a/experimental/kubernetes/README.md
+++ b/experimental/kubernetes/README.md
@@ -43,7 +43,7 @@ parameter if you skip this step.
 copying files to the node, so for simplicity, we will only use 1 host.
 
   ```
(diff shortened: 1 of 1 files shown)


experimental/kubernetes/README.md
```diff
@@ -43,7 +43,7 @@ parameter if you skip this step.
 copying files to the node, so for simplicity, we will only use 1 host.
 
   ```
-  $ gcloud container clusters create my-spinnaker-on-kubernetes --num-nodes 1 --scopes storage-rw,compute-rw
+  $ gcloud container clusters create my-spinnaker-on-kubernetes --num-nodes 1 --scopes storage-rw,compute-rw --machine-type n1-standard-8
```
ttomsu: I think we should be able to get away with an n1-standard-4. That's what C2D uses, no?

duftler:
PTAL.

PR #674, opened 2015-12-17
Fix startup local install

ewiseblatt:
Uses package manager where it can and uses direct download where it cannot of DOWNLOAD is true

Fixed a couple other bugs when performing a non-standard installation

@skorten

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -128,7 +128,6 @@ function process_args() {
           ;;
       --local-install)
           DOWNLOAD="true"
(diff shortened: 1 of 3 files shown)


ewiseblatt:
Note that I want to take away the requires openjdk-8 from spinnaker.deb. We'd removed this everywhere else a while ago, and this still deploys [on gce] but I dont know if there are other implications.

skorten:
LGTM

Should be no issue with removing the openjdk-8 dependency from the deb configuration.

PR #679, opened 2016-01-08
Related to pull request in deck, move install dir from /var/www to /var/www/html

gardleopard:
This change is related to this pull request in deck:
https://github.com/spinnaker/deck/pull/1851

More information in the deck pull request. This should not be merged unless the deck pull request is merged.

@cfieber here is the change you requested.


anotherchrisberry:
@cfieber are we good with this?

PR #681, opened 2016-01-11
Handle bintray 409

ewiseblatt:
When publishing InstallSpinnaker.sh to bintray from dev/build_spinnaker.sh, it
is not uncommon to get a 409 error. I think this is because the content changes but
version numbers do not (until the project is explicitly released). To work around
this, when encountering a 409 publising an update, try to delete the exiting content
and retry.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -248,12 +248,12 @@ class Builder(object):
       debian_tags = ';' + debian_tags
 
     url = ('https://api.bintray.com/content'
-               '/{subject}/{repo}/{package}/{version}/{path}'
-               '{debian_tags}'
-               ';publish=1;override=1'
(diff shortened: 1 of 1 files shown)


ewiseblatt:
@lwander

dev/build_release.py
```diff
@@ -266,35 +266,56 @@ def publish_to_bintray(self, source, package, version, path, debian_tags=''):
         try:
             result = urllib2.urlopen(put_request, data)
         except HTTPError as put_error:
-            if put_error.code != 400:
+            if put_error.code == 409:
```
lwander: Looks like this is actually the suggested workaround: https://github.com/pledbrook/lazybones/issues/111#issuecomment-46432234

lwander:
LGTM

PR #693, opened 2016-01-20
Dont touch existing spinnaker-local.yml file when upgrading

ewiseblatt:
@dpeterka

### DIFF
--- a/pkg_scripts/postInstall.sh
+++ b/pkg_scripts/postInstall.sh
@@ -9,8 +9,10 @@ if [ `readlink -f /opt/spinnaker/scripts` != "/opt/spinnaker/bin" ]; then
   ln -s /opt/spinnaker/bin /opt/spinnaker/scripts
 fi
 
-# Create master config
-cp /opt/spinnaker/config/default-spinnaker-local.yml /opt/spinnaker/config/spinnaker-local.yml
+if [ ! -f /opt/spinnaker/config/spinnaker-local.yml ]; then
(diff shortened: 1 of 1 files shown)


dpeterka:
This should probably be handled by the package's handler for config files.

https://github.com/spinnaker/spinnaker/blob/master/build.gradle#L88

https://wiki.debian.org/ConfigPackages
http://www-uxsup.csx.cam.ac.uk/~jw35/docs/rpm_config.html

I thought @duftler added a file for this already.

ewiseblatt:
While I'm not sure I'm doing the right thing here, I'm intentionally not
calling it a config file.
The default-spinnaker-local was never intended to be copied on install as
it is.

spinnaker-local was intended to be user-generated and managed -- only
overrides need to go in that file. And it was intended to be put in
user-controlled directory (~spinnaker) not an system/install controlled one
(/opt/spinnaker).

I'm suggesting that it is not really a "config file" from the installer
perspective. You added the original cp to make it more obvious where and
how to configure. On an upgrade, that step is already done. The
spinnaker-local.yml is likely to have some changes in it. Future changes to
the file, are likely to reiterate existing defaults (as the current one
does) so will cause MD5 conflicts that are not really meaningful since no
action would have been needed anyway.

Does that change your recommendation on how to fix the issue?
Or do you still disagree and think it should be called a config file?

dpeterka:
You make a good argument I won't disagree :) Merge away.

PR #699, opened 2016-01-24
Fail install on errors.

ewiseblatt:
@dpeterka 
We are seeing occasional failures where a repository is temporarily unavailable (usually its the packer download) so want the caller to know the install failed. I'm not signing my private repository so I'm allowing there not to be a key. If you think we should require it to be signed then I'll make the other changes I need to sign it and take that change out of this PR.


dpeterka:
Other than what I think are some leftovers from testing (mayb not).  Seems good.

PR #701, opened 2016-01-25
Add AWS remote docker-machine instructions

jason-riddle:
I see that there are instructions for creating a docker-machine on Azure and GCP for the docker-compose project. I went ahead added instructions for creating a docker-machine on AWS and I cleaned up the markdown formatting a bit as well.


tomaslin:
Thanks

PR #702, opened 2016-01-25
Sync settings.js with deck

AMeng:
The current settings.js file is outdated compared to the one in deck. This pulls in most of the latest changes. Some of the things in the deck version appear to be for internal use only? I wasn't sure if everything should be copied directly.

https://github.com/spinnaker/deck/blob/master/settings.js

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -94,6 +94,12 @@ services:
     # "igor" subsystem.
     enabled: false
 
+  deck:
+    # Frontend configuration.
+    timezone: 'America/Los_Angeles' # See: http://momentjs.com/timezone/docs/#/data-utilities/
+    auth:
+      enabled: false
+
   rush:
     # Spinnaker's "rush" subsystem is used by the "rosco" bakery.
     # You need to provide the fully-qualified path to the directory containing the packer templates.
@@ -101,7 +107,7 @@ services:
     configDir: /opt/rosco/config/packer
 
   echo:
-    # allow pipeline triggers to run periodically via cron expressions  
+    # allow pipeline triggers to run periodically via cron expressions
     cron:
       enabled: true
     # If you wish for spinnaker to send notifications, configure the following blocks
@@ -119,10 +125,8 @@ services:
         account: # twilio account id
         token: # twilio auth token
         from: # phone number by which sms messages are sent
-      # See https://api.slack.com/incoming-webhooks for details about using incoming webhooks 
+      # See https://api.slack.com/incoming-webhooks for details about using incoming webhooks
       # and how to register your own webhook.
       slack:
         enabled: false
         token: # the part after http://hooks.slack.com/services/
-
-
--- a/config/settings.js
+++ b/config/settings.js
@@ -11,14 +11,22 @@
 
 // var gateUrl = ${services.gate.baseUrl};
 // var bakeryBaseUrl = ${services.bakery.baseUrl};
+// var authEnabled = ${services.deck.auth.enabled};
+// var defaultTimeZone = ${services.deck.timezone};
 // var awsDefaultRegion = ${providers.aws.defaultRegion};
 // var awsPrimaryAccount = ${providers.aws.primaryCredentials.name};
 // var googleDefaultRegion = ${providers.google.defaultRegion};
 // var googleDefaultZone = ${providers.google.defaultZone};
 // var googlePrimaryAccount = ${providers.google.primaryCredentials.name};
+// var azureDefaultRegion = ${providers.azure.defaultRegion};
+// var azurePrimaryAccount = ${providers.azure.primaryCredentials.name};
 // var cfDefaultRegion = ${providers.cf.defaultOrg};
 // var cfDefaultZone = ${providers.cf.defaultSpace};
 // var cfPrimaryAccount = ${providers.cf.primaryCredentials.name};
+// var titanDefaultRegion = ${providers.titan.defaultRegion};
+// var titanPrimaryAccount = ${providers.titan.primaryCredentials.name};
+// var kubernetesDefaultNamespace = ${providers.kubernetes.primaryCredentials.namespace};
+// var kubernetesPrimaryAccount = ${providers.kubernetes.primaryCredentials.name};
 
 // END reconfigure_spinnaker
 /**
@@ -29,9 +37,16 @@
 window.spinnakerSettings = {
   gateUrl: gateUrl,
   bakeryDetailUrl: bakeryBaseUrl + '/api/v1/global/logs/{{context.status.id}}?html=true',
+  authEndpoint: gateUrl + '/auth/info',
   pollSchedule: 30000,
-  defaultTimeZone: 'America/New_York', // see http://momentjs.com/timezone/docs/#/data-utilities/
+  defaultTimeZone: defaultTimeZone, // see http://momentjs.com/timezone/docs/#/data-utilities/
   providers: {
+    azure: {
+      defaults: {
+        account: azurePrimaryAccount,
+        region: azureDefaultRegion
+      },
+    },
     gce: {
       defaults: {
         account: googlePrimaryAccount,
@@ -50,12 +65,27 @@ window.spinnakerSettings = {
         account: cfPrimaryAccount,
         region: cfDefaultRegion
       },
+    },
+    titan: {
+      defaults: {
+        account: titanPrimaryAccount,
+        region: titanDefaultRegion
+      },
+    },
+    kubernetes: {
+      defaults: {
+        account: kubernetesPrimaryAccount,
+        namespace: kubernetesDefaultNamespace
+      },
     }
   },
-  authEnabled: false,
+  authEnabled: authEnabled,
   feature: {
     pipelines: true,
     notifications: false,
+    fastProperty: false,
+    vpcMigrator: false,
+    clusterDiff: false,
     rebakeControlEnabled: true,
     netflixMode: false,
   },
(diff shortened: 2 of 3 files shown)


config/settings.js
```diff
@@ -27,11 +27,20 @@
...
-  bakeryDetailUrl: bakeryBaseUrl + '/api/v1/global/logs/{{context.status.id}}?html=true',
+  bakeryDetailUrl: bakeryBaseUrl + '/api/v1/global/logs/{{context.status.id}}?html=true'
+  authEndpoint: process.env.AUTH_ENDPOINT,
   pollSchedule: 30000,
-  defaultTimeZone: 'America/New_York', // see http://momentjs.com/timezone/docs/#/data-utilities/
+  defaultTimeZone: process.env.TIMEZONE || 'America/Los_Angeles', // see http://momentjs.com/timezone/docs/#/data-utilities/
   providers: {
+    azure: {
+      defaults: {
+        account: 'azure-test',
```
AMeng: Should this be added to the top as something like this:
```javascript
// var azurePrimaryAccount = ${providers.azure.primaryCredentials.name};
```

config/settings.js
```diff
@@ -51,12 +60,31 @@ window.spinnakerSettings = {
...
+    gistId: '32526cd608db3d811b38',
+    fileName: 'news.md',
   },
-  authEnabled: false,
+  authEnabled: process.env.AUTH === 'enabled',
   feature: {
     pipelines: true,
     notifications: false,
-    rebakeControlEnabled: true,
+    fastProperty: true,
```
duftler: These 3 values should all be `false`, and `rebakeControlEnabled` should be `true`.

config/settings.js
duftler: @ewiseblatt Can you please point @AMeng in the right direction as to what (if anything) would need to be modified in the checked-in 'centralized config' stuff to enable the additional accounts to be rewritten here?

config/settings.js
```diff
@@ -27,11 +27,20 @@
...
 window.spinnakerSettings = {
+  defaultProviders: ['aws', 'gce', 'azure', 'cf', 'kubernetes'],
+  feedbackUrl: process.env.FEEDBACK_URL,
   gateUrl: gateUrl,
-  bakeryDetailUrl: bakeryBaseUrl + '/api/v1/global/logs/{{context.status.id}}?html=true',
+  bakeryDetailUrl: bakeryBaseUrl + '/api/v1/global/logs/{{context.status.id}}?html=true'
+  authEndpoint: process.env.AUTH_ENDPOINT,
   pollSchedule: 30000,
-  defaultTimeZone: 'America/New_York', // see http://momentjs.com/timezone/docs/#/data-utilities/
+  defaultTimeZone: process.env.TIMEZONE || 'America/Los_Angeles', // see http://momentjs.com/timezone/docs/#/data-utilities/
```
AMeng: Will this actually pull values from `/etc/default/spinnaker`? I'm not sure how the individual processes startup. If not, I think new values should be added to the YAML config for things like timezone.

config/settings.js
ewiseblatt: @AMeng 
If you want to bind a variable to a spring configuration value (from spinnaker-local.yml + spinnaker.yml + /etc/default/spinnaker) then declare a var comment for it as you did, then reference that variable.

config/settings.js
ewiseblatt: settings.js is used by deck, which is typically run within apache or some other webserver neither of which knows about spinnaker. So I dont think you can rely on it. @anotherchrisberry would know.

config/settings.js
ewiseblatt: I dont think you need to change anything to add new providers and have them
get picked up by the centralized config stuff. Ping me if it doesnt seem to
"just work".

config/settings.js
```diff
@@ -27,11 +27,20 @@
  */
 
 window.spinnakerSettings = {
+  defaultProviders: ['aws', 'gce', 'azure', 'cf', 'kubernetes'],
+  feedbackUrl: process.env.FEEDBACK_URL,
   gateUrl: gateUrl,
-  bakeryDetailUrl: bakeryBaseUrl + '/api/v1/global/logs/{{context.status.id}}?html=true',
+  bakeryDetailUrl: bakeryBaseUrl + '/api/v1/global/logs/{{context.status.id}}?html=true'
```
anotherchrisberry: missing a comma at the end here.

config/settings.js
anotherchrisberry: It would be nice to move this value into one of the `spinnaker.yml` files and leave it as-is in Deck's settings.js file itself. I will talk to @ewiseblatt today and see if that makes sense.

AMeng:
So, because deck will not have access to environment variables, I've added a `deck` section under `services` for the configuration. The defaults have not changed, but they can now be easily overridden in the YAML configuration.

config/settings.js
```diff
@@ -27,11 +35,19 @@
  */
 
 window.spinnakerSettings = {
+  defaultProviders: ['aws', 'gce', 'azure', 'cf', 'kubernetes'],
```
AMeng: Should this list be constructed dynamically based on these services being enabled? This is just copied from deck's settings.js.

config/settings.js
AMeng: Create a `deck` section under services. Let me know what you think.

config/settings.js
anotherchrisberry: That would be nice - not sure how difficult it would be, but definitely would be nice.

anotherchrisberry:
LGTM - defer to @ewiseblatt on whether it's ready to be merged or not...

config/settings.js
ewiseblatt: We could potentially special case this by having the script that rewrites
this file derive the value for this based on the provider.enabled values it
encounters.

AMeng: Actually, looking at deck, I think this is just a fallback list in case no cloud providers are enabled. It seems like this can probably just be removed.

See [code](https://github.com/spinnaker/deck/blob/fee4c9839a465a614c105a52267e3ccb9f0e63e5/app/scripts/modules/core/account/account.service.js#L71) and [test](https://github.com/spinnaker/deck/blob/0a9ed6034363ee8c1c75958822b4e9009f6781b9/app/scripts/modules/core/account/account.service.spec.js#L130).

config/settings.js
```diff
@@ -51,11 +67,30 @@ window.spinnakerSettings = {
...
+      },
+    },
+    kubernetes: {
+      defaults: {
+        account: kubernetesPrimaryAccount,
+        namespace: kubernetesDefaultNamespace
+      },
+    }
+  },
+  whatsNew: {
```
ewiseblatt: what's this? Both the whatsNew container and the gistId.
Does this settings.js file need to change with every new release?
That doesnt seem desirable.

AMeng: https://github.com/spinnaker/deck/blob/master/settings.js#L50

I honestly just copied it from deck's `settings.js` file to try to sync them up more. Should I remove it?

config/settings.js
ewiseblatt: I assume you'll also verify the final form in a running deployment.

config/settings.js
ewiseblatt: @anotherchrisberry please advise.

anotherchrisberry: We can probably `whatsNew` from both here and from deck's settings.js - it's Netflix-specific.

AMeng:
Ran through a full test of this, and noticed that currently booleans don't work with the reconfigure script. I changed deck.auth to a string of either "disabled" or "enabled". Feels weird, but currently fields have to be strings.

ewiseblatt:
I'll patch your CL in and try myself to see what's going on.
I wont get to it until later tonight though.
Ping me in the morning if you dont hear back.

ewiseblatt:
When you say "dont work", what is the behavior you are seeing?
And where does it not work (e.g. writing the values into settings.js)?
Thanks.

ewiseblatt:
Alex, I didnt get to this last night but looking now.
The problem I am seeing is that the boolean values are rendering as strings.
Is the problem they should be boolean? I'll fix that now and send you a PR.

AMeng:
@ewiseblatt that is correct. I hadn't looked into fixing that Python script, but that's exactly what I was seeing.

AMeng:
This requires https://github.com/spinnaker/spinnaker/pull/704 to be merged first.

ewiseblatt:
LGTM. I'm not familiar with the actual usage. I know @anotherchrisberry LGTM'd it yesterday, but I'm not sure on what's different from when he LGTMd it. I'm going to defer back to him. He's at an offsite so wont get to it before later today. I submitted the dependency so this is good to go once he (or someone more deck savvy) reconfirms. If you can point me to the version he approved as a reference point, I can take another look as well.

config/default-spinnaker-local.yml
```diff
@@ -94,6 +94,12 @@ services:
     # "igor" subsystem.
     enabled: false
 
+  deck:
+    # Frontend configuration.
+    timezone: 'America/Los_Angeles' # See: http://momentjs.com/timezone/docs/#/data-utilities/
+    auth:
+      enabled: true
```
anotherchrisberry: Thinking this should be `false` by default, eh?

config/spinnaker.yml
```diff
@@ -53,6 +53,11 @@ services:
         enabled: false
         token: # the part after http://hooks.slack.com/services/
 
+  deck:
+    timezone: 'America/Los_Angeles'
+    auth:
+      enabled: true
```
anotherchrisberry: Default should probably be `false` here, too.

config/default-spinnaker-local.yml
AMeng: Yep, good catch.

anotherchrisberry:
LGTM - thanks, @AMeng !

PR #704, opened 2016-01-27
Fixed yaml replace() to return typed values rather than forcing strings

ewiseblatt:
@AMeng 
I havent fully confirmed this yet. It is going to take me a while. However I believe this works.
You can review it and try it out to verify it fixes your problem.


ewiseblatt:
I think I need to emit javascript true/false rather than True/False. Hold on.

ewiseblatt:
@AMeng
not quite right. works with bools in the yaml now, but environment variables are always strings. hold on..

AMeng:
Yep, this is working for booleans now. I've updated my PR.

Issue #709, opened 2016-02-02
Setting zone in Bake stage


hugoboos:
Hi,

I'm following the Codelab at [Spinnaker.io](http://spinnaker.io/documentation/source-to-prod-codelab.html).

Instead of using `ZONE=us-east1-b `, I have used `ZONE=europe-west1-b` (described in the section 'Provision an instance of Spinnaker and Jenkins').

The issue is that the bake stage is building the Packer vm in the `us-central1` zone and not in the `europe-west1` zone.

Also there is no option in the GUI to change the baking zone.

duftler:
What is the issue with the zone used by packer? It will shutdown the instance after configuring it and capture the image. The image itself is global, so it shouldn't matter much which zone the packer instance is provisioned in.

Should be straightforward to configure the zone, I'll grab those details now. Just curious about your use-case.

duftler:
There are two options here:


1) Edit the rosco.yml to choose a different zone (this will affect all bakes):
`sudo vi /opt/rosco/config/rosco.yml`
And change this property: https://github.com/spinnaker/rosco/blob/master/rosco-web/config/rosco.yml#L128
Then you'll need to cycle rosco to pick up the config changes: http://spinnaker.io/documentation/troubleshooting.html#i-changed-my-configuration-how-do-i-get-spinnaker-to-pick-up-the-modified-configuration


2) Edit the pipeline config json in the ui to specify the value just for that bake stage:

In the pipeline editor, select Pipeline Actions->Edit as JSON. Then add an "extendedAttribute" block into whichever stage has "type":"bake". You can add mappings for any user variable you have defined in your packer template (https://github.com/spinnaker/rosco/blob/master/rosco-web/config/packer/gce.json#L3-L8):
```
      "extendedAttributes": {
        "gce_zone": "europe-west1-b",
        "gce_project_id": "my-project",
        "gce_account_file": "/some/path/to/creds.json"
      }
```


The issue with (2) is that support for this was just merged late yesterday and is not yet present in the codelab image. So, at the moment, (1) is the only real option.

Also, we should probably update the "unified" config to respect the ZONE specified at provisioning time. I'll look into that.

hugoboos:
I'll try option 1.

There is also an key `region` in the JSON:

```json
{
  "type": "bake",
  "name": "Bake",
  "cloudProviderType": "gce",
  "region": "global",
  "user": "[anonymous]",
  "baseOs": "precise",
  "baseLabel": "release",
[...]

PR #715, opened 2016-02-02
Override GCE default bakery zone with zone specified as part of unifi…

duftler:
…ed configuration.


duftler:
Closes #709.

duftler:
@ewiseblatt or @ttomsu please review.

PR #725, opened 2016-02-05
rosco.yml had incorrect value for redis

cfieber:
From https://github.com/spinnaker/rosco/blob/master/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/persistence/config/JedisConfig.groovy#L32

@duftler PTAL


duftler:
LGTM.

Issue #678, opened 2016-01-07
Some details of document to run spinnaker on kubernetes


larry-fuy:
When I run spinnaker on kubernetes in Google Cloud according to README, an issue of clouddriver always rebooting hits me painfully. After hours digging, I found it is only fixed only when I set provider/aws/enabled as false in spinnaker-local.yml. Although this should be a tiny problem for someone familiar with spinnaker, as a new user I spend hours to figure out. So I suggest to add this detail into README. 

Another thing I noticed when I scramble on spinnaker-local.yml is the header comments ask to copy this file to .spinnaker/, but the README file said it should be copied to .spinnaker/config. This may be another inconsistency between code and documents.

tomaslin:
@ttomsu could you comment on this?

ttomsu:
Thanks for the feedback! I'm sorry that something so small caused you hours of pain :-(

I've removed the AWS sections to keep the setup simple.

PR #727, opened 2016-02-05
Disables AWS in experimental/kubernetes by default

ttomsu:
Per https://github.com/spinnaker/spinnaker/issues/678

### DIFF
--- a/experimental/kubernetes/spinnaker-local.yml
+++ b/experimental/kubernetes/spinnaker-local.yml
@@ -7,12 +7,7 @@ providers:
     # If you want to deploy some services to Amazon Web Services (AWS),
     # set enabled and provide primary credentials for deploying.
     # Enabling AWS is independent of other providers.
-    enabled: true
-    defaultRegion: us-east-1
-    defaultIAMRole: BaseIAMRole
-    primaryCredentials:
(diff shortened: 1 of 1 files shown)


experimental/kubernetes/spinnaker-local.yml
```diff
@@ -7,12 +7,7 @@ providers:
     # If you want to deploy some services to Amazon Web Services (AWS),
     # set enabled and provide primary credentials for deploying.
     # Enabling AWS is independent of other providers.
-    enabled: true
-    defaultRegion: us-east-1
-    defaultIAMRole: BaseIAMRole
-    primaryCredentials:
-      name: my-aws-account
-      # Store actual credentials in /home/spinnaker/.aws/credentials
+    enabled: false
```
duftler: Would it make sense to leave the rest of the aws config in place, and just set `enabled` to `false`? Otherwise one would have to go hunting around for the other attributes that need to be configured when flipping that flag.

ttomsu: I'd say keep it simple - there's already a lot going on in this file and additional, disabled junk just makes it harder to understand what's really going on.

PR #731, opened 2016-02-05
By default, disable AWS and Google

gregturn:
Supply default values of false in the event SPINNAKER_AWS_ENABLED and SPINNAKER_GOOGLE_ENABLED aren't set.


duftler:
@ewiseblatt This change looks good to me. Make sense to you as well?

PR #741, opened 2016-02-12
adding redis connection property for igor

diananana:
Looks like igor is unable to find redis or connect to it without this added configuration. We had a similar commit recently to the rosco config.


cfieber:
Thanks!

PR #758, opened 2016-02-23
Packer url moved

TheDigitalNinja:
Packer files are no longer on bin tray.


duftler:
Thanks for the fix.
LGTM.

PR #763, opened 2016-02-25
Fixed path for cf account name + password

gastaldi:
Feel free to accept it or reject it in case it doesn't look right


tomaslin:
@gregturn can you comment on this?

gregturn:
LGTM

PR #773, opened 2016-03-02
Fixed run_dev startup

ewiseblatt:
I introduced a bug into developer run_dev earlier today when I took out
AWS credentials from the spinnaker.yml. This fixes that.

@duftler


duftler:
Lgtm.

PR #778, opened 2016-03-03
Add clouddriver baseUrl to igor to enable docker polling

lwander:
@duftler 

Useful for https://github.com/spinnaker/igor/pull/64


lwander:
Hey @tomaslin, this should be enough to get the docker-compose setup to work, right? The current kubernetes sample deployment doesn't support Igor, and should probably be deprecated (I'll hopefully have a better one compatible with spinnaker soon).

tomaslin:
need to add clouddriver host and link to igor in the docker config - https://github.com/spinnaker/spinnaker/blob/master/experimental/docker-compose/docker-compose.yml#L64

lwander:
Added to docker.yml

PR #780, opened 2016-03-07
Remove circular dependency on services

lwander:
@ewiseblatt

### DIFF
--- a/config/igor.yml
+++ b/config/igor.yml
@@ -12,10 +12,11 @@ jenkins:
 
 redis:
   connection: ${services.redis.connection:redis://localhost:6379}
+  
 
-services:
-  clouddriver:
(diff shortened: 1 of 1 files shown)


config/igor.yml
```diff
@@ -12,10 +12,11 @@ jenkins:
 
 redis:
   connection: ${services.redis.connection:redis://localhost:6379}
+  
 
-services:
-  clouddriver:
-    baseUrl: ${services.clouddriver.baseUrl}
+# services:
```
ewiseblatt: Maybe change this to a comment that looks something like
# Also depends on "services.clouddriver.baseUrl", and "services.echo.baseUrl"
# which are normally declared directly in spinnaker[-local].yml

PR #784, opened 2016-03-09
Force subsystem services to manual upstart when spinnaker is installed.

ewiseblatt:
This is so microservices can assume they are standalone and auto upstart when they install.
Then if spinnaker is installed, it can oversee their startup collectively. The main benefit
here is to preserve the service start/stop/restart spinnaker to operate on the collection
and keep the current behaviors while supporting standalone service usage.

### DIFF
--- a/pkg_scripts/postInstall.sh
+++ b/pkg_scripts/postInstall.sh
@@ -32,15 +32,10 @@ cqlsh -f "/opt/spinnaker/cassandra/create_echo_keyspace.cql"
 cqlsh -f "/opt/spinnaker/cassandra/create_front50_keyspace.cql"
 cqlsh -f "/opt/spinnaker/cassandra/create_rush_keyspace.cql"
 
-# Start all the services
-
-# start 'clouddriver'
-# start 'orca'
-# start 'front50'
-# start 'rush'
-# start 'rosco'
-# start 'echo'
-# start 'gate'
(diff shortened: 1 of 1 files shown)


ewiseblatt:
@dpeterka, @peterpd
I think this change would allow https://github.com/spinnaker/rush/pull/46 to modify the rush.conf to startup instead of a special standalone and check for spinnaker.

peterpd:
Agree with Eric. that is a good solution when all installed together
I also modified PR spinnaker/rush#46

ewiseblatt:
ping @dpeterka.

@cfieber, FYI looks like when this goes in, the other subsystems will autostart on boot by default but installing spinnaker on the VM will suppress that behavior, and instead spinnaker will start them up (and will remain auto start on boot). Please confirm you are ok with this and we wont break you (my concern is changing clouddriver, etc).

ewiseblatt:
@cfieber the auto start isnt from this change, but allowing it is enabled by this change, so I will start approving peter's other PRs.

dpeterka:
I have been combating the flu sorry for my absence I need to take a look at this tomorrow.

ewiseblatt:
@skorten can you have a look since David is still out?

skorten:
@ewiseblatt Looking now.

ewiseblatt:
thanks

PR #785, opened 2016-03-09
Avoid breaking existing infrastructure dependent on igor polling.

lwander:
@ewiseblatt 

I'm making the assumption that the user has configured their local file to set `igor.enabled:  true`. If none of `dockerRegistry` or `jenkins` are enabled, then Igor can't start polling and will fail to start. By making Jenkins dependent on Igor, we won't break existing Igor installations.


lwander:
Rebased

lwander:
ping @ewiseblatt

PR #787, opened 2016-03-10
Replaced obsolete gcbundleimage to create tar image.

ewiseblatt:
Also, pass in service_account to gcloud rather than relying on the default account.
If no account is passed in, it will use the default account as before.

@skim1420


skim1420:
LGTM

PR #789, opened 2016-03-11
Fixed Google Cloud Logging configuration

ewiseblatt:
@lwander or @duftler

### DIFF
--- a/google/google_cloud_logging/spinnaker.conf
+++ b/google/google_cloud_logging/spinnaker.conf
@@ -3,8 +3,7 @@
   format multiline
   # Match the date at the beginning of each entry
   format_firstline /^(\d+-\d+-\d+\s)/
-  # The "not_sure" is always '----' in the logs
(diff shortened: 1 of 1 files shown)


google/google_cloud_logging/spinnaker.conf
```diff
@@ -76,7 +75,7 @@
   format multiline
   # Match the date at the beginning of each entry
   format_firstline /^(\d+-\d+-\d+\s)/
-  format1 /(?<time>\d+-\d+-\d+ \d+:\d+:\d+\.\d+) +(?<severity>[^ ]+) +(?<pid>\d+) +--- + \[ *(<thread>.*) *\] (?<logger>[^ :]+) *:\s*(?<message>.*)/
```
lwander: What does removing that space accomplish?

ewiseblatt: It wasnt matching the regular expression because the " +" eagarly matches
everything so the next " " doesnt match. I'd changed the expression shortly
before submitting. When I tested it, I'd looked to see that entries in the
console updated. But I wasnt paying attention that they were coming from
other instances running with my older format (where I thought the '---' was
a value field rather than just a separator).

ewiseblatt: At least I think that's what's happening. This matches whereas the other
didnt. Is that inconsistent with "\[ *(<thread>.*) *\]" ?

duftler:
LGTM.

PR #794, opened 2016-03-12
Typo while stopping the spinnaker.

murarisumit:
Output was : `Stopping apache server while starting Spinnaker.` changed
to `Stopping apache server while stopping Spinnaker.`


duftler:
@ewiseblatt

PR #812, opened 2016-03-18
Removed bakeryDefaults for AWS keys

ewiseblatt:
These were removed from spinnaker.yml because clouddriver never references them
so it doesnt make sense to have them here. AWS credentials are typically set in ~/.aws/credentials

@duftler


duftler:
LGTM.

PR #813, opened 2016-03-18
Fixed test targets

ewiseblatt:


ewiseblatt:
@duftler

duftler:
LGTM.
The 'kato' tests should really be named 'clouddriver' tests now.

ewiseblatt:
Yea. I started to do that when I renamed the "google" tests but I kept it
kato so that I could use the "kato" test prefix and have it map.
"clouddriver" is too long given the name length restrictions.

Issue #814, opened 2016-03-18
spinnaker_runner.py is missing import


cholin-bmc:
*Title*
Spinnaker repo: spinnaker_runner.py is missing import

*Cloud Provider*
N/A

*Environment*
```I am running Spinnaker development stack using the Amazon images to deploy into AWS ```

*Feature Area*
Dev Environment

*Description*
Starting the dev stack using the spinnaker/dev/run_dev.sh script should start the Spinnaker stack with no errors.
When there are '-local' configuration files for services in the config file search path such as in ~/.spinnaker then the script fails in the spinnaker_runner.py when attempting to use a yaml_util library. This is due to a missing import.

*Steps to Reproduce*
Start the dev stack via the run_dev.sh script.

*Additional Details*
I have a fix and will submit a PR.

PR #815, opened 2016-03-18
Add missing import yaml_util for spinnaker_runner.py as the startup f…

cholin-bmc:
Fix for issue #814


duftler:
@ewiseblatt PTAL.

PR #816, opened 2016-03-19
Add spinnaker user to sudoers

gardleopard:
This is to support the aws-chroot builder that requires sudo.

Related to https://github.com/spinnaker/rosco/pull/79

### DIFF
--- a/etc/sudoers.d/spinnaker
+++ b/etc/sudoers.d/spinnaker
@@ -1,0 +1,2 @@
+# User rules for spinnaker
+spinnaker ALL=(ALL) NOPASSWD: /usr/bin/packer
--- a/etc/sudoers.d/spinnaker
+++ b/etc/sudoers.d/spinnaker
@@ -1,2 +1,5 @@
+# An entry in /etc/sudoers.d is needed to grant spinnaker sudo rights for packer.
+# This is a security risk that can be exploited.
+# Uncomment the last line if you are using a template needing root in rosco.
 # User rules for spinnaker
-spinnaker ALL=(ALL) NOPASSWD: /usr/bin/packer
+# spinnaker ALL=(ALL) NOPASSWD: /usr/bin/packer
--- a/build.gradle
+++ b/build.gradle
@@ -72,15 +72,22 @@ ospackage {
   from('etc/apache2/sites-available/spinnaker.conf') {
     into '/etc/apache2/sites-available'
     user = 'root'
-    permissionGroup = 'root'    
+    permissionGroup = 'root'
     fileMode = 0644
   }
 
   from('etc/default/spinnaker') {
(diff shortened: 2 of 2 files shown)


duftler:
This looks reasonable to me. @ewiseblatt, @dpeterka, @skorten, any concerns?

ewiseblatt:
why do you need this?

duftler:
It's linked in the PR conversation thread:
https://github.com/spinnaker/rosco/pull/79

Gard recently added a new packer template to rosco that supports aws-chroot
builds. Several folks have tried it, but it needs to be run as root.

Gard is now hardening and upstreaming the changes they made in-house to
make this all work.

ewiseblatt:
Please hold off merging this. Let me inquire about the security
implications.

etc/sudoers.d/spinnaker
```diff
@@ -0,0 +1,2 @@
+# User rules for spinnaker
+spinnaker ALL=(ALL) NOPASSWD:ALL
```
duftler: If we change this to the following, it will only allow `sudo` for the `packer` command:
`spinnaker ALL=(ALL) NOPASSWD: /usr/bin/packer`

ewiseblatt: right. I dont know if he needs anything else. Even giving packer root
access is scary because packer has powerful scripting in it, but
restricting to packer keeps the other services from gaining root so this is
definitely a significant improvement.

I might still get other recommendations on Monday so lets give it another
day.

duftler: sgtm

gardleopard: One thing is that if we do this for only /usr/bin/packer we should comment or change https://github.com/spinnaker/rosco/blob/master/rosco-web/config/rosco.yml#L7 so that it is clear that there is a convention for where we install packer. If it is set like this it will work like magic on some installations, while others will have a hard time finding out why it is not working.

gardleopard:
I have limited scope and squashed the commits now

ewiseblatt:
There was an offline suggestion to rid the sudo need for mount was to add an fstab(5) entry with the "user" option in the fourth field, rather than using sudo. Another offline suggestion to use http://linux.die.net/man/1/fakechroot to remove the sudo need for chroot. Gard is looking into the fakechroot suggestion.

gardleopard:
Im deleting this pull request. It is not needed. We encourage manual update of sudoers to enable sudo access to packer.

ewiseblatt:
I think either you still want to at least add the config/rosco.yml with the entry about the template and command with warning about changing it, or we should update a document somewhere about that potential. Having the comment just in the embedded rosco yml wont be visible to users in practice.

PR #817, opened 2016-03-20
Install Google Cloud Logging from the Web.

ewiseblatt:
@duftler 
I think this command could be used (once submitted) to install google cloud logging from the web.

curl -s -q https://raw.githubusercontent.com/spinnakerspinnaker/google_cloud_logging/google/google_cloud_logging/add_google_cloud_logging.sh | bash

That will add google cloud logging to any spinnaker instance, not just those in GCE baked images
(i.e. for those that install spinnaker directly, have existing deployments, or are somewhere other than GCE).


duftler:
LGTM.

PR #819, opened 2016-03-21
Install Google Cloud Logging and Monitoring when installing on GCE.

ewiseblatt:
@lwander 

This has to be explicitly requested with arguments to InstallSpinnaker.

The build_google_image script builds both these in by default, but an explicit
install (calling InstallSpinnaker.sh directly) would not include these by default even on GCE.

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -53,6 +53,7 @@ function print_usage() {
 usage: $0 [--cloud_provider <aws|google|none|both>]
     [--aws_region <region>] [--google_region <region>]
     [--quiet] [--dependencies_only]
+    [--google_cloud_logging] [--google_cloud_monitoring]
     [--repository <debian repository url>]
     [--local-install] [--home_dir <path>]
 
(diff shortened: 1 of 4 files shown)


ewiseblatt:
This can be reviewed and merged now.

InstallSpinnaker.sh
```diff
@@ -311,6 +328,43 @@ function install_java() {
...
+        || [[ "x$GOOGLE_CLOUD_MONITORING" == "xtrue" ]]; then
+          if [[ ! -f /etc/google/auth/application_default_credentials.json ]];
+          then
+            echo "You may need to add Google Project Credentials."
+            echo "See https://developers.google.com/identity/protocols/application-default-credentials"
+          fi
+     fi
+  fi
+
+  if [[ "x$GOOGLE_CLOUD_LOGGING" == "xtrue" ]]; then
```
lwander: Why do you prefix both sides with "x"?

lwander:
LGTM

InstallSpinnaker.sh
ewiseblatt: I was following the practices elsewhere in the file. It's archaic and shouldnt be needed these days. in fact, it's probably only the -z test that [would have] needed it.

PR #824, opened 2016-03-24
Fixed url

ewiseblatt:
@lwander
I accidentally checked in the URL pointing to my repository from when I tested this.


lwander:
retroactive LGTM

PR #826, opened 2016-03-25
Multi-node storage pods for Spinnaker on Kubernetes

lwander:
This is step 1 in getting Spinnaker to run on Kubernetes. The two scripts `redis.sh` and `cassandra.sh` create multi-node redis and cassandra clusters respectively. I'm sticking to naming conventions that allow Spinnaker to manage itself once everything is running.

The redis image configuration was taken from the Kubernetes sample repository, so the License headers were left intact. 

@duftler @ttomsu


ttomsu:
LGTM

PR #844, opened 2016-04-12
Changed flow around getting cassandra to avoid failing, added flags.

jtk54:
Changed flow around getting cassandra to avoid failing and exiting the script if `apt-get install` fails. It now uses `apt-cache search` to check if any cassandra packages are available in APT. Otherwise, it pulls the debs explicitly and installs. ~~Also added flags to populate `/etc/default/spinnaker` with defaults for Google project id and service account credential path.~~

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -310,7 +310,7 @@ function add_apt_repositories() {
     gpg=$(curl -s -f "https://bintray.com/user/downloadSubjectPublicKey?username=$REPOSITORY_ORG") || true
     if [ ! -z "$gpg" ]; then
         echo "$gpg" | apt-key add -
-    fi   
+    fi
   fi
   echo "deb $REPOSITORY_URL $DISTRIB_CODENAME spinnaker" | tee /etc/apt/sources.list.d/spinnaker-dev.list > /dev/null
   # Java 8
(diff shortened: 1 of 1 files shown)


jtk54:
@ewiseblatt PTAL

InstallSpinnaker.sh
```diff
@@ -68,6 +70,8 @@ usage: $0 [--cloud_provider <aws|google|none|both>]
 
     --google_region <arg>       default region for google
     --google_zone <arg>         default zone for google
+    --google_project_id <arg>   default project for google
+    --google_creds_path  <arg>  path to google service account credentials
```
ewiseblatt: I dont think these belong here.
I'd like to handle configuration as a separate task.
Especially because there are lots of providers and it is going to get complicated.
Plus I think it will be a fairly common use case to want to change this post install anyway (e.g. add another provider down the road).

InstallSpinnaker.sh
```diff
@@ -448,26 +460,27 @@ EOF
...
-          curl -L -O $package_url/${pkg}_2.1.11_all.deb
-        done
-        dpkg -i *.deb
-        apt-mark hold cassandra cassandra-tools
-        popd
-        rm -rf /tmp/casspkgs
+  local cassandra_packages=$(apt-cache search cassandra)
+  if [ -z "$cassanrda_packages" ]; then
+    if [ "$DOWNLOAD" == "true" ]; then
+      echo "cassanrda not found in apt-cache, installing from debs..."
```
ewiseblatt: s/cassanrda/cassandra/

Maybe s/installing from debs/downloading from $package_url/?

InstallSpinnaker.sh
```diff
@@ -571,12 +584,10 @@ unzip -u -o -q packer_0.8.6_linux_amd64.zip -d /usr/bin
 popd
 rm -rf /tmp/packer
 
-if [[ "x$DEPENDENCIES_ONLY" != "x" ]]; then
-    exit 0
-fi
-
 ## Spinnaker
-install_spinnaker
+if [[ "x$DEPENDENCIES_ONLY" == "x" ]]; then
```
ewiseblatt: remove the superfluous x's here.

InstallSpinnaker.sh
```diff
@@ -612,6 +623,10 @@ if [ "x$GOOGLE_PROJECT_ID" != "x" ]; then
   write_default_value "SPINNAKER_GOOGLE_PROJECT_ID" $GOOGLE_PROJECT_ID
 fi
 
+if [ "x$GOOGLE_CREDS_PATH" != "x" ]; then
```
ewiseblatt: I'm not yet comfortable with this being here.

Also this is a little different than what we discussed.
deps_only is saying only to install the dependencies for spinnaker, but not do other stuff. I think that is legit. For example, developer builds execute that to prepare the host machine to run spinnaker, but spinnaker itself will never get installed because you are running it from source. There's no reason to add the other data installations (etc/default/spinnaker) either.

In your use case, you want to prepare both the host and the runtime data, but will be installing spinnaker separately. So I'd guard that (the data setup) with a separate flag. When you just say deps-only it should be off. Otherwise it should be on by default. But you could explicitly turn it on, so you can say deps-only +flag to meet your use case.

InstallSpinnaker.sh
jtk54: Removed flags and google initialization.

InstallSpinnaker.sh
jtk54: Done.

InstallSpinnaker.sh
jtk54: I removed this gate and made the script where it did before since we aren't setting any google initializations.

InstallSpinnaker.sh
jtk54: Removed flags and google initialization, let's just point them to /etc/default/spinnaker and have them put in the defaults. I will update my doc to reflect that.

PR #848, opened 2016-04-13
Treat spinnaker.conf as protected config file when dpkg installing.

### LINKED ISSUE #845: spinnaker.conf should be protected
When I upgraded my spinnaker packages with the upstream debs from you guys, it wiped out my spinnaker.conf which included all of my SSL configuration.

I expected behavior like spinnaker-local.yml where dpkg would ask me if I want to use my config file or the package maintainer's version.

duftler:
Closes #845.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -84,6 +84,7 @@ ospackage {
   }
 
   configurationFile('/etc/default/spinnaker\n')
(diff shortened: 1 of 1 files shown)


duftler:
@lwander please review.

lwander:
LG2M

PR #850, opened 2016-04-14
Configure notifications in Deck

anotherchrisberry:
Configures the notifications displayed in the UI based on Echo's configuration.

@ewiseblatt PTAL - I'm guessing this will work as expected, but am not sure how to verify it...

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -191,13 +191,14 @@ services:
         enabled: false
         url: # the hipchat server to connect to
         token: #the hipchat auth token
+        botName: # the username of the bot
       sms:
         enabled: false
         account: # twilio account id
         token: # twilio auth token
         from: # phone number by which sms messages are sent
-      # See https://api.slack.com/incoming-webhooks for details about using incoming webhooks
-      # and how to register your own webhook.
+      # See https://api.slack.com/bot-users for details about using bots and how to create your own bot user
       slack:
         enabled: false
-        token: # the part after http://hooks.slack.com/services/
+        token: # the API token for the bot
+        botName: # the username of the bot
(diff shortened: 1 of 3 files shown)


ewiseblatt:
I dont have deck tests, only for backend (REST) services.

One way you can test this is by running the spinnaker/runtime/reconfigure_spinnaker.sh script from your build directory, then looking at deck/settings.js (unfortunately, this will overwrite your deck/settings.js file as we've discussed before, so be sure you have a backup or a local commit you can restore).

If all goes well, the spring variable references should be resolved.
An easy way to do this is to simply grep for the reference, and filter our the commented templates
grep "\\$" settings.js | grep -v "// var"

When I run this command on what's already submitted in my client, it seems there is a bug here
var kubernetesDefaultNamespace = '${providers.kubernetes.primaryCredentials.namespace}';

@lwander either this is the wrong property, or you need to add a default to spinnaker.yml (and should also add it to default-spinnaker-local.yml for ease of user-configuration)

I'm going to modify the reconfigure_spinnaker.sh script to add this test in now that you've brought it up and I'm thinking about it.

LGTM otherwise.

ewiseblatt:
I mistyped the command, appearently you need to double-escape the $.
 grep "\\$" settings.js | grep -v "// var"

ewiseblatt:
Hmm, I didnt mistype it, github is interpreting it somehow. The grep command should be double-escaping the $ with two slashes.

lwander:
Would 
` var kubernetesDefaultNamespace = '${providers.kubernetes.primaryCredentials.namespace:default}';` work? Not sure how the string interpolation works with javascript + spring

ewiseblatt:
There is no interpolation with javascript. The mapping is done with our
script. Our script would do the substitution so this would work.

However, I do not think it would be desirable because it is a subtle
configuration that could easily go unnoticed.
It would be best to expose it in spinnaker-local.yml so that people can see
and know it is there and easily change it.
You can set the default in spinnaker.yml even if it is only to support the
deck settings (conceptually settings.js gets this configuration value from
spinnaker.yml even though it needs reconfigure_spinnaker.sh to achieve this)

lwander:
Ah I see, submitted a PR https://github.com/spinnaker/spinnaker/pull/852

anotherchrisberry:
@ewiseblatt Just to be sure before I push the button - this is okay to merge?

ewiseblatt:
I pushed it for you (but did not create a release).

PR #851, opened 2016-04-15
Check for unresolved spring variables in settings.js

ewiseblatt:
@anotherchrisberry 

This is to support testing your earlier PR.

Dont merge this yet until we resolve the issue or it is going to prevent people from being able to run Spinnaker at all. For that matter, you can argue this should be a soft error, but I think a hard error forcing a fix is appropriate.


ewiseblatt:
@anotherchrisberry or @lwander

lwander:
LGTM

PR #858, opened 2016-04-22
Configuration file portion of enabling azure in InstallSpinnaker.sh

rguthriemsft:
submitting configuration file portion for getting Azure into InstallSpinnaker.sh. We are planning to move credentials into file based credentials similar to google approach in near future but for now will set in yaml.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -15,6 +15,19 @@ providers:
       # Store actual credentials in $HOME/.aws/credentials. See spinnaker.yml for
       # more information, including alternatives.
 
+  azure:
+    # If you want to deploy some services to Microsoft Azure (azure),
+    # set enabled and provide primary credentials for deploying.
+    # Enabling azure is independent of other providers.
+    enabled: ${SPINNAKER_AZURE_ENABLED:false}
+    defaultRegion: ${SPINNAKER_AZURE_DEFAULT_REGION}
+    primaryCredentials:
+      name: my-azure-account
+      clientId:
+      appKey:
+      tenantId:
+      subscriptionId:
+    # To set azure credentials, enter values in spinnaker-local.yml 
   google:
     # If you want to deploy some services to Google Cloud Platform (google),
     # set enabled and provide primary credentials for deploying.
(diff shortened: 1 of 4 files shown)


tomaslin:
@spinnaker/reviewers PTAL

config/default-spinnaker-local.yml
```diff
@@ -15,6 +15,19 @@ providers:
...
+  azure:
+    # If you want to deploy some services to Microsoft Azure (azure),
+    # set enabled and provide primary credentials for deploying.
+    # Enabling azure is independent of other providers.
+    enabled: false
+    defaultRegion: westus
+    primaryCredentials:
+      name: my-azure-account
+      clientId: my-client-id
+      appKey: my-app-key
```
ewiseblatt: Same remark as for spinnaker.yml below, but if they are empty by default then you should add a comment indicating they need to be populated properly when enabling azure.

config/default-spinnaker-local.yml
```diff
@@ -15,6 +15,19 @@ providers:
       # Store actual credentials in $HOME/.aws/credentials. See spinnaker.yml for
       # more information, including alternatives.
 
+  azure:
+    # If you want to deploy some services to Microsoft Azure (azure),
+    # set enabled and provide primary credentials for deploying.
+    # Enabling azure is independent of other providers.
+    enabled: false
```
ewiseblatt: Since you added the environment variables to default spinnaker, these should be ${SPINNAKER_AZURE_ENABLED:false} and similar reference for the region. Otherwise you arent utilizing those environment variables.

config/spinnaker.yml
```diff
@@ -212,9 +212,13 @@ providers:
     # set enabled and provide primary credentials for deploying.
     # Enabling azure is independent of other providers.
     enabled: false
-    defaultRegion: West US
+    defaultRegion: westus
     primaryCredentials:
       name: my-azure-account
+      clientId: my-client-id
```
ewiseblatt: I'm not familiar with Azure yet -- are these valid values or placeholders? If they are not valid, then I think you want them to be empty (even if empty is not valid either) and treat empty values as being undefined. Otherwise, if these are valid as is, then that's fine.

PR #859, opened 2016-04-23
Configuration file portion of enabling azure in InstallSpinnaker.sh

rguthriemsft:
Updated config files based on feedback.  We will start work on file based credentials once we have installer integrated.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -15,6 +15,19 @@ providers:
       # Store actual credentials in $HOME/.aws/credentials. See spinnaker.yml for
       # more information, including alternatives.
 
+  azure:
+    # If you want to deploy some services to Microsoft Azure (azure),
+    # set enabled and provide primary credentials for deploying.
+    # Enabling azure is independent of other providers.
+    enabled: ${SPINNAKER_AZURE_ENABLED:false}
+    defaultRegion: ${SPINNAKER_AZURE_DEFAULT_REGION}
+    primaryCredentials:
+      name: my-azure-account
+      clientId:
+      appKey:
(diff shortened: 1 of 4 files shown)


config/spinnaker.yml
```diff
@@ -212,9 +212,14 @@ providers:
...
     enabled: false
-    defaultRegion: West US
+    defaultRegion: westus
     primaryCredentials:
       name: my-azure-account
+      clientId:
+      appKey:
+      tenantId:
+      subscriptionId:
+    # To set azure credentials, enter values for: clientId, appKey, tenantId, subscriptionId.  See spinnaker documentation - cloud provider setup for instructions on how to retrieve these values.
```
ewiseblatt: I was trying to keep 80 columns in these files if practical.
If you could link to the docs here that would be better (the URL could go beyond 80). I'd also add the link to the default-spinnaker-local because that is the point of the typical user interaction. The expectation is to start in spinnaker-local and not need to look here for common needs.

config/default-spinnaker-local.yml
```diff
@@ -15,6 +15,19 @@ providers:
       # Store actual credentials in $HOME/.aws/credentials. See spinnaker.yml for
       # more information, including alternatives.
 
+  azure:
+    # If you want to deploy some services to Microsoft Azure (azure),
+    # set enabled and provide primary credentials for deploying.
+    # Enabling azure is independent of other providers.
+    enabled: ${SPINNAKER_AZURE_ENABLED:false}
+    defaultRegion: ${SPINNAKER_AZURE_DEFAULT_REGION}
```
ewiseblatt: + :westus

PR #860, opened 2016-04-23
Unify timezone configuration across deck/orca/echo.

duftler:


duftler:
@ewiseblatt please review.

ewiseblatt:
I like how this is expressed.

PR #864, opened 2016-04-25
Adding support for Azure as well as option to install all clouds to s…

rguthriemsft:
Added support for Azure to installation script.  I have also added an option to install all three clouds called all.

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -50,8 +50,8 @@ fi
 
 function print_usage() {
   cat <<EOF
-usage: $0 [--cloud_provider <aws|google|none|both>]
-    [--aws_region <region>] [--google_region <region>]
+usage: $0 [--cloud_provider <aws|google|azure|both|none>]
+    [--aws_region <region>] [--google_region <region>] [--azure_region <region>]
     [--quiet] [--dependencies_only]
     [--google_cloud_logging] [--google_cloud_monitoring]
     [--repository <debian repository url>]
@@ -60,14 +60,20 @@ usage: $0 [--cloud_provider <aws|google|none|both>]
 
     If run with no arguments you will be prompted for cloud provider and region
 
-    --cloud_provider <arg>      currently supported are google, amazon and none
-                                if "none" is specified you will need to edit
-                                /etc/default/spinnaker manually
+    --cloud_provider <arg>     currently supported are amazon, azure, and google
+				you can install more than one provider by specifying
+				provider1 provider2 ... (Ex. aws google).
+                               if "none" is specified you will need to edit
+                               /etc/default/spinnaker manually
+				NOTE: "both" is deprecated and will be removed in
+				the future
 
     --aws_region <arg>          default region for aws
 
     --google_region <arg>       default region for google
     --google_zone <arg>         default zone for google
+
+    --azure_region <arg>	default region for Azure
 
     --quiet                     sets cloud provider to "none", you will need to
                                 edit /etc/default/spinnaker manually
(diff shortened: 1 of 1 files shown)


rguthriemsft:
@spinnaker/google-reviewers and @spinnaker/netflix-reviewers PTAL

InstallSpinnaker.sh
```diff
@@ -50,8 +50,8 @@ fi
 
 function print_usage() {
   cat <<EOF
-usage: $0 [--cloud_provider <aws|google|none|both>]
-    [--aws_region <region>] [--google_region <region>]
+usage: $0 [--cloud_provider <aws|google|azure|none|both|all>]
```
ewiseblatt: we should deprecate "both".
I'm not sure "all" is going to make sense in practice, but I guess would not hurt.

InstallSpinnaker.sh
```diff
@@ -60,7 +60,8 @@ usage: $0 [--cloud_provider <aws|google|none|both>]
 
     If run with no arguments you will be prompted for cloud provider and region
 
-    --cloud_provider <arg>      currently supported are google, amazon and none
+    --cloud_provider <arg>      currently supported are google, azure, amazon and none
```
ewiseblatt: is spacing off here?
dont document both. mark it deprecated below where it is handled.

InstallSpinnaker.sh
```diff
@@ -523,7 +541,7 @@ set_defaults_from_environ
 process_args "$@"
 
 if [ "x$CLOUD_PROVIDER" == "x" ]; then
-  read -e -p "Specify a cloud provider (aws|google|none|both): " -i "$DEFAULT_CLOUD_PROVIDER" CLOUD_PROVIDER
+  read -e -p "Specify a cloud provider (aws|azure|google|none|both|all): " -i "$DEFAULT_CLOUD_PROVIDER" CLOUD_PROVIDER
```
ewiseblatt: remove both

InstallSpinnaker.sh
```diff
@@ -537,15 +555,26 @@ case $CLOUD_PROVIDER in
...
       ;;
+  azure)
+      CLOUD_PROVIDER="azure"
+      set_azure_region
+      ;;
   n|no|none)
       CLOUD_PROVIDER="none"
       ;;
-  both|all)
+  both)
```
ewiseblatt: comment both as deprecated

InstallSpinnaker.sh
```diff
@@ -589,6 +618,11 @@ if [[ "${CLOUD_PROVIDER,,}" == "amazon" || "${CLOUD_PROVIDER,,}" == "google" ||
         write_default_value "AWS_VPC_ID" $AWS_VPC_ID
         write_default_value "AWS_SUBNET_ID" $AWS_SUBNET_ID
         ;;
+    azure)
+	write_default_value "SPINNAKER_AZURE_ENABLED" "true"
```
ewiseblatt: is spacing off?

InstallSpinnaker.sh
```diff
@@ -604,6 +638,17 @@ if [[ "${CLOUD_PROVIDER,,}" == "amazon" || "${CLOUD_PROVIDER,,}" == "google" ||
         write_default_value "AWS_VPC_ID" $AWS_VPC_ID
         write_default_value "AWS_SUBNET_ID" $AWS_SUBNET_ID
         ;;
+    all)
```
ewiseblatt: I'm guessing at this point it would be simpler to have independent checks for each and have "both" set google and aws, "all" set each, and the explicit flags set its corresponding one. Then at the end iterate over each check and set as needed.

InstallSpinnaker.sh
```diff
@@ -537,15 +555,26 @@ case $CLOUD_PROVIDER in
       set_google_region
       set_google_zone
       ;;
+  azure)
```
ewiseblatt: Could you fix these so they are orthogonal? Could you refactor it so the provider defaults to none and you can add each independently? That would make it easy to have arbitrary combinations. Then "both" can go away and be replaced by naming aws and google.

PR #867, opened 2016-04-27
Updating Install script to include azure support. Also enabling abili…

rguthriemsft:
Added support for Azure to installation script. I have also added an option to install all three clouds called in any combnation desired.  Deprecating both.

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -50,8 +50,8 @@ fi
 
 function print_usage() {
   cat <<EOF
-usage: $0 [--cloud_provider <aws|google|none|both>]
-    [--aws_region <region>] [--google_region <region>]
+usage: $0 [--cloud_provider <aws|google|azure|both|none>]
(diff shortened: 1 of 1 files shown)


InstallSpinnaker.sh
```diff
@@ -109,6 +115,43 @@ function echo_status() {
   fi
 }
 
+function process_provider_list() {
+  AWS_ENABLED=false
+  AZURE_ENABLED=false
+  GOOGLE_ENABLED=false
+
+  #check if aws enabled
```
ewiseblatt: Could you make a loop here that processes the next token in the list?
    a=$(CLOUD_PROVIDER) should turn the space delimited list into an array
then iterate over the array and process each element individually.
    for provider in ${a[@]}; do ...; done
That way we'll be able to detect malformed strings.

ewiseblatt:
Just the one change and then I think it is ok. I'll look to clean it up after.

You dont need to close and re-push. You can just squash this one the git -f push it.

PR #868, opened 2016-04-28
Added first boot script specific to GCE codelab image.

jtk54:
Added first boot script specific to GCE codelab image. It performs the config difference between what we want in the general case for GCE and the configuration specific for the codelab. @duftler @ewiseblatt PTAL.

### DIFF
--- a/install/first_gce_codelab_boot.sh
+++ b/install/first_gce_codelab_boot.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


install/first_gce_codelab_boot.sh
```diff
@@ -0,0 +1,31 @@
...
+
+if [[ `/usr/bin/id -u` -ne 0 ]];then
+  echo "$0 must be executed with root permissions; exiting"
+  exit 1
+fi
+
+/opt/spinnaker/install/first_google_boot.sh
+
+echo "Setting hostname for rosco deb repo"
+sed -i "s/debianRepository:.*/debianRepository: http:\/\/$(hostname):9999\/ trusty main/" /opt/rosco/config/rosco.yml
```
ewiseblatt: Does this need to be tied to "trusty"? Shouldnt it be whatever the host is running?

install/first_gce_codelab_boot.sh
```diff
@@ -0,0 +1,31 @@
...
+
+/opt/spinnaker/bin/reconfigure_spinnaker.sh
```
ewiseblatt: Do you need to restart spinnaker, or just rosco?

jtk54: Most likely just rosco, good point.

jtk54:
This startup script doesn't belong with the distribution -- I'll move it to another location in this repo so that it doesn't ship with the distribution, and modify the GCE codelab accordingly.

PR #871, opened 2016-04-29
Cleaned up install script to make more internally consistent.

ewiseblatt:
Use consistent bash syntax.
Cleaned up some comments.
Declare and document some global variables (using boolean values).
Fixed azure config with --quiet.
Deprecated "both" --cloud_provider option.

@dpeterka @rguthriemsft

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -4,6 +4,24 @@ set -o pipefail
 
 REPOSITORY_URL="https://dl.bintray.com/spinnaker/debians"
 
+
+# This script uses the following global variables.
+AWS_ENABLED=false     # set by --cloud_providers
+AZURE_ENABLED=false   # set by --cloud_providers
+GOOGLE_ENABLED=false  # set by --cloud_providers
+
+# Set by --local-install, means download packages
+# rather than adding additional external repositories.
+DOWNLOAD=false        
+
+# Set by --quiet, wont prompt or echo status output.                      
+QUIET=false
+
+# Set by --dependencies only,
+# Install dependencies but not spinnaker itself.
+DEPENDENCIES_ONLY=false
+
+
 ## This script install pre-requisites for Spinnaker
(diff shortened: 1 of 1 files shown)


InstallSpinnaker.sh
```diff
@@ -384,7 +404,7 @@ function add_apt_repositories() {
 }
 
 function install_java() {
-  if [ "$DOWNLOAD" != "true" ];then
+  if [[ $DOWNLOAD != true ]]; then
```
dpeterka: boolean instead of string compare just need ```if [ $DOWNLOAD]```

dpeterka:
Double brackets make me nervous if script is going to run on various systems but that is more of a preference....
http://stackoverflow.com/questions/669452/is-preferable-over-in-bash-scripts

ewiseblatt:
It was my impression that double brackets are explicitly part of bash, and
the script is specifically specified as a bash script so this is
safe/reliable. I hadnt considered that we might need a shell other than
bash, but I think it's pretty low risk to assume that either bash is
available, or there would be bigger problems with the script (e.g. apt
package manager isnt available either) for that particular system.

I propose we keep them, but if not, we should remove them everywhere and
I'll change the PR to do that.

dpeterka:
I think so, if we are extending our ability to test which OS the installer runs on at some point we should probably re-visit however.

LGTM otherwise. Merge!

InstallSpinnaker.sh
ewiseblatt: Actually, turns out that it is "if $DOWNLOAD" without the brackets because 'false' is treated as a non-0 length string rather than a keyword or command returning a non-0 result.

ewiseblatt:
I dont follow what "I think so" refers to. Are you ok with double brackets?
I changed the boolean "if" expressions to your earlier suggestion (minus
the brackets).

dpeterka:
I am okay with double brackets.

PR #872, opened 2016-04-29
Cleaned up config file comments and ensure default values for env vars.

ewiseblatt:
@rguthriemsft FYI I linked to a non-existent section for Azure. When you update http://www.spinnaker.io/docs/target-deployment-setup, use the section tag I'm assuming based on the pattern of the others.

@lwander Please review. Particularly the mention of kubernetes dependencies.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -1,4 +1,7 @@
-# This file is intended to override the default configuration in the spinnaker.yml file.
+# This file is intended to override the default configuration in the
+# spinnaker.yml file while providing guidance on the mostly likely
+# configuration parameters to be changed.
+#
 # In order for Spinnaker to discover it, it must be copied to a file named
 # "spinnaker-local.yml" and placed in the $HOME/.spinnaker directory.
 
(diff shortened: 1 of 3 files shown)


config/default-spinnaker-local.yml
```diff
@@ -7,82 +10,88 @@ global:
...
+      # account:
+      #   name: my-cf-username
+      #   password: my-cf-password
 
   kubernetes:
-    # If you want to deploy containers to Kubernetes,
-    # set enabled and provide primary credentials for deploying.
-    # Enabling kubernetes is dependent on dockerRegistry being enabled.
-    enabled: ${SPINNAKER_KUBERNETES_ENABLED:false}
+    # For more information on configuring Kubernetes clusters (kubernetes), see
```
lwander: Just wondering why you put added `(kubernetes)` right here.

config/default-spinnaker-local.yml
```diff
@@ -93,12 +102,13 @@ providers:
   dockerRegistry:
     # If you want to deploy containers to a container management solution,
     # you must specifiy where these container images exist first.
-    # Enabling dockerRegistry is independent of other providers.
+
+    # NOTE: Enabling dockerRegistry is independent of other providers.
+    # However, for convienience, we tie docker and kubernetes together
+    # since kubernetes (and only kubernetes) depends on this docker provider
```
lwander: Ah good idea to point this out.

config/default-spinnaker-local.yml
lwander: Ah nvm, I see how it's consistent with the other providers.

lwander:
LGTM, these comments are really helpful.

Issue #874, opened 2016-04-29
support scripts (e.g. reconfigure_spinnaker) are not aware "yaml" extension


ewiseblatt:
It turns out that spinnaker's ".yml" files can also be named ".yaml". The support scripts do not consider that.

PR #875, opened 2016-04-29
Look for either .yml or .yaml in scripting support

ewiseblatt:
Fixes https://github.com/spinnaker/spinnaker/issues/874

@lwander

### DIFF
--- a/pylib/spinnaker/yaml_util.py
+++ b/pylib/spinnaker/yaml_util.py
@@ -16,6 +16,27 @@ import os
 import re
 import yaml
 
+def yml_or_yaml_path(basedir, basename):
+  """Return a path to the requested YAML file.
+
+  Args:
+    basedir [string]: The directory path containing the file.
+    basename [string]: The base filename for the file
+
+  Returns:
+    Path to the YAML file with either a .yml or .yaml extension,
+    depending on which if any exists.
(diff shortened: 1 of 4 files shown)


pylib/spinnaker/yaml_util.py
```diff
@@ -16,6 +16,27 @@
...
+  """Return a path to the requested YAML file.
+
+  Args:
+    basedir [string]: The directory path containing the file.
+    basename [string]: The base filename for the file
+
+  Returns:
+    Path to the YAML file with either a .yml or .yaml extension,
+    depending on which if any exists.
+    If neither exists, return the .yml. If both, then raise an exception.
```
lwander: Why not check if `.yml` exists?

ewiseblatt: what do you mean? I thought that's what I do.

Oh, it doesnt matter if it exists or not since the default path with be the .yml whether or not it exists. It only matters if it exists if the .yaml exists as a check for an ambiguous file which technically might be permitted by the spring boot library but I'm flagging it as an error since it doesnt seem necessary to permit and more likely suggests a typo or future maintainence problem.

lwander:
Sorry - thought this was merged already. LGTM

PR #876, opened 2016-05-03
Adds script for configuring nested igor properties for codelab.

jtk54:
Adds script for configuring nested igor properties for codelab. @ewiseblatt PTAL. The full codelab provision script needs to run this script remotely (i.e. curl it from github). Hence I am committing this first and the full script will follow after testing.

### DIFF
--- a/pylib/spinnaker/codelab_config.py
+++ b/pylib/spinnaker/codelab_config.py
@@ -1,0 +1,39 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from spinnaker.yaml_util import YamlBindings
+
+def configure_codelab_igor_jenkins():
+  """Configures Igor to be enabled and to point to the codelab jenkins instance.
+
+  """
+  YamlBindings.update_yml_source(
+    '/opt/spinnaker/config/spinnaker-local.yml',
+    {
+      'jenkins': {
+        'defaultMaster': {
+          'name': 'CodelabJenkins',
+          'baseUrl': 'http://localhost:9090',
+          'name': 'admin',
+          'password': 'admin'
+        }
+      },
+      'igor': {
+        'enabled': 'true'
+      }
+    }
+  )
+
+if __name__ == '__main__':
+  configure_codelab_igor_jenkins()
--- a/pylib/spinnaker/yaml_util.py
+++ b/pylib/spinnaker/yaml_util.py
@@ -29,7 +29,7 @@ class YamlBindings(object):
 
   def __getitem__(self, field):
     return self.__get_field_value(field, [], original=field)
-    
+
   def get(self, field, default=None):
     try:
       return self.__get_field_value(field, [], original=field)
@@ -74,7 +74,7 @@ class YamlBindings(object):
 
   def __typed_value(self, value_text):
     """Convert the text of a value into the YAML value.
-    
+
     This is used for type conversion for default values.
     Not particularly efficient, but there doesnt seem to be a direct API.
     """
@@ -130,6 +130,42 @@ class YamlBindings(object):
 
   def replace(self, text):
     return self.__resolve_value(text, [], text)
+
+
+  def __get_flat_keys(self, container):
+    flat_keys = []
+    for key,value in container.items():
+      if isinstance(value, dict):
+        flat_keys.extend([key + '.' + subkey for subkey in self.__get_flat_keys(value)])
+      else:
+        flat_keys.append(key)
+    return flat_keys
+
+
+  @staticmethod
+  def update_yml_source(path, update_dict):
+    """Update the yaml source at the path according to the update dict.
+
+    All the previous bindings not in the update dict remain unchanged.
+    The yaml file at the path is re-written with the new bindings.
+
+    Args:
+      path [string]: Path to a yaml source file.
+      update_dict [dict]: Nested dictionary corresponding to
+          nested yaml properties, keyed by strings.
+    """
+    bindings = YamlBindings()
+    bindings.import_dict(update_dict)
+    updated_keys = bindings.__get_flat_keys(bindings.__map)
+    source = '' # declare so this is in scope for both 'with' blocks
+    with open(path, 'r') as source_file:
+      source = source_file.read()
+      for prop in updated_keys:
+        source = bindings.transform_yaml_source(source, prop)
+
+    with open(path, 'w') as source_file:
+      source_file.write(source)
+
 
   def transform_yaml_source(self, source, key):
     """Transform the given yaml source so its value of key matches the binding.
--- a/unittest/yaml_util_test.py
+++ b/unittest/yaml_util_test.py
@@ -249,7 +249,7 @@ unique:
 
      self.assertEqual(expect, bindings.transform_yaml_source(expect, 'bogus'))
      self.assertEqual(expect, got)
-                      
+
 
   def test_transform_fail(self):
      bindings = YamlBindings()
@@ -278,6 +278,49 @@ a:
      self.assertEqual([{'elem': True}, {'elem': True}, {'elem': False}, {'elem': False}],
                       bindings.get('root'))
      self.assertEqual(bindings.get('root'), bindings.get('copy'))
+
+  def test_update_yml_source(self):
+    yaml = """
+a: A
+b: 0
+c:
+  - A
+  - B
+d:
+  child:
+    grandchild: x
+e:
+"""
+    fd, temp_path = tempfile.mkstemp()
+    os.write(fd, yaml)
+    os.close(fd)
+
+    update_dict = {
+      'b': 'Z',
+      'd': {
+        'child': {
+          'grandchild': 'xy'
+        }
+      },
+      'e': 'AA'
+    }
+
+    expect = {'a': 'A',
+              'b': 'Z',
+              'c': ['A','B'],
+              'd': {
+                'child': {
+                  'grandchild': 'xy'
+                }
+              },
+              'e': 'AA'}
+
+    YamlBindings.update_yml_source(temp_path, update_dict)
+
+    comparison_bindings = YamlBindings()
+    comparison_bindings.import_path(temp_path)
+    self.assertEqual(expect, comparison_bindings.map)
+    os.remove(temp_path)
 
 if __name__ == '__main__':
   loader = unittest.TestLoader()


pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import spinnaker.yaml_util as yaml_util
+
+JENKINS_DEFAULT_MASTER_NAME = 'jenkins.defaultMaster.name'
+JENKINS_DEFAULT_MASTER_BASE = 'jenkins.defaultMaster.baseUrl'
+JENKINS_DEFAULT_MASTER_USERNAME = 'jenkins.defaultMaster.username'
+JENKINS_DEFAULT_MASTER_PASSWORD = 'jenkins.defaultMaster.password'
+IGOR_ENABLED = 'igor.enabled'
```
ewiseblatt: You dont use IGOR_ENABLED.

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,59 @@
...
+PROP_KEYS = [
+    JENKINS_DEFAULT_MASTER_NAME,
+    JENKINS_DEFAULT_MASTER_BASE,
+    JENKINS_DEFAULT_MASTER_USERNAME,
+    JENKINS_DEFAULT_MASTER_PASSWORD
+]
```
ewiseblatt: Since you dont reference these keys anywhere else, I'd just enumerate them here rather than having symbolic names. Simpler, cleaner, clearer.

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,59 @@
...
+    """Configures Igor to be enabled and to point to the codelab jenkins instance.
+
+    """
+    yaml_binds = yaml_util.YamlBindings()
+    update_dict = {
+        'jenkins': {
+            'defaultMaster': {
+                'name': 'Jenkins',
+                'baseUrl': 'http://localhost:9090',
+                'username': 'admin',
```
ewiseblatt: Maybe password should be a required command-line parameter?
You can make the name and port command line parameters as well, but with the default values.

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,59 @@
...
+JENKINS_DEFAULT_MASTER_NAME = 'jenkins.defaultMaster.name'
+JENKINS_DEFAULT_MASTER_BASE = 'jenkins.defaultMaster.baseUrl'
+JENKINS_DEFAULT_MASTER_USERNAME = 'jenkins.defaultMaster.username'
+JENKINS_DEFAULT_MASTER_PASSWORD = 'jenkins.defaultMaster.password'
+IGOR_ENABLED = 'igor.enabled'
+PROP_KEYS = [
+    JENKINS_DEFAULT_MASTER_NAME,
+    JENKINS_DEFAULT_MASTER_BASE,
+    JENKINS_DEFAULT_MASTER_USERNAME,
+    JENKINS_DEFAULT_MASTER_PASSWORD
```
ewiseblatt: 'igor.enabled'

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,59 @@
...
+
+def configure_igor_jenkins():
+    """Configures Igor to be enabled and to point to the codelab jenkins instance.
+
+    """
+    yaml_binds = yaml_util.YamlBindings()
+    update_dict = {
+        'jenkins': {
+            'defaultMaster': {
+                'name': 'Jenkins',
```
ewiseblatt: Since you are providing a jenkins server that has a particular configuration and installs for the tutorial, maybe you want to give this a more specific name, like CodelabJenkins?
Jenkins is fine too.

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,59 @@
...
+IGOR_ENABLED = 'igor.enabled'
+PROP_KEYS = [
+    JENKINS_DEFAULT_MASTER_NAME,
+    JENKINS_DEFAULT_MASTER_BASE,
+    JENKINS_DEFAULT_MASTER_USERNAME,
+    JENKINS_DEFAULT_MASTER_PASSWORD
+]
+
+def configure_igor_jenkins():
+    """Configures Igor to be enabled and to point to the codelab jenkins instance.
```
ewiseblatt: I think you can generalize this to update_yml(path, update_dict) and put it in yaml_util.

The PROPS are the keys from the yml_binds.
In main you call this function (which lives in yaml_util) with the local_yml and update_dict values you are using below.

if so, add a test.

jtk54: Added this generalized code to yaml_util and test.

pylib/spinnaker/codelab_config.py
jtk54: Updated name.

pylib/spinnaker/codelab_config.py
jtk54: This script is specific to the codelab image, which will have this Jenkins admin user configured. Note that this shouldn't be used anywhere near production -- this is meant to try out Spinnaker.

pylib/spinnaker/codelab_config.py
jtk54: Refactored this into yaml_util.

pylib/spinnaker/codelab_config.py
jtk54: Refactored this into yaml_util.

pylib/spinnaker/codelab_config.py
jtk54: Refactored this into yaml_util.

pylib/spinnaker/yaml_util.py
```diff
@@ -131,6 +131,44 @@ def __resolve_value(self, value, saw, original):
...
+
+    flat_keys = []
+    for key,value in container.items():
+      subkeys = self.__get_flat_keys(value)
+      if subkeys == []:
+        flat_keys.append(key)
+      else:
+        flat_keys.extend([key + '.' + subkey for subkey in subkeys])
+    return flat_keys
+
```
ewiseblatt: How about:

  def __get_flat_keys(self, container):
    flat_keys = []
    for key,value in container.items():
      if isinstance(value, dict):
        flat_keys.extend([key '.' subkey for subkey in self.__get_flat_keys(value)])
      else:
        flat_keys.append(key)
    return flat_keys

pylib/spinnaker/yaml_util.py
```diff
@@ -131,6 +131,44 @@ def __resolve_value(self, value, saw, original):
...
+
+  def update_yml_source(self, path, update_dict):
```
ewiseblatt: I think this is a static method that creates a binding dictionary within.

pylib/spinnaker/yaml_util.py
jtk54: Changed the impl.

pylib/spinnaker/yaml_util.py
jtk54: Done.

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import spinnaker.yaml_util as yaml_util
+
+def configure_codelab_igor_jenkins():
+    """Configures Igor to be enabled and to point to the codelab jenkins instance.
+
+    """
+    yaml_binds = yaml_util.YamlBindings()
```
ewiseblatt: dont need this.

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,39 @@
...
+                'username': 'admin',
+                'password': 'admin'
+            }
+        },
+        'igor': {
+            'enabled': 'true'
+        }
+    }
+    local_yml = '/opt/spinnaker/config/spinnaker-local.yml'
+    yaml_binds.update_yml_source(local_yml, update_dict)
```
ewiseblatt: YamlBindings.update_yml_source(...)

ewiseblatt: And you can even inline the update dict and local yml, pass by name since that's all this function is.

pylib/spinnaker/yaml_util.py
```diff
@@ -131,6 +131,42 @@ def __resolve_value(self, value, saw, original):
...
+  @staticmethod
+  def update_yml_source(path, update_dict):
+    """Update the yaml source at the path according to the update dict.
+
+    All the previous bindings not in the update dict remain unchanged.
+    The yaml file at the path is re-written with the new bindings.
+
+    Args:
+      path [string]: Path to a yaml source file.
+      update_dict [dict(string -> dict)]: Nested dictionary corresponding to
```
ewiseblatt: The type is a dict of object values keyed by strings. It isnt necessarily a dict of dicts (and since it is recursive, at some point it should not be!)

unittest/yaml_util_test.py
```diff
@@ -279,6 +279,49 @@ def test_bool(self):
...
+c:
+  - A
+  - B
+d:
+  child:
+    grandchild: x
+e:
+"""
+    fd, temp_path = tempfile.mkstemp()
+    os.write(fd, yaml)
```
ewiseblatt: Does this leak temp_path or is the file tied to the process lifetime? The docs arent clear. I would think they persist beyond the test. If so at least os.remove them at the end of this method on success.

pylib/spinnaker/codelab_config.py
jtk54: Removed.

pylib/spinnaker/yaml_util.py
jtk54: Changed just to `dict`.

unittest/yaml_util_test.py
jtk54: Removed after success.

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,39 @@
...
+# limitations under the License.
+
+from spinnaker.yaml_util import YamlBindings
+
+def configure_codelab_igor_jenkins():
+    """Configures Igor to be enabled and to point to the codelab jenkins instance.
+
+    """
+    YamlBindings.update_yml_source(
+      '/opt/spinnaker/config/spinnaker-local.yml',
```
ewiseblatt: This is ok, but I'm wondering if you'd want to be able to run the codelab using a developer build?
If so then this should be a parameter so you can override it with something like $HOME/.spinnaker/spinnaker-local.yml

pylib/spinnaker/codelab_config.py
```diff
@@ -0,0 +1,39 @@
...
+      {
+        'jenkins': {
+          'defaultMaster': {
+            'name': 'CodelabJenkins',
+            'baseUrl': 'http://localhost:9090',
+            'name': 'admin',
+            'name': 'admin'
```
ewiseblatt: name is twice. One of these is probably password. I think password should be injected in as a commandline argument so you can generate it at runtime for unique and private access only to the user. You dont need to worry about generation here, but the caller would generate the password, attach it to the instance metadata, then pass it to this script. The user would learn the password looking at the instance metadata. That's how the marketplace instances seem to work.

ewiseblatt:
LGTM

pylib/spinnaker/codelab_config.py
jtk54: Acked.

pylib/spinnaker/codelab_config.py
jtk54: Fixed.

PR #880, opened 2016-05-06
Add publicBakeryBaseUrl

tgracchus:
View bakery log link in Deck always point to bakeryBaseUrl. So, if for deployment reason the bakery has two url. One public and one used internally, behing a LB, the links will point to the internal one. So, this links will not work when trying to access them in Deck. This change allows you to specify a public bakery url to be used by deck. If it is no provided, it will fallback to baseBakeryUrl

### DIFF
--- a/config/settings.js
+++ b/config/settings.js
@@ -10,7 +10,8 @@
 // BEGIN reconfigure_spinnaker
 
 // var gateUrl = ${services.deck.gateUrl};
-// var bakeryBaseUrl = ${services.deck.bakeryUrl};
+// var bakeryBaseUrl = ${services.deck.bakeryUrl}; 
+// var bakeryPublicBaseUrl = ${services.bakery.publicBaseUrl:fallbackToBakeryBaseUrl};
 // var authEnabled = ${services.deck.auth.enabled};
 // var defaultTimeZone = ${services.deck.timezone};
 // var awsDefaultRegion = ${providers.aws.defaultRegion};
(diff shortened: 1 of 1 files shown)


duftler:
I think the way to accomplish what you describe here is to override `services.deck.bakeryUrl` in your `spinnaker-local.yml` file: https://github.com/spinnaker/spinnaker/blob/master/config/default-spinnaker-local.yml#L201

While `services.deck.bakeryUrl` points to the bakery's base url by default (https://github.com/spinnaker/spinnaker/blob/master/config/spinnaker.yml#L77), it can be overridden as linked to above.

Would like @ewiseblatt 's opinion here as well.

ewiseblatt:
What is services.bakery.publicBaseUrl? I dont see them in either
spinnaker.yml or default-spinnaker-local.yml.
If this is a user-specific problem, then you probably want to do what Matt
suggests.
If you added services.bakery.publicBaseUrl then you definitely would do as
he suggests.
Your definition looks fishy so suspect this is the case.

I think the expectations for your use case is to set up a reverse proxy and
edit the existing deck attributes in spinnaker-local.yml.

If you were to add a new property such as this in the future, the way to do
it would be to add a default value in spinnaker.yml
(e.g. services.bakery.publicBaseUrl: services.bakery.baseUrl)
And then change deck to use that instead of baseUrl.

However in this particular case, which admittedly I dont fully understand,
I think you want to be using a reverse proxy. That's the expectation in
Spinnaker's design. As noted elsewhere, you generally do not want to have
spinnaker services (or bakery) be "publically" accessable.

If it turns out that a reverse proxy is in fact not appropriate then we can
discuss this further.

tgracchus:
@duftler is completely right. 
We can simply override the deck.bakeryUrl property with spinnaker-local.yml

@ewiseblatt In our case, we have the bakery as available as a service, in case someone wants to use it exclusively. So, we have a public url to be used outside spinnaker and another one to be used by spinnaker. Yes, it's a user-specific problem

Going to close the PR
Thank you for your time

PR #889, opened 2016-05-16
cassandra.enabled: true (by default)

ajordens:
The default value for `cassandra.enabled` (in Front50) was changed from `true` to `false`.

I believe that `false` is the correct code-level default but that yml configuration should be explicitly enabled.

PTAL @jtk54 

Relates to spinnaker/front50#92


jtk54:
SGTM.

PR #902, opened 2016-05-23
Added codelab startup script to gcs bucket, copies to machine.

jtk54:
Added codelab startup script to gcs bucket, copies to machine during startup. @ewiseblatt FYI.

### DIFF
--- a/google/codelab/provision_spinnaker_codelab_image.sh
+++ b/google/codelab/provision_spinnaker_codelab_image.sh
@@ -85,6 +85,9 @@ server {
 EOF
 service nginx restart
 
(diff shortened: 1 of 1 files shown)


google/codelab/provision_spinnaker_codelab_image.sh
```diff
@@ -85,6 +85,9 @@ server {
 EOF
 service nginx restart
 
+wget https://storage.googleapis.com/codelab-startup-script/first_codelab_boot.sh -O /opt/spinnaker/install/first_codelab_boot.sh
```
ewiseblatt: Maybe curl below should be wget.

PR #916, opened 2016-06-01
docker & kubernetes: Update install scripts & config locations

lwander:
@duftler @ewiseblatt PTAL

If the user provides the contents of a service account json file for the GCP project containing their registry, and the name of a repository in that registry, the values

`SPINNAKER_DOCKER_REPOSITORY`
`SPINNAKER_DOCKER_REGISTRY`
`SPINNAKER_DOCKER_USERNAME`
`SPINNAKER_DOCKER_PASSWORD_FILE`

will be populated with the correct values to be loaded by `spinnaker-local.yml` for configuring GCR with Kubernetes.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -96,12 +96,11 @@ providers:
       # These credentials use authentication information at ~/.kube/config
       # by default.
       name: my-kubernetes-account
-      namespace: default
       dockerRegistryAccount: ${providers.dockerRegistry.primaryCredentials.name}
 
   dockerRegistry:
-    # If you want to deploy containers to a container management solution,
-    # you must specifiy where these container images exist first.
+    # For more information on configuring Docker registries, see
+    # http://www.spinnaker.io/v1.0/docs/target-deployment-configuration#section-docker-registry
 
     # NOTE: Enabling dockerRegistry is independent of other providers.
     # However, for convienience, we tie docker and kubernetes together
@@ -110,9 +109,12 @@ providers:
     enabled: ${SPINNAKER_KUBERNETES_ENABLED:false}
 
     primaryCredentials:
-      name: my-docker-registry-account
-      address: https://index.docker.io/
-      repository: library/nginx
+      name: my-docker-registry
+      address: ${SPINNAKER_DOCKER_REGISTRY:https://index.docker.io/}
+      repository: ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}
+      username: ${SPINNAKER_DOCKER_USERNAME}
+      # A path to a plain text file containing the user's password
+      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE} 
 
 services:
   default:
--- a/install/first_google_boot.sh
+++ b/install/first_google_boot.sh
@@ -24,6 +24,8 @@ set -u
 # We're running as root, but HOME might not be defined.
 AWS_DIR=/home/spinnaker/.aws
 KUBE_DIR=/home/spinnaker/.kube
+# Google Container Registry (GCR) password file directory.
+GCR_DIR=/home/spinnaker/.gcr
 SPINNAKER_INSTALL_DIR=/opt/spinnaker
 LOCAL_CONFIG_DIR=$SPINNAKER_INSTALL_DIR/config
 
@@ -147,6 +149,7 @@ function extract_spinnaker_credentials() {
     extract_spinnaker_google_credentials
     extract_spinnaker_aws_credentials
     extract_spinnaker_kube_credentials
+    extract_spinnaker_gcr_credentials
 }
 
 function extract_spinnaker_google_credentials() {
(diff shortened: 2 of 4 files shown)


lwander:
I also enable Igor polling when kubernetes is enabled

config/default-spinnaker-local.yml
```diff
@@ -96,7 +96,6 @@ providers:
       # These credentials use authentication information at ~/.kube/config
       # by default.
       name: my-kubernetes-account
-      namespace: default
```
lwander: This is a useless default since it's never consumed in `clouddriver.yml`, it overrides whats in `.kube/config`, and is already the default if nothing is specified in `.kube/config`. So I removed it.

config/default-spinnaker-local.yml
```diff
@@ -110,9 +109,11 @@ providers:
     enabled: ${SPINNAKER_KUBERNETES_ENABLED:false}
 
     primaryCredentials:
-      name: my-docker-registry-account
-      address: https://index.docker.io/
-      repository: library/nginx
+      name: my-docker-registry
```
ewiseblatt: What is the "name"? For all the others, it is the account name within spinnaker -- it has no meaning outside spinnaker. The default ends in "account" to help clarify this. Here, it sounds like name is the registry name, which might be a real name. If so, this isnt standard so perhaps should use a different name. Otherwise, perhaps it should decorate with "account" for consistency.

config/default-spinnaker-local.yml
```diff
@@ -110,9 +109,11 @@ providers:
...
+      address: ${SPINNAKER_DOCKER_REGISTRY:https://index.docker.io/}
+      repository: ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}
+      username: ${SPINNAKER_DOCKER_USERNAME}
+      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE}
```
ewiseblatt: What is a passwordFile? Is this standard? Is it keyed or flat?

You might want to modify reconfigure_spinnaker to check permissions on it.

install/first_google_boot.sh
```diff
@@ -24,6 +24,7 @@ set -u
 # We're running as root, but HOME might not be defined.
 AWS_DIR=/home/spinnaker/.aws
 KUBE_DIR=/home/spinnaker/.kube
+GCR_DIR=/home/spinnaker/.gcr
```
ewiseblatt: Perhaps comment what GCR is.

install/first_google_boot.sh
```diff
@@ -233,6 +235,35 @@ function extract_spinnaker_kube_credentials() {
   fi
 }
 
+function extract_spinnaker_gcr_credentials() {
+  local config_path="$GCR_DIR/gcr.json"
+  mkdir -p $(dirname $config_path)
+  chown -R spinnaker:spinnaker $(dirname $config_path)
+
+  if clear_metadata_to_file "gcr_json" $config_path; then
+    # This is a workaround for difficulties using the Google Deployment Manager
```
ewiseblatt: Not sure if this is still true. This comment dates back to last year.

install/first_google_boot.sh
```diff
@@ -233,6 +235,35 @@ function extract_spinnaker_kube_credentials() {
...
+    # to express no value. We'll use the value "None". But we don't want
+    # to officially support this, so we'll just strip it out of this first
+    # time boot if we happen to see it, and assume the Google Deployment Manager
+    # got in the way.
+    sed -i s/^None$//g $config_path
+    if [[ -s $config_path ]]; then
+      chmod 600 $config_path
```
ewiseblatt: This could be 400 since you arent going to edit a json file.
If you copied this from somewhere else, you can fix that too.
The configuration files are 600 since presumably you'll edit them.
The generated key files may eventually be replace when keys are redacted, etc, but would not be edited in place.

config/default-spinnaker-local.yml
ewiseblatt: I just noticed I made a typo at line 103. Also, there isnt a link to documentation url like the other sections. You might want to add one, but the docs are outside this PR.

config/default-spinnaker-local.yml
lwander: It's a file containing the password in plaintext. I'll write a comment here.

install/first_google_boot.sh
lwander: Sounds good, I grabbed this from the kubeconfig setup above and will change it there too.

config/default-spinnaker-local.yml
lwander: I removed `account`, since it's not an account like the others (you don't deploy to your docker account), instead it's `registry` in the sense that it's used to pull stored images out of. Since `primaryCredentials` is singular, I don't think the name needs to reflect the specific registry being used. Otherwise `my-kubernetes-account` should include the name of the cluster in use, and `my-google-account` should include the name of the project.

I'll link to the configuration docs in the comment above.

install/first_google_boot.sh
lwander: I'd rather keep it (or a reference to this comment elsewhere) since I'm hopefully adding first class GCR & Kubernetes support to the DM marketplace deployer soon and will get to see firsthand if this works

ewiseblatt:
LGTM

PR #917, opened 2016-06-02
Use noninteractive when performing dist-upgrade on startup.

duftler:
We encountered a package that was prompting for user input on upgrade.

### DIFF
--- a/dev/build_google_image.packer
+++ b/dev/build_google_image.packer
@@ -44,7 +44,7 @@
     "type": "shell",
     "inline": [
         "dpkg --list | grep linux-image | awk '{ print $2 }' | sort -V | sed -n '/'`uname -r`'/q;p' | xargs sudo apt-get -y purge",
-        "if [ '{{user `update_os`}}' = 'true' ]; then apt-get update -y; apt-get -y dist-upgrade; sed -i 's/start_rpc: false/start_rpc: true/' /etc/cassandra/cassandra.yaml; fi",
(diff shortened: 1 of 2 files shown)


duftler:
@lwander or @ewiseblatt please review.

dev/build_google_image.packer
```diff
@@ -44,7 +44,7 @@
     "type": "shell",
     "inline": [
         "dpkg --list | grep linux-image | awk '{ print $2 }' | sort -V | sed -n '/'`uname -r`'/q;p' | xargs sudo apt-get -y purge",
-        "if [ '{{user `update_os`}}' = 'true' ]; then apt-get update -y; apt-get -y dist-upgrade; sed -i 's/start_rpc: false/start_rpc: true/' /etc/cassandra/cassandra.yaml; fi",
+        "if [ '{{user `update_os`}}' = 'true' ]; then apt-get update -y; DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade; sed -i 's/start_rpc: false/start_rpc: true/' /etc/cassandra/cassandra.yaml; fi",
```
ewiseblatt: Why is this only an issue with upgrade and not the normal installs too?
Not as critical to answer since this would fail on our side rather than customers.

ewiseblatt:
LGTM

dev/build_google_image.packer
duftler: I believe it was an issue with both.

PR #919, opened 2016-06-02
deploy/kubernetes: Avoid recreating failed pods

lwander:
@duftler FYI
See more details on [`restartPolicy](http://kubernetes.io/docs/user-guide/jobs/#handling-pod-and-container-failures).


duftler:
A belated LGTM.

PR #922, opened 2016-06-06
notifications: spinnaker.baseUrl should default to services.deck.base…

duftler:
Echo relies on this url when constructing links to send in notifications.

### DIFF
--- a/config/echo.yml
+++ b/config/echo.yml
@@ -6,6 +6,9 @@ cassandra:
   embedded: ${services.cassandra.embedded:false}
   host: ${services.cassandra.host:localhost}
 
+spinnaker:
+  baseUrl: ${services.deck.baseUrl}
+
 front50:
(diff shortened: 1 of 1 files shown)


duftler:
@ewiseblatt please review

duftler:
Overrides this setting: https://github.com/spinnaker/echo/blob/master/echo-web/config/echo.yml#L5

duftler:
or @lwander

lwander:
Makes sense, LGTM

ewiseblatt:
Sorry, I missed this earlier when I scanned my inbox.
I know it was merged, but how is this used?

duftler:
It is used in constructing links in notifications. Prior to this change, if
somebody were to expose deck via http://some-host-name.io, the echo
notifications would still include links based on http://localhost:9000.

PR #927, opened 2016-06-07
surfaces defaultKeyPairTemplate in yml

danielpeach:
surfaces defaultKeyPairTemplate in yml. @duftler PTAL

### DIFF
--- a/config/spinnaker.yml
+++ b/config/spinnaker.yml
@@ -192,6 +192,9 @@ providers:
       # /home/spinnaker). The primaryCredentials.name will identify which profile
       # to use (for .aws/credentials).
 
+    # {{name}} will be interpolated with the aws account name (e.g. "my-aws-account-keypair").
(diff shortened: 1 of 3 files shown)


config/spinnaker.yml
```diff
@@ -191,6 +191,7 @@ providers:
       # spinnaker installation runs as user "spinnaker" whose $HOME is
       # /home/spinnaker). The primaryCredentials.name will identify which profile
       # to use (for .aws/credentials).
+    defaultKeyPairTemplate: "{{name}}-keypair"
```
duftler: Probably worth a comment here to explain the context within which the template will be expanded. A reader may be curious about the `{{name}}` bit without some explanation. Please put the same comment here and in `default-spinnaker-local.yml`.

danielpeach:
@duftler added comment

PR #938, opened 2016-06-13
Inject non-superfluous quotes when rewriting yaml

ewiseblatt:
@jtk54 
Write back quotes into YAML files when they are required.

### DIFF
--- a/pylib/spinnaker/transform_old_config.py
+++ b/pylib/spinnaker/transform_old_config.py
@@ -20,6 +20,11 @@ from run import check_run_quick
 
 
 class Processor(object):
+  def __collect_environ(self, content):
+    return {match.group(1): match.group(2).strip()
+            for match in re.finditer('^([^\s#]+)=([^\n#]*)',
+                                     content, re.MULTILINE)}
+    
   def __init__(self, config, environ_path, yml_path, aws_path):
       with open(environ_path, 'r') as f:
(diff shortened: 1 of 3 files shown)


ewiseblatt:
@jtk54, added the other change I mentioned. ready for you to review.

pylib/spinnaker/transform_old_config.py
```diff
@@ -20,6 +20,13 @@
 
 
 class Processor(object):
+  def __collect_environ(self, content):
+    result = {}
+    for match in re.finditer('^([^\s]+?)=([^\n]*)(?:\#.+)?', content,
```
jtk54: `[^\s]` should go to `[\S]` in the first group. Are you trying to cut off `#` comments with the second two groups? If so, I would change `[^\n]` to `[^\n\#]` to not match comments before the non-capturing group.

pylib/spinnaker/transform_old_config.py
```diff
@@ -20,6 +20,13 @@
...
+                             re.MULTILINE):
+      result[match.group(1)] = match.group(2)
```
jtk54: You could do some pythonic dictionary construction like `result = {match.group(1): match.group(2) for match in re.finditer(...)}`. If there are no matches this dictionary comprehension would still return `{}`.

jtk54:
LGTM

PR #940, opened 2016-06-14
Support external storage bucket configuration from spinnaker-local.yml

ewiseblatt:
@ajordens I'm sharing the config attributes for S3 and Google Cloud Storage and just having different enable flags. Front50 still uses explicit attributes for each of the services. Since they are disjoint, I thought you didnt need the different config parameters in spinnaker-local so thought this might make sense. Feel free to object.


ewiseblatt:
@duftler

duftler:
LGTM.

ajordens:
Belated LGTM

PR #942, opened 2016-06-17
Remove references to deprecated rush service.

duftler:


duftler:
@ewiseblatt please review.
@lwander This touches the k8s setup as well.
@tomaslin This touches the docker-compose setup.

experimental/packer/configure.sh
```diff
@@ -130,8 +130,6 @@ gem install bundler
 
 #deb-s3 upload --bucket my-bucket my-deb-package-1.0.0_amd64.deb
 
-docker run -d -p 5000:5000 --name registry registry:2
```
ewiseblatt: Is this change for some other reason? Ok, just curious.

ewiseblatt:
LGTM assuming port 5000 reference was intentional and that was the only occurance.

lwander:
LGTM, thanks for doing the k8s setup as well

experimental/packer/configure.sh
duftler: That related to the (deprecated) mechanism for running jobs within rush as docker containers.

Issue #949, opened 2016-06-23
validate_configuration.sh gives bad perm advice for spinnaker-local.yml


wgjohnson:
Running `validate_configuration.sh` gives a suggestion for non-owner access for `spinnaker-local.yml`
```
ubuntu@ip-172-31-197-212:~$ sudo /opt/spinnaker/bin/validate_configuration.sh 
/opt/spinnaker/config/spinnaker-local.yml has configuration errors:
   * "/opt/spinnaker/config/spinnaker-local.yml" should not have non-owner access. Mode is 644.
```

I've interpreted 'non-owner access' to mean no read access for group or other. Upstart log file for `gate` is empty.
```
ubuntu@ip-172-31-197-212:~$ sudo chmod 600 /opt/spinnaker/config/spinnaker-local.yml
ubuntu@ip-172-31-197-212:~$ sudo cat /var/log/upstart/gate.log
ubuntu@ip-172-31-197-212:~$ 
```

Starting Spinnaker fails to start, and the Upstart log file for `gate` has permission errors for `spinnaker-local.yml`.
```
ubuntu@ip-172-31-197-212:~$ sudo start spinnaker
spinnaker start/running
ubuntu@ip-172-31-197-212:~$ sudo cat /var/log/upstart/gate.log
Exception in thread "main" java.lang.IllegalStateException: java.io.FileNotFoundException: /opt/spinnaker/config/spinnaker-local.yml (Permission denied)
	at org.springframework.beans.factory.config.YamlProcessor.handleProcessError(YamlProcessor.java:187)<snip>

```

Expected behavior is for validate_configuration.sh to provide useful suggestions :) 

Cloud provider is AWS, with latest build per http://www.spinnaker.io/docs/amazon-ami-ids Spinnaker-Ubuntu-14.04-40	14.04 LTS us-west-2 ami-80718ce0

duftler:
@ewiseblatt

ewiseblatt:
@wgjohnson Could the problem here be that your files are owned by root rather than by spinnaker?
I think the behavior is correct, and I think you want tight control over spinnaker-local.yml because it can contain clear text passwords in it.

As for your remedy "Expected behavior is for validate_configuration.sh to provide useful suggestions :)" I am inclined to fix this by checking the ownership of the file and directing you to chown spinnaker:spinnaker the file. Would that be acceptable?

wgjohnson:
[...]

PR #950, opened 2016-06-23
Check ownership of sensitive files.

ewiseblatt:
Fix for https://github.com/spinnaker/spinnaker/issues/949

@wgjohnson verify this satisfies your needs and expectations.
@duftler review

### DIFF
--- a/pylib/spinnaker/validate_configuration.py
+++ b/pylib/spinnaker/validate_configuration.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 import os
+import pwd
 import re
 import sys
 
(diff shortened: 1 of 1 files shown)


duftler:
LGTM once @wgjohnson verifies the change.

wgjohnson:
I'm using AWS AMI 'Spinnaker-Ubuntu-14.04-40' from Spinnaker.io. The fixed version seems a lot newer[1], and fails to start because of missing module "Configurator"[2]. How important is it that I personally test this?

Also, per new comments for #949, other local files can be expected to have sensitive information. If it isn't issue-creep, I would suggest verify_security() includes at least gate-local.yml, and possibly more (Sorry, I'm not familiar enough with Spinnaker to know which files are implied to have sensitive information). If it is creep, and you're okay with me not explicitly testing it, LGTM.


1 - 
```
ubuntu@ip-172-31-197-212:/opt/spinnaker$ wc -l bin/validate_configuration.*
  329 bin/validate_configuration.ewiseblatt:check_ownership.py
   19 bin/validate_configuration.sh
```

2 - 
```
ubuntu@ip-172-31-197-212:/opt/spinnaker/bin$ sudo ./validate_configuration.ewiseblatt\:check_ownership.py 
Traceback (most recent call last):
  File "./validate_configuration.ewiseblatt:check_ownership.py", line 22, in <module>
    from configurator import Configurator
ImportError: No module named configurator
```

ewiseblatt:
The python file is intended to be used a s library in support of the bash
script, which sets the python path.

Checking -local files makes sense, though I dont know what passwords would
be in gate.
Normally clouddriver-local and spinnaker-local would be the main places.
Maybe igor.
What is the gate password for?

wgjohnson:
Java keystore password per http://www.spinnaker.io/docs/securing-spinnaker

ewiseblatt:
I'm not addessing the other -local files here because of potential unanticipated consequences before I leave on vacation. But will open an issue for it.

PR #951, opened 2016-06-23
Removed unnecessary apache conf removal, listener overwrites.

jtk54:
Removed unnecessary apache conf removal, listener overwrites. @ewiseblatt PTAL.

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -17,9 +17,9 @@ GOOGLE_ENABLED=false  # set by --cloud_providers
 
 # Set by --local-install, means download packages
 # rather than adding additional external repositories.
-DOWNLOAD=false        
+DOWNLOAD=false
 
-# Set by --quiet, wont prompt or echo status output.                      
+# Set by --quiet, wont prompt or echo status output.
 QUIET=false
 
 # Set by --dependencies only,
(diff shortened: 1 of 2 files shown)


jtk54:
@dpeterka Seems like these changes were yours, can you take a look as well? I'm not sure which assumptions were made here that might be broken by this change.

dpeterka:
Hmm the local proxy used to listen on :80 I believe the `rm -f *` can go, echo will append repeated lines if you run the installer more than once. I would prefer to keep the sed.

jtk54:
If a user actually is using a listener on 80, this won't be good. Maybe we can check if the "127.0.0.1:9000" line exists before appending?

jtk54:
@dpeterka Good catch on the multiple appends. Which local proxy was listening on :80? Was it forwarding to :9000?

ewiseblatt:
FYI I dont know why the rm was there before, however if we are introducing
apache (was not already in the deployed instance) then we should not be
opening up any new vectors into the instance. I dont know what the default
apache installation does. A postInstall wouldnt know what we did. I think
this (if anything) would need to happen as part of the installation action
itself. Ultimately I'd defer to standard practices presuming they are
accepted as being secure for someone not intending on exposing a public
service.

jtk54:
I moved the apache2 configuration into the `install_apache2` function in the main install script, and removed it from the post install. If there is an existing installation (via `dpkg -s`) then we assume we should protect the config, and append "Listen 127.0.0.1:9000" if it's not there. Otherwise, we replace the default "Listen 80" with our localhost listener. Sorry for the whitespace noise.

dpeterka:
This looks good to me.  I think this step has always been a bit of overkill (installing apache). At some point we should probably just consider creating nginx/apache2 example configurations in /usr/share/ and make it a manual step or check and do some symlinks.  That being said LGTM thanks @jtk54

PR #954, opened 2016-06-23
Make it possible to shut off cassandra.

ewiseblatt:
@duftler These changes should allow you to manually configure out spinnaker with standard configs and have things run

@lwander When you add redis to echo, you need to add similar enabled entries for the forthcoming scripts to allow you to chose that, at least for the time being.


ewiseblatt:
@duftler

duftler:
LGTM.

PR #955, opened 2016-06-23
Automate removing cassandra (and bringing it back)

ewiseblatt:
@duftler This script allows you to turn cassandra on and off, and defaults the google image to off. It only supports prod builds right now. I dont have a developer build path yet. To run it, in a spinnaker

sudo /opt/spinnaker/install/change_cassandra --echo=<sys> --front50=<sys> [--change_local=true|false] [--change_global=true|false] [--bucket=<name>]
where sys is a supported storage system for echo or front50. You need both. If neither is cassandra, it will also shut down cassandra. If either is cassandra it will also start up cassandra and initialize the keyspace. Either way, it will modify the yaml files. By default --change_local=true and --change_global=false where global is spinnaker.yml and local is spinnaker-local.yml.

If you specify a bucket it will configure that storage bucket as well. I modified first_google_boot to guess at a unique storage bucket based on the project name. If you deploy multiple spinnaker deployments into the same project then this will not work as intended (but you can override the name). Otherwise I think it is reasonable and easy to experiment with.

@lwander To support changing to redis in echo, add it to the _ECHO_CHOICES and ECHO_KEYS and the desired value to the bindings config in the constructor. The yml files needs to have existing values from the prior PR.

### DIFF
--- a/install/change_cassandra.sh
+++ b/install/change_cassandra.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+#
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+PYTHONPATH=$(dirname $0)/../pylib \
+python $(dirname $0)/../pylib/spinnaker/change_cassandra.py $*
--- a/pylib/spinnaker/change_cassandra.py
+++ b/pylib/spinnaker/change_cassandra.py
@@ -1,0 +1,219 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


ewiseblatt:
relized i'd overriden my local in deployment. Pushed a fix but cannot test it vanilla until later.

ewiseblatt:
@duftler - this is ready now.

install/change_cassandra.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+#
+# Copyright 2015 Google Inc. All Rights Reserved.
```
duftler: s/2015/2016

pylib/spinnaker/change_cassandra.py
```diff
@@ -0,0 +1,212 @@
...
+
+"""Tweak an installation to enable cassandra or not.
+
+This module/program lets you change a deployment's persistence mechanisms
+for front50 and echo. If using cassandra for either, then a cassandra
+server will be started (if cassandra was configured to run on this host).
+Otherwise, cassandra will be shut down. Either way, the yaml configuration
+files will be updated to persist the configuration.
+
+By default only spinnaker-local will be updated. However if you can specify
```
duftler: Strike 'if'.

pylib/spinnaker/change_cassandra.py
```diff
@@ -0,0 +1,212 @@
...
+whether to update spinnaker.yml and/or spinnaker-local.yml.
+
+When using s3 or gcs storage (for front50) you can also override the bucket
+that is used.
+
```
duftler: Please add a comment regarding the idempotency of this script.

pylib/spinnaker/spinnaker_runner.py
```diff
@@ -240,6 +240,12 @@ def start_subsystem(self, subsystem, environ=None):
                      .format(command=command, log=base_log_path)],
                      environ=environ)
 
+  def needs_cassandra(self):
+    """Check if configuration requires cassandra ornot."""
```
duftler: s/ornot/or not

pylib/spinnaker/spinnaker_runner.py
```diff
@@ -240,6 +240,12 @@ def start_subsystem(self, subsystem, environ=None):
...
+    front50 = self.__bindings.get('services.front50.cassandra.enabled')
+    echo = self.__bindings.get('services.front50.cassandra.enabled')
```
duftler: Shouldn't this be `services.echo.cassandra.enabled`?

duftler:

PR #959, opened 2016-06-27
Fixed yaml_util writing captalized booleans to config.

jtk54:
Fixed yaml_util writing captalized booleans to config. @lwander PTAL. @ewiseblatt FYI.


lwander:
LGTM

PR #960, opened 2016-06-27
provider/kubernetes Remove cassandra as a dependency

lwander:
@skim1420, @jtk54, @ttomsu PTAL

Removes Cassandra from front50 & echo in the Kubernetes deployment. This is a key step in "productionizing" this deployment method.


jtk54:
In the yaml blocks, I would mention which properties should be verbatim the string that you put there (e.g.` __json_key`), and which properties can be more freely chosen.

experimental/kubernetes/README.md
```diff
@@ -37,14 +27,118 @@ can authenticate with the cluster you want to deploy Spinnaker to.
 Once that is all squared away, make sure that running `$ kubectl config
 current-context` refers to the cluster you want to have Spinnaker running in.
 (You may need to update `kubectl` for the previous command to work).
+There are two paths to take here
+
+## Configuring your Docker registry
+
+Once you've decided which registries you will use and they are configured,
+please make sure that the following list in `./config/clouddriver-local.yml` is
```
jtk54: I might add a note that `./config/` is whichever config directory they're using, i.e. `.spinnaker` or `/opt/spinnaker/config`, etc.

lwander: No, in this case `./config` is the existing subdirectory in this directory containing all spinnaker config. It's not user chosen but exists as a place to collect all spinnaker config before it is packaged into a Kubernetes Secret

jtk54: Got it.

PR #963, opened 2016-06-29
Removed Cassandra waiting from upstart config.

jtk54:
Removed Cassandra waiting from upstart config. This hangs for people not using Cassandra on localhost. Replaced with reporting. @lwander PTAL. @ewiseblatt FYI.

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -678,7 +678,10 @@ fi
 ## Spinnaker
 install_spinnaker
 
-#write values to /etc/default/spinnaker
+# Touch a file to tell other scripts we installed Cassandra.
+touch /opt/spinnaker/cassandra/SPINNAKER_INSTALLED_CASSANDRA
+
+# Write values to /etc/default/spinnaker.
 if [[ $AWS_ENABLED || $AZURE_ENABLED || $GOOGLE_ENABLED ]] ; then
   if [[ $AWS_ENABLED == true ]] ; then
     write_default_value "SPINNAKER_AWS_ENABLED" "true"
--- a/etc/init/spinnaker.conf
+++ b/etc/init/spinnaker.conf
@@ -4,13 +4,22 @@ start on filesystem or runlevel [2345]
 stop on shutdown
 
 pre-start script
-  if ! nc -z localhost 9042; then
-    echo "Waiting for cassandra"
-    while ! nc -z localhost 9042; do
-      sleep 1
-    done
-    echo "Cassandra is ready"
-  fi
+    if [ -f /opt/spinnaker/cassandra/SPINNAKER_INSTALLED_CASSANDRA ]; then
+        if ! nc -z localhost 9042; then
+            echo "Waiting for Cassandra"
+            for i in `seq 1 30`; do
+                if nc -z localhost 9042; then
+                    echo "Cassandra is ready"
+                    break
+                fi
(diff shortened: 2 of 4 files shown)


etc/init/spinnaker.conf
```diff
@@ -4,12 +4,12 @@ start on filesystem or runlevel [2345]
...
-    echo "Waiting for cassandra"
-    while ! nc -z localhost 9042; do
-      sleep 1
-    done
-    echo "Cassandra is ready"
+  if nc -z localhost 9042; then
+    echo "Cassandra server is ready on localhost."
+  else
+    echo "Cassandra does not appear to be running on localhost.
+If Spinnaker is configured to use Cassandra, start it and restart Spinnaker.
```
lwander: I'm not familiar with .conf format, but is this how comments work?

duftler:
The original looping/blocking behavior makes me think that cassandra will sometimes (usually?) not be ready the first time it is pinged. Isn't there a risk that installations using cassandra will see the warning message almost every time this is executed?

etc/init/spinnaker.conf
jtk54: It's a multi-line string to keep line length < 80.

jtk54:
Discussed this out of band, pushed new changes.

InstallSpinnaker.sh
```diff
@@ -678,6 +678,9 @@ fi
 ## Spinnaker
 install_spinnaker
 
+# touch a file to tell other scripts we installed Cassandra
```
duftler: Please capitalize the comment and terminate it with a period.

pylib/spinnaker/change_cassandra.py
```diff
@@ -104,12 +104,19 @@ def __init__(self, options):
...
+      open('/opt/spinnaker/cassandra/SPINNAKER_DISABLED_CASSANDRA', 'w').close()
     with open('/etc/init/cassandra.override', 'w') as f:
         f.write('manual')
     print 'Stopping cassandra service...'
     os.system('service cassandra stop')
 
   def enable_cassandra(self):
+    if os.path.exists('/opt/spinnaker/cassandra/SPINNAKER_DISABLED_CASSANDRA'):
+      os.remove('/opt/spinnaker/cassandra/SPINNAKER_DISABLED_CASSANDRA')
+      (open('/opt/spinnaker/cassandra/SPINNAKER_INSTALLED_CASSANDRA', 'w')
```
duftler: What's the reason this line has parens around it and the similar line in `disable_cassandra()` does not?

pylib/spinnaker/change_cassandra.py
```diff
@@ -104,12 +104,19 @@ def __init__(self, options):
     self.__bindings.import_dict(config)
 
   def disable_cassandra(self):
+    if os.path.exists('/opt/spinnaker/cassandra/SPINNAKER_INSTALLED_CASSANDRA'):
```
duftler: Could these 2 paths be constants instead?

pylib/spinnaker/change_cassandra.py
jtk54: This makes each line < 80 chars.

duftler:
LGTM once comments are addressed.
@lwander PTAL as well.

lwander:
LGTM

PR #975, opened 2016-07-06
Redis baseUrl host comment

edwinavalos:
baseUrl doesn't seem to be used in spinnaker.yml, it should be host instead. Also added a comment clarifying that the redis:// is not needed in this case because spinnaker.yml contains that prefix.


medyagh:
+1

PR #976, opened 2016-07-07
Moved keyspace creation into install script.

jtk54:
We should create keyspaces only if we install Cassandra, and because we only install Cassandra in InstallSpinnaker, we can move the keyspace creation. There is no guarantee that postInstall will run after InstallSpinnaker. @lwander PTAL, @ewiseblatt @danielpeach FYI.

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -680,6 +680,8 @@ install_spinnaker
 
 # Touch a file to tell other scripts we installed Cassandra.
 touch /opt/spinnaker/cassandra/SPINNAKER_INSTALLED_CASSANDRA
+cqlsh -f "/opt/spinnaker/cassandra/create_echo_keyspace.cql"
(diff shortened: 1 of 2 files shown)


InstallSpinnaker.sh
```diff
@@ -680,6 +680,8 @@ install_spinnaker
 
 # Touch a file to tell other scripts we installed Cassandra.
 touch /opt/spinnaker/cassandra/SPINNAKER_INSTALLED_CASSANDRA
```
lwander: Is this still necessary (are there other checks relying on this) because below you deleted one.

jtk54: Yes, the upstart config and enable/disable Cassandra scripts check for this.

lwander:
Out of band @jtk54 told me he means "because we only install Cassandra in Install Spinnaker" he can make these changes, which SGTM

lwander:
LGTM, mergearooni

PR #979, opened 2016-07-11
kubernetes: show how to update port & host

lwander:


experimental/kubernetes/rcs/spin-deck.yaml
```diff
@@ -25,6 +25,11 @@ spec:
             - -c
           args:
             - "cp /opt/spinnaker/config/settings.js .; npm start"
+          env:
+            - name: DECK_HOST
+              value: localhost
```
janwillies: I think it'd make sense to default to "0.0.0.0". Otherwise the provided service "spin-deck" can't be used (and should be deleted, because it confuses people)

It would be still internal, for external access you would need to configure the spin-deck service with Type: Loadbalancer/NodePort

lwander: That's fair, I'll update it here.

PR #999, opened 2016-07-19
provider/consul: Start creating consul install path

lwander:
The idea is that this is run once to create a base VM image with consul installed. Since it's not in any apt repository (AFAIK), this is the easiest install path.

@duftler  PTAL


duftler:
LGTM.

Issue #1002, opened 2016-07-20
Front50 should allow configuring regions for GCS 


leelasharma:
Front50 should provide the ability to configure location/region for the gcs bucket.

ttomsu:
I'll be in that area for the auth stuff. I'll add this to the list.

PR #1003, opened 2016-07-20
Add bucket location to default GCS config

ttomsu:
@ewiseblatt PTAL

tag: #1002


ewiseblatt:
squash. I assume you added bucket location into spinnaker.yml already or else this is malformed and you need to do that too.

LGTM

PR #1011, opened 2016-07-23
Remove reference to Cassandra

lwander:
@duftler - I forgot to remove this reference when I remove cassandra as a pipeline store option.


duftler:
Lgtm.

PR #1026, opened 2016-07-28
Allow InstallSpinnaker.sh to ommit cassandra

ewiseblatt:
@duftler 

When omitting cassandra, it will still add the package manager location
but will not actually install it. I can remove that too, but figured I'd
wait until we completely remove cassandra. In the meantime adding the location
will make it easier to manually install cassandra later should the deployer change
their mind.

The default behavior for GCP is to leave out cassandra. The other platforms are unchanged
for the time being. Adding a --noinstall_cassandra will force it off and --install_cassandra
will force it on.

Removed explicit cassandra (and redis) management entirely from developer mode (running from source). To use cassandra from source, you'll need to ensure it is running yourself.
If it is an upstart service (as is normal) then it will already be running anyway.

I left the stop scripts in for now because I'm not sure if anyone is using the stop dependencies
script, but expect to remove this as well.


etc/default/spinnaker
```diff
@@ -12,16 +39,17 @@ SPINNAKER_AWS_DEFAULT_REGION=us-west-2
 # If you plan on using Google Compute Engine set this to true
 SPINNAKER_GOOGLE_ENABLED=false
 # GCP project you intend to manage
-SPINNAKER_GOOGLE_PROJECT_ID=
+SPINNAKER_GOOGLE_PROJECT_ID=`__get_google_metadata_value "projects/project-id"`
```
duftler: What is the behavior when not running on GCE?

duftler:
LGTM after addressing comment.

etc/default/spinnaker
ewiseblatt: For etc/default/spinnaker, this is problematic as I'd argued offline when we came up with this solution. What will happen is that it will still attempt to get the metadata value, which will fail and return "" as the value. For google, the failure is essentially immediate (the cost of a DNS lookup) because the hostname wont be found (I suppose you could spoof DNS by creating an artificial entry). For AWS, which uses internal IP addresses, it will likely have to time out trying to connect to a server at that address, unless there happens to be a server there, which will likely fail because it wont know the URL. The function will return "" so as not to bind a value. This isnt an issue at this exact moment because I left AWS to default to cassandra so there are no actual AWS call references here. However moving forward, that's what will happen unless there are other environment signals saying we're on AWS that can guard the metadata server attempt.

I have thoughts about some little optimizations that can be done in that file, but am waiting for more than the one call to be introduced.

When not running on GCE, for the time being, the behavior will be to install cassandra unless you run InstallSpinnaker with --noinstall_cassandra.

@duftler, ptal at default/spinnaker one last time. I made a change to how the default bucket is declared. It is ready for merge so you can do that after you look.

duftler:
LGTM.

PR #1036, opened 2016-07-28
Fix wrong(?) key name on default front50.yml file

fgbreel:
Hi, I come with this and took few hours debugging how to have `front50` backed with `s3` as storage persistence layer.

Looks like the latest AMI for the regions `eu-west-1` have a wrong key name on the default `front50.yml` file.

After this change I got `spinnaker` working and configured just in one place: `spinnaker-local.yml`

Is that right?

btw, great project! 👍

### DIFF
--- a/config/front50.yml
+++ b/config/front50.yml
@@ -32,7 +32,7 @@ spinnaker:
     project: ${providers.google.primaryCredentials.project}
     jsonPath: ${providers.google.primaryCredentials.jsonPath}
 
-  aws:
+  s3:
     enabled: ${services.front50.s3.enabled:false}
     bucket: ${services.front50.storage_bucket:}
     rootFolder: ${services.front50.bucket_root:front50}


duftler:
@ewiseblatt Can you please take a look at this?

ajordens:
LGTM

The key name should be `s3` not `aws`.

fgbreel:
❤️

ewiseblatt:
Thanks. Sorry about that!

fgbreel:
@ewiseblatt no problem! 😄 

Just one more thing:

ewiseblatt:
Your message ended after "one more thing:". If you asked something, I did
not get it.

On Thu, Jul 28, 2016 at 3:13 PM, Gabriel Francisco <notifications@github.com

fgbreel:
@ewiseblatt sorry for that, I hit ⌘+⏎ to fast.

ewiseblatt:
I see your question here in github (i originally got it through email).
Check the /var/log/spinnaker/front50/front50.log file for more details on the 500.
I'm familiar with the gcs implementation, not s3.

One thing it could be is that the data could be too big for the post request. We saw that on one of our internal migrations using a similar strategy. If that is the case you'll need to break it up.

duftler:
For that internal migration, it ended up not being an issue with the size.

fgbreel:
I'll check! 

I have 5 applications in `applications.json`, the longest one have 16 lines.
Also tried with just one using the minimum structure (from swagger docs http://localhost:8080/v2/api-docs) and got the same error.

This is the sanitized response from `curl http://localhost:8080/global/applications | json_pp`: 
```
[
    {
       "name" : "TEST",
       "createTs" : "1466677326975",
       "description" : null,
       "updateTs" : "1468910395462",
       "platformHealthOnly" : true,
       "email" : "test@example.com",
       "repoType" : "github",
       "instancePort" : null,
       "accounts" : "default",
       "providerSettings" : {
          "aws" : {
             "useAmiBlockDeviceMappings" : false
          }
       },
       "cloudProviders" : "aws",
       "platformHealthOnlyShowOverride" : true
    }
]
```

Thank you so much!

ewiseblatt:
Did you get any info from the log file about the 500?
The application definition looks ok to me.
Re-reading your original message, I'm not familiar with the notation of what you are posting, what it means or where it is from. It is suggesting you are posting a message whose content type is application/json but the message body is not a json body. It looks like something intended to direct to another tool, but I'm not familiar with it. Could you clarify?

fgbreel:
Somehow is working through curl now too.

Interesting is that only failed with `applications.json` and not with the others (pipelines, and so on) even using the same code, but I can use plain curl too.

The behavior was:
- posting `pipelines.json`: √
- when posting the `applications.json` `front50` returns `http 500`, then crash and try to restart itself but the previous still running, so it receive `address already in use` exception. So I believe that is related to the change of this PR, and now fixed. (this is working now)
- now I just receiving `202` for `projects.json`, but I believe that is because of `uuid` or something. (don't worry about this)

About the tool, it's Ansible. I'm using to automate the configuration of an official Netflix AMI after launching, then I terminate the instance, so can't get the log from old ones, but if happen again I can open another issue with the stack trace and logs.

Thank you so much!

PR #1054, opened 2016-08-10
Drop propagateCloudProviderType flag.

duftler:
This was dropped from orca here: https://github.com/spinnaker/orca/pull/922


duftler:
@ewiseblatt please review.

duftler:
@lwander fyi

PR #1080, opened 2016-08-23
fix a bug with upstart where override files are overwritten, these fi…

JohnTheodore:
…les should be sticky and not be overwritten once created. Add sane defaults for open file descriptors. For the recommended instance types, this is a better default. igor coughs out errors about open file descriptors with the upstart defaults. Our spinnaker instance is only 10 applications, so afaict others get this error too


duftler:
@dstengle Can you PTAL?

dstengle:
@duftler LGTM

PR #1082, opened 2016-08-25
Fixed postInstall

ewiseblatt:
@duftler 
postInstall is running in bourne shell, with built-in echo that doesnt support -e.
This was causing the config files to be malformed (containing -e) and preventing spinnaker
from running via upstart.

Also make first_google_boot resilient for configuring in the case when spinnaker could not start.


duftler:
LGTM.

PR #1105, opened 2016-09-05
(consul) default config

lwander:
@ewiseblatt PTAL, @duftler FYI


ewiseblatt:
I dont understand why this is tied to gce. We can talk about it in the
morning.

ewiseblatt:
LGTM pending environment variable name discussed offline

PR #1110, opened 2016-09-07
(kubernetes): simplify install

lwander:


lwander:
Sorry I missed the commit message here. 

There are two changes in metadata read by the install script. 

The first is that the user no longer provides a `~/.kube/config`, but instead only a cluster name to manage. Providing the whole `~/.kube/config` is unwieldy. 

The second is that the user provides the name of the service account the authenticate against GCR rather than the json key itself. Unfortunately, there is a bit of a catch-22 when it comes to generating service accounts with sufficient permission out of thin air. To create a service account on a VM and assign it a new role, the VM needs to be created with a service account that has `project/owner`-level permissions. But to create a VM with `project/owner`-level permissions, you need the ability to assign service accounts new roles. So the only solution I see is to require the user of the script to have a canned service account name ready.

@duftler @ewiseblatt PTAL

config/clouddriver.yml
```diff
@@ -75,9 +75,7 @@ dockerRegistry:
     - name: ${providers.dockerRegistry.primaryCredentials.name}
       address: ${providers.dockerRegistry.primaryCredentials.address}
       username: ${providers.dockerRegistry.primaryCredentials.username:}
-      passwordFile: ${providers.dockerRegistry.primaryCredentials.passwordFile:}
-      repositories: 
-        - ${providers.dockerRegistry.primaryCredentials.repository}
+      passwordFile: ${providers.dockerRegistry.primaryCredentials.passwordFile}
```
lwander: The motivation to remove the repos field is that GCR now supports `/_catalog` (reading all owned repos of the authenticated user).

lwander:
The motivation here is to enable the google cloud launcher to easily deploy to kubernetes: https://cloud.google.com/launcher?q=spinnaker

The same way optional jenkins & AWS credentials can supplied, the kubernetes cluster & gcr account will be configurable as well: 
sz8kk1ddori

PR #1118, opened 2016-09-08
manually set $HOME, and pin a gcloud version

lwander:
@duftler Pinning a gcloud version makes this script more consistent. Also, apparently the startup script does not have a fully configured set of environment variables (`$HOME` isn't set).


duftler:
Lgtm.

PR #1142, opened 2016-09-13
Adds fiatEnabled flag to autoconfigured settings.js

ttomsu:
@danielpeach - This hooks up the feature flag you added in https://github.com/spinnaker/deck/pull/2662
@danielpeach, @ewiseblatt - PTAL


danielpeach:
LGTM

PR #1147, opened 2016-09-14
Fix the key that aws.defaults.iamRole is set up with by default

edwinavalos:
There was a typo provider -> providers, there is no provider key in configs


edwinavalos:
@spinnaker/reviewers PTAL

edwinavalos:
@duftler if you can take a look, I tried to ping the spinnaker/reviewers team but I don't seem to have rights to ping them.

Issue #694, opened 2016-01-20
Allow an extensible set of parameters to be configured in rosco and dynamically rendered in the bake stage configuration ui.


duftler:
These should be propagated back to the selected template.

duftler:
The Bake Stage now supports an optional`extendedAttributes` map. Any key/value pairs specified in this map will be propagated all the way through to the packer job. They will each be passed to packer using `-var`.

There is no first-class support in the ui for this yet, but it can be added to the pipeline config via the `Edit as JSON` action.

ajordens:
Satisfied with where this is at @duftler?

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 89 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

Issue #11, opened 2014-07-02
Kato: Create and Tag ENIs


danveloper:
spinnaker/kato#20

PR #1168, opened 2016-09-24
add -p to mkdir to fix errors when re-run Installer

tony612:

Please refer to [Contributing to Spinnaker](http://spinnaker.io/online_docs/community/contributing_guidelines.html) and [On Collaborative Development](http://spinnaker.io/documentation/collab_dev.html).

* Provide a descriptive title for your changes.


InstallSpinnaker.sh
```diff
@@ -689,7 +689,7 @@ if [[ "$INSTALL_CASSANDRA" != "false" ]]; then
 fi
 
 ## Packer
-mkdir /tmp/packer && pushd /tmp/packer
+mkdir -p /tmp/packer && pushd /tmp/packer
```
ewiseblatt: Perhaps this should be something like:

PACKER_DIR=$(mktemp -d)
pushd $PACKER_DIR
..
rm -rf $PACKER_DIR

And I see there are other mkdir calls in this script, which are susceptable to the same issue.
Maybe wrap the whole script in a TEMPDIR=$(mktemp -d installspinnaker.XXXX) up top and rm -rf $TEMPDIR below and have each of the internal mkdir calls (including this) use $TEMPDIR instead of /tmp

tony612: Thanks for your advice! Please review again

InstallSpinnaker.sh
```diff
@@ -3,6 +3,7 @@
 # This script installs Spinnaker and its dependencies.
 # See http://www.spinnaker.io/docs/creating-a-spinnaker-instance
 
+TEMPDIR=$(mktemp -d installspinnaker.XXXX)
```
ewiseblatt: Move this to before #11 (so it is after the -e and other shell config).
On second thought, I'd like to move it around #679 which is the block where all the installs happen that use this. Then the delete can go near by around #694 and the two pair "together".

ewiseblatt: I'd made another comment which I dont think went through.
I'd suggested leaving the rm calls where they were to do partial cleanup in the event of failure. The return at the end wouldnt get called if you only installed dependencies, which terminates early. So the rmdir may as well move up to the end of the dependency installation block, which is where it is used.

ewiseblatt:
ping

tony612:
@ewiseblatt Done

duftler:
@ewiseblatt Is this ready to be merged?

PR #1170, opened 2016-09-26
(gcr) simplify lookup of service account email

lwander:
@ewiseblatt PTAL

If the user now provides "gcr_enabled" as instance metadata, but does not
supply an email address, the default compute service account is used.

Also the user can optionally provide which GCR registry to use.

The next set of changes will allow a user to specify where their kube cluster is.

### DIFF
--- a/install/first_google_boot.sh
+++ b/install/first_google_boot.sh
@@ -249,26 +249,43 @@ function extract_spinnaker_gcr_credentials() {
   mkdir -p $(dirname $config_path)
   chown -R spinnaker:spinnaker $(dirname $config_path)
 
+  local gcr_enabled=$(get_instance_metadata_attribute "gcr_enabled")
   local gcr_account=$(get_instance_metadata_attribute "gcr_account")
-  if [ -n "$gcr_account" ]; then
-    # This downloads the JSON key for the specified service account
+
+  if [ -n "$gcr_enabled" ] || [ -n "$gcr_account" ]; then
+    echo "GCR enabled"
+
+    if [ -z "$gcr_account" ]; then
+      gcr_account=$(gcloud iam service-accounts list --filter="displayName:'Compute Engine default service account'" --format="value(email)")
+      clear_instance_metadata "gcr_account"
+    fi
+
+    echo "Extracting GCR credentials for email $gcr_account"
     gcloud iam service-accounts keys create $config_path --iam-account=$gcr_account
 
     if [[ -s $config_path ]]; then
       echo "Extracted GCR credentials to $config_path"
-
+  
       chmod 400 $config_path
       chown spinnaker:spinnaker $config_path
 
+      local gcr_location=$(get_instance_metadata_attribute "gcr_location")
+
+      if [ -z "$gcr_location" ]; then
+        gcr_location="https://gcr.io"
+      fi
+  
       write_default_value "SPINNAKER_DOCKER_PASSWORD_FILE" $config_path
       write_default_value "SPINNAKER_DOCKER_USERNAME" "_json_key"
-      write_default_value "SPINNAKER_DOCKER_REGISTRY" "https://gcr.io"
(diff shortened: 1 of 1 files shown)


lwander:
@andreasevers once this is surfaced in the launcher it will help avoid the `eu.gcr` issue we ran into.

lwander:
@jtk54 PTAL

jtk54:
LGTM

install/first_google_boot.sh
```diff
@@ -249,26 +249,43 @@ function extract_spinnaker_gcr_credentials() {
...
+      clear_instance_metadata "gcr_account"
+    fi
+
+    echo "Extracting GCR credentials for email $gcr_account"
     gcloud iam service-accounts keys create $config_path --iam-account=$gcr_account
 
     if [[ -s $config_path ]]; then
       echo "Extracted GCR credentials to $config_path"
-
+
```
ewiseblatt: extra spaces

install/first_google_boot.sh
```diff
@@ -249,26 +249,43 @@ function extract_spinnaker_gcr_credentials() {
...
+  if [ -n "$gcr_enabled" ] || [ -n "$gcr_account" ]; then
+    echo "GCR enabled"
+
+    if [ -z "$gcr_account" ]; then
+      gcr_account=$(gcloud iam service-accounts list --filter="displayName:'Compute Engine default service account'" --format="value(email)")
+      clear_instance_metadata "gcr_account"
+    fi
+
+    echo "Extracting GCR credentials for email $gcr_account"
```
ewiseblatt: if gcr_is_enabled and gcr_account is empty, this will still execute.
The log will look wrong. Will gcloud do the right thing?

install/first_google_boot.sh
```diff
@@ -249,26 +249,43 @@ function extract_spinnaker_gcr_credentials() {
...
 
+  local gcr_enabled=$(get_instance_metadata_attribute "gcr_enabled")
   local gcr_account=$(get_instance_metadata_attribute "gcr_account")
-  if [ -n "$gcr_account" ]; then
-    # This downloads the JSON key for the specified service account
+
+  if [ -n "$gcr_enabled" ] || [ -n "$gcr_account" ]; then
+    echo "GCR enabled"
+
+    if [ -z "$gcr_account" ]; then
```
ewiseblatt: why -n above and -z here?

install/first_google_boot.sh
```diff
@@ -249,26 +249,43 @@ function extract_spinnaker_gcr_credentials() {
...
   chown -R spinnaker:spinnaker $(dirname $config_path)
 
+  local gcr_enabled=$(get_instance_metadata_attribute "gcr_enabled")
   local gcr_account=$(get_instance_metadata_attribute "gcr_account")
-  if [ -n "$gcr_account" ]; then
-    # This downloads the JSON key for the specified service account
+
+  if [ -n "$gcr_enabled" ] || [ -n "$gcr_account" ]; then
+    echo "GCR enabled"
+
```
ewiseblatt: How unusual would it be to set a gcr_account but not enable it?
I assume this || is to go the other way around and use the default account.
Could you use an empty account to imply the default and gcr_enabled = ! [-z gcr_account] ? Just asking.

lwander: The idea is setting a gcr_account implicilty means you intend to use GCR. Not setting gcr_account, but setting gcr_enabled means you use the default compute account.

gcr_enabled = ! [ -z gcr_account ] only triggers when both flags are set, which is not the behavior we want.

install/first_google_boot.sh
lwander: -n is true when the string is non-empty, -z is true when the string is empty.

install/first_google_boot.sh
lwander: That's fine - above if gcr_account isn't set it will be set to Compute Engine default service account, which exists as long as you have an instance running.

PR #1173, opened 2016-09-27
(kubernetes) allow zone & project for cluster to be specified

lwander:
@ewiseblatt PTAL

Note the behavior is unaffected when no zone or project is specified.

### DIFF
--- a/install/first_google_boot.sh
+++ b/install/first_google_boot.sh
@@ -224,11 +224,22 @@ function extract_spinnaker_kube_credentials() {
   chown -R spinnaker:spinnaker $(dirname $config_path)
 
   local kube_cluster=$(get_instance_metadata_attribute "kube_cluster")
+  local kube_project=$(get_instance_metadata_attribute "kube_project")
+  local kube_zone=$(get_instance_metadata_attribute "kube_zone")
+
+  if [ -z "$kube_project" ]; then
+    kube_project=$MY_PROJECT
+  fi
+
+  if [ -z "$kube_zone" ]; then
+    kube_zone=$MY_ZONE
+  fi
+
   if [ -n "$kube_cluster" ]; then
-    echo "Downloading credentials..."
+    echo "Downloading credentials for cluster $kube_cluster in project $kube_project in zone $kube_zone..."
     export KUBECONFIG=$config_path
     gcloud config set container/use_client_certificate true
-    gcloud container clusters get-credentials $kube_cluster --zone $MY_ZONE
+    gcloud container clusters get-credentials $kube_cluster --zone $kube_zone --project $kube_project
 
     if [[ -s $config_path ]]; then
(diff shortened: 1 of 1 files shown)


install/first_google_boot.sh
```diff
@@ -224,11 +224,22 @@ function extract_spinnaker_kube_credentials() {
...
+
+  if [ -z "$kube_project" ]; then
+    kube_project=$MY_PROJECT
+  fi
+
+  if [ -z "$kube_zone" ]; then
+    kube_zone=$MY_ZONE
+  fi
+
   if [ -n "$kube_cluster" ]; then
```
ewiseblatt: what happens when you leave something out, misspell it or cluster isnt there? I think this will prevent the startup script from running.  I guess that is ok, except you should run "first_google_boot" after fixing the problem and, if the problem is in the metadata, you have to trash the VM and start over again. So I think as you add more parameters here, error recovery becomes more significant to address (and more complex).

With the spinnaker-local, if you mess it up, it will still extract out to disk, spinnaker will fail to start (or configure incorrectly) but you could edit the file and restart spinnaker using normal/standard practices.

Can you think through error scenarios and how to recover from them? I dont know how important that is. I assume that in practice use of these fields would be scripted so error rates might not be that high. But it would be easy to leave out a project/zone or add a wrong one when supplying script parameters. And C2D doesnt provide the means to check it. I think that will cause the script to abort early in a non-recoverable way. Although you could fix it manually, you'd never be able to complete the other parts of the script. That feels bad.

ewiseblatt: Or maybe just disable -e and allow these comands to fail without failing the script.
Think it though. I havent.

lwander: Well these aren't fields that can be supplied to spinnaker.yml, since they're needed before spinnaker.yml is ever read/interpreted, so that's not an option here unfortunately.

Also, the "immutable infrastructure" vision we are pushing encourages trashing VMs, even for config changes, so I'm not too worried about that.

From a C2D perspective, presenting the fact that the cluster needs to exist in the supplied project & zone seems to justify failing early.

lwander: If we see a lot of activity in Slack stating that C2D is unusable, or confusing we can go back and revisit this. But this here seems like a step in the right direction.

ewiseblatt: spinnaker.yml is just a file. Anyone can read it at any time. The thing producing the spinnaker.yml can also be the thing suppling this metadata. It's just a packaging.

The immutable infrastructure argument is a good point that probably negates this. At least for production use cases. For development, I anticipate it to be annoying. Like creating a VM then realizing you left a scope out and having to trash it and create another one (and wait for  the creation, spinup, etc).

ewiseblatt:
LGTM

PR #1174, opened 2016-09-27
Migrate citest

ewiseblatt:
@aglover this is the proposed migration of integration tests from github.com/google/citest/spinnaker to testing/citest (leaving room in that directory for other tests, such as on the UI)

@ashleykasim can you verify this on your end?
I rewrote the README. I havent published citest into a pip package yet because I havent had a chance to look into the maintenance process for it. It seems straightforward to install from a cloned citest repo for now so that's how it will be initially.


ewiseblatt:
@aglover In the original branch it had the individual log history and date for each entry. Now it appears that they are squashed into the first entry. That's not what I was going after. I'm not sure if I did this wrong, or it is a github-ism, or it is just showing up this way in the PR.

ewiseblatt:
The commit history looks right (although there were places where it referred to previous PR #s that dont resolve to the right PR. I dont think I need to worry about that but will see if it is easy to address.

ewiseblatt:
This is out of date. I'm going to push an update.

PR #1181, opened 2016-09-28
(gcr) adjust search for gcr email to work on GCE VMs

lwander:
getopts on the GCE vms parses `"` as a token, rather than the start/end of a string causing the filter command to gcloud to fail. This is a workaround. @ewiseblatt @jtk54 PTAL

### DIFF
--- a/install/first_google_boot.sh
+++ b/install/first_google_boot.sh
@@ -267,7 +267,9 @@ function extract_spinnaker_gcr_credentials() {
     echo "GCR enabled"
 
     if [ -z "$gcr_account" ]; then
-      gcr_account=$(gcloud iam service-accounts list --filter="displayName:'Compute Engine default service account'" --format="value(email)")
+      # the @developer address is reserved for the project's special service
+      # compute engine default service account.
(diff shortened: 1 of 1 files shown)


jtk54:
I'd document the workaround in a comment, then LGTM.

install/first_google_boot.sh
```diff
@@ -267,7 +267,9 @@ function extract_spinnaker_gcr_credentials() {
     echo "GCR enabled"
 
     if [ -z "$gcr_account" ]; then
-      gcr_account=$(gcloud iam service-accounts list --filter="displayName:'Compute Engine default service account'" --format="value(email)")
+      # the @developer address is reserved for the project's special service
+      # compute engine default service account.
+      gcr_account=$(gcloud iam service-accounts list --format='value(email)' | grep developer -m 1)
```
ewiseblatt: Does it matter that we are just picking a random one? Could it matter which (e.g. permissions may vary)

lwander: The docs say there is only one `@developer.gserviceaccount` - but I should probably strengthen the regex

lwander: You know actually I can get the exact email based on the active project number.

PR #1191, opened 2016-09-30
(consul) this captures the GCE setup side for getting consul to work with Spinnaker

lwander:
@duftler @skim1420 PTAL


skim1420:
lgtm

PR #1192, opened 2016-10-02
Fix readme command

matejkramny:
Btw, it also created the service account in the wrong project.


anotherchrisberry:
cc @lwander

lwander:
@matejkramny when you say wrong project, what project did you expect it to be created in?

matejkramny:
I had a few projects, the commands picked the first one available. You should either specify the project ID (starting with the first command) or be presented with a choice

PR #1201, opened 2016-10-05
Updating config for Openstack.

varikin:


config/clouddriver.yml
```diff
@@ -69,6 +69,29 @@ kubernetes:
       dockerRegistries:
         - accountName: ${providers.kubernetes.primaryCredentials.dockerRegistryAccount}
 
+openstack:
+  enabled: ${providers.openstack.enabled:false}
+  accounts:
+    - name: ${providers.openstack.primaryCredentials.name}
+      authUrl: ${providers.openstack.primaryCredentials.authUrl}
+      username: ${providers.openstack.primaryCredentials.username}
+      password: ${providers.openstack.primaryCredentials.password}
```
ewiseblatt: this is ok, but would be more secure to put these external if possible.

config/default-spinnaker-local.yml
```diff
@@ -121,6 +121,21 @@ providers:
...
+  openstack:
+    # For more information on configuring Openstack clusters, see
+    # TODO add link to docs when available
+
+    enabled: ${SPINNAKER_OPENSTACK_ENABLED:false}
+    primaryCredentials:
+      name: my-openstack-account
+      authUrl: ${SPINNAKER_OPENSTACK_AUTH_URL}
+      username: ${SPINNAKER_OPENSTACK_USERNAME}
+      password: ${SPINNAKER_OPENSTACK_PASSWORD}
```
ewiseblatt: how do these get populated?
Does openstack have a mechanism to store passwords externally so you can give a file reference here instead? Upper case is intended to be an environment variable. Usually these are in /etc/default/spinnaker. You definitely dont want passwords in there.

config/default-spinnaker-local.yml
```diff
@@ -121,6 +121,21 @@ providers:
...
       passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE}
 
+  openstack:
+    # For more information on configuring Openstack clusters, see
+    # TODO add link to docs when available
+
+    enabled: ${SPINNAKER_OPENSTACK_ENABLED:false}
+    primaryCredentials:
+      name: my-openstack-account
+      authUrl: ${SPINNAKER_OPENSTACK_AUTH_URL}
```
ewiseblatt: The default values are empty. Here you are overriding them. Did you mean to have a default?

config/spinnaker.yml
```diff
@@ -294,3 +294,18 @@ providers:
...
+    # TODO Added doc link when available
+
+    enabled: false
+    primaryCredentials:
+      name: my-openstack-account
+      authUrl:
+      username:
+      password:
+      projectName:
+      domainName:
```
ewiseblatt: In default-spinnaker-local you gave "Domain" as a default value. Did you mean to have that here?

config/spinnaker.yml
```diff
@@ -294,3 +294,18 @@ providers:
...
+      regions:
```
ewiseblatt: Is empty ok? Could there be default values?

config/spinnaker.yml
```diff
@@ -294,3 +294,18 @@ providers:
...
+
+  openstack:
+    # For more information on configuring Openstack, see
+    # TODO Added doc link when available
+
+    enabled: false
+    primaryCredentials:
+      name: my-openstack-account
+      authUrl:
+      username:
```
ewiseblatt: If you are using environment variables in default-spinanker-local, you can put them here instead (e.g. projectName: ${SPINNAKER_OPENSTACK_PROJECT:}
That would be safer because really spinnaker-local should just contain the minimal overrides. I'd like to get rid of default-spinnaker-local or at least prune it down to just the minimal things that *have* to be overwritten. It was intended to be a minimal summary but minimal isnt so small anymore so might as well defer to spinnaker.yml. You can leave what you have up there for consistency with everything else, but prepare for it not being there.

ewiseblatt:
Do you use InstallSpinnaker.sh? You might want to accomodate some initial configuration there. For example populating etc/default/spinnaker with runtime defaults (e.g. region, projectName or whatever else you might be able to infer).

You can also add static constants to etc/default/spinnaker if appropriate (ok if there are none).

config/default-spinnaker-local.yml
varikin: The Openstack driver that we are using does not have a way to put the passwords in a file.

config/spinnaker.yml
varikin: Not really. Openstack admins decide on project & regions. Domain is the only thing that has a reasonable default.

config/spinnaker.yml
varikin: OK

varikin:
I don't use InstallSpinnaker.sh.  I will look into that.

config/clouddriver.yml
varikin: Nevermind, I saw your other comment about passwords in an external file.

config/default-spinnaker-local.yml
varikin: Many of these don't have worthwhile default value. They are very environment specific.

config/default-spinnaker-local.yml
ewiseblatt: ok. there's precedent for them being here (e.g. jenkins). FYI azure and cf had started out with them here but moved them out. They might have added mechanisms for it or maybe their platforms had them and they werent using them initially. Having passwords here makes it easy to forget and do something like check the file into a revision control system or forget to chmod 600 (or 400) the file.

varikin:
@ewiseblatt I am looking through `InstallSpinnaker.sh`. I am happy to add OpenStack to that script, but my concern is that Aws and GCE look like they support credential files in default locations, so this script doesn't need to handle credentials at all, and like you mentioned, passwords then don't get added to `/etc/defaults/spinnaker`. I don't want to have this script only partially configure OpenStack support either.

varikin:
@ewiseblatt Have you had a chance to look at my latest comments and changes?

config/default-spinnaker-local.yml
```diff
@@ -121,6 +121,21 @@ providers:
...
+    # TODO add link to docs when available
+
+    enabled: ${SPINNAKER_OPENSTACK_ENABLED:false}
+    primaryCredentials:
+      name: my-openstack-account
+      authUrl:
+      username:
+      password:
+      projectName:
+      domainName:
```
ewiseblatt: This is overriding the defaults in spinnaker.yml with the empty string. And all the others are also overriding the environment variables in the spinnaker.yml with empty. I think you either want nothing here or to copy what you have in spinnaker.yml.

For now copying would probably be consistent with all the other platforms, but I'd like to get rid of this file or at least only put the absolute minimal required things here because they have no meaningful default values.

config/spinnaker.yml
```diff
@@ -323,12 +323,28 @@ providers:
...
+openstack:
+    # This default configuration uses the same environment variable names set in
+    # the OpenStack RC file. See
+    # http://docs.openstack.org/user-guide/common/cli-set-environment-variables-using-openstack-rc.html
+    # for details on the OpenStack RC file.
+    enabled: false
+    primaryCredentials:
+      name: my-openstack-account
+      authUrl: ${OS_AUTH_URL}
+      username: ${OS_USERNAME}
```
ewiseblatt: On one hand I see this seems to be standard environment variable naming for open stack based on the link you provided. So this makes sense when you are running on openstack. I'm not as sure if this is ideal if you were running elsewhere and deploying into openstack. I guess it is ok for now -- it's easy enough to resolve and seems reasonable when deploying on openstack which is your primary concern. I'm just asking.

I'm still not entirely sure about how the password/username work. Where are they set in practice? In the other environment variable precedents here, they are set in /etc/default/spinnaker. However that file is not protected (it is 644 where passwords would suggest no more than 600). We'll have to take some additional actions to protect that file. Or do this come from some other user configuration where these variables are set outside the scope of spinnaker?

ewiseblatt:
Overall I think this is ok -- it isnt going to break anyone else. I'm just trying to understand how I'd set the password for openstack if I were running spinnaker outside openstack and how that is protected.

varikin:
My expectation would that the `OS_PASSWORD` would be set in the environment externally from Spinnaker, such as via consul and vault or placed in a file that limited access controls which gets sourced.

ewiseblatt:
where and how would it get sourced?
The microservices in spinnaker source /etc/default/spinnaker already.
I presume this would go before that. Just asking where and how that works.

PR #1203, opened 2016-10-05
(consul) added sample consul service

lwander:


lwander:
@duftler @skim1420 PTAL

consul/hello-consul/src/main.rs
```diff
@@ -0,0 +1,28 @@
+extern crate iron;
```
duftler: bah

PR #1222, opened 2016-10-11
publish InstallSpinnaker.sh during release process

cfieber:
@dstengle @skorten FYI


cfieber:
@spinnaker/netflix-reviewers @spinnaker/google-reviewers FYI

(we've had issues with InstallSpinnaker.sh being wildly out of date, this will publish it every time we cut a release of this repo)

PR #1237, opened 2016-10-14
Add environment variable references to spinnaker.yml

ewiseblatt:
@lwander can you test this against kubernetes? I forgot to do that and woudnt hurt for an independent test.

@danielpeach 

I'm pretty sure this fixes the breakage. 

This is a pretty big PR for a conceptually small fix because I tried to be comprehensive and fix the issue on multiple levels. There are many different deployment scenarios.

default-spinnaker-local added a number of environment variables for provider configuration.
These are set in /etc/default/spinnaker. When we dont use a spinnaker-local, these dont get
picked up (from spinnaker.yml) so are ignored. Here we introduce the default environment
variable bindings in the spinnaker.yml itself.

I write out some other variables, like the json path that was written.
I threw in some additional stackdriver changes while here. These arent used yet, but will likely be tomorrow.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -281,15 +281,3 @@ services:
         enabled: false
         token: # the API token for the bot
         botName: # the username of the bot
-
-spectator:
-  applicationName: ${spring.application.name}
-  endpoint:
-    enabled: false
-    prototypeFilter:
-      path:
- 
-  stackdriver:
-    projectName: ${providers.google.primaryCredentials.project}
-    credentialsPath: ${providers.google.primaryCredentials.jsonPath}
-    enabled: false
--- a/config/spinnaker.yml
+++ b/config/spinnaker.yml
@@ -231,9 +231,10 @@ providers:
     # For more information on configuring Amazon Web Services (aws), see
     # http://www.spinnaker.io/v1.0/docs/target-deployment-setup#section-amazon-web-services-setup
 
-    enabled: false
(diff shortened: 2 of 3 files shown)


ewiseblatt:
I see primaryAccountName isnt used anymore in the microservice configs.
If so, I'm going to take it out since configuring it is complex.
The tests use it so I'll need to refactor them first. I'd like to do that before submitting this PR if I can get it done in the morning.

lwander:
Works for me, but I didn't package this into a .deb & run first_google_boot to fully validate. I validated by placing these the config files on a VM with a correctly configured /etc/default/spinnaker (produced by our nightly Jenkins).

ewiseblatt:
ok. I already did the end to end plumbing validation, but only had gce/aws
credentials in that deployment.

FWIW, the scopes required for kubernetes is why I leave it out of my
testing because I dont want the scope creep causing me to miss scope errors
of which I am sensitive since I'm working on new APIs and hit those kinds
of errors. If I could just pass in the existing config and not need the
scopes to create a certificate...

lwander:
Ok, in that case let me add the kubeconfig parameter back to the install
metadata so you can just push the credentials to the VM for future use

InstallSpinnaker.sh
```diff
@@ -734,6 +734,7 @@ else
   echo "Not enabling a cloud provider"
 fi
 
+
```
danielpeach: Seems extraneous

danielpeach:
LGTM

PR #1238, opened 2016-10-14
Specify credentials for testing front50 GCS separate from Compute cre…

ewiseblatt:
@jtk54 When merging this we'll need to check jenkins configuration to pass in a --gcs_json_path.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -129,7 +129,6 @@ services:
     # directly under 'services'.
     protocol: http    # Assume all spinnaker subsystems are using http
     host: localhost   # Assume all spinnaker subsystems are on localhost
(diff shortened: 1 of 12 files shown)


ewiseblatt:
I'm going to add the refactoring to remove primaryAccountName to this PR.

ewiseblatt:
@jtk54 I'll do them as separate PRs. This is ready as is.

ewiseblatt:
Changed the theme of this PR to removing the primaryAccountName afterall.

@lwander I removed references from experimental as well. Not sure why there are a lot of config files there, I didnt look at the details but those files are likely quite stale. Either we should discuss a different way to manage them or you should look into purging them.

lwander:
You're right about the config files in experimental, but they'll be purged soon<sup>tm</sup>.

PR #1239, opened 2016-10-14
(kubernetes) provide first_google_boot with kube_config

lwander:
If the contents of your `~/.kube/config` file are passed as the instance
metadata parameter `kube_config`, and you do not specify the GKE cluster you
wish to manage with `kube_cluster`, that file's contents will be used as the
credentials to manage your kubernetes cluster with Spinnaker.

@ewiseblatt PTAL

### DIFF
--- a/install/first_google_boot.sh
+++ b/install/first_google_boot.sh
@@ -44,7 +44,7 @@ SPINNAKER_SUBSYSTEMS="spinnaker-clouddriver spinnaker-deck spinnaker-echo spinna
 
 # By default we'll tradeoff the utmost in security for less startup latency
 # (often several minutes worth if there are any OS updates at all).
-# Note that this is only the initial boot off the image in this script, which is 
+# Note that this is only the initial boot off the image in this script, which is
 # intended to configure spinnaker from metadata, which is typically an adhoc or trial
 # scenario. A production instance would probably be using a different means to configure
 # spinnaker, or might be baked in, or might have different security policies not requiring
@@ -148,7 +148,7 @@ function extract_spinnaker_google_credentials() {
       chown spinnaker $json_path
       echo "Extracted google credentials to $json_path"
     else
-       rm $json_path
+      rm $json_path
     fi
     write_default_value "SPINNAKER_GOOGLE_PROJECT_CREDENTIALS_PATH" "$json_path"
   else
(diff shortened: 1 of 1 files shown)


install/first_google_boot.sh
```diff
@@ -222,17 +222,38 @@ function extract_spinnaker_kube_credentials() {
...
     fi
-    write_default_value "SPINNAKER_KUBERNETES_ENABLED" "true"
+
+    # Safely return since we have attempted to configure Kubernetes with the
+    # provided metadata.
+    return
   else
     clear_instance_metadata "kube_cluster"
   fi
+
```
lwander: Logically, you can't specify both a cluster & a config file since the credentials for one would overwrite the other.

ewiseblatt: Ah, I missed that return statement. That looks problematic. Minimally it would leave the other credentials hanging in the metadata. If you overspecify then you should know that. You are giving the cluster preference. Why would you do that since it is more work? Significantly more stepping back when you consider that you need additional scopes and are also forcing an upgrade to gcloud in order to use those APIs (which is affecting boot times -- we should move that to the bake).

What's the harm in giving precedence the other way around? You can interpret the cluster parameter as being available if needed and turns out that it isnt currently needed for anything else. So the implementation is superfluous but the interface is well defined and not overspecified (though could be surprising if the config was from a different cluster than the one you gave, but definitions still feel more robust).

lwander: Yeah that makes sense - should we move the gcloud update into this function when no kube_config is provided but kube_cluster is specified?

ewiseblatt:
I was thinking that, unless it is needed otherwise.
Last night I was considering how we could add it into the packer image bake
at least, but the version is coded here (which looks like another
maintainence issue) so I bailed on attempting that for now. It's timely
that it came up today (or not since it was on my mind).

lwander:
What would be wrong with removing the coded version here and putting it into a packer template?

ewiseblatt:
Can you assume that all images came through that template?

ewiseblatt:
Perhaps we could justify upgrading gcloud and all its components in the
packer template regardless.
That would make this a no-op in practice (or will it downgrade)?

lwander:
It will downgrade, so I'd prefer to remove this gcloud update here. Who are you worried about affecting with the version change?

ewiseblatt:
nobody in particular. just gcloud sometimes has breaking apis. or maybe
that's just alpha beta.
I'm heading off to physical therapy. Dont know when I'll be back.
This change looks ok in principal other than the precedence things we
discussed.

lwander:
Ok, I'll fix the precedence and dangling credentials. We can talk later about the gcloud updates.

lwander:
Comments addressed

install/first_google_boot.sh
```diff
@@ -207,6 +207,10 @@ function extract_spinnaker_kube_credentials() {
   local kube_project=$(get_instance_metadata_attribute "kube_project")
   local kube_zone=$(get_instance_metadata_attribute "kube_zone")
 
+  clear_instance_metadata "kube_cluster"
```
ewiseblatt: I like to clear them at the end so that if you encounter a failure you can fix it then run first_google_boot again and it will automatically pick up where it left off. If you clear this first and hit an error downloading gcloud below (for example) then there is no recovery path because you've destroyed the evidence. If you want until later then next time through it will see the attributes and attempt [again] to process them.

install/first_google_boot.sh
```diff
@@ -215,21 +219,42 @@ function extract_spinnaker_kube_credentials() {
     kube_zone=$MY_ZONE
   fi
 
+  if clear_metadata_to_file "kube_config" $config_path; then
```
ewiseblatt: This function still feels complicated the way it flows. It appears this function is responsible for determining if kubernetes should be enabled or not and, if so, configuring it. This involves three things:
  0) determining if kubernetes should be enabled or not
if so then:
  1) writing the credentials to .kube/config
  2) writing the SPINNAKER_KUBERNETES_ENABLED flag to /etc/default/spinnaker

(0) is inferred by having credentials (which happens in (1))
(1) either gets explicit credentials from "kube_config" or creates its own credentials by interacting with the cluster specified by "kube_(project|cluster|zone)".

So the flow should read more like
   if (have_kube_credentials) {
        write ENABLED
   }
where have_kube_credentials is a helper function that:
    if (have "kube_config") {
        write config_path
        clear kube_(cluster|zone|project) since we dont need them
        return true
    }
   if can generate credentials from others
        return true
   return false

You can unroll this without the helper function but the flow is similar, especially isolating the two different ways to produce credentials from everything else, which is the same regardless of how you got the credentials.

Especially since you dont use the other parameters. If you were configuring the actual kubernetes cluster, zone, project even if you had kube_config then the flow might look different. So I think it's that much more important to clarify.

lwander: Right now I only enable Kubernetes if the credentials are successfully extracted to avoid producing noise in clouddriver when the culprit/cause is this script. But I don't feel strongly about this approach over yours.

install/first_google_boot.sh
lwander: That's a good point, I can fix that. I originally took the stance of "immutable infrastructure, re-provision" on error but this makes it a little easier to use.

install/first_google_boot.sh
ewiseblatt: I was doing the same -- enable kube if you were able to extract out a path.
Since you mentioned it, if you specified kube but could not extract a path, that should manifest itself somewhere as a visible error. Either failure to startup (crash here) or from error logs in clouddriver saying kube is enabled but you dont have vaild credentials.

install/first_google_boot.sh
```diff
@@ -215,21 +219,42 @@ function extract_spinnaker_kube_credentials() {
...
+    return
+  else
+    clear_instance_metadata "kube_config"
+  fi
+
   if [ -n "$kube_cluster" ]; then
     echo "Downloading credentials for cluster $kube_cluster in project $kube_project in zone $kube_zone..."
     export KUBECONFIG=$config_path
     gcloud config set container/use_client_certificate true
     gcloud container clusters get-credentials $kube_cluster --zone $kube_zone --project $kube_project
```
ewiseblatt: FYI, the gcloud call doesnt work to reach into another project because you dont have the credentials for it.
If the google_credentials passed in are also for that project, then you can register them into an account with gcloud and then use that --account flag here and I think it would work. To give the credentials to gcloud, see https://cloud.google.com/sdk/gcloud/reference/auth/activate-service-account

install/first_google_boot.sh
lwander: Not quite, you're enabling Kubernetes, then attempting to write credentials (which can fail due to permissions or gcloud errors). This leaves the spinnaker config in an inconsistent state.

install/first_google_boot.sh
lwander: Ah I agree with the first point.

I've explained that startup script warning in an internal bug - it's related to package management difficulties between apt-get and gcloud that are being worked on.

install/first_google_boot.sh
lwander: Take a look here :https://github.com/spinnaker/spinnaker/pull/1239/files/974284040f4e9941aeb38b4793b22956afa2a168#diff-6fe926a9170979345c6d40b1d595015aR253 
the enabled bit is only set after the call the gcloud produces the credentials file.

ewiseblatt:
I cant find the context of this comment in the log.
The block you refer to is functionally correct as you describe.
I'm concerned about maintainability and thought I was simplifying how to
think about it.

Sorry I'm not around. We can do a hangout to talk about it.

lwander:
PTAL @ewiseblatt 

I'm leaning against allowing multi-project config by cluster name. It's enabled by passing a kubeconfig, which should be enough. For the sake of click-to-deploy, it would require another field (the current GCE credentials don't need to correspond to your GKE cluster's project), and just add more error cases.

ewiseblatt:
I dont understand the multi-project config message. Was this pertaining to
managing kubernetes in a different cluster than spinnaker? I wasnt
suggesting you need to do that here. Only that you cannot do that, so the
project becomes superfluous because it is either your project or I dont
think you'd have access to it without setting up credentials for the
gcloud/kubectl commands.

lwander:
No, I was referring to providing `"kube_cluster"` (for GKE), and `"kube_project"` for a project the default service account doesn't have access to.

ewiseblatt:
LGTM

PR #1249, opened 2016-10-18
Linking Rosco to Orca and Gate in docker compose env.

varikin:
In testing out baking, found that the docker-compose.yml was missing these links. I have been using this change for a couple weeks now.

### DIFF
--- a/experimental/docker-compose/docker-compose.yml
+++ b/experimental/docker-compose/docker-compose.yml
@@ -49,6 +49,7 @@ gate:
     - SERVICES_FRONT50_HOST=front50
     - SERVICES_IGOR_HOST=igor
     - SERVICES_ORCA_HOST=orca
+    - SERVICES_ROSCO_HOST=rosco
   image: quay.io/spinnaker/gate:master
   links:
     - redis
@@ -57,6 +58,7 @@ gate:
     - front50
     - igor
     - orca
+    - rosco
   ports:
     - "8084:8084"
 igor:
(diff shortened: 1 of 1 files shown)


experimental/docker-compose/docker-compose.yml
```diff
@@ -49,6 +49,7 @@ gate:
     - SERVICES_FRONT50_HOST=front50
     - SERVICES_IGOR_HOST=igor
     - SERVICES_ORCA_HOST=orca
+    - SERVICIES_ROSCO_HOST=rosco
```
ewiseblatt: SERVICES

ewiseblatt:
@lwander can you have a look. I'm not familiar with as much context.

experimental/docker-compose/docker-compose.yml
varikin: Bah! I had copy my changes I have been running on a demo instance of spinnaker so obviously I mistyped it as when trying to created this PR :(

lwander:
I know very very little about docker compose. Maybe @gbougeard could take a peak?

gbougeard:
I never used rosco with docker-compose because it did not work due to missing links. I thought it was already fixed since Bake stage was available (at least its UI correctly displayed).
@varikin did you experiment circular references between links or does it boot correctly with the links you added?

varikin:
@gbougeard The missing links that this PR adds prevented Gate and Orca from connecting to Rosco. You are correct that the UI showed the bake stage, but it failed when attempting to start one. This does not introduce circular references since Rosco doesn't call back to anything that I have seen. I have a demo environment running with this modified docker-compose setup for several weeks now without a problem.

varikin:
@gbougeard, @lwander, @ewiseblatt  Can this be merged?

Issue #1058, opened 2016-08-11
docker provider configuration not merged correctly


colemickens:
**Cloud Provider**: Kubernetes

**Environment**: Spinnaker running on an Azure VM pointed at a Kubernetes cluster

**Feature Area**: Configuration

**Description**

*Expected:*

* I can easily add Docker Hub credentials to the `spinnaker-local.yml` and reference them from the `kubernetes` provider section.

*Actual:*

* a couple hours spent only to realize that I had the fields right, they just weren't being merged together properly

**Steps to Reproduce**

* (in `spinnaker-local.yml`) try to add a new docker account, then reference it from the kubernetes section, it will complain it can't find it

* (in `spinnaker-local.yml`) try to remove the default "repository" from the primary credentials. The "library/nginx" image still gets set as the requested one

* try to remove the repository field, in *any* way, including direct edits to `clouddriver.yml`, pointed at Docker Hub. It seems like it access the images... but when the the web UI queries for containers in the autocomplete, an array of empty objects is returned

**Workaround**

* for my needs, I just wanted to show it working, so I picked a repository and hard coded it and the docker credentials into `clouddriver.yml`

lwander:
Hey @colemickens, could you post the docker registry section of your `spinnaker-local.yml` config that failed?

JargoonPard:
I have validated that these same values placed into clouddriver.yml will work but not when they only exist in spinnaker-local.yml

  dockerRegistry:
    # For more information on configuring Docker registries, see
    # http://www.spinnaker.io/v1.0/docs/target-deployment-configuration#section-docker-registry

    # NOTE: Enabling dockerRegistry is independent of other providers.
    # However, for convienience, we tie docker and kubernetes together
    # since kubernetes (and only kubernetes) depends on this docker provider
    # configuration.
    enabled: true

    primaryCredentials:
      name: my-docker-registry
      address: https://index.docker.io/
      repository: larryguger/web-lab
      username: larryguger
      # A path to a plain text file containing the user's password
      #passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE}
      password: <removed>

JargoonPard:
What shows up for a list of accounts when I try to create a new application is "my-docker-registry" but not "my-kubernetes-account".  Below is the K8S config section.
  kubernetes:
    # For more information on configuring Kubernetes clusters (kubernetes), see
[...]

PR #1251, opened 2016-10-19
Adds blank defaults, even when disabled

ttomsu:
Unresolvable ENV vars break /resolvedEnv endpoint, so I added blank string defaults.

@lwander @varikin PTAL


lwander:
Having those colons break config: https://github.com/spinnaker/spinnaker/issues/1058

This should be fixed in /resolveEnv

varikin:
Do you want the same in `default-spinnaker-local.yml`?

varikin:
@lwander would `${ENV_VAR:''} be any different in regards to #1058?

ttomsu:
Fixing in Kork. Closing this PR

PR #1262, opened 2016-10-24
L7 test deletes security group and updated health check.

jtk54:
The L7 test should not leak resources anymore. @ewiseblatt please review.

### DIFF
--- a/testing/citest/tests/google_http_lb_upsert_scenario.py
+++ b/testing/citest/tests/google_http_lb_upsert_scenario.py
@@ -226,7 +226,7 @@ class GoogleHttpLoadBalancerTestScenario(sk.SpinnakerTestScenario):
             jp.STR_EQ(self._get_bs_link(pm['defaultService']['name'])),
         'pathRules':  jp.LIST_MATCHES(path_rules_spec)
         }
-      
+
       url_map_clause_builder.contains_match({
           'name': jp.STR_EQ(self.__lb_name),
           'defaultService':
@@ -564,6 +564,37 @@ class GoogleHttpLoadBalancerTestScenario(sk.SpinnakerTestScenario):
 
     return st.OperationContract(
       self.new_post_operation(title='create security group',
+                              data=sec_group_payload, path='tasks'),
+      contract=builder.build()
+    )
+
+  def delete_security_group(self):
+    '''Deletes a security group.
+    '''
+    bindings = self.bindings
+    sec_group_payload = self.agent.make_json_payload_from_kwargs(
+      job=[
+        {
+          'cloudProvider': 'gce',
+          'credentials': bindings['SPINNAKER_GOOGLE_ACCOUNT'],
+          'regions': ['global'],
+          'securityGroupName': self.__lb_name + '-rule',
+          'type': 'deleteSecurityGroup',
+          'user': '[anonymous]'
+        }
+      ],
+      description='Delete a Security Group.',
+      application=self.TEST_APP
+    )
+
+    builder = gcp.GcpContractBuilder(self.gcp_observer)
+    (builder.new_clause_builder('Security Group Deleted',
+                                retryable_for_secs=30)
+     .list_resource('firewalls')
+     .excludes_path_value('name', self.__lb_name + '-rule'))
+
+    return st.OperationContract(
+      self.new_post_operation(title='delete security group',
                               data=sec_group_payload, path='tasks'),
       contract=builder.build()
     )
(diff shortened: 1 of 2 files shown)


ewiseblatt:
I want to look into something before commenting. And also I'm going to use this to test out my leak detetector configuration. I should get to this by the end of the day.

jtk54:
Ok, fine with me.

ewiseblatt:
Security group looks ok, but where is the health check created? Is that implicit when you changed it? How would the health check get reclaimed in practice?

jtk54:
The health check is created new when we update the name of the health check in Upsert. I'm re-inserting it in the load balancer during the certificate tests, and then the final delete will remove all the components included in the load balancer. If it's not clear I can add more detailed comments or shift this around a bit.

jtk54:
I updated the comments to be clearer, and added one into the actual test case enumeration. This works because we do the tests in a certain order, so I commented as such.

testing/citest/tests/google_http_lb_upsert_test.py
```diff
@@ -168,6 +168,8 @@ def test_l_add_cert(self):
     self.run_test_case(self.scenario.add_cert(self.__class__.FIRST_CERT,
                                               'add cert'))
 
+  # Note: We include the updated health check in this upsert description.
+  # If the order of tests 'm' and 'o' change, we will leak resources.
```
ewiseblatt: name the tests since letters can change.

jtk54:
The 'delete hc' is the 'delete load balancer' command. The load balancer delete deals with all the components in the load balancer, including health checks, backend services, forwarding rules, etc.

testing/citest/tests/google_http_lb_upsert_scenario.py
```diff
@@ -480,7 +480,17 @@ def update_port_range(self):
   def add_cert(self, certname, title):
     '''Add cert to targetHttpProxy to make it a targetHttpsProxy.
     '''
+    # Note: We add this health check to the upsert description because
+    # this health check would be orphaned otherwise. This allows the 'delete'
+    # test to delete all the resources we allocated.
```
ewiseblatt: I'm still hung up on this.
If before the healthCheck was unchanged, so why would it be orphaned? Now you have a different one, so what happens to the original? If anything I'd think now would lead to something orphaned. Please explain.

jtk54:
```python
  def test_c_upsert_min_lb(self): 
    self.run_test_case(self.scenario.upsert_min_load_balancer()) # => Use proto health check for all backend services.

  def test_d_upsert_full_lb(self):
    self.run_test_case(self.scenario.upsert_full_load_balancer()) # => Use proto health check for all backend services.

  def test_e_add_security_group(self):
    self.run_test_case(self.scenario.add_security_group())

  # Test letters f and n are reserved for another test derived from this one.

  def test_g_change_hc(self):
    self.run_test_case(self.scenario.change_health_check()) # => Create new, updated health check and use for all backend services. Old proto health check still exists in GCP and is orphaned.

  def test_h_change_bs(self):
    self.run_test_case(self.scenario.change_backend_service()) # => Re-use proto health check for all backend services. Now the updated health check is orphaned.

  def test_i_add_host_rule(self):
    self.run_test_case(self.scenario.add_host_rule()) # => Re-use proto health check for all backend services. Now the updated health check is orphaned.

  def test_j_update_host_rule(self):
    self.run_test_case(self.scenario.update_host_rule()) # => Re-use proto health check for all backend services. Now the updated health check is orphaned.

  def test_k_update_port_range(self):
    self.run_test_case(self.scenario.update_port_range()) # => Re-use proto health check for all backend services. Now the updated health check is orphaned.

  def test_l_add_cert(self):
    self.run_test_case(self.scenario.add_cert(self.__class__.FIRST_CERT, 'add cert')) # => Add the updated health check to just one of the backend services, so both the updated health check and the proto health check are included in the load balancer.

def test_m_change_cert(self):
    self.run_test_case(self.scenario.add_cert(self.__class__.SECOND_CERT,
                                              'update cert')) # => Add the updated health check to just one of the backend services, so both the updated health check and the proto health check are included in the load balancer.

  # Test letters f and n are reserved for another test derived from this one.

  def test_o_delete_lb(self):
    self.run_test_case(self.scenario.delete_http_load_balancer()) # => Delete all components currently associated with the load balancer.
 
```

In one of the upsert operations, we create a _new_ health check that is an 'updated' health check to the L7 load balancer. Because we replace _all_ instances of the old health check in the L7 load balancer, the old (and thus unused) health check is now orphaned in GCP.

We then switch back to using the original health check for all the health checks in the L7 for a few upserts, which then makes the 'updated' health check orphaned in GCP. Before this PR, we would no longer use the updated health check in any of the rest of the operations so it would leak. This PR now makes it so that we add the 'updated' health check back into the load balancer so we can delete it.

ewiseblatt:
My concern here is that I'd call this a spinnaker bug.
You shouldnt have to delete that. If spinnaker created it implicitly and it
is no longer needed, then spinnaker should delete it as well. Otherwise as
a normal user, you'd be leaking these as well.

jtk54:
Ping -- I now delete the orphaned health check with an explicit API call in the the test cleanup function.

PR #1268, opened 2016-10-26
Run apt-get update non-interactively as well

skim1420:
@duftler @ewiseblatt 

An update to sudo was hanging waiting for user input. This possibly opens us up to other updates squeezing into update, but is the only way I understand how to prevent updating to sudo hanging.


duftler:
LGTM2.

PR #1269, opened 2016-10-26
Edited references to $HOME/.spinnaker

ewiseblatt:
@ttomsu 
I went on a hunt to erradicate $HOME/.spinnaker/spinnaker-local references from production deployments but turned up empty handed. There are a few references not visible in this PR but they are strictly and explicitly when running from source, and that is still intentional.

These are other related cleanups I made during that expedition.

- eric

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -77,16 +77,23 @@ so is performed here.
 
 == Configure Spinnaker
 
-We will create a directory for Spinnaker configuration overrides, copy the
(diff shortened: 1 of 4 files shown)


ttomsu:
LGTM. My image must be old then.

ewiseblatt:
It would have to be really old. Really, really old.
If you still have it, could you point me to it?

PR #1273, opened 2016-10-28
Modified tests for front50 migration to v2 API

ewiseblatt:
@jtk54 
front50 is being refactored. When adam submits then we'll need these changes to complement it.

### DIFF
--- a/testing/citest/spinnaker_testing/spinnaker_test_scenario.py
+++ b/testing/citest/spinnaker_testing/spinnaker_test_scenario.py
@@ -75,6 +75,23 @@ class SpinnakerTestScenario(sk.AgentTestScenario):
                                        status_class=status_class)
 
   @classmethod
+  def new_patch_operation(cls, title, data, path, status_class=None):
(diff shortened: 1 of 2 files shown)


jtk54:
As far as I know, v1 will still be around for a while even after v2 is merged down and all the Google parts of Clouddriver won't cut over to the new version for a bit. Should we make a temporary separate test for v2 so we can validate that and still make sure v1 isn't broken?

ajordens:
`v1` applications were removed in clouddriver a month or so back ... as my merge this morning, they're also gone from `front50`. 

Is there something else you guys are aware of that was calling the old `/default/applications` endpoints?

jtk54:
Actually I'm thinking of the Orca v2 by Rob F... It think we're ok here.

jtk54:
LGTM

ewiseblatt:
Adam, I'm not aware of anything. I dont know what jacob was refering to.
I know v1 wont work anymore since I had to rewrite the tests for v2.

- eric

PR #1277, opened 2016-10-31
(google) first boot & codelab image installs apt fix

lwander:
There was a duplicate apt upgrade in first_google_boot, and we have to set noninteractive mode for the codelab setup otherwise we might get stuck during the update process.

### DIFF
--- a/google/codelab/provision_spinnaker_codelab_image.sh
+++ b/google/codelab/provision_spinnaker_codelab_image.sh
@@ -35,9 +35,14 @@ service apache2 stop
 
 set -e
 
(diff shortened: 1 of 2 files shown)


lwander:
@duftler PTAL

google/codelab/provision_spinnaker_codelab_image.sh
```diff
@@ -35,9 +35,11 @@ service apache2 stop
 
 set -e
 
+export DEBIAN_FRONTEND=noninteractive
+
 # update apt
 apt-get update
```
ewiseblatt: Can you document / clarify what this means?

PR #1278, opened 2016-11-01
(google) bump node version for bootstrap_dev

danielpeach:
@ewiseblatt or @ttomsu 

Deck has new build scripts that will not run with a node version <4. Deck's `build.gradle` and `Dockerfile` both use node 4.4.1.


ttomsu:
LGTM

PR #1279, opened 2016-11-01
Can't nest ${..${}} in config

lwander:
This won't fail to parse, but it won't resolve the nested ${} expression. I'm dropping the env reference, since it's more likely that `primaryCredentials` is correctly defined. We can talk about a better way to resolve which project handles stack driver, but I figure we should leave this in an intermediate, working state.

There is another fix to kork incoming separate from this that is actually causing everything on kork 1.77 to fail that has stackdriver enabled (so far clouddriver, echo & front50).

@ewiseblatt, @duftler PTAL

### DIFF
--- a/config/spinnaker.yml
+++ b/config/spinnaker.yml
@@ -231,8 +231,8 @@ services:
 
   stackdriver:
     enabled: ${SPINNAKER_STACKDRIVER_ENABLED:false}
-    projectName: ${SPINNAKER_STACKDRIVER_PROJECT_NAME:${providers.google.primaryCredentials.project}}
-    credentialsPath: ${SPINNAKER_STACKDRIVER_CREDENTIALS_PATH:${providers.google.primaryCredentials.jsonPath}}   
+    projectName: ${providers.google.primaryCredentials.project}
+    credentialsPath: ${providers.google.primaryCredentials.jsonPath} 
 
 
 providers:


config/spinnaker.yml
```diff
@@ -231,8 +231,8 @@ services:
 
   stackdriver:
     enabled: ${SPINNAKER_STACKDRIVER_ENABLED:false}
-    projectName: ${SPINNAKER_STACKDRIVER_PROJECT_NAME:${providers.google.primaryCredentials.project}}
```
cfieber: I *think* you could do this with `${SPINNAKER_STACKDRIVER_PROJECT_NAME:#{providers.google.primaryCredentials.project}}`

config/spinnaker.yml
```diff
@@ -231,8 +231,8 @@ services:
...
-    credentialsPath: ${SPINNAKER_STACKDRIVER_CREDENTIALS_PATH:${providers.google.primaryCredentials.jsonPath}}   
+    projectName: ${providers.google.primaryCredentials.project}
+    credentialsPath: ${providers.google.primaryCredentials.jsonPath}
```
ewiseblatt: The effect I was after was to have stackdriver write into the deployed spinnaker project, not the project spinnaker was managing.
The proposal would work, but would prefer finding a different way so taht when on GCE it would default to the "other" project presuming they are differe.t

ewiseblatt:
what's that other incoming kork fix? If that is cameron's, I think he
worked it out over the weekend.
At least he informed me and I validated it.
I had setup a VM for him to validate on, and I validated it again in the
same build so if you are still seeing something, let us know.

config/spinnaker.yml
lwander: I'm back at my desk now and playing around with all the different options - I'm guess the above won't work because it will resolve to that string literal `${...}` rather than the expansion of the expression, but I'll try that one too.

ewiseblatt:
That's what the \ was for.

lwander:
I was wrong, you can nest these expressions.

ewiseblatt:
Good. I thought I'd been taking advantage of it.
So what was the original problem?

lwander:
I found the cause of the problem in the config created for the codelab instance, I'll try and produce a fix which will make it easier to explain here.

ewiseblatt:
thanks.

lwander:
Hmm the fix won't actually reveal the problem as well as I had hoped. A few things came together that caused a lot of grief:

1. The  image published & staged for marketplace is a few versions behind the release debians, so when trying to generate an image for the codelab the spinnaker services were all updated.

2. Yesterday, when the codelab build originally failed, it had gotten stuck waiting for input on whether or not to override a packages configuration files, so I added that non-interactive flag. (See #1277)

3. The stackdriver config update was spread across several files. (See #1253)

Now what happened was every services config already had #1253's configuration, except for spinnaker. It's old configuration was silently kept the same (rather than overridden) meaning there were a few subtle errors in the config that manifested themselves in a strange way (complaining about duplicate `{` in the configuration properites of `StackdriverConfig.class`). 

My fix will be to simply put all spinnaker packages on hold during the codelab install script so we don't run into these update errors again. PR incoming soon.

@duftler & @jtk54 FYI (since you guys were interested in the resolution to this).

ewiseblatt:
Regarding stackdriver config, I hope I am doing it in a reasonable way. You
probably want it all or none, but took an individual approach like
everything else (i.e. there is a services.stackdriver that sets the system
properties for each individual servce. And FWIW the system properties are
actually spectator.stackdriver and not stackdriver).

lwander:
I think the stackdriver config was done great, it was just the new config
addition in conjunction with a bad update that caused the problem.

PR #1280, opened 2016-11-01
Disable cassandra by default for echo.

duftler:
Disable Cassandra and enable in-memory by default for echo.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -261,9 +261,9 @@ services:
   echo:
     # Persistence mechanism to use
     cassandra:
-      enabled: true
-    inMemory:
       enabled: false
+    inMemory:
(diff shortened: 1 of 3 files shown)


duftler:
@lwander or @danielpeach please review.

danielpeach:
LGTM

ewiseblatt:
this has production implications.
I'm not sure you want to do that.

duftler:
We discussed this offline. The changes only affect the default behavior of echo and not of front50. Our nightly images have also been produced with these settings in effect for some time already.

ewiseblatt:
The nightly images use InstallSpinnaker which runs
pylib/spinnaker/change_cassandra.sh to modify spinnaker.yml as if these
changes had been made.

PR #1281, opened 2016-11-01
(google) codelab should not update spinnaker subsystems

lwander:
See the motivation for this in #1279


lwander:
@duftler @ewiseblatt PTAL

PR #1290, opened 2016-11-04
(Openstack) Adding default config for user data files.

varikin:
(Openstack) Adding default config for user data files.

Default being, not set, but clouddriver.yml will read it from
spinnaker.yml if set.


varikin:
@spinnaker/reviewers PTAL

I will be following up with updated docs on readme.io for this config option.

PR #1297, opened 2016-11-11
refactored metric collector tool.

ewiseblatt:
@lwander 

Segregated stackdriver stuff into its own modules to make it easier to add another option.
Support text or html so command line tool has readable output.
Other cleanups while I'm here.

### DIFF
--- a/google/stackdriver_monitoring/metric_collector.py
+++ b/google/stackdriver_monitoring/metric_collector.py
@@ -1,21 +1,37 @@
 #!/usr/bin/env python
+
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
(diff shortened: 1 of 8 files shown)


google/stackdriver_monitoring/metric_collector.py
```diff
@@ -69,27 +85,17 @@ def get_options():
   return parser.parse_args()
 
 
-def process_command(command, spectator, stackdriver, options):
+def process_command(command, registry):
```
lwander: What is `registry` here?

google/stackdriver_monitoring/metric_collector_handlers.py
```diff
@@ -1,24 +1,47 @@
...
+from datetime import datetime
+
+
+def accepts_html(request):
+  if not hasattr(request, 'headers'):
+    return False
+  accept = request.headers.get('accept', None)
+  if not accept:
+    return None
+  return 'text/html' in accept.split(',')
```
lwander: This probably isn't sufficient if you look at all the ways your header can be formatted: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -0,0 +1,162 @@
...
+        self.num_ok = 0
+
+      def handle_batch_response(self, index_str, good, bad):
+        index = int(index_str)
+        if bad:
+          self.batch_response[index] = 'ERROR {0}'.format(cgi.escape(str(bad)))
+          logging.error(bad)
+        else:
+          self.num_ok += 1
+          if not good:
```
lwander: What's the difference between "bad", "ok", and "good" here?

ewiseblatt:
updated

PR #1298, opened 2016-11-12
Added control loop and helper functions for monitoring metrics.

ewiseblatt:
Added skeleton into metric_collector for externally monitoring metrics.

### DIFF
--- a/google/stackdriver_monitoring/metric_collector.py
+++ b/google/stackdriver_monitoring/metric_collector.py
@@ -17,21 +17,79 @@
 
 # pylint: disable=missing-docstring
 import argparse
+import collections
 import logging
 import logging.config
-import collections
+import socket
+import threading
+import time
 
 from http_server import (HttpServer, StdoutRequestHandler)
 
 import metric_collector_handlers as handlers
 import stackdriver_handlers as stackdriver_handlers
 
-from spectator_client import SpectatorClient
-from stackdriver_client import StackdriverClient
+import spectator_client
+import stackdriver_client
 
 
 HandlerDefinition = collections.namedtuple(
   'HandlersDefinition', ['handler', 'url_path', 'command_name', 'description'])
+
+
+class DummyMetricService(object):
+  def store(self, service_metrics):
+    spectator_client.foreach_metric_in_service_map(service_metrics, self.dump)
+
+  def dump(self, service, name, metadata, instance):
+    name, tags = spectator_client.normalize_name_and_tags(
+        name, metadata, instance)
+    if tags is None:
+      print 'IGNORE {0}.{1}'.format(service, name)
+    else:
+      print '{0}.{1} {2}'.format(service, name,
+                                 [(tag['key'], tag['value']) for tag in tags])
+
+
+def __determine_host(host):
+  if host in ['localhost', '127.0.0.1', None, '']:
+    host = socket.getfqdn()
+  return host
+
+
+class Monitor(object):
+  def __init__(self, spectator, metric_service, options):
+    self.__period = options.period
+    self.__spectator = spectator
+    self.__metric_service = metric_service
+    self.__services = options.services
+    self.__params = {}
+
+  def __data_map_to_service_metrics(self, data_map):
+    result = {}
+    count = 0
+    for service, metrics in data_map.items():
+      actual_metrics = metrics.get('metrics', None)
+      if actual_metrics is None:
+        logging.error('Unexpected response from "%s"', service)
+      else:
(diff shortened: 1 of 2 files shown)


ewiseblatt:
@lwander

google/stackdriver_monitoring/metric_collector.py
```diff
@@ -17,23 +17,81 @@
...
 
+class DummyMetricService(object):
+  def store(self, service_metrics):
+    spectator_client.foreach_metric_in_service_map(service_metrics, self.dump)
+
+  def dump(self, service, name, metadata, instance):
+    name, tags = spectator_client.normalize_name_and_tags(
+        name, metadata, instance)
+    if tags is None:
+      print 'IGNORE {0}.{1}'.format(service, name)
```
lwander: Is this formatting (and some in the last PR) in a format required by stackdriver?

google/stackdriver_monitoring/spectator_client.py
```diff
@@ -19,6 +19,38 @@
...
+  tags = metric_instance.get('tags', None)
+  if not tags:
+    return name, None   # signal this metric had no tags so we can ignore it.
+
+  is_timer = metric_metadata['kind'] == 'Timer'
+  if is_timer:
+    for index, tag in enumerate(tags):
+      if tag['key'] == 'statistic':
+        name = name + '__{0}'.format(tag['value'])
+        del tags[index]
```
lwander: I have a hunch that this could crash/misbehave since python's iterators aren't made for deleting elements during iteration.

google/stackdriver_monitoring/spectator_client.py
```diff
@@ -19,6 +19,38 @@
...
+def normalize_name_and_tags(name, metric_metadata, metric_instance):
+  tags = metric_instance.get('tags', None)
+  if not tags:
+    return name, None   # signal this metric had no tags so we can ignore it.
+
+  is_timer = metric_metadata['kind'] == 'Timer'
+  if is_timer:
+    for index, tag in enumerate(tags):
+      if tag['key'] == 'statistic':
+        name = name + '__{0}'.format(tag['value'])
```
lwander: What made you pick this as the format for `name`?

google/stackdriver_monitoring/metric_collector.py
ewiseblatt: no, this is not stackdriver related at all.
I dont know what you mean by format.
Here I'm just ignoring metrics that have no tags. This is a spectator thing. It is picking up a number of spring metrics which are mostly not interesting and those that are have other equivalents. The spring metrics cause problems because they have names that stackdriver doesnt like.

I put this in just as a placeholder because I dont have a stackdriver implementation here. I threw it out a while ago because it became obsolete. The current stackdriver implementation is in kork-stackdriver that is java embedded into the runtime services.

I'm preparing this server to make it accomodating to other metric services and perhaps another stab at stackdriver here with a different strategy for dealing with custom metrics.

google/stackdriver_monitoring/spectator_client.py
ewiseblatt: Arbitrary. It is what is in kork. I wanted to convert existing names into something new and wanted to avoid collisions with other metrics since it is a blind search/replace.

In spinnaker, the Timer metrics have a tag called "statistic" with is either "count" or "totalTime" where count is the number of calls and totalTime is the nanosecond duration.
So you might have a timer metric "invocations" with values
   name='invocations', tags=[{'statistic': 'count'}] value={'time':T, 'value':C}
   name='invocations', tags=[{'statistic': 'totalTime'}] value={'time':T, 'value':N}

This feels particularly awkward. Especially because count and totalTime are different things with different units and semantics so should be treated in different ways. So I split them into two different metrics

name='invocations__count', tags=[], value={'time':T, 'value':C}
name='invocations__totalTime', tags=[], value={'time':T, 'value':N}

Note that here there are no tags. Thats why above I explicitly use 'None' as a return value indicating there were never any tags as opposed to [] which is derived to no longer have any.

google/stackdriver_monitoring/spectator_client.py
ewiseblatt: It seems to work, and I'd expect it to. I would not be surprised for the iterator itself to break as you suggest, however I break immediately after doing this so never access the iterator again. Therefore, this is equivalent to breaking out of the loop and then deleting but simpler since I dont need to remember the index and carry it out of scope nor distinguish between found and unfound cases.

PR #1303, opened 2016-11-15
remove `.svc.cluster.local` from simple Kubernetes setup

rickypai:
Hi, I've been trying to setup spinnaker with the simple Kubernetes setup, and have been finding in the position of needing to clean up the setup scripts a bit. Hope contributions like this are accepted.

The current setup uses `*.spinnaker.svc.cluster.local` for hostnames for components. It assumes the cluster uses the default cluster domain of `cluster.local`, and breaks if the cluster does not use such setting. To support such setting, we can either resort to configuration management, or remove `.svc.cluster.local` to `*.spinnaker`, which would also resolve correctly.

@lwander

### DIFF
--- a/experimental/kubernetes/simple/README.md
+++ b/experimental/kubernetes/simple/README.md
@@ -130,7 +130,7 @@ is mostly filled out:
 
 ```yaml
   front50:
-    host: spin-front50.spinnaker.svc.cluster.local
+    host: spin-front50.spinnaker
     port: 8080
     baseUrl: ${services.default.protocol}://${services.front50.host}:${services.front50.port}
 
--- a/experimental/kubernetes/simple/config/spinnaker-local.yml
+++ b/experimental/kubernetes/simple/config/spinnaker-local.yml
@@ -20,7 +20,7 @@ services:
     protocol: http
 
   clouddriver:
-    host: spin-clouddriver.spinnaker.svc.cluster.local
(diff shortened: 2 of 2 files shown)


lwander:
Ah I see this is a problem. I'm having trouble finding docs that would assure me this will resolve, could you provide some?

Thanks for this and all the other fixes btw

rickypai:
No problem. Here are some examples of the DNS naming scheme:

http://kubernetes.io/docs/user-guide/services/#dns
> For example, if you have a Service called "my-service" in Kubernetes Namespace "my-ns" a DNS record for "my-service.my-ns" is created. Pods which exist in the "my-ns" Namespace should be able to find it by simply doing a name lookup for "my-service". Pods which exist in other Namespaces must qualify the name as "my-service.my-ns". The result of these name lookups is the cluster IP.

http://kubernetes.io/docs/admin/dns/#what-things-get-dns-names
> Assume a Service named foo in the Kubernetes namespace bar. A Pod running in namespace bar can look up this service by simply doing a DNS query for foo. A Pod running in namespace quux can look up this service by doing a DNS query for foo.bar.

http://kubernetes.io/docs/admin/dns/#validate-dns-works
> `kubectl exec busybox -- nslookup kubernetes.default`

where the namespace is `default` and the service is `kubernetes`.

PR #1304, opened 2016-11-15
sync `../../../config/*.yml` from `startup-config.sh` in simple Kubernetes setup

rickypai:
Hi, I've been trying to setup spinnaker with the simple Kubernetes setup, and have been finding in the position of needing to clean up the setup scripts a bit. Hope contributions like this are accepted.

`startup-config.sh` contains a step to copy all the configuration files from `config/` to the simple Kubernetes setup directory. Since there's been updates to the files in the root config directory, it produces unclean git state after `startup-config.sh` or `startup-all.sh` is run.

This PR commits the results from running `startup-config.sh`. Note that because all ReplicationController specs in https://github.com/spinnaker/spinnaker/tree/master/experimental/kubernetes/simple/rcs reference master images rather than locking to a version, there seems to be divergence of code and configuration compatibility. Particularly, whether we use the configurations in this branch or in master, not all spinnaker components will startup correctly. Particularly, `front50` produces error with both configuration files from master and this branch.

@lwander

### DIFF
--- a/experimental/kubernetes/simple/config/clouddriver.yml
+++ b/experimental/kubernetes/simple/config/clouddriver.yml
@@ -24,7 +24,7 @@ aws:
   # info on setting environment variables.
   enabled: ${providers.aws.enabled:false}
   defaults:
-    iamRole: ${provider.aws.defaultIAMRole:BaseIAMRole}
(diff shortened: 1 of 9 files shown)


ewiseblatt:
last call

PR #1308, opened 2016-11-16
Added stackdriver monitoring service

ewiseblatt:
@lwander 

This is an alternative implementation of the stackdriver integration that polls the server from the outside rather than pushing from the inside. It also employs a different strategy for mitigrating stackdriver's brittleness, which ultimately I'll put into the embedded kork implementation as well.

This is doing some awkward and complicated things because of stackdriver's constraint on custom metric descriptors being immutable. Whereas before I tried to proactively anticipate what descriptors I would need up front and inject labels, etc, this one reactively diagnoses the 400 errors it sees and then attempts to re-create the metric descriptor to accomodate the new usage pattern. Besides being complicated, the downside to this is that in doing so there is a chance you will lose your historic data for that metric. This seems stackdriver is forgiving for retaining data after you delete the descriptors now (it wasnt when I first tried a few months ago). I have a couple days worth of data despite having wiped my repository several times to start fresh and test through some real-world spinnaker scenarios.

In the tool, I can capture and restore the lists to a json file. This can be used for an install to populate the metric descriptors up front with what's likely needed to reduce the number of runtime adaptations that need to be made.


google/stackdriver_monitoring/spectator_client.py
```diff
@@ -147,12 +151,30 @@ def scan_by_service(self, service_list, params=None):
...
-      port = self.SERVICE_PORT_MAP[service]
+    start = time.time()
+    service_time = {service: 0 for service in service_list}
+    result = {service: None for service in service_list}
+    threads = {}
+
+    def timed_collect(self, service): #, params, service_time, result):
+      now = time.time()
       try:
+        port = self.SERVICE_PORT_MAP[service]
```
lwander: These are hard-coded, right? Why not pull these out of the config?

google/stackdriver_monitoring/spectator_client.py
```diff
@@ -147,12 +151,30 @@ def scan_by_service(self, service_list, params=None):
...
+
+    for service in service_list:
+      threads[service] = threading.Thread(
+          target=timed_collect,
+          args=(self, service))#, params, service_time, result))
+      threads[service].start()
+    for service in service_list:
+      threads[service].join()
+
+    logging.info('Collection times %d (ms): %s',
```
lwander: Maybe import `pprint` to properly format `service_time`

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -13,24 +13,110 @@
...
+    self.__num_data = len(self.__data_list)
+    self.__invocation_factory = invocation_factory
+    self.__get_name = get_name
+
+    self.batch_response = [None] * self.__num_data
+    self.num_bad = 0  # number of unsuccessful responses.
+    self.num_ok = 0   # number of successful responses.
+    self.was_ok = [None] * self.__num_data
+
+  def handle_batch_response(self, index_str, good, bad):
```
lwander: It might be nicer to wrap good & bad into a named tuple, so that accidentally switching the params here doesn't cause grief.

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -13,24 +13,110 @@
...
+      batch.add(invocation, callback=self.handle_batch_response,
+                request_id=str(count))
+      count += 1
+      if count % max_batch == 0:
+        logging.info('Executing batch of %d', max_batch)
+        batch.execute()
+        batch = self.__stackdriver.service.new_batch_http_request()
+
+    if count % max_batch:
+      logging.info('Executing final batch of %d', count % max_batch)
```
lwander: This logging is potentially misleading, since this won't be hit if `len(self.__data_list) % max_batch == 0`. If the user doesn't see "final batch" in their logging after expecting it, this serves as a red herring.

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -13,24 +13,110 @@
...
+      batch.execute()
+
+  def make_response(self, request, as_html, action, title):
+    """Create a response for the caller to ultimately send."""
+    if as_html:
+      html_rows = [('<tr><td>{0}</td><td>{1}</td></tr>'
```
lwander: Since you're inserting newlines below for (i'm guessing) readability, why not insert some here between `<tr>` elements? Especially since the number of rows could be quite large.

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -98,78 +187,206 @@ def __init__(self, options, stackdriver):
...
+    handler = BatchProcessor(
+        project, self.__stackdriver,
+        type_map.values(), delete_invocation, get_descriptor_name)
+    handler.process()
 
-    all_names = [descriptor['name'] for descriptor in type_map.values()]
-    batch = self.__stackdriver.service.new_batch_http_request()
-    max_batch = 100
-    count = 0
+    response_code = 200 if handler.num_ok == len(type_map) else 500
```
lwander: The builtin [http library](https://docs.python.org/2/library/httplib.html) allows you to replace these response codes with constants.

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -98,78 +187,206 @@ def __init__(self, options, stackdriver):
...
+          project, update_list, create_list, failed_list, delete_errors)
+
+    if create_list:
+      self.__do_batch_update_create_helper(
+        project, create_list, success_list, restore_list, create_errors)
+      restore_list = [original_type_map[elem['type']] for elem in restore_list]
+
+    if restore_list:
+      self.__do_batch_update_create_helper(
+        project, restore_list, not_updated_list, lost_list, restore_errors)
```
lwander: Why is `not_updated_list` the `success` parameter here? I'm just asking from a semantic perspective to understand what the intent is.

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -98,78 +187,206 @@ def __init__(self, options, stackdriver):
...
+
+    response_code = 200 if len(failed_list) == 0 else 500
```
lwander: It seems like there are more failure conditions here than just the original `failed_list`. What if either of the batch update creates fail?

google/stackdriver_monitoring/stackdriver_handlers.py
```diff
@@ -98,78 +187,206 @@ def __init__(self, options, stackdriver):
...
+        create_list.append(elem)
+
+    headers = {}
+    response_code = 200
+    response_body = []
+
+    if create_list:
+      create_response_code, create_headers, create_response_body = (
+          self.__do_batch_create(project, create_list))
+      response_code = max(response_code, create_response_code)
```
lwander: This feels off, is this typical response code handling?

lwander: I get it works here, but still feels strange

google/stackdriver_monitoring/spectator_client.py
ewiseblatt: what's wrong with the way it is? It's just for reference and better to keep on one line for this use case.

google/stackdriver_monitoring/stackdriver_handlers.py
ewiseblatt: The good and bad are actually different types. This is the interface from the batch callback so I cant control their structure. I renamed them.

google/stackdriver_monitoring/stackdriver_handlers.py
ewiseblatt: ok

google/stackdriver_monitoring/stackdriver_handlers.py
ewiseblatt: added comment

google/stackdriver_monitoring/stackdriver_handlers.py
ewiseblatt: Agreed. I'm not sure what to do here and other parts need cleaned up as well so saving it for a rainy day to go back over.

google/stackdriver_monitoring/spectator_client.py
lwander: Sounds good

PR #1318, opened 2016-11-26
Refactored metric tool for multiple mechanisms to make more sense.

ewiseblatt:
@lwander 

I renamed a couple modules that github doesnt seem to track.
metric_collector_handlers.py -> spectator_handlers.py
metric_collector.py -> spinnaker_metric_tool.py

I pulled out the HTTP and Monitor server into server_handlers.py

I changed the CommandHandler to distinguish between web and commandline.
I use argparse subcommands and have each primary module inject its commands
(the main tool asks each module for its commands).

Added datadog integration as another alternative, requirements.txt
to facilitate setup, and a bash script for execution because of the complex pythonpath.


google/stackdriver_monitoring/command_processor.py
```diff
@@ -40,23 +64,121 @@ def __init__(self, content_type='text/plain', options=None):
...
+      subparsers: The subparsers from the argparser parser.
+
+    Returns:
+      A new parser for to add additional custom parameters for this command.
+    """
+    parser = subparsers.add_parser(self.command_name, help=self.description)
+    parser.add_argument('--quiet', default=False, action='store_true',
+                        help='Dont print output to stdout.')
+    parser.add_argument('--output_path', default='',
+                        help='If provided, write command output to this path.')
```
lwander: Does this work if `quiet` is supplied?

lwander: Ah, I read the behavior below. The help text reads like I expected, looks good

google/stackdriver_monitoring/command_processor.py
```diff
@@ -40,23 +64,121 @@ def __init__(self, content_type='text/plain', options=None):
...
+    if do_print:
+      print content
+    output_path = options.get('output_path', None)
+
+    if output_path:
+      # pylint: disable=invalid-name
+      with open(options['output_path'], 'w') as f:
+        f.write(content)
+      if do_print:
+        print 'Wrote {0}'.format(output_path)
```
lwander: This feels a little noisy, how often do you expect to call `output`?

google/stackdriver_monitoring/datadog_handlers.py
```diff
@@ -0,0 +1,137 @@
...
+
+    json_text = json.JSONEncoder(indent=2).encode(get_method(artifact_id))
+    self.output(options, json_text)
+
+
+def add_handlers(handler_list, subparsers):
+  """Registers all the CommandHandlers for interacting with Datadog."""
+  command_handlers = [
+      ListArtifactsHandler('screenboard',
+                           None, 'list_datadog_screenboards',
```
lwander: I'm missing something here: where does the `command_name` get used?

google/stackdriver_monitoring/datadog_service.py
```diff
@@ -0,0 +1,105 @@
...
+        'points': [(elem['t'] / 1000, elem['v'])
+                   for elem in instance['values']],
+        'tags': ['{0}:{1}'.format(tag['key'], tag['value']) for tag in tags]
+    })
+
+  def publish_metrics(self, service_metrics):
+    """Writes time series data to Datadog for a metric snapshot."""
+    points = []
+    spectator_client.foreach_metric_in_service_map(
+        service_metrics, self.__append_timeseries_point, points)
```
lwander: I missed when `foreach_metric_in_service_map` was introduced - but it looks like it takes a pretty hefty function signature as an argument when I look at the spec for `__append_timeseries_point`, is that right?

google/stackdriver_monitoring/server_handlers.py
```diff
@@ -0,0 +1,200 @@
...
+    """Constructor.
+
+    Args:
+      all_handlers: [list of CommandHandler] Determines the page contents.
+        The entries that have url_paths will be displayed.
+    """
+    super(HomePageHandler, self).__init__(url_path, command_name, description)
+    self.__all_handlers = all_handlers
+
+  def process_web_request(self, request, path, params, fragment):
```
lwander: Any need to check if request accepts html?

google/stackdriver_monitoring/command_processor.py
ewiseblatt: The help explicitly said stdout.

google/stackdriver_monitoring/datadog_handlers.py
ewiseblatt: The command_name is the value of the 'command' option. It makes sense to call it command in the options. Should I call it command here? command_name seems to make more sense as an attribute name because it is a name (string) whereas a command sounds like a complex class.

The name is used in spinnaker_metric_tool.main to dispatch the command to the appropriate handler (and in command_processor.process_command to correlate the command against the handler instance to dispatch to).

google/stackdriver_monitoring/datadog_service.py
ewiseblatt: yes. it kind of grew.
It takes the metric source, name, and value as you'd expect but then it takes some metadata values, then any additional values you want to inject in.

The metadata values is what's been growing. These are
   service level metadata  (mainly its start time which stackdriver uses as a reference point for CUMULATIVE values)
   metric level metadata (mainly the type of metric this is (e.g. Timer, Gauge, Counter))

google/stackdriver_monitoring/server_handlers.py
ewiseblatt: This is for the home page. I dont care. The command is intended to support a web browser in HTML. I'm not going to support an alternative. The reason I check other commands for HTML is that I'm reusing the same logic for command line and web invocation and only want HTML for web. I've been moving away from that (having split process_web_request vs process_commandline_request). This particular handler doesnt give itself a command name, so isnt even visible or invocable from the commandline (using the standard dispatch mechanism).

google/stackdriver_monitoring/datadog_handlers.py
lwander: Naming is up to you, I was just curious how it tied back to the actual command run. Sounds good.

lwander:
LGTM

PR #1319, opened 2016-11-28
added retry logic to build google tarball script

danielpeach:
`gsutil ls` will sometimes list a file that has just been deleted, which has caused the build to fail a few times. I was able to reproduce this locally and I don't think a sleep is necessary here.

@ewiseblatt or @jtk54 please review.


danielpeach:
Eric suggested making the GCS bucket regional instead of multi-regional. We'll see if that fixes the issue permanently.

PR #1324, opened 2016-12-01
Dont assume metric monitor tool is running on GCE

ewiseblatt:
Other small tweaks.

@ttomsu 
I think this fixes all the issues we ran into yesterday afternoon.

### DIFF
--- a/google/stackdriver_monitoring/stackdriver_service.py
+++ b/google/stackdriver_monitoring/stackdriver_service.py
@@ -35,6 +35,18 @@ import spectator_client
 logging.getLogger('googleapiclient.discovery_cache').setLevel(logging.ERROR)
 
 
+def get_aws_identity_document():
+  url = 'http://169.254.169.254/latest/dynamic/instance-identity/document'
+  request = urllib2.Request(url)
+  try:
+    response = urllib2.urlopen(request)
+  except IOError as ioex:
+    logging.info('Cannot read AWS Identity Document, probably not on Amazon Web Services.'
+                 ' url=%s: %s', url, ioex)
+    raise ioex
+  return json.JSONDecoder().decode(response.read())
+
+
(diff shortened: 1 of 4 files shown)


google/stackdriver_monitoring/stackdriver_service.py
```diff
@@ -35,15 +35,27 @@
 logging.getLogger('googleapiclient.discovery_cache').setLevel(logging.ERROR)
 
 
+def get_aws_identity_document():
+  url = 'http://169.254.169.254/latest/dynamic/instance-identity/document'
+  request = urllib2.Request(url)
+  try:
+    response = urllib2.urlopen(request)
+  except IOError as ioex:
+    logging.error('Cannot read AWS Identity Document, probably not on Amazon Web Services.'
```
ttomsu: Seems like this would be an the info level, rather than error.

google/stackdriver_monitoring/stackdriver_service.py
```diff
@@ -35,15 +35,27 @@
...
 def get_google_metadata(attribute):
   url = os.path.join('http://169.254.169.254/computeMetadata/v1', attribute)
   request = urllib2.Request(url)
   request.add_header('Metadata-Flavor', 'Google')
   try:
     response = urllib2.urlopen(request)
   except IOError as ioex:
-    print 'CAUGHT ' + str(ioex)
-    print 'request {0}'.format(url)
+    logging.error('Cannot read google metadata, probably not on Google Cloud Platform.'
```
ttomsu: Same here.

ttomsu:
LGTM

google/stackdriver_monitoring/stackdriver_service.py
ewiseblatt: agreed.

PR #1326, opened 2016-12-07
(consul) bootstrap script for starting consul

lwander:
This works by either joining the current MIG, or joining a metadata-provided list of known consul members

@ewiseblatt or @duftler PTAL

Note the install scripts automatically package this onto the consul base images.

### DIFF
--- a/consul/config/bootstrap-consul.sh
+++ b/consul/config/bootstrap-consul.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+
+# Note: for instances running in GCP only.
+#
+# This script is meant to be run by any instance either joining the consul 
+# network, or bootstrapping a consul network. It does so by first looking for
+# an explicit network to join in the instance metadata under "consul-members",
+# and if it doesn't find any, it will attempt to join other instances in the
+# same instance group.
+
+METADATA_URL="http://metadata.google.internal/computeMetadata/v1"
+INSTANCE_METADATA_URL="$METADATA_URL/instance"
+INSTANCE_LIST_METADATA="consul-members"
+
+function get_instance_metadata_attribute() {
+  local name="$1"
+  local value=$(curl -s -f -H "Metadata-Flavor: Google" \
+                     $INSTANCE_METADATA_URL/attributes/$name)
+  if [[ $? -eq 0 ]]; then
+    echo "$value"
+  else
+    echo ""
+  fi
+}
+
+function get_instance_list_from_metadata() {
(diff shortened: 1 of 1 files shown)


consul/config/bootstrap-consul.sh
```diff
@@ -0,0 +1,60 @@
...
+  get_instance_metadata_attribute $INSTANCE_LIST_METADATA
+}
+
+function get_instance_list_from_mig() {
+  # projects/<projectid>/zones/<zone>/instanceGroupManagers/<igm>
+  createduri=$(get_instance_metadata_attribute "created-by")
+
+  if [ -z $createduri ]
+  then
+    echo "Failed to retrieve who created" $HOSTNAME
```
duftler: I think "Failed to identify creator of" might sound better.

consul/config/bootstrap-consul.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "Make sure it is running in a managed instance group"
+    exit 1
+  fi
+
+  IFS='/' read -r -a createdlist <<< "$createduri"
+
+  mig=${createdlist[-1]}
+  zone=${createdlist[-3]}
+
+  gcloud compute instance-groups list-instances $mig --zone $zone --format='value(instance)'
```
duftler: Might not be a very important distinction, but I think this particular operation is operating on an _instance group_, not a _managed instance group_.

consul/config/bootstrap-consul.sh
```diff
@@ -0,0 +1,60 @@
...
+
+if [ -z "$instances" ]
+then
+  echo "Failed to retrieve consul instances" 
+  exit 1
+fi
+
+echo "Waiting" $WAIT_TIME_SECONDS "before joining nodes" $instances
+
+sleep $WAIT_TIME_SECONDS
```
ewiseblatt: what's the issue here? why two minutes?
Should you have a bounded loop here instead?

lwander: I'm waiting for other instances in the MIG to come up, but now that you mention it, the joins should be symmetric so there is no need to wait.

ewiseblatt: if you are waiting on a condition, you should try to test for that condition rather than assume it will be ready on an upper bound. I'd be aggressive testing earlier since presumably this is an upper bound that you have expected to nearly always exceed the actual required time.

lwander:
Added a top-level comment, fixed up some wording, and replaced the sleep with a while loop that waits for the consul agent to start.

PR #1328, opened 2016-12-10
Specify v2 execution engine for all test pipelines.

duftler:


duftler:
fyi @ewiseblatt @lwander 

Running the jobs now...

PR #1343, opened 2016-12-17
Clean up default docker config

EricJizbaMSFT:
‘repository’ should be ‘repositories’ based on DockerRegistryConfigurationProperties (I kept getting 401 Unauthorized and discovered it was because the repository wasn’t set right and it was trying to get the catalog for the ENTIRE registry at ‘index.docker.io’)

Add ‘:’ to the username and passwordFile environment vars so that they default to null (and thus you can connect to public repos). Otherwise it was using the literal string “${SPINNAKER_DOCKER_USERNAME}” as my username

### DIFF
--- a/config/clouddriver.yml
+++ b/config/clouddriver.yml
@@ -110,7 +110,8 @@ dockerRegistry:
     - name: ${providers.dockerRegistry.primaryCredentials.name}
       address: ${providers.dockerRegistry.primaryCredentials.address}
       username: ${providers.dockerRegistry.primaryCredentials.username:}
-      passwordFile: ${providers.dockerRegistry.primaryCredentials.passwordFile}
+      passwordFile: ${providers.dockerRegistry.primaryCredentials.passwordFile:}
+      repositories: ${providers.dockerRegistry.primaryCredentials.repository:}
 
 credentials:
   primaryAccountTypes: ${providers.aws.primaryCredentials.name}, ${providers.google.primaryCredentials.name}, ${providers.cf.primaryCredentials.name}, ${providers.azure.primaryCredentials.name}
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -126,10 +126,10 @@ providers:
     primaryCredentials:
       name: my-docker-registry
       address: ${SPINNAKER_DOCKER_REGISTRY:https://index.docker.io/}
-      repository: ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}
-      username: ${SPINNAKER_DOCKER_USERNAME}
+      repository: ${SPINNAKER_DOCKER_REPOSITORY:}
+      username: ${SPINNAKER_DOCKER_USERNAME:}
       # A path to a plain text file containing the user's password
-      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE}
+      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE:}
 
   openstack:
     # For more information on configuring Openstack clusters, see
--- a/config/spinnaker.yml
+++ b/config/spinnaker.yml
@@ -332,10 +332,10 @@ providers:
     primaryCredentials:
       name: my-docker-registry-account
       address: ${SPINNAKER_DOCKER_REGISTRY:https://index.docker.io/}
-      repository: ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}
-      username: ${SPINNAKER_DOCKER_USERNAME}
+      repository: ${SPINNAKER_DOCKER_REPOSITORY:}
+      username: ${SPINNAKER_DOCKER_USERNAME:}
       # A path to a plain text file containing the user's password
-      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE}
+      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE:}
 
   openstack:
     # This default configuration uses the same environment variable names set in


EricJizbaMSFT:
@rguthriemsft @scotmoor @lwander @duftler PTAL

experimental/kubernetes/simple/config/clouddriver.yml
```diff
@@ -74,8 +74,7 @@ dockerRegistry:
       address: ${providers.dockerRegistry.primaryCredentials.address}
       username: ${providers.dockerRegistry.primaryCredentials.username:}
       passwordFile: ${providers.dockerRegistry.primaryCredentials.passwordFile:}
-      repositories: 
-        - ${providers.dockerRegistry.primaryCredentials.repository}
+      repositories: ${providers.dockerRegistry.primaryCredentials.repositories}
```
lwander: Spring config can't refere to structs or lists like this, only scalar values & strings.

config/spinnaker.yml
```diff
@@ -332,10 +332,11 @@ providers:
...
       name: my-docker-registry-account
       address: ${SPINNAKER_DOCKER_REGISTRY:https://index.docker.io/}
-      repository: ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}
-      username: ${SPINNAKER_DOCKER_USERNAME}
+      username: ${SPINNAKER_DOCKER_USERNAME:}
       # A path to a plain text file containing the user's password
-      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE}
+      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE:}
+      repositories:
+        - ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}
```
lwander: This won't be referenceable by `clouddriver.yml`

config/default-spinnaker-local.yml
```diff
@@ -127,9 +127,9 @@ providers:
       name: my-docker-registry
       address: ${SPINNAKER_DOCKER_REGISTRY:https://index.docker.io/}
       repository: ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}
-      username: ${SPINNAKER_DOCKER_USERNAME}
+      username: ${SPINNAKER_DOCKER_USERNAME:}
       # A path to a plain text file containing the user's password
-      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE}
+      passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE:}
```
EricJizbaMSFT: If I don't have the colons in 'spinnaker-local', I get this in 'localhost:8084/resolvedEnv':
> "providers.dockerRegistry.primaryCredentials.passwordFile": "Exception occurred: Could not resolve placeholder 'SPINNAKER_DOCKER_PASSWORD_FILE' in string value \"${SPINNAKER_DOCKER_PASSWORD_FILE}\""

If I also don't have the colons in 'clouddriver.yml', I get this exception when I check the clouddriver log:

> java.io.FileNotFoundException: ${providers.dockerRegistry.primaryCredentials.passwordFile} (No such file or directory)

There's many reasons the config could've failed in #1058 and I'm not convinced the colon was a problem. From what I can see the colon is used in many cases throughout these files

lwander: My interpretation was that the `:` was inserting an empty string , which Spring wasn't overriding correctly. If you can verify that this change still allows registries to authenticate when presenting `password` instead of overriding `passwordFile` in `clouddriver-local.yml` with this change in place this should be good.

EricJizbaMSFT: Let's see if I understand your scenario correctly. I left clouddriver.yml and spinnaker.yml with the colons like in this pull request. I then added my username and password to my spinnaker-local.yml file like this:
> username: ${SPINNAKER_DOCKER_USERNAME:erijiz}
> password: examplePassword
> passwordFile: ${SPINNAKER_DOCKER_PASSWORD_FILE:}

That worked and I was able to authenticate to a private repo. My resolvedEnv looked like this:

> "providers.dockerRegistry.primaryCredentials.password": "******",
> "providers.dockerRegistry.primaryCredentials.passwordFile": "",
> "providers.dockerRegistry.primaryCredentials.repository": "erijiz/skuservice",
> "providers.dockerRegistry.primaryCredentials.username": "erijiz",

The resolvedEnv displays an empty string for passwordFile, which is better than getting the exception I listed above.

lwander: ok neat, this looks good - one last comment and this is ready to merge

config/clouddriver.yml
```diff
@@ -101,7 +101,9 @@ dockerRegistry:
     - name: ${providers.dockerRegistry.primaryCredentials.name}
       address: ${providers.dockerRegistry.primaryCredentials.address}
       username: ${providers.dockerRegistry.primaryCredentials.username:}
-      passwordFile: ${providers.dockerRegistry.primaryCredentials.passwordFile}
+      passwordFile: ${providers.dockerRegistry.primaryCredentials.passwordFile:}
+      repositories:
+        - ${providers.dockerRegistry.primaryCredentials.repository}
```
lwander: I omit this because I prefer to depend on the `/v2/_catalog` endpoint to retrieve all repos

lwander: Could you delete these two lines? They would break our public image

EricJizbaMSFT: What registry are you getting the entire catalog for? When I try to get the entire catalog for 'index.docker.io', I get a '401 Unauthorized' every single time.

This is at the crux of why I had trouble connecting to docker. Without these two lines, the repository specified in spinnaker-local.yml is ignored and it wasn't obvious that it was trying to get the whole catalog

EricJizbaMSFT: @lwander Hope you had a good holidays! What if I leave these two lines, but modify 'spinnaker.yml' and 'default-spinnaker-local.yml' from this:

> repository: ${SPINNAKER_DOCKER_REPOSITORY:library/nginx}

to this:

> repository: ${SPINNAKER_DOCKER_REPOSITORY:}

That way '/v2/_catalog' is still the default behavior, but the repository won't be ignored if it's specified.

Would that still break your public image?

lwander: Nice! Let me try that out 

We use `/v2/_catalog` for GCR, hosted docker registries, and quay

lwander: Hmm, when I leave this line empty in clouddriver.yml

```yaml
repositories:
  - 
```
to emulate the non-resolved repository name, Spinnaker doesn't query the catalog endpoint. When you ran it, did 

```yaml
repository: ${SPINNAKER_DOCKER_REGISTRY:}
```
cause the catalog endpoint to be queried?

EricJizbaMSFT: What about if you do this in clouddriver instead of list it on the next line?

> repositories: ${providers.dockerRegistry.primaryCredentials.repository}

This one seems to work for me in both cases (it queries the catalog if the repository isn't specified and has a list of 1 repository if I do specify it)

EricJizbaMSFT: @lwander PTAL

lwander: Nice, that does the trick

lwander:
:confetti_ball:

PR #1345, opened 2016-12-19
Refactored to support multiple endpoints/replicas for a given service.

ewiseblatt:
This is currently configurable by command line parameters with fixed service
names so is restricted to the pre-existing service names (not HA names).
However except for that one function that converts the options toa list of
endpoints, that is the only place for that restriction. So in the interim
you can replace that method with a different mechanism if you need different
names (such as 'clouddriver-ro', etc) and it should work.

--service_hosts=<comma delimited list of IPs, defaulting to "localhost">
--<service>=<comma listed list of network locations> where <service> is
"clouddriver", etc and <network location> is <host[:port]> where the
standard port will be assumed if not provided. A host of "*" means
use all the hosts in --service_hosts.

If --service_hosts is empty then by default no services will be used.
If --<service>=<empty> then that service will be ignored. If non empty
it will override --service_hosts (unless it is "*" in which case it is
service_hosts).

By default each service specifies its endpoints as "*" so the net effect
is that it will look at the standard ports for localhost if you say nothing,
or the standard ports for --service_hosts if you just specify that.

@ttomsu  please review
@gardleopard  FYI

### DIFF
--- a/google/stackdriver_monitoring/command_processor.py
+++ b/google/stackdriver_monitoring/command_processor.py
@@ -106,7 +106,7 @@ class CommandHandler(object):
     query_list = ['{0}={1}'.format(key, value) for key, value in params.items()]
     return '?{0}'.format('&'.join(query_list)) if query_list else ''
 
-  def process_commandline_request(self, options):
+  def process_commandline_request(self, options, **kwargs):
     """Processes the command from a commandline request.
 
     Output (as opposed to logging) should be emitted using output()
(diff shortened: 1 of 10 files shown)


ttomsu:
I don't think doing this via command line args is going to solve the issue - the list of hosts and ports is going to be dynamic and potentially changing over time. It makes more sense to me to have a module that pulls down hostnames from [Kubernetes API, Consul] and a web endpoint to modify the list manually (RESTful HTTP verbs make sense here).

ewiseblatt:
The mechanism in this PR is commandline args. As I mentioned last week, the
intent is to add other ways within the implementation to generate the list
of dependencies other than those args (e.g. querying kubernetes).

ttomsu:
LGTM

PR #1346, opened 2016-12-19
Updated L7 LB tests for read-only forwarding rules.

jtk54:
This updates the nightly L7 tests to handle the recent changes making forwarding rules read-only. We no longer can "upgrade" a HTTP LB to an HTTPS LB, so I took those test cases out. The tests now look like the following:

* upsert HTTP LB
* delete HTTP LB
* upsert HTTPS LB
... manipulate HTTPS LB
* delete HTTPS LB

This makes sure we test both L7 code paths.

### DIFF
--- a/testing/citest/tests/google_http_lb_upsert_test.py
+++ b/testing/citest/tests/google_http_lb_upsert_test.py
@@ -144,8 +144,11 @@ class GoogleHttpLoadBalancerTest(st.AgentTestCase):
     return citest.base.TestRunner.global_runner().get_shared_data(
         GoogleHttpLoadBalancerTestScenario)
 
-  def test_c_upsert_min_lb(self):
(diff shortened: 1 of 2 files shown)


testing/citest/tests/google_http_lb_upsert_test.py
```diff
@@ -144,9 +144,12 @@ def scenario(self):
     return citest.base.TestRunner.global_runner().get_shared_data(
         GoogleHttpLoadBalancerTestScenario)
 
-  def test_c_upsert_min_lb(self):
+  def test_a_upsert_min_lb(self):
```
duftler: These test case names read like a (failed) sobriety check.

ewiseblatt:
lgtm.

PR #1351, opened 2016-12-25
(google) Document how to index and use images from additional GCP pro…

duftler:
…jects.

### DIFF
--- a/config/clouddriver.yml
+++ b/config/clouddriver.yml
@@ -51,6 +51,13 @@ google:
       jsonPath: ${providers.google.primaryCredentials.jsonPath}
       consul:
         enabled: ${providers.google.primaryCredentials.consul.enabled:false}
+      # Specify the project id of each GCP project whose images you want indexed and
+      # made available to this account. Each project must have granted the IAM role
+      # compute.imageUser to the service account associated with the json key referenced
+      # above, as well as to the 'Google APIs service account' automatically created for
+      # the project being managed (should look similar to
+      # 12345678912@cloudservices.gserviceaccount.com).
+      # imageProjects:
 
 cf:
   enabled: ${providers.cf.enabled:false}


duftler:
@lwander or @ewiseblatt please review.

duftler:
@danielpeach please review.

ewiseblatt:
FWIW, I see I never sent my comments

config/clouddriver.yml
```diff
@@ -51,6 +51,13 @@ google:
       jsonPath: ${providers.google.primaryCredentials.jsonPath}
       consul:
         enabled: ${providers.google.primaryCredentials.consul.enabled:false}
+      # Specify the project id of each GCP project whose images you want indexed and
```
duftler: Good suggestion. Have done so here: http://www.spinnaker.io/v1.0/docs/target-deployment-configuration#section-google-cloud-platform

I'm assuming by 'built-in defaults' you're referring to the public cloud projects. Those are configured here: https://github.com/spinnaker/clouddriver/blob/master/clouddriver-web/config/clouddriver.yml#L137-L145, are available to _all_ Google accounts, and are not affected by the `imageProjects` attribute. Note that the `imageProjects` attribute has been in place for a long time and we're just adding some documentation in response to a user query.
> The "should look similar to" isnt clear. Do you mean it is usually in the form?
Yes. I did several experiments and each account I observed that worked properly had that form.

lwander: Do you think it's worth adding this link? https://cloud.google.com/compute/docs/images/sharing-images-across-projects#granting_a_user_access_to_images

duftler: Sure. PR created here: https://github.com/spinnaker/spinnaker/pull/1355

ewiseblatt:
LGTM.

FWIW, Ultimately configuration should probably be documented outside the
yaml.
At that time, the yaml could point to the spinnaker documentation (and the
spinnaker docs would link).
At least I think. While it is an extra level of indirection, it allows
spinnaker to present the concepts in its own context and bring attention to
the important issues and considerations (e.g. awareness that this even
needs to happen).

PR #1360, opened 2017-01-05
Added tool to facilitate deleting GCP resources.

ewiseblatt:
@jtk54

### DIFF
--- a/google/dev/delete_resources.py
+++ b/google/dev/delete_resources.py
@@ -1,0 +1,402 @@
+#!/usr/bin/python
+#
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# pylint: disable=import-error
+# pylint: disable=star-args
+"""
+Implements program for deleting Google API resources by their name and/or age.
+
+Example usage:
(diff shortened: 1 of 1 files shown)


google/dev/delete_resources.py
```diff
@@ -0,0 +1,399 @@
...
+  python delete_resources.py \
+      storage.objects \
+      --delete_kwargs=bucket=mybucket \
+      --list_kwargs=versions=true \
+      --dry_run
+
+  # Delete all the instances from this project that have 'xyz' in the name.
+  # Use the list_aggregated method, which considers all zones.
+  python delete_resources.py \
+      compute.instances \
```
jtk54: Are the options for resources enumerated anywhere? A link to the docs or just a list here would suffice.

google/dev/delete_resources.py
```diff
@@ -0,0 +1,399 @@
...
+
+
+def get_metadata(relative_url):
+  """Return metadata value.
+
+  Args:
+    relative_url: [string] Metadata url to fetch relative to base metadata URL.
+  """
+  base_url = 'http://metadata/computeMetadata/v1/'
+  url = os.path.join(base_url, relative_url)
```
jtk54: Probably want `urljoin` instead of `os.path.join` here: [urljoin](https://docs.python.org/2/library/urlparse.html#urlparse.urljoin). I don't know if anyone would execute this script on windows, but it wouldn't work.

google/dev/delete_resources.py
ewiseblatt: added in help message

google/dev/delete_resources.py
```diff
@@ -0,0 +1,402 @@
...
+                           ' The default is application default credentials')
+  parser.add_argument('--aggregated', default=False, action='store_true',
+                      help='Use aggregated_list() method.')
+  parser.add_argument('--dry_run', default=False, action='store_true',
+                      help='Show proposed delete, dont actually do them.')
+  parser.add_argument(
+      '--delete_kwargs', default=None,
+      help='The extra arguments to pass to delete.'
+           ' This is a comma-delimted list of name=value.'
+           ' For more info, ee https://developers.google.com/apis-explorer/#p/')
```
jtk54: s/ee/see

PR #1365, opened 2017-01-11
Optionally log metric deltas across polls in metric_tool.

ewiseblatt:
Keep stats on number of data points collected.

@jtk54

### DIFF
--- a/google/stackdriver_monitoring/spectator_client.py
+++ b/google/stackdriver_monitoring/spectator_client.py
@@ -13,14 +13,15 @@
 # limitations under the License.
 
 # pylint: disable=missing-docstring
+# pylint: disable=star-args
 
+import copy
 import json
 import logging
 import threading
 import time
 import socket
 import urllib2
-
 
 def __foreach_metric_tag_binding(
     service, metric_name, metric_data, service_data,
@@ -138,6 +139,14 @@ class SpectatorClient(object):
                         ' containing the services.')
     parser.add_argument('--prototype_path', default='',
                         help='Optional filter to restrict metrics of interest.')
+    parser.add_argument(
+        '--log_metric_diff',
+        default=False, action='store_true',
+        help='Keep track of the last set of metrics/bindings were'
+             ' and show the differences with the current metric/bindings.'
+             ' This is to show a change in what metrics are available, not'
+             ' the values of the metrics themselves.')
+
     for service in SpectatorClient.DEFAULT_SERVICE_PORT_MAP.keys():
       add_microservice(parser, service)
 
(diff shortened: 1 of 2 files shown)


google/stackdriver_monitoring/spectator_client.py
```diff
@@ -151,12 +160,67 @@ def __init__(self, options):
...
+
+    previous_keys = set(previous_metrics.keys())
+    keys = set(metrics.keys())
+    new_keys = keys.difference(previous_keys)
+    same_keys = keys.intersection(previous_keys)
+    lost_keys = previous_keys.difference(keys)
+    lines = []
+
+    if lost_keys:
+      lines.append('Lost metrics for:\n  - {0}\n'
```
jtk54: What does 'Lost metrics' mean? It makes sense where the `lost_keys` come from, but the terminology seems weird.

google/stackdriver_monitoring/spectator_client.py
```diff
@@ -151,12 +160,67 @@ def __init__(self, options):
     self.__host = options['host']
     self.__prototype = None
     self.__default_scan_params = {'tagNameRegex': '.+'}
+    self.__previous_scan_lock = threading.Lock()
+    self.__previous_scan = {} if options.get('log_metric_diff') else None
```
jtk54: Does `options.get('log_metric_diff')` throw a `KeyError` if `'log_metric_diff'` isn't present? Make it default to `None` if it does.

google/stackdriver_monitoring/spectator_client.py
```diff
@@ -178,12 +242,31 @@ def collect_metrics(self, host, port, params=None):
     url = 'http://{host}:{port}/spectator/metrics{query}'.format(
         host=host, port=port, query=query)
     response = urllib2.urlopen(url)
+
     all_metrics = json.JSONDecoder(encoding='utf-8').decode(response.read())
+    try:
+      self.__log_scan_diff(host, port, all_metrics.get('metrics', {}))
```
jtk54: Seems like control flow here to call this only if `--log_metric_diff` would be clearer than calling this each time and setting internal object state based on the flag.

google/stackdriver_monitoring/spectator_client.py
ewiseblatt: get doesnt throw key error. It returns None. [] throws key error.

google/stackdriver_monitoring/spectator_client.py
ewiseblatt: Chanted to Started/Stopped rather than Added/Lost

PR #1366, opened 2017-01-11
Factored out CI image validation script from Jenkins.

jtk54:


dev/jenkins_task_validate_image.sh
```diff
@@ -0,0 +1,624 @@
...
+#
+# This script configures and deploys Spinnaker on GCP and runs an integration
+# test suite for validation that manages resources on AWS, GCP, and Kubernetes.
+#
+# The following environment variables are expected to be populated:
+#
+# IMAGE_TO_VALIDATE [String] - The name of the image to validate.
+# IMAGE_PROJECT [String] - The GCE project containing the $IMAGE_TO_VALIDATE.
+# If changing this, you will also need to make sure that the $TESTER_SERVICE_ACCOUNT has permission to View (or Edit) the project so that it can download the image.
+# MY_INSTANCE_NAME [String] - The name of the instance to deploy when testing. Must be unique.
```
ewiseblatt: can you rename this to SPINNAKER_INSTANCE_NAME and SPINNAKER_PROJECT and SPINNAKER_ZONE? You may also need a SPINNAKER_SERVICE_ACCOUNT. I dont notice one here because we use the same account for both. But if you are pulling this out as a script then you should decouple that.

Regroup these to collcate the jenkins config from the deplment config from the managed config from the citest config.

dev/jenkins_task_validate_image.sh
```diff
@@ -0,0 +1,624 @@
...
+# PARALLELIZE_TESTS [Boolean] - Run each of the top-level test suites in parallel.
+# MY_CLUSTER_NAME [String] - The name of the Kubernetes cluster to be tested against.
+# AWS_ACCESS_KEY [String] - The value of the AWS access key.
+# AWS_SECRET_KEY [String] - The value of the AWS secret key.
+# JENKINS_PASSWORD [String] - Password to Jenkins instance configured for Spinnaker instance.
+# JENKINS_USER [String] - User for Jenkins instance configured for Spinnaker instance.
+# HOME [String] - Home directory for the user running this script.
+# BASE_ARCHIVE_PATH [String] - Path to GCS bucket for build archival.
```
ewiseblatt: Maybe let this be empty and only archive if it is set.
The data isnt actually used for anything [yet].

dev/jenkins_task_validate_image.sh
```diff
@@ -0,0 +1,624 @@
...
+# MY_SPINNAKER_PROJECT [String] - Name of the project Spinnaker is running in.
+# MY_ZONE [String] - GCP zone Spinnaker is running in.
+# TESTER_SERVICE_ACCOUNT [String] - Service account used for managing Spinnaker resources.
+# REPORT_EVERY [Integer] - Period of print statements when waiting for servers (in seconds).
+# PROJECT_JSON_PATH [String] - Path to service account credentials file for GCP project managed by Spinnaker.
+# SPINNAKER_LOCAL_YML_PATH [String] - Path to spinnaker-local.yml used to configure Spinnaker microservices.
+# OPT_FILE_METADATA [String] - Metadata entry specifying manage project credentials of the form: 'managed_project_credentials=$PROJECT_JSON_PATH'.
+# JENKINS_URL [String] - External URL for Jenkins instance configured in the Spinnaker instance. Must be firewalled so the VM running this script can make an HTTP request.
```
ewiseblatt: Jenkins url should not be firewalled against this script. So, you might need to write a firewall rule or something to permit this jenkins access to the jenkins url.

Also, it is not clear that the jenkins url is for testing the deployed spinnaker against, which is most likely not the same jenkins server as the production server used to build and run these candidates.

dev/jenkins_task_validate_image.sh
```diff
@@ -0,0 +1,624 @@
...
+          --delete_added \\
+          --delete_for_real
+  done
+EOF
+  chmod +x cleanup_leaks.sh
+}
+
+######################################################################
+
+cleanup_workspace
```
ewiseblatt: Do you think this should be wrapped in a function too?
If the whole thing wraps into functions with no side effects you can source it to get the definitions then run main() yourself or create your own flow out of the steps you want.

jtk54: Let's do this if/when we port this to python.

jtk54:
Tested this script again successfully in Jenkins after the small changes from comments.

PR #1367, opened 2017-01-12
Factored 'ImageFromDebianRepo' Jenkins task.

jtk54:


jtk54:
Moved to `google/dev` and added a comment stating this script is for GCP only.

google/dev/jenkins_task_image_from_deb_repo.sh
```diff
@@ -0,0 +1,91 @@
...
+}
+
+
+function build_spinnaker_google_image() {
+  # Normally these BINTRAY variables arent needed or used.
+  # However, this can give us access to passing a private bintray DEBIAN_REPO_URL
+  export BINTRAY_KEY
+  export BINTRAY_USER
+
+  gcloud config set account $BUILDER_SERVICE_ACCOUNT
```
ewiseblatt: I thought we took out dependencies on the gcloud service account. I see build_google_image.sh uses gcloud to list the current images in order to tease out the base image. That could be a parameter to the script instead and you can compute that here (maybe leaving the default in the script as it is if the parameter isnt set, which is convienent to use by hand as a developer where gcloud account credentials could be assumed). The benefit is that the base image becomes a parameter (i.e. can easly switch to ubuntu 16 or something else entirely as long as it supports debians).

The desire is that running this script should not have a side effect of changing the gcloud account that is configured.

google/dev/jenkins_task_image_from_deb_repo.sh
```diff
@@ -0,0 +1,91 @@
...
+#
+# A minimal Jenkins job configuration to run this script:
+#
+# rm -rf spinnaker/
+# git clone https://github.com/spinnaker/spinnaker.git
+# ./spinnaker/dev/jenkins_task_image_from_deb_repo.sh
+
+
+function clone_spinnaker_with_owner() {
+  rm -rf spinnaker
```
ewiseblatt: ok, but this is a very specialized use case. This is a generic script in the debian repository, but its intention is that it is running in the CWD of a jenkins server task workspace. Otherwise this rm -rf can ruin your day.

Maybe main should ensure that you are running this from a jenkins workspace?

jtk54: We can check that the current directory starts with `/var/lib/jenkins/jobs` in this main function as well as the `validate_image` script. Do we care that the use case for this script is restricted to Jenkins, or are we going to solve that issue later if we port/refactor this into python?

ewiseblatt:
For the time being just make sure the script is used as intended because it
is making serious assumptions (e.g. the "rm -rf").
I'm open to other ideas too. For example not rm -rf'ing the directory (and
let the jenkins script calling this one do that) or checking that (if it
exists) it is a git directory and there are no local changes in it.

What I dont want to happen is that I check this out to make a change, then
I test the change by running it, and I loose my git repository (with all
the other branches of changes I am working on too but never synced back to
github because they are local experiments). That would be a bad day.

- eric

jtk54:
I modified this to check that the `cwd` is in a Jenkins workspace and only run the `rm -rf` if that's the case. Otherwise, it just checks out the branch supplied by `$GITHUB_REPOSITORY_BRANCH`. I also removed the dependency on the build service account pointed out. This tested fine in our Jenkins instance.

jtk54:
Updated and tested again after the last batch of comments, Jenkins succeeded.

jtk54:
@ewiseblatt any more comments on this or are we good to merge?

ewiseblatt:
lgtm

PR #1372, opened 2017-01-17
Switch from deploying replication controllers to replica sets

lwander:
I also dropped the startup command in the deck image that's no longer necessary since deck now runs on apache.

### DIFF
--- a/experimental/kubernetes/simple/rs/data-redis-master.yaml
+++ b/experimental/kubernetes/simple/rs/data-redis-master.yaml
@@ -1,5 +1,5 @@
-apiVersion: v1
-kind: ReplicationController
+apiVersion: extensions/v1beta1
+kind: ReplicaSet
 metadata:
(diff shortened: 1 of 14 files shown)


lwander:
@duftler @skim1420 @jtk54 PTAL

skim1420:
Nice, thanks.

dforrest88:
Are replication controllers still supported for auto-discovering applications deployed into the cluster via kubectl, or do they need to be deployed as replica sets in order to be picked up by Spinnaker? I've deployed a replication controller via kubectl to a cluster that my Spinnaker is connected to, and it doesn't appear to be loading into the Applications view properly. I'm wondering if I need to convert from Replication Controllers to Replica Sets.

lwander:
It should still be picked up by Spinnaker - this commit just changes the default installation of Spinnaker into Kubernetes.

PR #1376, opened 2017-01-20
Added source Annotator class for Spinnaker git repos.

jtk54:
Adds a class for determining semantic version bumps based on conventional-changelog commit messages and `sX.Y.Z` Spinnaker product tags. Tested against https://github.com/jtk54/changelog-test, which is set up how Spinnaker's components would look like (tags and commit messages).

### DIFF
--- a/dev/annotate_source.py
+++ b/dev/annotate_source.py
@@ -83,6 +83,18 @@ class VersionBump:
     self.__minor = minor
     self.__patch = patch
 
+  def __repr__(self):
+    return 'version_str: %s, major: %s, minor: %s, patch: %s' % (self.version_str,
+                                                                 self.major,
+                                                                 self.minor,
+                                                                 self.patch)
+
+  def __eq__(self, other):
+    return (self.version_str == other.version_str
+            and self.major == other.major
+            and self.minor == other.minor
+            and self.patch == other.patch)
+
   @property
   def version_str(self):
     return self.__version_str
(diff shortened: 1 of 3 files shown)


dev/annotate_source.py
```diff
@@ -0,0 +1,235 @@
...
+  def __init__(self, message):
+    self.message = message
+
+
+class Annotator(Refresher):
+  """Provides semantic version tagging for Spinnaker repositories.
+
+  Each Spinnaker repository has tags that denote releases. These tags follow
+  semantic versioning. At the present time, there are two sets of tags in use
+  for the Spinnaker repositories: 'vX.Y.Z' for Netflix releases and 'sX.Y.Z'
```
jtk54: 'sX.Y.Z' is what I was suggesting as a Spinnaker product version tag. We could start with our own version number, and not have any interaction with the existing `vX.Y.Z` tags at all.

The issue with using a 'v*' tag is how gradle resolves the tag to use for the next version. When we use `-PuseLastTag=true` in gradle, it will sort the tags of the form 'X.Y.Z' and 'vX.Y.Z*' (including '-rc.*') and use the latest tag by semver sort, favoring tags without the '-rc.*' if there's a collision.

If Netflix uses the '-rc.*' namespace, and we use the non-rc namespace and keep rolling the current microservice versions forward in the Spinnaker product releases I believe these issues wind up being resolved. We can use 'vX.Y.Z-$build' to publish unique artifacts to our private repo when testing, and push a 'vX.Y.Z' tag to master on a formal validated release, which TravisCI will ignore.

@ewiseblatt @cfieber @ajordens What do you think? Any issues/misconceptions here? Is Netflix willing to use the '-rc.*' namespace for internal releases?

ewiseblatt:
-rc for netflix doesnt make sense.

Why cant we use something like 'version-' and let netflix keep using 'v'?

jtk54:
Actually you're right, conceptually 'version-X.Y.Z' works the same way as 'sX.Y.Z'. I think I misunderstood what you were suggesting. Disregard my comments above.

jtk54:
@ewiseblatt updated to use 'version-*' and did a bit of refactoring.

PR #1380, opened 2017-01-20
Redis RCs -> RS

ttomsu:
@lwander is this all that's needed to make Replication Controllers into ReplicaSets?


lwander:
Almost - you need to update the labels -> matchLabels as well: https://github.com/spinnaker/spinnaker/pull/1372/files#diff-5409647037fc1048b678fbdcef7a3259R14

ttomsu:
Updated. PTAL

lwander:
LGTM

PR #1381, opened 2017-01-23
Updated experimental monitoring support.

ewiseblatt:
@jtk54


ewiseblatt:
It was intentional. I renamed the Prometheus sample to avoid future
confusion. They are sample dashboards for the different systems.

PR #1385, opened 2017-01-26
Copy over production default configs.

judy:
Say you have a kubernetes cluster set up, and running the simple k8s install for Spinnaker, from the experimental/kubernetes/simple folder. When you run `bash scripts/startup-config.sh`, the first line copies over the latest default config files from the root. These files are all out of date. All this commit does is copy over the newest ones.

The alternative, I suppose, is to take these default configs out altogether and add them to the `.gitignore` file, so when you're deploying you can easily see changes to your own `*-local.yml` files. Because otherwise, this will have to be done all the time. Please let me know if you would rather see a PR with that done.

### DIFF
--- a/experimental/kubernetes/simple/config/clouddriver.yml
+++ b/experimental/kubernetes/simple/config/clouddriver.yml
@@ -24,7 +24,7 @@ aws:
   # info on setting environment variables.
   enabled: ${providers.aws.enabled:false}
   defaults:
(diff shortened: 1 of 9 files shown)


judy:
Disregard this. I didn't see this [matching PR](https://github.com/spinnaker/spinnaker/pull/1304). I'm going to submit a new one with the stale configs deleted. This should be cleaner going forward, if you think?

PR #1387, opened 2017-01-26
Clean up replica sets during cleanup.

judy:
In commit 3338be8718c20a, the k8s install switched from using
replication controllers to using replication sets. This updates the
cleanup script to delete the replication sets too. Note that I left in
deleting replication controllers for older Spinnaker installs.


ewiseblatt:
@lwander ping

lwander:
This was fixed in another commit

PR #1389, opened 2017-01-26
Fix experimental docker-compose to be able to run reconfigure

tecnobrat:
This is a workaround due to some changes made in the `deck` Dockerfile about 10 days ago.  The `settings.js` file is no longer automatically regenerated when you load deck for the first time.

So we need to allow a way to regenerate the `settings.js` file.  With these changes in place, if you run:

```
DOCKER_IP=`docker-machine ip` docker-compose run deck /opt/spinnaker/runtime/reconfigure_spinnaker.sh
```

Then you should end up with a working settings.js file.  I'm not sure where in the README this will fit in, and I know it won't work for remote docker machines.

### DIFF
--- a/experimental/docker-compose/docker-compose.override.yml
+++ b/experimental/docker-compose/docker-compose.override.yml
@@ -6,9 +6,12 @@ clouddriver:
 echo: 
   volumes: 
     - "../../config:/opt/spinnaker/config"
-#deck:
-#  volumes: 
-#    - "../../config/settings.js:/deck/settings.js"
+deck:
+  volumes: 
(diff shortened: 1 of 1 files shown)


lwander:
Good catch, thanks. This LGTM - @tomaslin knows more about docker-compose than me though.

ewiseblatt:
@lwander - did you mean to submit this or were you asking @tomaslin?

lwander:
I was hoping Tomas could double check

ewiseblatt:
@tomaslin ping

Issue #1146, opened 2016-09-13
Add user data and user data type to server group details pane


drmaas:
Need to add an 'Advanced Settings' tab, where userData and userDataType will go.

PR #1390, opened 2017-01-27
(Azure Windows) Add 'object_id' as a configuration parameter as it's required by packer when baking Windows images

robumberger:
Related PR [#1146](https://github.com/spinnaker/orca/pull/1146) has been opened in Ocra. Another related PR will be opened in Rosco that requires this change

### DIFF
--- a/config/spinnaker.yml
+++ b/config/spinnaker.yml
@@ -297,12 +297,13 @@ providers:
     primaryCredentials:
       name: my-azure-account
 
-      # To set Azure credentials, enter your Azure supscription values for:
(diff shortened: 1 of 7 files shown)


config/spinnaker.yml
```diff
@@ -298,11 +298,12 @@ providers:
       name: my-azure-account
 
       # To set Azure credentials, enter your Azure supscription values for:
```
duftler: s/supscription/subscription (Since you're in the neighborhood.)

robumberger: Fixed

PR #1397, opened 2017-02-02
Deleting replicaset as well as replicationcontrollers

olvesh:
Still deleting both in case someone has not migrated yet.


lwander:
Thanks @olvesh

PR #1404, opened 2017-02-03
Fixed tag regex to handle large build numbers.

jtk54:


dev/annotate_source.py
```diff
@@ -136,7 +136,7 @@ class Annotator(Refresher):
   """
 
   # regex for 'version-X.Y.Z-$build' versions
-  TAG_MATCHER = re.compile('^version-[0-9]+\.[0-9]+\.[0-9]+-[0-9]$')
+  TAG_MATCHER = re.compile('^version-[0-9]+\.[0-9]+\.[0-9]+-[0-9]+$')
```
ewiseblatt: you can use \d instead of [0-0]

PR #1411, opened 2017-02-07
fixing simple cleanup script for replica sets

emjburns:
@lwander updating this script to cleanup the replica sets.


lwander:
Thanks :) could you keep `rc` in that list in case people are migrating?

emjburns:
@lwander fixed to keep rc!

lwander:
Thanks

PR #1415, opened 2017-02-08
Add basic azure storage to change_cassandra script

EricJizbaMSFT:


EricJizbaMSFT:
@spinnaker/reviewers @duftler @scotmoor @rguthriemsft PTAL

ewiseblatt:
LGTM
Do you guys use InstallSpinnaker.sh?

EricJizbaMSFT:
Yes we have some 'quickstart templates' that use InstallSpinnaker.sh. I noticed that if you pass '--noinstall_cassandra', it will default to gcs, which is fine. However, I wanted to leverage this script in our templates to switch to Azure

PR #1419, opened 2017-02-10
Dont require --jenkins_master or --jenkins_url for bake_and_deploy_test.

ewiseblatt:
@jtk54


ewiseblatt:
The defaults were migrated to be the runtime values unless overriden by the
args.

jtk54:
SGTM

PR #1423, opened 2017-02-10
(appengine) config

danielpeach:
@lwander or @duftler please review.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -145,6 +145,28 @@ providers:
       projectName: ${OS_PROJECT_NAME}
       regions: ${OS_REGION_NAME:RegionOne}
 
+  appengine:
+    enabled: ${SPINNAKER_APPENGINE_ENABLED:false}
+
+    primaryCredentials:
+      name: my-appengine-account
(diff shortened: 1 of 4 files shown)


config/default-spinnaker-local.yml
```diff
@@ -145,6 +145,28 @@ providers:
...
+
+    primaryCredentials:
+      name: my-appengine-account
+      # The project is the Google Project ID for the project to manage with
+      # Spinnaker.
+      project: ${SPINNAKER_APPENGINE_PROJECT_ID:}
+      # The jsonPath is a path to the JSON service credentials downloaded
+      # from the Google Developer's Console.
+      jsonPath: ${SPINNAKER_APPENGINE_PROJECT_CREDENTIALS_PATH:}
+      # The path and password to an SSH private key to be used when connecting with
```
lwander: I'm guessing at most one of `ssh` or `https` credentials are required? I might highlight this here.

danielpeach: You can specify at deploy time which credential type to use (https://github.com/spinnaker/deck/pull/3231), so you can specify as many credentials as you want.

PR #1426, opened 2017-02-14
feat(annotator) Removed build number and add flag to force rebuild.

jtk54:


jtk54:
Refactored a bit to clean the logical primitives up. Reads much cleaner now I think, PTAL.

dev/annotate_source.py
```diff
@@ -170,27 +173,36 @@ def __partition_tags_on_pattern(self):
...
+                                           self.__build_number)
     # This tag is for gradle to use as the package version. It incorporates the
     # build number for uniqueness when publishing. This tag is of the form
     # 'X.Y.Z-$build_number' for gradle to use correctly. This is not pushed
     # to the upstream git repository.
-    first_dash_idx = next_tag.index('-')
+    first_dash_idx = next_tag_with_build.index('-')
     if first_dash_idx != -1:
-      gradle_version = next_tag[first_dash_idx + 1:]
+      gradle_version = next_tag_with_build[first_dash_idx + 1:]
```
ewiseblatt: I dont understand this. This takes version-x.y.z and makes gradle_version 'x.y.z' right?
I thought gradle_version ended in '-build'.

jtk54: `next_tag_with_build` has the `-$build` ending, so `gradle_version` has the requisite info.

dev/annotate_source.py
```diff
@@ -170,27 +173,36 @@ def __partition_tags_on_pattern(self):
...
+      version_bump [VersionBump]: Semver to add as a gradle build tag.
+    """
+    next_tag_with_build = '{0}-{1}'.format(version_bump_tag,
+                                           self.__build_number)
     # This tag is for gradle to use as the package version. It incorporates the
     # build number for uniqueness when publishing. This tag is of the form
     # 'X.Y.Z-$build_number' for gradle to use correctly. This is not pushed
     # to the upstream git repository.
-    first_dash_idx = next_tag.index('-')
+    first_dash_idx = next_tag_with_build.index('-')
```
jtk54: Yeah that makes more sense.

jtk54: Actually no it doesn't, that's a trivial check since there's always a string with at least '-'. I just removed that check.

PR #1429, opened 2017-02-16
Remove azure objectId from clouddriver.yml

EricJizbaMSFT:
This property is only used in Rosco for baking Windows images, not clouddriver.


EricJizbaMSFT:
@robumberger @rguthriemsft @scotmoor @duftler PTAL

(ObjectId was added in PR #1390 as a part of windows bake support)

robumberger:
@EricJizbaMSFT Correct, only used in Rosco. I don't see an issue with removing from those 2 files.

PR #1430, opened 2017-02-16
feat(build_release): Use gradle to publish builds instead of manual c…

jtk54:
…opy.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -185,6 +185,7 @@ class Builder(object):
       self.__background_processes = []
 
       os.environ['NODE_ENV'] = os.environ.get('NODE_ENV', 'dev')
+      self.__build_number = options.build_number
       self.__options = options
       self.refresher = refresh_source.Refresher(options)
       if options.bintray_repo:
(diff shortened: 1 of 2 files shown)


dev/build_release.py
```diff
@@ -211,28 +212,62 @@ def determine_gradle_root(self, name):
       gradle_root = name if name != 'spinnaker' else self.__project_dir
       return gradle_root
 
-  def start_build_target(self, name, target):
-      """Start a subprocess to build the designated target.
+  def start_candidate_release(self, name):
```
ewiseblatt: Use terminology we established earlier.

candidate releases dont start until builds have been verified.
The build part is initiating an unverified build.

However this function is merely building a subsystem.
Where it falls into the release process is a matter the caller's context.
Dont assume that here. This is just the primitive operator that the
release process is composed with.

dev/build_release.py
```diff
@@ -211,28 +212,62 @@ def determine_gradle_root(self, name):
...
+    ]
+    if name == 'deck' and not 'CHROME_BIN' in os.environ:
+      extra_args.append('-PskipTests')
+
+    # Currently spinnaker is in a separate location
+    gradle_root = self.determine_gradle_root(name)
+    print 'Building and publishing {name}...'.format(name=name)
+    return BackgroundProcess.spawn(
+      'Building and publishing {name}...'.format(name=name),
+      'cd "{gradle_root}"; ./gradlew {extra} candidate'.format(
```
ewiseblatt: I dont like "candidate" here. Or if it is we need to change our terminology. However calling these things candidates prematurely causes confusion later as we discovered earlier on defining the process.

dev/build_release.py
```diff
@@ -211,28 +212,62 @@ def determine_gradle_root(self, name):
...
+    package version for each component.
+
+    Args:
+      name [string]: Name of the subsystem repository.
+
+    Returns:
+      BackgroundProcess
+    """
+    jarRepo = self.__options.jar_repo
+    parts = self.__options.bintray_repo.split('/')
```
ewiseblatt: Do these come from gradle properties?
I'm ok with environment variables.
I thought the current gradle plugins used properties.

dev/build_release.py
```diff
@@ -529,6 +511,14 @@ def init_argument_parser(cls, parser):
         '--bintray_repo', default='',
         help='Publish to this bintray repo.\n'
              'This requires BINTRAY_USER and BINTRAY_KEY are set.')
+      parser.add_argument(
```
ewiseblatt: jar repo shouldnt have a default.
Maybe this should be a property?

Is this jar repo write only or is it also read?
When other people run this script, they shouldnt publish to the official repo, but should be able to read form it rather than building everything themselves. But if they do build things, they should read those from their repo. So this should be for writing, and there should be a path set somewhere that this has precedence of the existing external path for jar files in this repo but not elsewhere.

dev/build_release.py
jtk54: `candidate` is the Nebula gradle task name. The only other choice that behaves the way we want is `final` at the moment, and that doesn't seem better. I'm not too sure about our ability to rename these.

dev/build_release.py
jtk54: These are flags passed into this script, which can be populated by environment variables.

dev/build_release.py
jtk54: Ok, I can rename the function but unfortunately the gradle task that is executed has to be either `candidate` or `final`.

ewiseblatt:
ok. add a comment.

dev/build_release.py
jtk54: First part makes sense. I'm confused at the second.

The jar repo is read/write as long as you have a valid key with permission to that repo or it is public. I agree that everyone shouldn't be able to publish to the official repo, but I don't understand why users would want to read from the official repo in this case. This is for building from source, not copying upstream, prebuilt artifacts. As you put it, this is for building artifacts and writing to `bintray_repo` and `jar_repo`. Consuming artifacts from those repos is a separate concern.

ewiseblatt: I could be confused.

If I am building jars I put them in this repository. Fine. But I dont build every jar I depend on, only the ones defined in this [git] repository. I dont know how this works under the covers, but isnt clear how the precedence and scoping works (or how things get removed so that I can use official external jars even if I may have built those projects myself earlier so the same versions are in my repo (built were built with a different tool chain).

jtk54: This `jar_repo` is the target repo for the microservice jars, not the dependency jars. For example, the jars produced from testing this script: https://bintray.com/spinnaker-team/jacobkiefer-jars. My understanding is all the dependency jars are downloaded to a local repository when the microservices are being built, and are pulled at whatever version is specified in `build.gradle`.

ewiseblatt:
So what happens if I want to build my own kork?

jtk54:
Kork would need to be built and published and the up-to-date version specified in [spinnaker dependencies](https://github.com/spinnaker/spinnaker-dependencies/blob/master/src/spinnaker-dependencies.yml#L4)

dev/build_release.py
```diff
@@ -529,6 +518,14 @@ def init_argument_parser(cls, parser):
...
+        '--jar_repo', default='',
+        help='Publish produced jars to this repo.\n'
+             'This requires BINTRAY_USER and BINTRAY_KEY are set.')
+      parser.add_argument(
+        '--build_number', default='',
```
ewiseblatt: Is this required? What's the default or is it left out if not specified?

dev/build_release.py
```diff
@@ -185,6 +185,7 @@ def __init__(self, options):
       self.__background_processes = []
 
       os.environ['NODE_ENV'] = os.environ.get('NODE_ENV', 'dev')
+      self.__build_number = options.build_number or os.environ.get('BUILD_NUMBER', '')
```
ewiseblatt: set the argparser default to os.environ.get('BUILD_NUMBER', '')

dev/build_release.py
```diff
@@ -211,36 +212,77 @@ def determine_gradle_root(self, name):
...
+      raise ValueError(
+          'Expected --bintray_repo to be in the form <owner>/<repo>')
+    org, packageRepo = parts[0], parts[1]
+    bintray_key = os.environ['BINTRAY_KEY']
+    bintray_user = os.environ['BINTRAY_USER']
+
+    extra_args = [
+      '--stacktrace',
+      '-Prelease.useLastTag=true',
+      '-PbintrayPackageBuildNumber={number}'.format(number=self.__build_number),
```
ewiseblatt: build_number could be empty. is that ok?

jtk54: The gradle task interprets an empty string as 'no build number' and names the package accordingly.

PR #1431, opened 2017-02-16
Allow server group tests to test either zonal or regional.

ewiseblatt:
@danielpeach 
Please review. After this goes in, could you look at cleaning up the payloads to be more concise -- not using zone and/or region where they arent needed, etc?

@duftler FYI per my suggestion last night to have caught the recent regional MIG bug


danielpeach:
LGTM. I'll take a look at the payloads.

PR #1432, opened 2017-02-16
Remove rootFolder from AZS config

EricJizbaMSFT:
For a few reasons:
1. Simplify the config
2. Match the folder structure for other providers, which are only 2 levels deep not 3 (in front50.yml, s3.bucket == azs.storageAccountName and s3.rootFolder == azs.storageContainerName)
3. Fix a bug where the 'last_modified' files were stored at the container level and everything else was stored at the (now removed) rootFolder level. Now everything is at the container level.

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -291,8 +291,7 @@ services:
       enabled: false
       storageAccountName:
       storageAccountKey:
-      storageContainerName: spinnaker
-      rootFolder: front50
+      storageContainerName: front50
 
   echo:
     # Persistence mechanism to use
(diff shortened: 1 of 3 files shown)


EricJizbaMSFT:
@scotmoor @rguthriemsft @duftler @ewiseblatt PTAL

We realize this would be a breaking change for existing users of azure storage, but want to fix while it's new and before 1.0. When this and it's companion front50 PR https://github.com/spinnaker/front50/pull/199 are released, I will post in slack telling any possible users to follow the migration steps so that their data is not lost (download all metadata, update & restart front50, upload new metadata)

EricJizbaMSFT:
The storageAccountName is analogous to your "bucket" (aka it's in a global namespace and has to be unique). The container is just a folder in the "storageAccount" and has no global or user uniqueness requirement (it just has to be unique within the storageAccount). I think my comment in the other PR with an example will help clarify.

PR #1433, opened 2017-02-16
added app engine smoke test

danielpeach:
@ewiseblatt please review.

### DIFF
--- a/testing/citest/spinnaker_testing/spinnaker_test_scenario.py
+++ b/testing/citest/spinnaker_testing/spinnaker_test_scenario.py
@@ -280,6 +280,20 @@ class SpinnakerTestScenario(sk.AgentTestScenario):
         '--kube_credentials',
         dest='spinnaker_kubernetes_account',
         help='DEPRECATED. Replaced by --spinnaker_kubernetes_account')
+    
+  @classmethod
+  def _initAppengineOperationConfigurationParameters(cls, parser, defaults):
+    """Initialize arguments for configuring operations for App Engine.
+    
+      parser: [argparse.ArgumentParser]
+      defaults: [dict] Default binding value overrides.
+         This is used to initialize the default commandline parameters.      
+    """
+    parser.add_argument(
+      '--spinnaker_appengine_account',
+      default=defaults.get('SPINNAKER_APPENGINE_ACCOUNT', None),
+      help='Spinnaker account name to use for test operations against'
+           'App Engine. Only used when managing resources on App Engine.')
 
   @classmethod
   def _initOperationConfigurationParameters(cls, parser, defaults):
@@ -300,6 +314,7 @@ class SpinnakerTestScenario(sk.AgentTestScenario):
     cls._initGoogleOperationConfigurationParameters(parser, defaults)
     cls._initAwsOperationConfigurationParameters(parser, defaults)
     cls._initKubeOperationConfigurationParameters(parser, defaults)
+    cls._initAppengineOperationConfigurationParameters(parser, defaults)
 
   @classmethod
   def _initObservationConfigurationParameters(cls, parser, defaults):
@@ -318,6 +333,13 @@ class SpinnakerTestScenario(sk.AgentTestScenario):
     parser.add_argument(
         '--aws_profile', default=defaults.get('AWS_PROFILE', None),
         help='aws command-line tool --profile parameter when observing AWS.')
+    
+    parser.add_argument(
+        '--appengine_credentials_path',
+        default=defaults.get('APPENGINE_CREDENTIALS_PATH', None),
+        help='A path to the JSON file with credentials to use for observing'
+             ' tests run against Google App Engine. Defaults to the value set for'
+             '--gce_credentials_path, which defaults to application default credentials.')
 
 
   @classmethod
@@ -351,6 +373,11 @@ class SpinnakerTestScenario(sk.AgentTestScenario):
   def aws_observer(self):
     """The observer for inspecting AWS platform state, if configured."""
     return self.__aws_observer
+  
+  @property
+  def appengine_observer(self):
+    """The observer for inspecting App Engine platform state, if configured."""
+    return self.__appengine_observer
 
   def __init__(self, bindings, agent=None):
     """Constructor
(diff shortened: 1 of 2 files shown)


testing/citest/spinnaker_testing/spinnaker_test_scenario.py
```diff
@@ -318,6 +333,12 @@ def _initObservationConfigurationParameters(cls, parser, defaults):
     parser.add_argument(
         '--aws_profile', default=defaults.get('AWS_PROFILE', None),
         help='aws command-line tool --profile parameter when observing AWS.')
+    
+    parser.add_argument(
+        '--appengine_credentials_path',
```
ewiseblatt: Mention the default here is to use --gce_credentials_path whose default is application credentials.

I'm not sure how you'd use application credentials for appengine but an explicit service account for GCE.

testing/citest/tests/appengine_smoke_test.py
```diff
@@ -0,0 +1,314 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Integration test for App Engine. Primarily tests the deploy operation and upsert load balancer pipeline stage,
+which are relatively complex and not well covered by unit tests.
+
+Sample Usage:
+    Assuming you have created $PASSPHRASE_FILE (which you should chmod 400)
+    and $CITEST_ROOT points to the root directory of this repository
```
ewiseblatt: you can use ssh-agent in the shell instead of the passphrase file.

testing/citest/tests/appengine_smoke_test.py
```diff
@@ -0,0 +1,314 @@
...
+  def __init__(self, bindings, agent=None):
+    super(AppengineSmokeTestScenario, self).__init__(bindings, agent)
+    
+    self.TEST_APP = bindings['TEST_APP']
+    self.TEST_STACK = bindings['TEST_STACK']
+    self.__path = 'applications/%s/tasks' % self.TEST_APP
+    self.__gcp_project = bindings['APPENGINE_PRIMARY_MANAGED_PROJECT_ID']
+    self.__cluster_name = frigga.Naming.cluster(self.TEST_APP, self.TEST_STACK)
+    self.__server_group_name = frigga.Naming.server_group(self.TEST_APP, self.TEST_STACK)
+    self.__lb_name = 'citest'
```
ewiseblatt: If this is under your control, dont use a constant name. If it is derived, dont have it derive constant names. You want multiple tests to be able to run without necessarily conflicting with one another. If one test leaks the load balancer you dont want that to influence another run.

The google_smoke_test does this:
   self.__lb_name = '{app}-{stack}-{detail}'.format(
	app=bindings['TEST_APP'], stack=bindings['TEST_STACK'],
        detail=self.__lb_detail)

It should be updated to use frigga as you are elsewhere already.

testing/citest/tests/appengine_smoke_test.py
```diff
@@ -0,0 +1,314 @@
...
+    self.__repo_url = 'https://github.com/danielpeach/python-docs-samples.git'
```
ewiseblatt: Not sure what this is for, are you sure you want to reference your repo to anonymous people that may run this and copy and paste future things from it?

testing/citest/tests/appengine_smoke_test.py
```diff
@@ -0,0 +1,314 @@
...
+    self.__app_yaml_path = 'appengine/standard/hello_world/app.yaml'
+    self.__branch = 'master'
+    
+    self.pipeline_id = None
+    
+  def create_app(self):
+    contract = jc.Contract()
```
ewiseblatt: mention you arent testing create/delete app because they are well covered elsewhere so assume they work.

danielpeach:
I updated this. I've added flags to pass in the hard-coded values you were concerned with, though `lb_name` can't be dynamic, since it's determined by the App Engine application and not Spinnaker. 

However, just by the nature of how load balancers work in App Engine, there should be no issue between tests if there is a resource leak (or at least there is no danger of a name collision).

ewiseblatt:
What is the App Engine application? Is that this test? If so, then this
test could determine the name. Bake the test_id into the name, or like
everyone else, the APP_NAME (and STACK_NAME). In turn those bake the
test_id. At least this way you correlated the LB instance with the test
instance that created it.

danielpeach:
What you're asking will be possible after https://github.com/spinnaker/clouddriver/pull/1461 has been merged.

danielpeach:
@ewiseblatt updated.

testing/citest/spinnaker_testing/spinnaker_test_scenario.py
```diff
@@ -318,6 +333,13 @@ def _initObservationConfigurationParameters(cls, parser, defaults):
...
+        default=defaults.get('APPENGINE_CREDENTIALS_PATH', None),
+        help='A path to the JSON file with credentials to use for observing'
+             ' tests run against Google App Engine. Defaults to the value set for'
+             '--gce_credentials_path, which defaults to application credentials.')
```
ewiseblatt: application default credentials

testing/citest/tests/appengine_smoke_test.py
```diff
@@ -0,0 +1,339 @@
...
+  def new_agent(cls, bindings):
+    return gate.new_agent(bindings)
+
+  @classmethod
+  def initArgumentParser(cls, parser, defaults=None):
+    """Initialize command line argument parser."""
+    super(AppengineSmokeTestScenario, cls).initArgumentParser(
+          parser, defaults=defaults)
+    parser.add_argument('--source_repo_url', default=None, 
+                        help='URL of a source code repository used by Spinnaker to deploy to App Engine.')
```
ewiseblatt: Are these required? Can the test just run without this ?

testing/citest/tests/appengine_smoke_test.py
```diff
@@ -0,0 +1,339 @@
...
+    # Python is clearly hard-coded as the runtime here, but we're just asking App Engine to be a static file server.
+    self.__app_yaml = ('\n'.join(['runtime: python27',
+                                  'api_version: 1',
+                                  'service: {service}',
+                                  'handlers:',
+                                  ' - url: /.*',
+                                  '   static_dir: .']).format(service=self.__lb_name))
+    
+    self.__repo_url = bindings['SOURCE_REPO_URL']
+    self.__app_directory_root = bindings['APP_DIRECTORY_ROOT']
```
ewiseblatt: is it ok these are empty by default

testing/citest/tests/appengine_smoke_test.py
danielpeach: Yes - this argument and the one you've pointed out below are required. The test now throws an error if you haven't supplied these.

PR #1434, opened 2017-02-17
Dont update gcloud on boot

ewiseblatt:
@lwander or @spinnaker/google-reviewers


lwander:
Just wondering, why not? 

I added that when the default installation was a few versions behind of what we needed feature-wise to get Kubernetes configured - I'm guessing now everything works with what comes in a GCE VM?

PR #1440, opened 2017-02-24
Make the update_run_scripts in refresh_source.py default to True.

dsimonto:
And added `--noupdate_run_scripts` to allow for disabling it.

This fixes an issue observed yesterday where refresh_source no longer updates the run scripts by default, which broke the "Setting Up Spinnaker for Development" instructions in the spinnaker/README.

FYI @jtk54

### DIFF
--- a/dev/refresh_source.py
+++ b/dev/refresh_source.py
@@ -538,9 +538,13 @@ bash -c "(npm start >> '$LOG_DIR/{name}.log') 2>&1\
                           help='Pull from this github user\'s repositories.'
                                ' If the user is "default" then use the'
                                ' authoritative (upstream) repository.')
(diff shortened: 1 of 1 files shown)


lwander:
Thanks. Could you take a look at this @ewiseblatt?

jtk54:
@ewiseblatt that's correct. We can manually flag this in our build scripts.

jtk54:
... I had one change to this that needed addressed -- the action of that flag is `store_true`, so we needed to change that for this to work properly. This is broken after the merge.

jtk54:
Couldn't type fast enough.

ewiseblatt:
I thought the one was true and the other was false.

PR #1444, opened 2017-02-28
feat(monitoring): Optionally install spinnaker-monitoring from Instal…

ewiseblatt:
…lSpinnaker

This is disabled by default for now but can be enabled with --monitoring.
Note that monitoring installs the monitoring components but still does not
activate them because they need further deployment customization.

Other internal changes:
 - Deprecated the old --google_monitoring flag.
 - Install --monitoring when baking GCE images and added hook to trigger on
first_google_boot.
  - Updated the dev scripts in order to build spinnaker-monitoring.
    In doing so I added back the old curl upload as an option because the
    nebula upload mechanism requires a clean git client, which isnt developer
    friendly. The spinnaker-monitoring package is currently considered optional.

### DIFF
--- a/InstallSpinnaker.sh
+++ b/InstallSpinnaker.sh
@@ -29,6 +29,7 @@ QUIET=false
 # Install dependencies but not spinnaker itself.
 DEPENDENCIES_ONLY=false
 
+MONITORING_CHOICES=(datadog, prometheus, stackdriver, undef)
 
 # We can only currently support limited releases
 # First guess what sort of operating system
@@ -75,7 +76,7 @@ function print_usage() {
 usage: $0 [--cloud_provider <aws|google|azure|none>]
     [--aws_region <region>] [--google_region <region>] [--azure_region <region>]
     [--quiet] [--dependencies_only]
-    [--google_cloud_logging] [--google_cloud_monitoring]
+    [--google_cloud_logging] [--monitoring]
     [--repository <debian repository url>]
     [--local-install] [--home_dir <path>]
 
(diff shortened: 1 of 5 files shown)


ewiseblatt:
@jtk54

InstallSpinnaker.sh
```diff
@@ -217,8 +221,13 @@ function process_args() {
...
         GOOGLE_CLOUD_LOGGING="true"
         ;;
-      --google_cloud_monitoring)
-        GOOGLE_CLOUD_MONITORING="true"
+      --monitoring)
+        MONITORING="$1"
+        shift
+        if [[ ! ${MONITORING_CHOICES[*]} =~ "$MONITORING" ]]
+        then
+            echo "Unknown monitoring choice '$MONITORING'."
```
jtk54: Exit here?

InstallSpinnaker.sh
```diff
@@ -650,6 +645,16 @@ function install_spinnaker() {
     fi
   fi
 
+  if [[ "$MONITORING" != "" ]] && [[ "$MONITORING" != "none" ]]; then
+     sudo apt-get install -y --force-yes spinnaker-monitoring-daemon
```
jtk54: Does `-y` do something different than `--force-yes` here?

dev/build_release.py
```diff
@@ -356,6 +372,11 @@ def publish_to_bintray(self, source, package, version, path, debian_tags=''):
                   package=package, pkg_url=pkg_url)
 
               # All the packages are from spinnaker so we'll hardcode it.
+              # Not spinnaker-monitoring is a github repo with two packages in it.
```
jtk54: s/Not/Note/

dev/build_release.py
```diff
@@ -434,6 +454,51 @@ def start_copy_file(self, source, target):
         shutil.copy(source, target)
         return NO_PROCESS
 
+  def start_copy_debian_target(self, name):
```
jtk54: Does this use the 'simple' debian index or the structured one when it copies to the remote debian repository?

PR #1446, opened 2017-02-28
feat(bom): BOM generation tool.

jtk54:
Note that this script requires `hal` to be installed. This generates the BOM file and writes it to local. `hal` will be responsible for publishing the BOM to the correct bucket.

### DIFF
--- a/dev/annotate_source.py
+++ b/dev/annotate_source.py
@@ -138,9 +138,9 @@ class Annotator(Refresher):
   # regex for 'version-X.Y.Z' versions
   TAG_MATCHER = re.compile('^version-[0-9]+\.[0-9]+\.[0-9]+$')
 
-  def __init__(self, options):
+  def __init__(self, options, path=None):
     self.__next_tag = options.next_tag
-    self.__path = options.path
+    self.__path = path or options.path
     self.__initial_branch = options.initial_branch
     self.__stable_branch = options.stable_branch
     self.__build_number = options.build_number or os.environ.get('BUILD_NUMBER', '')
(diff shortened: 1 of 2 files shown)


dev/annotate_source.py
```diff
@@ -138,18 +138,16 @@ class Annotator(Refresher):
   # regex for 'version-X.Y.Z' versions
   TAG_MATCHER = re.compile('^version-[0-9]+\.[0-9]+\.[0-9]+$')
 
-  def __init__(self, options):
+  def __init__(self, options, path=None):
     self.__next_tag = options.next_tag
-    self.__path = options.path
+    self.path = path or options.path
```
ewiseblatt: If this was just to expose it, you could @property it with a getter.
The difference is that it would not be settable from the outside.

dev/generate_bom.py
```diff
@@ -0,0 +1,116 @@
...
+
+  @classmethod
+  def init_argument_parser(cls, parser):
+    """Initialize command-line arguments."""
+    parser.add_argument('--base_dir', default='.',
+                        help="Base directory containing the component's git repositories as subdirectories.")
+    super(BomGenerator, cls).init_argument_parser(parser)
+
+  def write_bom(self):
+    output_yaml = {SERVICES: {}}
```
ewiseblatt: I think the bom should contain some kind of traceability to who/how it was generated, especially if there are different sources of releases (official one vs customer generated)

jtk54: Added a comment that says hostname and timestamp.

dev/annotate_source.py
jtk54: Added properties and a setter for the prop I set.

PR #1453, opened 2017-03-01
feat(bom): Publish BOM using Halyard.

jtk54:


dev/generate_bom.py
```diff
@@ -91,13 +91,34 @@ def write_bom(self):
     toplevel_with_build = '{0}-{1}'.format(toplevel_version, self.build_number)
     output_yaml[VERSION] = toplevel_with_build
     bom_file = '{0}.yml'.format(toplevel_with_build)
-    with open(os.path.join(self.__base_dir, bom_file), 'w') as output_file:
+    self.__write_bom(bom_file, output_yaml)
+    self.__publish_bom(bom_file)
+    self.__write_bom('nightly.yml', output_yaml) # Publish a 'nightly' BOM for easy promotion later.
```
lwander: Make sure the version in `output_yaml` is also `nightly`

dev/generate_bom.py
```diff
@@ -91,13 +91,35 @@ def write_bom(self):
...
-      print 'Wrote BOM to {0}.'.format(bom_file)
+      print 'Wrote BOM to {0}.'.format(filename)
+
+  def __publish_bom(self, bom_path):
+    """Publishes the BOM using Halyard.
+
+    Assumes that Halyard is installed and correctly configured on the current
+    machine.
+    """
+    run_quick('hal admin publish bom --color false --bom-path {0}'
```
ewiseblatt: how, where, and when does hal get installed?

jtk54: I added a comment to this script that Halyard is a dependency, and added another commit that installs Halyard as a part of `install_development.sh`.

ewiseblatt:
LGTM

PR #1455, opened 2017-03-01
fix(bom): Fail if Halyard fails.

jtk54:


dev/generate_bom.py
```diff
@@ -80,8 +80,12 @@ def write_bom(self):
       output_yaml[SERVICES][comp] = version_entry
 
     # Current publicly released version of Spinnaker product.
-    curr_version = run_quick('hal versions latest --color false', echo=False).stdout.strip()
-    print curr_version
+    result = run_quick('hal versions latest --color false', echo=False)
```
lwander: Why not `echo=true`? The return code might not be descriptive enough here.

jtk54: Made the logging output stronger by including the `stdout`, which has `stderr` appended to it in failures.

lwander: Ah fantastic

PR #1464, opened 2017-03-07
feat(halyard): Added halyard to list of built artifacts

lwander:


dev/build_release.py
```diff
@@ -63,7 +63,7 @@
 from spinnaker.run import run_quick
 
 
-SUBSYSTEM_LIST = ['clouddriver', 'orca', 'front50',
+SUBSYSTEM_LIST = ['clouddriver', 'orca', 'front50', 'halyard',
```
jtk54: Please add halyard to https://github.com/spinnaker/spinnaker/blob/master/dev/refresh_source.py also. I'd probably add this to `ADDITIONAL_SUBSYSTEMS` since it's not a microservice.

lwander: Is this enough? https://github.com/spinnaker/spinnaker/blob/master/dev/refresh_source.py#L94

jtk54: Yep

PR #1465, opened 2017-03-07
feat(build): Generate Docker images during build script.

jtk54:


dev/build_release.py
```diff
@@ -287,11 +288,32 @@ def start_subsystem_build(self, name):
...
+
+    Args:
+      name [string]: Name of the subsystem repository.
+
+    Returns:
+      BackgroundProcess
+    """
+    gradle_root = self.determine_gradle_root(name)
+    return BackgroundProcess.spawn(
+      'Building and publishing container image for {name}...'.format(name=name),
```
jtk54: If the other registries follow the pattern of `${provider}/${project}/${repo}:${tag}` then I'm all for adding the registry as an argument to this script. Also I believe we'd need a `cloud-builder/docker` for the registry provider as well: https://github.com/jtk54/spinnaker/blob/e983cc878419f3ca64bc48ca4eaf8ec56ff52b9b/dev/generate_bom.py#L38 -- I'll check which providers are supported with the Container Builder team.

ewiseblatt: it's still using gcloud right? isnt that an issue?

lwander: That's a good point - the only problem is that each registry (that has a builder) has slightly different semantics for triggering a build. We could provide an optional local docker builder that would work for any registry. i.e. if gcloud isn't installed `docker build -t <name> .` in the project's directory does the trick.

ewiseblatt: If that isnt hard, I think we should do it. I dont know for sure this is important, but I suspect it will be since this is an essential activity, most of the user base wouldnt normally have gcloud installed, dont need it otherwise, and probably do have a normal docker tooling and the support community that comes with it.

If it is hard or complicated, then we can file a bug and come back to it later. Getting this end to end running is the short term priority, but it's easier and better to fix this now if it isnt a big detour.

jtk54:
Added and tested the option to use Docker for building the container images, PTAL.

dev/generate_bom.py
```diff
@@ -49,17 +64,68 @@ class BomGenerator(Annotator):
...
-    parser.add_argument('--base_dir', default='.',
+    parser.add_argument('--base_dir', default='.', required=True,
                         help="Base directory containing the component's git repositories as subdirectories.")
+    parser.add_argument('--gcr_project', default='',
+                        help="GCR project we use in Google Container Builder configuration files.")
+    parser.add_argument('--docker_registry', default='',
+                        help='Docker registry to push the container images to.')
     super(BomGenerator, cls).init_argument_parser(parser)
 
+  def write_container_builder_config(self):
```
ewiseblatt: could you rename this to be clear this is for gcr?

dev/generate_bom.py
```diff
@@ -49,17 +64,68 @@ class BomGenerator(Annotator):
...
     super(BomGenerator, self).__init__(options)
 
   @classmethod
   def init_argument_parser(cls, parser):
     """Initialize command-line arguments."""
-    parser.add_argument('--base_dir', default='.',
+    parser.add_argument('--base_dir', default='.', required=True,
                         help="Base directory containing the component's git repositories as subdirectories.")
+    parser.add_argument('--gcr_project', default='',
+                        help="GCR project we use in Google Container Builder configuration files.")
```
ewiseblatt: help text should indicate these are disjoint depending on where you are publishing. --docker_registry is probably generic. Different providers will replciate --gcr_project with more variations. You dont need to indicate that (maybe in a comment) but keep that in mind.

dev/build_release.py
```diff
@@ -287,11 +288,42 @@ def start_subsystem_build(self, name):
...
+
+    Returns:
+      BackgroundProcess
+    """
+    gradle_root = self.determine_gradle_root(name)
+    print self.__options
+    if self.__options.google_container_builder:
+      return BackgroundProcess.spawn(
+        'Building and publishing container image for {name} with Google Container Builder...'.format(name=name),
+        'cd "{gradle_root}"; gcloud container builds submit --account={account} --config="../{name}.yml" .'
```
lwander: Are `{name}.yml` the same file in both cases?

dev/generate_bom.py
lwander: I think another (maybe more future-proof) way of doing this would be to have `--docker_registry` be required for any docker build (in gcr world it would be `gcr.io/<project>`, in docker it would be `index.docker.io/<user>`, etc...), and then have a second flag indicating which builder to use. Right now, that second flag would only support `gcr` and `local`, but could be expanded to work for quay, dockerhub, etc...

dev/build_release.py
jtk54: They are. I can differentiate them.

jtk54:
Responded to comments and re-tested, everything looks ok. Merging if there are no complaints.

dev/generate_bom.py
```diff
@@ -49,17 +64,67 @@ class BomGenerator(Annotator):
...
     super(BomGenerator, self).__init__(options)
 
   @classmethod
   def init_argument_parser(cls, parser):
     """Initialize command-line arguments."""
-    parser.add_argument('--base_dir', default='.',
+    parser.add_argument('--base_dir', default='.', required=True,
                         help="Base directory containing the component's git repositories as subdirectories.")
+    parser.add_argument('--container_builder', default='gcb',
+                        help="Type of builder to use. Currently, the supported options are {'gcb', 'docker'}.")
```
lwander: Why do we have two sets of flags, one with `container_builder={gcb, docker}`, and one with `--{google_container_builder, docker}`?

jtk54: It's two ways to do the same thing, I'll make it consistent.

jtk54: Updated.

lwander:
LGTM

PR #1469, opened 2017-03-09
chore(dev): Cleaned up build_release

ewiseblatt:
@jtk54

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -23,9 +23,8 @@ The gradle script does not yet coordinate a complete build, so
 this script fills that gap for the time being. It triggers all
 the subsystem builds and then publishes the resulting artifacts.
 
-Publishing is typically to bintray. It is currently possible to publish
-to a filesystem or storage bucket but this option will be removed soon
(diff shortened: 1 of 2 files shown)


dev/build_release.py
```diff
@@ -196,26 +122,17 @@ def __init__(self, options):
 
       os.environ['NODE_ENV'] = os.environ.get('NODE_ENV', 'dev')
       self.__build_number = options.build_number
-      self.__google_service_account = options.google_service_account
+      self.__gcb_service_account = options.gcb_service_account
       self.__options = options
+      if (self.__options.container_builder
```
jtk54: Could add another check that `container_builder` is populated when `build_container_image` is true.

ewiseblatt: I removed build_container_image as a flag.
The default container_builder is None. Setting the container builder expresses you want to build a container image. Not setting it expresses you dont so the flag is superfluous. Setting it empty ensure it is off.

jtk54: Ok got it.

PR #1472, opened 2017-03-10
feat(bom): Generate and write changelog.

jtk54:
This generates the changelog using `clog-cli` and writes it to a file. I'll add the changelog publication step soon.

### DIFF
--- a/dev/generate_bom.py
+++ b/dev/generate_bom.py
@@ -67,6 +67,9 @@ class BomGenerator(Annotator):
     self.__docker_registry = options.docker_registry
     self.__bom_file = ''
     self.__component_versions = {}
+    self.__changelog_start_hashes = {} # Hashes to start from when generating changelogs.
+    self.__toplevel_version = ''
+    self.__changelog_output = options.changelog_output
     super(BomGenerator, self).__init__(options)
 
   @classmethod
@@ -78,6 +81,8 @@ class BomGenerator(Annotator):
                         help="Type of builder to use. Currently, the supported options are {'gcb', 'docker'}.")
     parser.add_argument('--docker_registry', default='',
                         help="Docker registry to push the container images to.")
+    parser.add_argument('--changelog_output', default='',
+                        help="Output file to write the changelog to.")
     super(BomGenerator, cls).init_argument_parser(parser)
 
   def write_container_builder_gcr_config(self):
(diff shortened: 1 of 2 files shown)


dev/generate_bom.py
```diff
@@ -125,6 +127,38 @@ def write_docker_version_files(self):
...
+        print "Changelog generation failed for {0} with \n{1}\n exiting...".format(comp, result.stdout)
+        exit(result.returncode)
+      # Capitalize
+      comp_cap = comp[0].upper() + comp[1:]
+      changelog += '\n# {0}\n{1}'.format(comp_cap, result.stdout)
+    print 'Writing changelog...'
+    # Write the changelog with the toplevel version without the build number.
+    # This is ok since the changelog is only published if the toplevel version is released.
+    changelog_file = '{0}-changelog.md'.format(self.__toplevel_version)
+    with open(changelog_file, 'w') as clog:
```
ewiseblatt: You shoudl be able to control where the output goes other than CWD.

dev/generate_bom.py
```diff
@@ -125,6 +127,38 @@ def write_docker_version_files(self):
...
         cfg.write(docker_tag)
 
+  def generate_changelog(self):
+    """Generate a release changelog and write it to a file.
+
+    The changelog contains a section per microservice that describes the
+    changes made since the last Spinnaker release. It also contains the
+    version information as well.
+    """
+    changelog = 'Spinnaker {0}\n\n'.format(self.__toplevel_version)
```
ewiseblatt: changelog shoudl be a [] that you append lines to.
I guess you can append a blank.
You should probably informally sign it with a date and tool so when someone has it, they know where it came from.

dev/generate_bom.py
```diff
@@ -125,6 +127,38 @@ def write_docker_version_files(self):
...
+    for comp, hash in self.__changelog_start_hashes.iteritems():
+      version_bump = self.__component_versions[comp]
+      next_tag_with_build = '{0}-{1}'.format(version_bump.version_str,
+                                             self.build_number)
+      first_dash_idx = next_tag_with_build.index('-')
+      gradle_version = next_tag_with_build[first_dash_idx + 1:]
```
ewiseblatt: where did gradle come from? These are from git tags, right?

dev/generate_bom.py
jtk54: Do you mean specify the file or the output path with the file name as I have it now?

jtk54:
Added option for changelog output file, cleaned some things up -- merging if no complaints.

PR #1473, opened 2017-03-13
feat(build): Build spinnaker distribution to validate with citest.

jtk54:
This refactors our Jenkins build script (*GithubToBintray) out into a python script that
* Encapsulates a few release operations into one container.
* Runs the primitive operations in an order that won't leave the public BOM bucket or artifact repositories in an inconsistent state, i.e. we never publish a BOM that contains a non-existent package.
* Refactors the primitive classes to separate the logic slightly better.

### DIFF
--- a/dev/build_prevalidation.py
+++ b/dev/build_prevalidation.py
@@ -1,0 +1,70 @@
+#!/usr/bin/python
+#
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import os
+import sys
+
+from annotate_source import Annotator
+from build_release import Builder
+from generate_bom import BomGenerator
+from refresh_source import Refresher
+
+
+def __annotate_component(annotator, component):
+  """Annotate the component's source but don't include it in the BOM.
+  """
+  annotator.path = component
+  annotator.parse_git_tree()
+  annotator.tag_head()
+  annotator.delete_unwanted_tags()
+
+def init_argument_parser(parser):
+  # Don't need to init args for Annotator since BomGenerator extends it.
+  BomGenerator.init_argument_parser(parser)
+  Builder.init_argument_parser(parser)
+
+def main():
+  """Build a Spinnaker release to be validated by Citest.
+  """
+  parser = argparse.ArgumentParser()
+  init_argument_parser(parser)
+  options = parser.parse_args()
+
+  annotator = Annotator(options)
+  __annotate_component(annotator, 'spinnaker')
+  __annotate_component(annotator, 'halyard')
+  __annotate_component(annotator, 'spinnaker-monitoring')
+
+  bom_generator = BomGenerator(options)
+  bom_generator.determine_and_tag_versions()
+  if options.container_builder == 'gcb':
+    bom_generator.write_container_builder_gcr_config()
+  elif options.container_builder == 'docker':
+    bom_generator.write_docker_version_files()
+  else:
+    raise NotImplementedError('container_builder="{0}"'
+                              .format(options.container_builder))
+  Builder.do_build(options, options.build_number, options.container_builder)
+  # Load version information into memory and write BOM to disk. Don't publish yet.
+  bom_generator.write_bom()
+  bom_generator.publish_microservice_configs()
+  bom_generator.publish_boms()
+  bom_generator.generate_changelog()
+
+
+if __name__ == '__main__':
+  sys.exit(main())
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -117,17 +117,16 @@ def determine_package_version(gradle_root):
 class Builder(object):
   """Knows how to coordinate a Spinnaker release."""
 
-  def __init__(self, options):
+  def __init__(self, options, build_number=None, container_builder=None):
       self.__package_list = []
       self.__build_failures = []
       self.__background_processes = []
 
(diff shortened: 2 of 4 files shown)


dev/build_prevalidation.py
```diff
@@ -0,0 +1,70 @@
...
+  if options.container_builder == 'gcb':
+    bom_generator.write_container_builder_gcr_config()
+  elif options.container_builder == 'docker':
+    bom_generator.write_docker_version_files()
+  else:
+    raise NotImplementedError('container_builder="{0}"'
+                              .format(options.container_builder))
+  Builder.do_build(options, options.build_number, options.container_builder)
+  bom_generator.write_bom()
+  bom_generator.publish_microservice_configs()
```
lwander: Does `write_bom` publish it as well? If so, it should happen after the configs are published.

jtk54: Yes it does. I factored the publishing out into another function and put it after the config push.

dev/build_prevalidation.py
```diff
@@ -0,0 +1,72 @@
...
+
+def main():
+  """Build a Spinnaker release to be validated by Citest.
+  """
+  parser = argparse.ArgumentParser()
+  init_argument_parser(parser)
+  options = parser.parse_args()
+
+  annotator = Annotator(options)
+  __annotate_component(annotator, 'spinnaker')
```
ewiseblatt: I'd prefer having the private functions appear first

dev/build_prevalidation.py
```diff
@@ -0,0 +1,72 @@
...
+  annotator.tag_head()
+  annotator.delete_unwanted_tags()
+
+def init_argument_parser(parser):
+  # Don't need to init args for Annotator since BomGenerator extends it.
+  BomGenerator.init_argument_parser(parser)
+  Builder.init_argument_parser(parser)
+
+def determine_project_root():
+  return os.path.abspath(os.path.dirname(__file__) + '/..')
```
ewiseblatt: os.dirname(os.path.dirname(__file__))

dev/build_release.py
```diff
@@ -117,17 +117,16 @@ def determine_package_version(gradle_root):
...
 
-  def __init__(self, options):
+  def __init__(self, options, build_number=None, container_builder=None):
       self.__package_list = []
       self.__build_failures = []
       self.__background_processes = []
 
       os.environ['NODE_ENV'] = os.environ.get('NODE_ENV', 'dev')
-      self.__build_number = options.build_number
+      self.__build_number = build_number or os.environ.get('BUILD_NUMBER')
```
ewiseblatt: this could still be None, right?

dev/generate_bom.py
```diff
@@ -202,9 +202,13 @@ def write_bom(self):
...
     self.write_bom_file(self.__bom_file, output_yaml)
-    self.publish_bom(self.__bom_file)
     output_yaml[VERSION] = 'nightly'
     self.write_bom_file('nightly.yml', output_yaml) # Publish a 'nightly' BOM for folks wanting to run bleeding-edge Spinnaker.
+
+  def publish_boms(self):
+    """Pushes the generated BOMs to a public GCS bucket for Halyard to use.
+    """
+    self.publish_bom(self.__bom_file)
     self.publish_bom('nightly.yml')
```
ewiseblatt: Why does this publish "nightly.yml" ?

dev/generate_bom.py
```diff
@@ -202,9 +202,13 @@ def write_bom(self):
     output_yaml[VERSION] = toplevel_with_build
     self.__bom_file = '{0}.yml'.format(toplevel_with_build)
     self.write_bom_file(self.__bom_file, output_yaml)
-    self.publish_bom(self.__bom_file)
     output_yaml[VERSION] = 'nightly'
```
ewiseblatt: where is "nightly" coming from and why is that hardcoded?
isnt a nightly build a specific context?

dev/build_release.py
jtk54: Yes, but the gradle task that ingests it handles None/empty string correctly.

dev/build_prevalidation.py
jtk54: I don't think this is used, I'm taking it out.

dev/generate_bom.py
jtk54: "nightly" is the bleeding edge, unstable set of microservices we build nightly. I added a flag to enable building and publishing this BOM, default to `False`.

dev/build_prevalidation.py
jtk54: Moved the functions around.

dev/generate_bom.py
```diff
@@ -83,6 +85,8 @@ def init_argument_parser(cls, parser):
                         help="Docker registry to push the container images to.")
     parser.add_argument('--changelog_output', default='',
                         help="Output file to write the changelog to.")
+    parser.add_argument('--nightly', default=False, action='store_true',
```
ewiseblatt: I was getting at "nightly" is an external policy, not part of this tool. Either the caller would alias the specific bom as a "nightly" or it would ask the tool to as you've done here. But again, "nightly" seems very one-off. This should be a means to ask for an alias. The default is None but the caller can pass in "nightly" -- or anything else.

I prefer having the tooling such that the caller can do this after the fact. That way you can have something more useful like "stable" (passed automated tests but not yet approved for release, or even release candidates for new major releases that might involve a communal approval process) which cannot be determine at the time this is called. Heck, you could even rollback a symbolic alias to point to an older version if you discover a problem after you set it.

It could be that this tool does the aliasing, but should be after the fact. You ask the tool to alias a specific version as another name. The reason for that is so the knowledge is encapsulated in this one tool. The control is left external as to when, why and what to alias as. Not a narrow special case hard wired into the tool. The special case is more efficient and convenient, but not by much. We're adding this switch to save one call from one client as opposed to support a mechanism that can be used for a more general range of use cases covering this special case as well as other more interesting and useful ones.

jtk54: Ok makes sense. For now, I'm going to add a flag to alias the 'real' BOM as `nightly.yml` at the time of creation so we can support that for 1.0. I like the after the fact aliasing, but I actually think that should go into `dev/promote_bom.py` since we're doing something similar there when we want to release.

jtk54: To be clear here, the flag will accept any string as the alias before '.yml'. We'll use this to do our 'nightly.yml' for the time being. Later this may exist elsewhere.

jtk54:
This tested successfully a few times, merging if there are no complaints.

PR #1474, opened 2017-03-14
(wip) feat(promote): Publish changelog as github gist.

jtk54:
**Work in progress, do not merge**

I wanted to get some feedback on this before I continue. This publishes a Spinnaker changelog as a Gist on behalf of a Github user. It uses `pygithub` to do so. A few questions I wanted to ask:
* Is it sufficient to publish a changelog to Gist and link to it from a 'Releases' page in readme.io (or another location) accompanying a version other release information (BOM, etc)? Note that this would require a dedicated Github 'Spinnaker release' user.
* Thoughts on adding `pygithub` as a dependency? I'd like to do this now, since I'm planning on using it to scrape Github issues to compile in the changelog sometime in the future.

I'm open to suggestions. @spinnaker/google-reviewers


duftler:
I don't have any real thoughts on using pygithub, other than that the license is ok. If it seems to be the dominant python->github library, then I'd say use it.

I think it would be better to embed the changelog in the releases page rather than link out to it. I did see some options for embedding gists directly in gitbook pages: https://www.npmjs.com/package/gitbook-plugin-gist

ewiseblatt:
I dont mind dependencies as long as they are easily and readily available.
However I think that the tools and artifacts should be self contained.
Publishers do publish information about releases somewhere to announce
them, etc. However I'm not that familiar with gists. What's the point? I'd
expect official spinnaker releases to be associated with
https://github.com/spinnaker/spinnaker/ or https://spinnaker.io/.

ttomsu:
+1 to direct embedding vs link out.

danielpeach:
One good reason for using a gist - which we've discussed in the past - is that we could pull the "What's New" button in Deck out of the Netflix module and use it to render the changelog for the latest release. When clicked, the "What's New" button opens a modal that pulls in and renders a gist.

Definitely not necessary, and it could consume another format, but it would be nice.

duftler:
+1 to trying to enable the "What's New?" feature.

So embedding via that plugin would allow us to both support the "What's
New?" feature and inline the contents in the Releases page.

...assuming we end up going with gitbook.

jtk54:
Ok so the sentiment here seems like publishing Gists is ok, as long as they're available on some central source (spinnaker.io or spinnaker/spinnaker) and embedded, not linked to separate places. Does that sound ok to you guys? I'll clone Steven's gitbook repo and try out the 'embed' feature.

jtk54:
Also, Eric to your point about dependencies being widely available -- `pygithub` is distributed by pip so that shouldn't be an issue.

ewiseblatt:
sure. we said from the beginning that publishing a release put artifacts
out to different places as appropriate, including accouncements to
spinnaker.io.  I dont see the point of adding yet another place to find
stuff. Having all our eggs in one basket seems fine in this case. But, I
dont object as long as there is a centralized place that has everything and
anybody can reproduce that (publishing to their own centralized place).

jtk54:
I added a `requirements.txt` for the `pygithub` dependency, PTAL.

PR #1481, opened 2017-03-17
fix(build): Updated Docker configs for spinnaker-monitoring.

jtk54:


dev/build_release.py
```diff
@@ -236,14 +236,24 @@ def start_container_build(self, name):
                 project=self.__options.gcb_project)
       )
     elif self.__options.container_builder == 'docker':
-      return BackgroundProcess.spawn(
-          'Build/publishing container image for {name} with Docker...'.format(
+      if name == 'spinnaker-monitoring':
```
ewiseblatt: It looks like both these are the same except for the path to the docker.yml file. Could you just have if/else on that path then run through a single line where the command is to emphasize how they are the same/different?

dev/generate_bom.py
```diff
@@ -155,9 +155,15 @@ def write_docker_version_files(self):
     """
     for comp in self.__component_versions:
       gradle_version = self.__version_from_tag(comp)
-      docker_tag = '{reg}/{comp}:{tag}'.format(reg=self.__docker_registry,
-                                               comp=comp,
-                                               tag=gradle_version)
+      docker_tag = ''
+      if comp == 'spinnaker-monitoring':
+        docker_tag = '{reg}/{comp}-daemon:{tag}'.format(reg=self.__docker_registry,
+                                                        comp=comp,
```
ewiseblatt: make a comment here (and in other file) indicating what makes spinnaker-monitoring special is that the deb name is different from the repo name (because it has multiple debs).

I'm not sure why third-party isnt here. It does have the dashboards and other monitoring support.

dev/build_release.py
ewiseblatt: also, how does third-party (for dashbaords and integrations) factor into docker?

dev/generate_bom.py
lwander: It's not in here yet (but probably should be soon). Halyard sends the CLI a command describing how to install the dashboards, which (for now) only involves pulling a debian and running the correct script. In the future, I'd imagine that if the monitoring is being setup on a remote k8s cluster, it would use a "run job" stage to create/update the dashboards.

ewiseblatt:
Sorry, not sure what happened with this. I assume it is obsolete and you should close it out. Otherwise let's have another look.

jtk54:
I'm pretty sure this is obsolete, closing.

PR #1482, opened 2017-03-17
fix(bom): Rename 'spinnaker-monitoring-*' to 'monitoring-*'.

jtk54:


lwander:
This probably won't run until this is merged: https://github.com/spinnaker/halyard/pull/246 right?

jtk54:
We can run this whenever the supporting tools are ready.

PR #1485, opened 2017-03-18
fix(google): Remove build data from google images.

ewiseblatt:
I still need to add the keys so I can ssh from a headless job.
And I have not verified the image, but believe it will be correct.

This is so you can preview and comment on the approach and technique.

@ttomsu, @duftler, @skim1420


dev/build_google_image_with_gcloud.sh
```diff
@@ -0,0 +1,380 @@
...
+PROJECT=$DEFAULT_PROJECT
+ZONE=$(gcloud config list 2>&1 \
+           | grep "zone =" | head -1 \
+           | sed "s/.* \(.*\)$/\1/")
+
+SSH_KEY_FILE=$HOME/.ssh/google_empty
+
+
+function fix_defaults() {
+  if [[ "$ZONE" = "" ]]; then
```
skim1420: Should this be ==?

dev/build_google_image_with_gcloud.sh
```diff
@@ -0,0 +1,380 @@
...
+    ZONE=us-central1-f
+  fi
+
+  local image_entry=$(gcloud compute images list 2>&1 | grep $BASE_IMAGE | head -1)
+
+  # If this was a family, convert it to a particular image for argument consistency
+  BASE_IMAGE=$(echo "$image_entry" | sed "s/\([^ ]*\) .*/\1/")
+  if [[ "$BASE_IMAGE_PROJECT" = "" ]]; then
```
skim1420: Should this be ==?

dev/build_google_image_with_gcloud.sh
```diff
@@ -0,0 +1,380 @@
...
+SCRIPT_DIR=$(dirname $0)
+ADD_CLOUD_LOGGING=true
+ADD_MONITORING=true
+UPDATE_OS=true
+
+TIME_DECORATOR=$(date +%Y%m%d%H%M%S)
+REPOSITORY_URL=https://dl.bintray.com/spinnaker/debians
+INSTALL_SCRIPT=https://dl.bintray.com/spinnaker/scripts/InstallSpinnaker.sh
+
+BASE_IMAGE=ubuntu-1404-lts
```
ewiseblatt: The idea here is you can pass in either so you can create off a specific image or off an image family. I called it BASE_IMAGE but maybe it should be something else (but not FAMILY)

dev/clean_google_image.sh
```diff
@@ -0,0 +1,59 @@
...
+disk_id=$1
+
+
+# Mount the dirty disk
+mkdir -p /mnt/disks/${disk_id}
+mount -o discard,defaults /dev/disk/by-id/google-${disk_id}-part1 /mnt/disks/${disk_id}
+cd /mnt/disks/${disk_id}
+
+
+# Remove user information from etc/ and their home directories.
```
skim1420: It's /mnt/etc so cwd etc/ is right

ewiseblatt:
@skim1420 

refactored to break apart script so part for cleaning a disk is reusable.

ewiseblatt:
@skim1420 
confirm the refactoring of the script

dev/build_google_image.sh
```diff
@@ -14,29 +14,320 @@
...
+  gcloud compute instances delete $INSTANCE \
+      --project $PROJECT \
+      --account $ACCOUNT \
+      --zone $ZONE \
+      --quiet
+}
+
+
+function cleanup_instances_on_error() {
+    delete_build_instance
```
skim1420: extra spaces

PR #1486, opened 2017-03-18
chore(install): Halyard bootscript

lwander:
@jtk54 PTAL

Here is a sample script to create a VM to invoke this (You still need to set this script as the startup script, though) 
```bash
#!/bin/bash

kube_config=../.kube/config
gcr_cred=../.gcp/nits-account.json
gce_cred=../.gcp/nits-account.json
project="nits-1132"
bucket_name="lwander-spin-halyard"
version="nightly"
jenkins_master="jmaster"
jenkins_address="localhost:8080"
jenkins_username="lwander"
jenkins_password="hunter2"

gcloud compute instances create "$1" \
    --scopes storage-full,compute-rw \
    --machine-type "n1-highmem-4" \
    --image-family "ubuntu-1404-lts" \
    --image-project "ubuntu-os-cloud" \
    --metadata project=$project,bucket_name=$bucket_name,version=$version,jenkins_master=$jenkins_master,jenkins_address=$jenkins_address,jenkins_username=$jenkins_username,jenkins_password=$jenkins_password  \
    --metadata-from-file kube_config="$kube_config",gcr_cred="$gcr_cred",gce_cred="$gce_cred"

```

### DIFF
--- a/install/first_halyard_boot.sh
+++ b/install/first_halyard_boot.sh
@@ -1,0 +1,255 @@
+#!/bin/bash
+#
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is specific to preparing a Google-hosted virtual machine
+# for running Spinnaker when the instance was created with metadata
+# holding configuration information, using Halyard.
+
+set -e
+set -u
+
+METADATA_URL="http://metadata.google.internal/computeMetadata/v1"
+INSTANCE_METADATA_URL="$METADATA_URL/instance"
+KUBE_FILE="/home/spinnaker/.kube/config"
+GCR_FILE="/home/spinnaker/.gcp/gce-account.json"
+GCE_FILE="/home/spinnaker/.gcp/gcr-account.json"
+
+DOCKER_ACCOUNT=my-gcr-account
+KUBERNETES_ACCOUNT=my-k8s-account
+GOOGLE_ACCOUNT=my-google-account
+AWS_ACCOUNT=my-aws-account
+
+function get_instance_metadata_attribute() {
+  local name="$1"
+  local value=$(curl -s -f -H "Metadata-Flavor: Google" \
+                     $INSTANCE_METADATA_URL/attributes/$name)
+  if [[ $? -eq 0 ]]; then
+    echo "$value"
+  else
+    echo ""
+  fi
+}
+
+function has_instance_metadata_attribute() {
+  local value=$(get_instance_metadata $1)
+
+  if [[ "$value" == "" ]]; then
+    return 1
+  else
+    return 0
+  fi
+}
+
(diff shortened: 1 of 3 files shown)


lwander:
the `version="nightly"` parameter does that. Ideally we'd be picking the exact number, though. Don't run this script until halyard is re-released & you've updated it.

install/first_halyard_boot.sh
```diff
@@ -0,0 +1,202 @@
...
+function replace_startup_script() {
+  # Keep the original around for reference.
+  # From now on, all we need to do is start_spinnaker
+  local original=$(get_instance_metadata_attribute "startup-script")
+  echo "$original" > "$SPINNAKER_INSTALL_DIR/scripts/original_startup_script.sh"
+  clear_instance_metadata "startup-script"
+}
+
+function install_halyard() {
+  wget https://raw.githubusercontent.com/spinnaker/halyard/master/InstallHalyard.sh
```
ewiseblatt: This means to test a new halyard you need to either build a different version of the bootscript that changes this or you need to create the VM without this as the startscript, then login, edit the script, and then manually run it.

Both these can still be automated but perhaps this could be parameterized.
The simplest thing to do would be to let someone pass the script through metadata then have this first check the metadata then do the wget. Now it's the same typical API but you have a hook to facilitate development and testing.

Also, why wget in some places and curl in others?

install/first_halyard_boot.sh
```diff
@@ -0,0 +1,202 @@
...
+METADATA_URL="http://metadata.google.internal/computeMetadata/v1"
+INSTANCE_METADATA_URL="$METADATA_URL/instance"
+KUBE_DIR="/home/spinnaker/.kube"
+GCR_DIR="/home/spinnaker/.gcr"
+GCE_DIR="/home/spinnaker/.gce"
+
+DOCKER_ACCOUNT=my-gcr-account
+KUBERNETES_ACCOUNT=my-k8s-account
+GOOGLE_ACCOUNT=my-google-account
+
```
ewiseblatt: Support AWS configuration and eventually others as well.

install/first_halyard_boot.sh
```diff
@@ -0,0 +1,202 @@
...
+    --json-path $config_path \
+    --project $project
+}
+
+function configure_storage() {
+  local bucket_name=$(get_instance_metadata_attribute "bucket_name")
+  hal config storage edit --account-name $GOOGLE_ACCOUNT --bucket $bucket_name
+}
+
+function configure_jenkins() {
```
ewiseblatt: This is starting to feel too tedious.
I recognize at least some of the value of individual attributes, but it means that this script is really brittle. I think you want to pass a configuration file through as metadata and then let hal interpret that. Doing so would eliminate the brittleness of the stack between the person specifying the configuration and hal in the deployment. Otherwise you are passing all these properties through everywhere which is a complicated and tedious interface.

For example, you are going to want to configure monitoring (and that could be one of a few different systems). There are probably echo notifications you want to configure too.

The file you pass through might be two -- a control and data. Or just one. One should be a tar file so you can attach different data files (credentials). And for that matter, maybe you can eventually encrypt it so as not to expose the passwords and credentials through the metadata and other parts of the stack between the creator and hal runtime.

In itself there is nothing here that isnt common practice on GCE (e.g. bitnami jenkins installation) but I dont think they scale to the diversity of Spinnaker. And there's too much independent evolution going on for this level of brittleness.

lwander:
@ewiseblatt This is a temporary solution for installing a specific spinnaker version for testing. The long-term solution doesn't require passing config metadata but instead uses vault as a secret store for all config. I realize it's brittle, and not really an improvement over `first_google_boot` aside for the ability to use a specific BOM.

lwander:
This is slowly being merged into the new C2D launcher script. It assumes your halconfig already exists & has had all components installed.

PR #1487, opened 2017-03-19
fix(build): Replaces removed workflow to build debs with main in

jtk54:
Builder.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -44,6 +44,7 @@ Usage:
 import argparse
 import base64
 import collections
+import datetime
 import glob
 import os
 import multiprocessing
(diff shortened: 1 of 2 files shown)


dev/build_release.py
```diff
@@ -594,3 +595,14 @@ def do_build(cls, options, build_number, container_builder):
...
       print '\nFINISHED writing release to {rep}'.format(
         rep=options.bintray_repo)
+
+  @classmethod
+  def main(cls):
+    parser = argparse.ArgumentParser()
+    cls.init_argument_parser(parser)
+    options = parser.parse_args()
+    # builds debians only
+    cls.do_build(options, '', None)
```
ewiseblatt: why not still a build number?

jtk54: We removed the `--build_number` flag from `Builder` (if it ever existed there) because of name collisions in https://github.com/spinnaker/spinnaker/blob/master/dev/build_prevalidation.py#L35. You can set it with an environment variable: https://github.com/spinnaker/spinnaker/blob/master/dev/build_release.py#L126

jtk54:
I think this is ready to merge, PTAL @ewiseblatt.

dev/build_release.py
ewiseblatt: Seems like build_number and container_builder should be keyword arguments that default to None so you only specify them when you want to override the defaults.

jtk54:
Let's merge after https://github.com/spinnaker/spinnaker/pull/1488 since going the other way is more work.

jtk54:
Rebased this on #1488, merging if there are no complaints.

PR #1488, opened 2017-03-19
feat(promote): Push tags and branches to remote repos during promotion.

jtk54:


dev/promote_bom.py
```diff
@@ -83,6 +135,7 @@ def main(cls):
     bom_promoter = cls(options)
```
ewiseblatt: I'm not sure what promote means or how the bom gets published. It looks like promote_bom publishes it with halyard, but it doesnt specify where, so I guess halyard is configured with somewhere and this script is hoping halyard was configured correctly rather than explicitly pushing the location through.

Error detection needs to be considered in these scripts. Error handling isnt as important, though when an error is detected it should be clear there was an error and at least which component it was in and what "phase" or step you were performing so that it can be troubleshooted, replicated, fixed, and verified. You dont want to have a bunch of errors and end up completing the script and publishing a partial release.

dev/promote_bom.py
```diff
@@ -91,10 +144,14 @@ def init_argument_parser(cls, parser):
                         help='The changelog to publish during this promotion.')
     parser.add_argument('--github_token', default='', required=True,
                         help="The GitHub user token with scope='gists' to write gists.")
+    parser.add_argument('--github_user', default='', required=True,
+                        help="The GitHub user to write gists as.")
     parser.add_argument('--rc_version', default='', required=True,
                         help='The version of the Spinnaker release candidate we are promoting.')
     parser.add_argument('--release_version', default='', required=True,
-                        help='The version for the new Spinnaker release.')
+                        help="The version for the new Spinnaker release. This needs to be of the form 'X.Y.Z'.")
```
jtk54: I can s/promote/publish/

dev/promote_bom.py
```diff
@@ -68,12 +84,48 @@ def publish_changelog_gist(self):
...
+
+    """
+    major, minor, patch = self.__release_version.split('.')
+
+    # The stable branch will look like <major>.<minor>.X since nebula
+    # enforces restrictions on what branches it does releases from.
+    # https://github.com/nebula-plugins/nebula-release-plugin#extension-provided
+    stable_branch = '.'.join([major, minor, 'X'])
+    for comp in COMPONENTS:
+      run_quick('git -C {0} checkout -b {1}'.format(comp, stable_branch))
```
jtk54: OK will add.

dev/promote_bom.py
```diff
@@ -91,10 +144,14 @@ def init_argument_parser(cls, parser):
                         help='The changelog to publish during this promotion.')
     parser.add_argument('--github_token', default='', required=True,
                         help="The GitHub user token with scope='gists' to write gists.")
+    parser.add_argument('--github_user', default='', required=True,
```
jtk54: The gist user would be different than the `--github_publisher` when we publish to the upstream Spinnaker repos. I think we wouldn't have access to publish a gist as the `spinnaker` user.

jtk54:
I fixed this up quite a bit, @ewiseblatt PTAL.

jtk54:
Ping -- made the changes we talked about in person @ewiseblatt. This is ready I think.

dev/publish_bom.py
```diff
@@ -0,0 +1,193 @@
...
+    self.write_bom_file('latest.yml', self.__bom_dict)
+    self.publish_bom('latest.yml')
+
+  def publish_changelog_gist(self):
+    """Publish the changelog as a github gist.
+    """
+    description = 'Changelog for Spinnaker {0}'.format(self.__release_version)
+    with open(self.__changelog_file, 'r') as clog:
+      raw_content_lines = clog.readlines()
+      spinnaker_version = '# Spinnaker {0}\n'.format(self.__release_version)
```
ewiseblatt: Should this be dated too? And signed with some kind of builder identifier so you can tell a bom is from a release as opposed to home-grown (default identifier is some kind of anonymous thing and we inject some other name in)?

dev/publish_bom.py
```diff
@@ -0,0 +1,193 @@
...
+               .format(tag=version_tag, owner=self.__github_publisher, comp=comp))
+        self.__exit_if_failure(
+          'git -C {comp} push release {tag}'.format(comp=comp,  tag=version_tag),
+          'git push tag'
+        )
+
+  def __exit_if_failure(self, command, task_name):
+    """Helper function to run_quick a command and exit if it fails.
+    """
+    result = run_quick(command)
```
ewiseblatt: this is check_run_quick (which raises RuntimeError but you can trap that top level)

dev/publish_bom.py
```diff
@@ -0,0 +1,193 @@
...
+    """Read, update, and publish a release candidate BOM.
+    """
+    self.__unpack_bom()
+    new_bom_file = '{0}.yml'.format(self.__release_version)
+    self.__bom_dict[VERSION] = self.__release_version
+    self.write_bom_file(new_bom_file, self.__bom_dict)
+    self.publish_bom(new_bom_file)
+    # Re-write the 'latest' Spinnaker version.
+    # TODO(jacobkiefer): Update 'available versions' with Halyard when that feature is ready.
+    self.write_bom_file('latest.yml', self.__bom_dict)
```
ewiseblatt: should this really be here or somewhere else higher level?
Is this scoped to a branch or a global name?

dev/publish_bom.py
jtk54: I think raising an RuntimError is fine here actually. I just want this script to exit somehow if one of those steps fail.

dev/publish_bom.py
jtk54: `latest.yml` is a global alias for the last Spinnaker version that was released by us. It lives in the BOM bucket that Halyard knows about. Since we're publishing the release BOM in this step and making the release official, I think it makes sense to replace the latest version in this step as well.

dev/publish_bom.py
jtk54: What sort of identifier are you suggesting here? I was thinking of appending something like:

```
Generated by ${hostname} at ${timestamp}.
```
like we did with the BOM comments. Thoughts?

dev/publish_bom.py
ewiseblatt: if we release a new minor version then release a new patched release of an old version, which is latest?

jtk54: As it stands, whichever was released last, which doesn't seem correct. I'll add an `alias` flag instead of hardcoding this.

jtk54:
Merging if there are no complaints.

dev/publish_bom.py
```diff
@@ -0,0 +1,186 @@
...
+
+      version_tag_build = ''
+      if comp == 'spinnaker-monitoring':
+        version_tag_build = 'version-{0}'.format(self.__bom_dict[SERVICES]['monitoring-daemon'][VERSION])
+      else:
+        version_tag_build = 'version-{0}'.format(self.__bom_dict[SERVICES][comp][VERSION])
+
+      last_dash = version_tag_build.rindex('-')
+      version_tag = version_tag_build[:last_dash]
+      repo_to_push = ('git@github.com:{owner}/{comp}.git'
```
lwander: Is having a running ssh agent to authenticate this push an implicit dependency?

lwander: Ah I see it is above, sounds good.

dev/publish_bom.py
lwander: It's a little nit-picky, but I think the hostname isn't as important to have at the top-level as the timestamp. The typical consumers of the changelog won't care (and might even think it's weird) to see "Generated by Jenkins at ${today}". I think putting the hostname at the bottom of the changelog is a good idea, but the date should probably live right near the top at the version, so the reader sees something like 

# Spinnaker 1.0.0
_Published on May 21st, 2017_

... contents ...

_This change log was auto generated by Jenkins and clog_

ewiseblatt:
sorry, I thought I published my comments last night. I didnt realize they were never sent.

dev/publish_bom.py
```diff
@@ -0,0 +1,186 @@
...
+    """Publish the changelog as a github gist.
+    """
+    description = 'Changelog for Spinnaker {0}'.format(self.__release_version)
+    with open(self.__changelog_file, 'r') as clog:
+      raw_content_lines = clog.readlines()
+      spinnaker_version = '# Spinnaker {0}\n'.format(self.__release_version)
+      # Re-write the correct Spinnaker version at the top of the changelog.
+      # Also add some identifying information.
+      raw_content_lines[0] = spinnaker_version
+      timestamp = '{:%Y-%m-%d %H:%M:%S}'.format(datetime.datetime.now())
```
jtk54: Check out 'datetime' in [this article](https://pyformat.info/). There are a bunch of ways to use `format` essentially.

ewiseblatt:
thanks!

jtk54:
I'm not sure how to distill the comments into a code change. It makes sense to add tracking information to the BOM and changelog, but the issue is that anyone can just copy that changelog and redistribute it with the same information. You shouldn't really trust changelogs and BOMs in file format. They have to be published to the correct place (spinnaker.io) for you to trust it, since a small group of people have access to publish content to that site.

ewiseblatt:
Sure, they have to be published to the right site. But if you are looking
at an installation or an image then having it in the file provides
traceability as to where it came from. Sure these things arent signed so
can be forged. But this isnt about security, it is just identity among
trusted friends. To actually verify you'd need to go back to the source,
but at least you'd know where too look.

When someone is getting help reporting a bug etc, how do they identify
their version? From the BOM right? Wouldnt that be a good time for them to
discover they arent actually running the version they thought they were?
And if there were a problem where they would report it?

So maybe this is just a URI (which can be the url to the official repo
location).

jtk54:
Ok, so I'll add a timestamp, the `--gist_user` and a link to either a nightly build or a release entry on the proper spinnaker.io page once we have that figured out (gitbook, github pages, etc).

PR #1490, opened 2017-03-20
fix(dev/build): Update path to log on startup.

duftler:


duftler:
@ewiseblatt please review.

PR #1493, opened 2017-03-21
Testing

ewiseblatt:
@jtk54


unittest/run_tests.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+TEST_DIR=$(readlink -f `dirname $0`)
```
jtk54: Was this tested on a linux VM or mac? `readlink -f` behaves differently on linux/mac, so one should work as intended but another might not.

PR #1495, opened 2017-03-21
fix(changelog): Fix broken URLs by invoking 'clog' properly.

jtk54:
@duftler @danielpeach FYI.

### DIFF
--- a/dev/generate_bom.py
+++ b/dev/generate_bom.py
@@ -180,8 +180,13 @@ class BomGenerator(Annotator):
 
       # Generate the changelog for the component.
       print 'Generating changelog for {comp}...'.format(comp=comp)
-      result = run_quick('cd {comp}; clog -f {hash} --setversion {version}; cd ..'
-                         .format(comp=comp, hash=hash, version=version))
+      # Assumes the remote repository is aliased as 'origin'.
+      component_url = run_quick('git -C {path} config --get remote.origin.url'
+                                .format(path=comp)).stdout.strip()
+      if component_url.endswith('.git'):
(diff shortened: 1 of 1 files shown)


dev/generate_bom.py
```diff
@@ -101,6 +102,9 @@ def init_argument_parser(cls, parser):
                         help="Output file to write the changelog to.")
     parser.add_argument('--bom_alias', default='',
                         help="Alias to rename the 'real' BOM as. This also sets the Spinnaker version as the alias.")
+    parser.add_argument('--changelog_repo_owner', default='spinnaker',
```
ewiseblatt: This should be the actual source you are using. If you need the owner, ti should be the origin in the git repo you already have. As noted below, this use case doesnt look right.

dev/generate_bom.py
```diff
@@ -180,8 +184,10 @@ def generate_changelog(self):
 
       # Generate the changelog for the component.
       print 'Generating changelog for {comp}...'.format(comp=comp)
-      result = run_quick('cd {comp}; clog -f {hash} --setversion {version}; cd ..'
-                         .format(comp=comp, hash=hash, version=version))
+      component_url = ('https://github.com/{owner}/{comp}'
```
ewiseblatt: I dont understand why you are doing this as opposed to using the local source.
The net effect seems to be that you are generating the changelog for something different from what you generated. Either because you put a different name here, or have a race condition in which the source updated on github different from the source you built.

jtk54:
This interprets the remote repository url from `origin` in the local repo now, PTAL.

PR #1498, opened 2017-03-22
Adds redis up check to upstart conf

skim1420:
@spinnaker/google-reviewers

### DIFF
--- a/etc/init/spinnaker.conf
+++ b/etc/init/spinnaker.conf
@@ -45,7 +45,3 @@ pre-start script
     start $i
   done
 end script
-
-respawn
-
-post-stop exec sleep 5
--- a/etc/init/spinnaker.conf
+++ b/etc/init/spinnaker.conf
@@ -4,22 +4,36 @@ start on filesystem or runlevel [2345]
 stop on shutdown
 
 pre-start script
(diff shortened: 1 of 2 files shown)


etc/init/spinnaker.conf
```diff
@@ -4,22 +4,36 @@ start on filesystem or runlevel [2345]
...
+        if [ $i -eq 30 ]; then
+          echo "Cassandra is not up on localhost.
+          Start it, and then restart Spinnaker."
+        fi
+        sleep 1
+      done
     fi
+  fi
+
+  if ! nc -z localhost 6379; then
```
lwander: Once someone takes their redis instance and moves it to another node they will no longer be able to run spinnaker

etc/init/spinnaker.conf
```diff
@@ -31,3 +45,7 @@ pre-start script
     start $i
   done
 end script
+
+respawn
```
lwander: I don't think the top-level spinnaker upstart entry will be aware of the subservice health. This needs to go in orca's upstart.

skim1420: Yes. Added orca upstart conf ptal.

skim1420:
Came full circle. Closing PR.

PR #1511, opened 2017-03-26
fix(docker-compose): Fixing typo in docker-compose.yml

alextbrandwatch:
As far as I am aware, these are the correct environment variable names for the Fiat service.  I have tested and only been able to get the app functioning with the variables set in this way.  Please let me know if I am missing something.


kkapelon:
This PR seems to be a partial fix for issue #1546.

PR #1514, opened 2017-03-27
feat(release): Deploy built spinnaker version to k8s.

jtk54:
This installs Halyard on an ubuntu:14.04 container and subsequently deploys Spinnaker on the same k8s cluster using Halyard. Once Spinnaker is running on k8s, we can tunnel using `kubectl` to run a citest suite.

The container is built using something like:
```
docker build --build-arg BUILD_PROJECT="${project}" \
   --build-arg  BUILD_ACCOUNT="${builder account}" \
   --build-arg BUILD_KEY="$(cat build.json)" \
   --build-arg DOCKER_PROJECT="${docker project}" \
   --build-arg DOCKER_ACCOUNT="${docker project account}" \
   --build-arg DOCKER_KEY="$(cat docker.json)" \
   --build-arg CLUSTER_NAME="halyard-cluster" \
   --build-arg KUBE_CONF="$(python -c 'import sys, yaml, json; json.dump(yaml.load(sys.stdin), sys.stdout, indent=4)' < ~/.kube/config)" \
   -t ${tag name} -f hal_install_spin.docker .
```

and subsequently run with `kubectl create -f ${pod spec}.yml` after the building and pushing the container.

### DIFF
--- a/dev/hal_install_spin.docker
+++ b/dev/hal_install_spin.docker
@@ -1,0 +1,26 @@
+FROM ubuntu:14.04
+
+# This is assumed to run in spinnaker/dev.
+# The container already has a '/dev' so we name it something different.
+COPY . runscript
+
+RUN apt-get install -y curl software-properties-common apt-transport-https wget
+
+# The container expects the following environment variables to be set.
+# These should be handled by creating k8s secrets and specifying
+# environment variables in the pod spec using those secrets.
+# ---
+# Parameters of the GCP project Spinnaker is maintaining.
+# BUILD_PROJECT
+# BUILD_KEY # json keyfile contents
+#
+# Parameters to configure the Docker registry.
+# DOCKER_KEY # json keyfile contents
+#
+# Kubernetes config in json format so we can pass it to the
+# container an an env variable.
+# KUBE_CONF
+
+WORKDIR runscript
+
+CMD /runscript/hal_k8s_run.sh
--- a/dev/hal_k8s_run.sh
+++ b/dev/hal_k8s_run.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The purpose of this script is to install Halyard at the latest released
+# version (output from our build process, not a Github release) and use
+# Halyard to deploy Spinnaker to Kubernetes at the versions specified in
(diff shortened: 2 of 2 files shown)


dev/hal_install_spin.docker
```diff
@@ -0,0 +1,34 @@
+FROM ubuntu:14.04
+# Parameters of the GCP project Spinnaker is maintaining.
+ARG BUILD_PROJECT
+ARG BUILD_ACCOUNT
+ARG BUILD_KEY # json keyfile contents
```
lwander: Secrets shouldn't be build-time args, since that would have us baking keys into images.

dev/hal_k8s_run.sh
```diff
@@ -0,0 +1,86 @@
...
+echo "Downloading Halyard..."
+wget https://raw.githubusercontent.com/spinnaker/halyard/master/InstallHalyard.sh
+
+# We need to make changes to the Halyard install script so it can run in a container.
+sed "s/^printf/echo/" -i InstallHalyard.sh
+sed "s/^start halyard/exit 0/" -i InstallHalyard.sh
+
+sudo bash InstallHalyard.sh -y
+echo "Configuring Halyard..."
+cat > /opt/spinnaker/config/halyard.yml  <<EOF
```
lwander: Are you publishing BOMs? if not, the default halyard.yml that comes with your installation should be enough

dev/hal_install_spin.docker
lwander: It's totally OK to assume (& document) what env vars need to be set in the container at runtime. You can pass these in either via the docker CLI, or by modifying the pod spec that's running this container in k8s.

lwander: You can also load env vars in k8s from secrets, which would probably be the golden-path here. 

`  kubectl create secret generic kubeconfig --from-file=relative/path/to/kube/config`

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: hal
spec:
  containers:
    - name: hal
      image: hal
      env:
        - name: KUBE_CONFIG
          valueFrom:
            secretKeyRef:
              name: kubeconfig
              key: config # this is the name of the file you published in the secret
```

dev/hal_k8s_run.sh
jtk54: Good point.

dev/hal_install_spin.docker
jtk54: Moved stuff to secrets.

jtk54:
I updated this to use secrets instead of injecting environment variables, and removed unnecessary environment variables. This container is now built with the usual `docker build` (no `--build-arg`s), but needs to be run with a pod spec similar to (via @lwander):

```
apiVersion: v1
kind: Pod
metadata:
  name: hal
spec:
  containers:
    - name: hal
      image: hal
      env:
        - name: KUBE_CONFIG
          valueFrom:
            secretKeyRef:
              name: kubeconfig
              key: config # this is the name of the file you published in the secret
```
with an `env` entry for each of the required environment variables.

lwander:
sweet

dev/hal_install_spin.docker
```diff
@@ -0,0 +1,26 @@
+FROM ubuntu:14.04
+
+# This is assumed to run in spinnaker/dev.
+# The container already has a '/dev' so we name it something different.
+COPY . runscript
+
+RUN apt-get install -y curl software-properties-common apt-transport-https wget
+
+# The container expects the following environment variables to be set.
```
lwander: looks good

PR #1518, opened 2017-03-29
fix(kube_smoke_test): Use passed k8s account name in pipeline.

jtk54:
Used to be hardcoded, now uses the passed k8s account and defaults to the previous string.


jtk54:
Actually I removed the default since it was already using the passed k8s account in another pipeline definition.

PR #1528, opened 2017-03-31
fix(google_front50_test): Default bucket and base path from test bind…

jtk54:
…ings.

### DIFF
--- a/testing/citest/tests/google_front50_test.py
+++ b/testing/citest/tests/google_front50_test.py
@@ -49,6 +49,12 @@ class GoogleFront50TestScenario(sk.SpinnakerTestScenario):
     parser.add_argument(
         '--gcs_json_path', default='',
         help='The path to the Google Cloud Storage service credentials JSON file.')
+    parser.add_argument(
+        '--gcs_bucket', default='',
+        help="The name of the Google Cloud Storage bucket to use, e.g. 'spinnaker-bucket'.")
+    parser.add_argument(
+        '--gcs_base_path', default='',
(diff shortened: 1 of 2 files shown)


testing/citest/tests/google_front50_test.py
```diff
@@ -49,6 +49,12 @@ def initArgumentParser(cls, parser, defaults=None):
     parser.add_argument(
         '--gcs_json_path', default='',
         help='The path to the Google Cloud Storage service credentials JSON file.')
+    parser.add_argument(
+        '--gcs_bucket', default='',
+        help='The name of the Google Cloud Storage bucket to use.')
```
ewiseblatt: It's not clear what these are. Is a bucket foo or gs://foo?
Is a base path gs://foo/bar or bar?

testing/citest/tests/google_front50_test.py
```diff
@@ -61,8 +67,8 @@ def _do_init_bindings(self):
     if not enabled:
       raise ValueError('spinnaker.gcs.enabled is not True')
 
-    self.BUCKET = config['spinnaker.gcs.bucket']
-    self.BASE_PATH = config['spinnaker.gcs.rootFolder']
+    self.BUCKET = config['spinnaker.gcs.bucket'] or self.bindings['GCS_BUCKET']
```
ewiseblatt: This is in the wrong order. Commandline flags have highest precedence.

jtk54:
Switched the precedence of the flags and make the help strings clearer, PTAL.

PR #1529, opened 2017-04-01
fix(citest): Dont perform superfluous platform initializations

ewiseblatt:
@rguthriemsft When the dependent PR goes in, you should take things you may be adding into spinnaker_test_scenario.py and instead add them to an azure_support.py so the only change to spinnaker_test_scenario is adding AzureScenarioSupport to the PLATFORM_SUPPORT_CLASSES. This PR should fix the other issues made me aware of so your workarounds should not be needed anymore.

@Roshan2017 I migrated all the OpenStack changes, but I did not test them because I still havent gotten around to reinstalling openstack. Could you test this for me?

@jtk54 or @lwander Please review.

----

This CL refactors SpinnakerTestScenario so that bindings requiring
platform interrogation are not performed until they are needed.
This allows tests to be decoupled from platforms they are not using.

This also refactors the SpinnakerTestScenario so that the platform
specific code is factored out into an encapsulated support module.
This should make it easier to add and maintain platforms, as
well as simplify what's going on.

Most of this PR is moving existing code from spinnaker_scenario_test.py into individual <platform>_support.py modules. The interface has a method for initializing the argparser (used before the class is constructed), initializing the bindings (used on the instance after constructed) and creating the observer for that platform. The appropriate code blocks were moved from spinnaker_testing without any major change except as follows:

1) The observers are not constructed until the property is referenced. So if the observer is never referenced, it will never be instantiated. Presumably if one is not needed it will not be referenced.

2) Binding values derived from interrogating the platform if not otherwise available are added using lazy initializers so that they will not incur the work if they are never needed. This also decouples the platform initialization from tests that do not need them.

3) I added a SPINNAKER_<platform>_ENABLED binding to easily check if a platform is enabled or not rather than merely seeing if there is an account configured for it. This is added in the base support class.

This PR depends on  https://github.com/google/citest/pull/252 which is not yet approved.

### DIFF
--- a/testing/citest/spinnaker_testing/aws_scenario_support.py
+++ b/testing/citest/spinnaker_testing/aws_scenario_support.py
@@ -1,0 +1,174 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Amazon Web Services platform and test support for SpinnakerTestScenario."""
+
+import logging
+
+from citest.base import ExecutionContext
+import citest.aws_testing as aws
+from spinnaker_testing.base_scenario_support import BaseScenarioPlatformSupport
+
+
+class AwsScenarioSupport(BaseScenarioPlatformSupport):
+  """Provides SpinnakerScenarioSupport for Amazon Web Services."""
+
+  @classmethod
+  def add_commandline_parameters(cls, scenario_class, builder, defaults):
+    """Implements BaseScenarioPlatformSupport interface.
+
+    Args:
+      scenario_class: [class spinnaker_testing.SpinnakerTestScenario]
+      builder: [citest.base.ConfigBindingsBuilder]
+      defaults: [dict] Default binding value overrides.
+         This is used to initialize the default commandline parameters.
+    """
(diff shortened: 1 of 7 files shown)


testing/citest/spinnaker_testing/aws_scenario_support.py
```diff
@@ -0,0 +1,174 @@
...
+    builder.add_argument(
+        '--test_aws_zone',
+        default=defaults.get('TEST_AWS_ZONE', 'us-east-1c'),
+        help='The AWS zone to test generated instances in (when managing AWS).'
+             ' This implies the AWS region as well.')
+
+    builder.add_argument(
+        '--test_aws_vpc_id',
+        default=defaults.get('TEST_AWS_VPC_ID', None),
+        help='Default AWS VpcId to use when creating test resources.')
```
lwander: Why not "VPC ID" rather than "VpcId"?

lwander:
I'm missing something - where is the loading from lazy bindings defined?

ewiseblatt:
Not sure what you mean by where loading is defined.

There's only one place it is currently used, which is in the AWS support
around here
https://github.com/spinnaker/spinnaker/pull/1529/commits/41e6aa3b2988af5d5512359526cb48e9db8b26a8#diff-7d5658c9672146f06d4d0722fc4675a0R123

The mechanism is in citest, the PR defined. It adds the injected function
map into the "key" search path.

testing/citest/spinnaker_testing/aws_scenario_support.py
ewiseblatt: VpcId is the identifier in the AWS schemas. Not that it matters.

lwander:
What I meant was I see you call `bindings.add_lazy_initializer`, and I see you read properties from the bindings that were added with that method, but I didn't see where the logic for that lives.

ewiseblatt:
It's in the other cited PR
https://github.com/google/citest/pull/252

Roshan2017:
@ewiseblatt Okay! I will test and will let you know.

ewiseblatt:
I submitted the citest change so this is all you need to patch in.

PR #1530, opened 2017-04-03
Aws smoke test

ewiseblatt:
@lwander

### DIFF
--- a/testing/citest/spinnaker_testing/spinnaker.py
+++ b/testing/citest/spinnaker_testing/spinnaker.py
@@ -338,7 +338,8 @@ class SpinnakerAgent(service_testing.HttpAgent):
       base_url = '{protocol}://{netloc}'.format(protocol=protocol,
                                                 netloc=netloc)
       logger.info('%s is available at %s', name, base_url)
-      deployed_config = scrape_spring_config(os.path.join(base_url, 'env'))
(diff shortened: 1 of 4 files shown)


testing/citest/spinnaker_testing/spinnaker.py
```diff
@@ -338,7 +338,8 @@ def new_gce_instance_from_bindings(
       base_url = '{protocol}://{netloc}'.format(protocol=protocol,
                                                 netloc=netloc)
       logger.info('%s is available at %s', name, base_url)
-      deployed_config = scrape_spring_config(os.path.join(base_url, 'env'))
+      deployed_config = scrape_spring_config(
+          os.path.join(base_url, 'resolvedEnv'))
       spinnaker_agent = cls(base_url, status_factory)
       spinnaker_agent.__deployed_config = deployed_config
```
lwander: Where does this get used now?

ewiseblatt: Same as before -- when initializing the spinnaker agent.
I wrote the original long before the resolvedEnv endpoint existed.
The resolvedEnv is more accurate. Originally I thought I was seeing a bug interpreting the env, but turned out that I hadnt restarted one of the internal services. However, it is better to use the resolvedEnv anyway.

PR #1538, opened 2017-04-05
fix(dev): fixes deck dev startup

danielpeach:
@duftler or @ewiseblatt please review.

### DIFF
--- a/dev/refresh_source.py
+++ b/dev/refresh_source.py
@@ -434,6 +434,7 @@ d=$(dirname "$0")
 cd "$d"
 LOG_DIR=${{LOG_DIR:-../logs}}
 
+npm install >& "$LOG_DIR/{name}.log"
 bash -c "(./start.sh >> '$LOG_DIR/{name}.log') 2>&1\
  | tee -a '$LOG_DIR/{name}.log' >& '$LOG_DIR/{name}.err' &"
 """.format(name=name))


dev/refresh_source.py
```diff
@@ -434,7 +434,8 @@ def write_deck_run_script(self, repository):
 cd "$d"
 LOG_DIR=${{LOG_DIR:-../logs}}
 
-bash -c "(./start.sh >> '$LOG_DIR/{name}.log') 2>&1\
+npm install >& "$LOG_DIR/deck.log"
```
ewiseblatt: what then is start.sh?
is it needed?

duftler: I can confirm that modifying `start_dev.sh` to contain what this routine will generate does result in deck starting up properly on a dev vm.
It does not appear that `start.sh` is used anywhere.

duftler: To clarify, it doesn't appear to be used when spinning up deck on a dev vm built this way. Not sure about where else it is used.

duftler: Also noticed it results in these log files being created:
```
 deck.log
{name}.err
{name}.log
```

danielpeach: `start.sh` tries to set the right node version based on Deck's `package.json`, but makes a wrong assumption (for dev vms) about where `nvm` is located. Even if it doesn't find `nvm`, it runs `npm run start-dev-server`. I can keep `start.sh` or get rid of it or fix it, it's just logging a confusing warning now.

Matt, I'm not seeing either of the `{name}` log files. How are you generating Deck's `start_dev.sh`?

duftler: I patched in your changes to my already-generated start-dev.sh file. It looks like so:
```
#!/bin/bash
d=$(dirname "$0")
cd "$d"
LOG_DIR=${LOG_DIR:-../logs}

npm install >& "$LOG_DIR/deck.log"
bash -c "(npm run start-dev-server >> '$LOG_DIR/{name}.log') 2>&1\
 | tee -a '$LOG_DIR/{name}.log' >& '$LOG_DIR/{name}.err' &"
```

duftler:
```
duftler@mjd-dev-2:~/build$ ls -l logs/
total 0
duftler@mjd-dev-2:~/build$ ./deck/start_dev.sh 
duftler@mjd-dev-2:~/build$ ls -l logs/
total 8
-rw-rw-r-- 1 duftler duftler 273 Apr  6 13:30 deck.log
-rw-rw-r-- 1 duftler duftler   0 Apr  6 13:30 {name}.err
-rw-rw-r-- 1 duftler duftler 712 Apr  6 13:30 {name}.log
```

danielpeach: Right, but `refresh_source.py` formats those `{name}` pieces. The one `deck.log` is inconsistent, so I'll change that.

I cloned this branch, then ran 
```
spinnaker/dev/refresh_source.sh --pull_origin --github_user spinnaker
spinnaker/dev/run_dev.sh
```
and it produced the right log files.

ewiseblatt: I'm running from HEAD and have the right log files, deck starts up, and I see stuff at 9000 though I havent interacted with it (though I have before but not sure of the build purity). I'm not sure what was broken.

I guess start.sh is for those using the repository without refresh_source, so it is still relevant.
Would be nice for run_dev to wrap start.sh and just add the consistent logging.

duftler: Ah, makes sense Dan.

duftler: If `start.sh` is just for use without refresh_source, why not just add some comments to it indicating that? Can't see the reason to make it more complicated. The nvm/npm/node stuff is already complicated enough.

ewiseblatt:
but why not wrap start.sh from within run_dev so run_dev is just
redirecting. Especially since it is complicated. Why maintain it in two
different scripts?

duftler:
Seems like they have half of one line in common.

I don't feel too strongly about it, just want to get the dev path finished up today. I can live with it either way.

ewiseblatt:
On Thu, Apr 6, 2017 at 10:06 AM, Matt Duftler <notifications@github.com>
wrote:

> Seems like they have half of one line in common.
>
That's the thing that bothers me. If start.sh is how you run it, run_dev is
just handling the logging.
If that extra stuff in start.sh is needed then it is left out. And if it
isnt needed, I'm not sure why it is there. Either way, when someone adds
some extra thing to start.sh it they arent going to modify refresh_source
to add it into run_dev.


>
>
I don't feel too strongly about it, just want to get the dev path finished

duftler:
Ok, works for me.

danielpeach:
Fine with me. I've verified that Deck starts up properly with `start.sh`.

PR #1540, opened 2017-04-06
fix(install): Inject YAML nodes to fix disabling cassandra.

ewiseblatt:
Rewrote YAML mutator so it uses parse tree rather than regex searching
for more robustness.

@duftler 
This also generates the spinnaker-local on install.
I didnt explicitly test that in the bootstrap, but did so outside bootstrap.

### DIFF
--- a/pylib/spinnaker/yaml_util.py
+++ b/pylib/spinnaker/yaml_util.py
@@ -168,7 +168,7 @@ class YamlBindings(object):
 
 
   @staticmethod
-  def update_yml_source(path, update_dict):
+  def update_yml_source(path, update_dict, add_new_nodes=True):
     """Update the yaml source at the path according to the update dict.
 
     All the previous bindings not in the update dict remain unchanged.
@@ -178,6 +178,8 @@ class YamlBindings(object):
       path [string]: Path to a yaml source file.
       update_dict [dict]: Nested dictionary corresponding to
           nested yaml properties, keyed by strings.
+      add_new_nodes [boolean]: If true, add nodes in update_dict that
+          were not in the original. Otherwise raise a KeyError.
     """
     bindings = YamlBindings()
     bindings.import_dict(update_dict)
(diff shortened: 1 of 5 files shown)


pylib/spinnaker/yaml_util.py
```diff
@@ -186,13 +188,96 @@ def update_yml_source(path, update_dict):
...
       for prop in updated_keys:
-        source = bindings.transform_yaml_source(source, prop)
+        source = bindings.transform_yaml_source(source, prop,
+                                                add_new_nodes=add_new_nodes)
 
     with open(path, 'w') as source_file:
       source_file.write(source)
 
+  def find_yaml_context(self, source, root_node, full_key, raise_if_not_found):
+    """Given a raml file and full key, find the span of the value for the key.
```
duftler: s/raml/yaml

pylib/spinnaker/yaml_util.py
```diff
@@ -186,13 +188,96 @@ def update_yml_source(path, update_dict):
...
+      raise_if_not_found: [boolean] Whether to raise a KeyError or return
+         additional context if key isnt found.
+
+    Returns:
+      context, span
+
+      where:
+          context:[string, string)]: Text to append before and after the cut.
+          span: [(int, int)]: The start/end range of source with existing value
+                               to cut.
```
duftler: Seems like this indentation should either back out one space or just have the 2 words on the previous line.

pylib/spinnaker/yaml_util.py
```diff
@@ -186,13 +188,96 @@ def update_yml_source(path, update_dict):
...
+
+    if found:
+      # value of closest_node is what we are going to replace.
+      # There is still a space between the token and value we write.
+      return (' ' if span_is_empty else '', ''), span
+
+    if raise_if_not_found:
+      raise KeyError('.'.join(parts[0:key_index + 1]))
+
+    # We are going to add a new child. This s going to be indented equal
```
duftler: s/s going/is going

unittest/yaml_util_test.py
```diff
@@ -285,27 +285,23 @@ def test_concat_default(self):
...
   b:
      space: A
-     nospace:B
      empty:
 """
-     source = template.format(space='SPACE', nospace='NOSPACE', empty='')
+     source = template.format(space='SPACE', empty='')
      expect = template.format(space='WithSpace',
-                              nospace=' WithoutSpace',
                               empty=' Empty')
```
duftler: Moving `empty` to the previous line would be more consistent with neighbors.

PR #1541, opened 2017-04-06
fix(front50_test): Don't inspect config if host platform is 'native'.

jtk54:


testing/citest/tests/google_front50_test.py
```diff
@@ -64,7 +64,8 @@ def _do_init_bindings(self):
     context = ExecutionContext()
     config = self.agent.deployed_config
     enabled = config.get('spinnaker.gcs.enabled', False)
-    if not enabled:
+    # We don't read the config files while using 'native' platform.
+    if self.bindings['HOST_PLATFORM'] != 'native' and not enabled:
```
ewiseblatt: Is there another way to do this? I want to get rid of HOST_PLATFORM. At least I think I do. I havent found a need until now.
Instead could you change 66 to default to None and make this a check against False?

jtk54: I'm using `HOST_PLATFORM` for the k8s tests currently, I'd be real sad if you removed it. I can do your suggestion though.

ewiseblatt:
I'm not removing it imminently. I marked it deprecated like a year ago.
I was waiting for other platforms to shake out to understand what the
issues are.
We can visit it after you are done.

testing/citest/tests/google_front50_test.py
```diff
@@ -63,8 +63,8 @@ def _do_init_bindings(self):
     """Hook to initialize custom test bindings so journaling is scoped."""
     context = ExecutionContext()
     config = self.agent.deployed_config
-    enabled = config.get('spinnaker.gcs.enabled', False)
-    if not enabled:
+    enabled = config.get('spinnaker.gcs.enabled', None)
+    if enabled == False:
       raise ValueError('spinnaker.gcs.enabled is not True')
 
     self.BUCKET = self.bindings['GCS_BUCKET'] or config['spinnaker.gcs.bucket']
```
ewiseblatt: Since you arent guaranteeing config, you should check for BUCKET to fail fast. If there is no BUCKET, I'd treat the whole test as misconfigured similar to GCS not being enabled.

PR #1552, opened 2017-04-07
config(provider/openstack): Added extra settings and some documentation

kkapelon:

### DIFF
--- a/config/default-spinnaker-local.yml
+++ b/config/default-spinnaker-local.yml
@@ -135,6 +135,9 @@ providers:
   openstack:
     # For more information on configuring Openstack clusters, see
     # http://www.spinnaker.io/v1.0/docs/target-deployment-configuration#section-openstack
+    # To obtain the correct settings log into your OpenStack UI (Horizon), click on "Access and Security" on the left
+    # sidebar, click on the API access tab and finally click on "Download Openstack RC file V3". The rc file
+    # contains the correct settings.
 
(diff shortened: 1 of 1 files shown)


kkapelon:
ping @spinnaker/reviewers

config/default-spinnaker-local.yml
```diff
@@ -143,7 +146,10 @@ providers:
       username: ${OS_USERNAME}
       password: ${OS_PASSWORD}
       projectName: ${OS_PROJECT_NAME}
+      domainName: ${OS_PROJECT_DOMAIN_NAME}
```
kkapelon: I have both properties OS_USER_DOMAIN_NAME and OS_PROJECT_DOMAIN_NAME defined locally (to the same value). Maybe you are correct and one of them is not needed. They are both valid however. https://docs.openstack.org/developer/python-openstackclient/authentication.html. Maybe add both of them? (With one of them commented out?)

ewiseblatt: while you may have both, it seems overspecified to be using both and seems wrong to have one file overwrite the other. Pick one as your interface, or create an attribute for each and do the right thing internally to support that. I'm not familiar with open stack to tell you what the right thing is.

Keep in mind that spinnaker.yml will always be present. That is your base interface.
The default-spinnaker-local.yml is purely optional and is only used to default an initial spinnaker-local.yml if one was not explicitly supplied. So you cannot count on anything in it showing up in the runtime.

kkapelon:
Updated setting.

Issue #1544, opened 2017-04-06
Spinnaker ./InstallSpinnaker.sh does not work on Ubuntu 16.04 LTS


kkapelon:
**Title:** The install scripts works only on Ubuntu 14.x and nothing else
**Cloud provider:** None
**Environment:** Ubuntu 16.04 LTS 64 bit (under Virtualbox in Windows 10)
**Feature area:** Installation

I downloaded the install script locally and examined it to see what it does (as I always do with these scripts that need root access). I saw the message that says "we require 14.04 or higher" and I assumed that I was good to go.

Unfortunately the installation script fails

1.  It tries to install package python-support that exists in Ubuntu 14.x but not in 16.x
2.  It tries to downgrade Cassandra to 2.x while Ubuntu 16.x comes with 3.x

I manually installed python-support and also tried with ./InstallSpinnaker.sh --noinstall_cassandra
This failed as well with error message

```
0 to upgrade, 0 to newly install, 0 to remove and 316 not to upgrade.
~/spin/spinnaker/installspinnaker.kynf/packer ~/spin/spinnaker
~/spin/spinnaker
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package spinnaker
```

I found out that  there is indeed no package for xenial at
https://dl.bintray.com/spinnaker/debians/dists/
Only trusty packages are offered.

**Steps to reproduce**
Run `./InstallSpinnaker.sh --noinstall_cassandra` in Ubuntu 16.04 LTS

**Expected result**

1. Either have Spinnaker installed OR
2. get a message that Ubuntu 16.x is not supported

I think that the installation script should be modified and rather than a "less than" check, perform an "exactly like" 14.x check.

Documentation pages of Spinakker should also be updated to say that Ubuntu 14.x is the only version supported and newer versions will not work (even LTS ones)

**Actual Result**

Spinnaker packages are not available for Ubuntu Xenial

ewiseblatt:
@lwander would halyard work?

lwander:
No, the install Halyard script has the same problem: https://github.com/spinnaker/halyard/issues/279

lwander:
We aren't publishing our spinnaker packages with the Xenial distribution enabled. @jtk54 might have some ideas on how to fix this.

jtk54:
[...]

PR #1553, opened 2017-04-07
fix(installation): make it clear that the current Ubuntu LTS is not supported

kkapelon:
Most of the documentation refers to Ubuntu 14.04 LTS. The new LTS version 16.04 does not work with Spinnaker. Version 15.04 was not an LTS and it has reached its EOL

Fixed Issue #1544


kkapelon:
ping @spinnaker/reviewers

ewiseblatt:
We are in the process of fixing this (and republishing the debs).

PR #1560, opened 2017-04-11
initial commit - Azure test (app/security group)

rguthriemsft:
Initial commit of azure support and test cases for app and security group.  Pull request https://github.com/google/citest/pull/257 is required for this to work properly.

@EricJizbaMSFT PTAL @ewiseblatt PTAL

### DIFF
--- a/testing/citest/README.md
+++ b/testing/citest/README.md
@@ -60,6 +60,7 @@ you may need additional tools.
 Platform | Tools | Installation Command
 ---------|-------|---------------------
 Amazon Web Services | awscli | ```sudo apt-get install -y awscli```
+Microsoft Azure | az | [See instructions](https://docs.microsoft.com/cli/azure/install-azure-cli)
 Google Cloud Platform | gcloud | ```curl https://sdk.cloud.google.com | bash```
 Kubernetes | kubectl | [See instructions](http://kubernetes.io/docs/user-guide/prereqs/)
 OpenStack | openstack | [See instructions](https://docs.openstack.org/user-guide/common/cli-install-openstack-command-line-clients.html)
@@ -194,6 +195,10 @@ when creating test instances requiring an image.
 test_aws_vpc_id' | The default AWS VpcId to use when creating test resources.
 test_aws_security_group_id: The default AWS SecurityGroupId to use when
 creating test resources.
+test_azure_rg_location | The azure location where the test resources should be created. Default to westus.
+azure_storage_account_key | The key used to read the content from the Azure storage account used by Spinnaker when Spinnaker is configured to use Azure Storage for front50.
+azure_storage_account_name | The name of the Azure storage account used by Spinnaker when Spinnaker is configured to use Azure Storage for front50.
+
 
 
 ### Account Information
@@ -203,6 +208,7 @@ spinnaker_google_credentials | The name of the Spinnaker [clouddriver] account t
 spinnaker_kubernetes_credentials |  The name of the Spinnaker [clouddriver] account that you wish to use for Kubernetes operations. If not specified, this will use the configured primary account.
 spinnaker_aws_credentials |  The name of the Spinnaker [clouddriver] account that you wish to use for Amazon Web Services operations. If not specified, this will use the configured primary account.
 spinnaker_os_account | The name of the Spinnaker [clouddriver] account that you wish to use for OpenStack operations. If not specified, this will use the configured primary account.
+spinnaker_azure_account | The name of the Spinnaker [clouddriver] account that you wish to use for the Azure operations. If not specified, this will use the configured primary account.
 
 
 ## Standard Parameters For Configuring Observers
@@ -211,6 +217,7 @@ Flag | Description
 gce_credentials_path | The path to a service account JSON credentials file used by the test to verify effects on GCE. The permissions needed on the account may vary depending on what the test is doing. You can use the same service account that you have configured spinnaker with,
 aws_profile | The name of the awscli profile to use when verifying effects on AWS. The permissions needed in the profile may vary depending on what the test is doing. You can use the same AWS credentials as those you configured spinnaker to use.
 os_cloud | The name of the cloud. OpenStack will look for a clouds.yaml file that contains a cloud configuration to use for authentication.
+azure | For Azure, you must login with the following command on the test environment ```az login -u SPN_Client_ID -p SPN_Application_key --tenant Tenant_Name --service-principal``` prior to test execution.
 
 
 ## Typical Invocations
@@ -255,6 +262,10 @@ things, and lastly, additional "observer" things.
 
 ### OpenStack
         --os_cloud=$OS_CLOUD
+
+### Azure
+        --azure_storage_account_key=Key to access the storage account used by Spinnaker
+        --azure_storage_account_name=Name of the Azure storage account used by Spinnaker
 
 # Usage Examples
 
--- a/testing/citest/spinnaker_testing/azure_scenario_support.py
+++ b/testing/citest/spinnaker_testing/azure_scenario_support.py
@@ -1,0 +1,72 @@
+# Copyright 2017 Microsoft Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Azure platform and test support for SpinnakerTestScenario."""
+
+import citest.azure_testing as az
+from spinnaker_testing.base_scenario_support import BaseScenarioPlatformSupport
+
+
+class AzureScenarioSupport(BaseScenarioPlatformSupport):
+  """Provides SpinnakerScenarioSupport for Azure."""
+
+  @classmethod
+  def add_commandline_parameters(cls, scenario_class, builder, defaults):
+    """Implements BaseScenarioPlatformSupport interface.
+
+    Args:
+      scenario_class: [class spinnaker_testing.SpinnakerTestScenario]
+      builder: [citest.base.ConfigBindingsBuilder]
+      defaults: [dict] Default binding value overrides.
+         This is used to initialize the default commandline parameters.
+    """
+    #
+    # Operation Parameters
+    #
+
+    # pylint: disable=line-too-long
+    builder.add_argument(
+        '--test_azure_rg_location', 
+        default = defaults.get('TEST_AZURE_RG_LOCATION', 'westus'),
+        help='The location of the azure resource group where test resources should be created.')
+    builder.add_argument(
+        '--azure_storage_account_name', 
+        dest='azure_storage_account_name',
+        help='The name of the Azure storage account used by Front50 in Spinnaker.')
+    builder.add_argument(
+        '--azure_storage_account_key', 
+        dest='spinnaker_azure_storage_account_key',
+        help='The key used to access storage account used by front50.')
+
+  def _make_observer(self):
+    """Implements BaseScenarioPlatformSupport interface."""
+    bindings = self.scenario.bindings 
+    if not bindings.get('TEST_AZURE_RG_LOCATION'):
+      raise ValueError('There is no location specified')
+
+    return az.AzAgent(bindings['TEST_AZURE_RG_LOCATION'])
+
+  def __init__(self, scenario):
+    """Constructor.
+
+    Args:
+      scenario: [SpinnakerTestScenario] The scenario being supported.
+    """
+    super(AzureScenarioSupport, self).__init__("azure", scenario)
+    self.__az_observer = None
+
+    bindings = scenario.bindings
+    if not bindings['SPINNAKER_AZURE_ACCOUNT']:
+      bindings['SPINNAKER_AZURE_ACCOUNT'] = (scenario.agent.deployed_config.get(
+          'providers.azure.primaryCredentials.name', None))
--- a/testing/citest/spinnaker_testing/spinnaker_test_scenario.py
+++ b/testing/citest/spinnaker_testing/spinnaker_test_scenario.py
@@ -33,16 +33,16 @@ from spinnaker_testing import appengine_scenario_support
 from spinnaker_testing import google_scenario_support
 from spinnaker_testing import kubernetes_scenario_support
 from spinnaker_testing import openstack_scenario_support
+from spinnaker_testing import azure_scenario_support
 
 PLATFORM_SUPPORT_CLASSES = [
     aws_scenario_support.AwsScenarioSupport,
-
     # appengine depends on google so order it after
     google_scenario_support.GoogleScenarioSupport,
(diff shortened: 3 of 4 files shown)


testing/citest/spinnaker_testing/azure_scenario_support.py
```diff
@@ -0,0 +1,96 @@
...
+        '--azure_location', 
+        dest='spinnaker_azure_location',
+        help='The azure location to be used.')
+    builder.add_argument(
+        '--test_azure_rg_location', 
+        default = defaults.get('TEST_AZURE_RG_LOCATION', 'westus'),
+        help='The location of the azure resource group where test resources should be created.')
+    # builder.add_argument(
+    #     '--spinnaker_azure_account_name', 
+    #     dest='spinnaker_azure_account_name',
```
EricJizbaMSFT: Uncomment this arg and change to `default=defaults.get('SPINNAKER_AZURE_ACCOUNT', None),`

testing/citest/spinnaker_testing/azure_scenario_support.py
```diff
@@ -0,0 +1,96 @@
...
+      defaults: [dict] Default binding value overrides.
+         This is used to initialize the default commandline parameters.
+    """
+    #
+    # Operation Parameters
+    #
+
+    # pylint: disable=line-too-long
+    # builder.add_argument(
+    #     '--azure_subscription_name',
```
EricJizbaMSFT: nit: remove unnecessary comments

testing/citest/tests/azure_smoke_test.py
```diff
@@ -0,0 +1,235 @@
...
+        'Security Group Created', retryable_for_secs=30)
+     .collect_resources(
+         az_resource='network',
+         command='nsg',
+         args=['show', '--name', self.TEST_SECURITY_GROUP, '--resource-group', self.TEST_SECURITY_GROUP_RG],
+         no_resources_ok=True)
+     .contains_pred_list([
+         jp.DICT_MATCHES({
+             'name': jp.STR_SUBSTR(self.TEST_SECURITY_GROUP)#,
+             #'securityRules': jp.STR_SUBSTR("protocol='tcp'")
```
ewiseblatt: I'm not sure what you were trying to do here.
Ideally your API returns back a json document so you have a dictionary with name value pairs (e.g. priority: 100) rather than a string (priority=100). If you have one big string you can either add a bunch of individual regexes into it or ideally have your agent be able to convert the string into a dictionary. I just checked in a python implementation of aws agent rather than CLI which is pretty simple. It's more simplistic than the gcp one. If your CLI cant give you json back then you might consider that route. It should be straightforward now that you've written the CLI one.

I think the multi regex solution here would look like
.contains_path_pred('securityRules',
    jp.AND(jp.STR_SUBSTR("protocol='tcp'"),
                 jp.STR_SUBSTR("priority=100"),
                 jp.STR_SUBSTR("destinationPortRange='80=80'")))

The reason a MATCH would be better, besides being simpler to read, is that the reporting is more straight forward, especially in the event of a failure.

testing/citest/README.md
```diff
@@ -60,6 +60,7 @@ you may need additional tools.
 Platform | Tools | Installation Command
 ---------|-------|---------------------
 Amazon Web Services | awscli | ```sudo apt-get install -y awscli```
+Microsoft Azure | az | [See instructions](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli)
```
rguthriemsft: done

testing/citest/README.md
```diff
@@ -203,6 +207,7 @@ spinnaker_google_credentials | The name of the Spinnaker [clouddriver] account t
 spinnaker_kubernetes_credentials |  The name of the Spinnaker [clouddriver] account that you wish to use for Kubernetes operations. If not specified, this will use the configured primary account.
 spinnaker_aws_credentials |  The name of the Spinnaker [clouddriver] account that you wish to use for Amazon Web Services operations. If not specified, this will use the configured primary account.
 spinnaker_os_account | The name of the Spinnaker [clouddriver] account that you wish to use for OpenStack operations. If not specified, this will use the configured primary account.
+spinnaker_azure_account | The name of the Spinnaker [clouddriver] account that you wish to use for the Azure operations. If not specified, this will use the configures primary account.
```
rguthriemsft: done

testing/citest/spinnaker_testing/azure_scenario_support.py
```diff
@@ -0,0 +1,96 @@
...
+        dest='spinnaker_azure_tenant_id',
+        help='The AAD tenant id.')
+    builder.add_argument(
+        '--azure_client_id', 
+        dest='spinnaker_azure_client_id',
+        help='The service principal id.')
+    builder.add_argument(
+        '--azure_app_key', 
+        dest='spinnaker_azure_app_key',
+        help='The service principal secret.')
```
rguthriemsft: No we don't support file method.  We could do environment variables but the way we plan to run these test is part of a jenkins job that deploys a templated spinnaker install and then scp's the -local.yaml file and grabs values and passes them into python/test case execution.

testing/citest/spinnaker_testing/azure_scenario_support.py
```diff
@@ -0,0 +1,96 @@
...
+    builder.add_argument(
+        '--azure_location', 
+        dest='spinnaker_azure_location',
+        help='The azure location to be used.')
+    builder.add_argument(
+        '--test_azure_rg_location', 
+        default = defaults.get('TEST_AZURE_RG_LOCATION', 'westus'),
+        help='The location of the azure resource group where test resources should be created.')
+    # builder.add_argument(
```
rguthriemsft: done

testing/citest/tests/azure_smoke_test.py
```diff
@@ -0,0 +1,235 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Smoke test to see if Spinnaker can interoperate with Microsoft Azure.
+
+----- Documentation to update --------
```
rguthriemsft: added documentation

testing/citest/README.md
rguthriemsft: done

testing/citest/spinnaker_testing/azure_scenario_support.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # pylint: disable=line-too-long
+    # builder.add_argument(
+    #     '--azure_subscription_name', 
+    #     dest='spinnaker_azure_account',
+    #     help='The name of your subscriptoin.')
+    builder.add_argument(
+        '--azure_subscription_id', 
+        dest='spinnaker_azure_subscription_id',
+        help='The subscription id of your subscriptoin.')
```
rguthriemsft: done

testing/citest/tests/azure_smoke_test.py
```diff
@@ -0,0 +1,235 @@
...
+      agent: [GateAgent] The agent for invoking the test operations on Gate.
+    """
+    super(AzureSmokeTestScenario, self).__init__(bindings, agent)
+    bindings = self.bindings
+
+    # We'll call out the app name because it is widely used
+    # because it scopes the context of our activities.
+    # pylint: disable=invalid-name
+    self.TEST_APP = bindings['TEST_APP']+AzureSmokeTestScenario.DEFAULT_TEST_ID
+    self.TEST_STACK = bindings['TEST_STACK']+AzureSmokeTestScenario.DEFAULT_TEST_ID
```
rguthriemsft: fixed

testing/citest/tests/azure_smoke_test.py
```diff
@@ -0,0 +1,235 @@
...
+    }]
+    builder = az.AzContractBuilder(self.az_observer)
+    (builder.new_clause_builder(
+        'Security Group Created', retryable_for_secs=30)
+     .collect_resources(
+         az_resource='network',
+         command='nsg',
+         args=['show', '--name', self.TEST_SECURITY_GROUP, '--resource-group', self.TEST_SECURITY_GROUP_RG],
+         no_resources_ok=True)
+     .contains_pred_list([
```
rguthriemsft: we plan to make the check more comprehensive after we get the initial set of test done.

testing/citest/spinnaker_testing/azure_scenario_support.py
ewiseblatt: Maybe default the value to an environment variable as an option?
If you pass keys on the command-line then they become visible in all the places that record command lines (process status, shell history, etc). You can still learn the key in your jenkins task doing the scp, but set an environment variable rather than passing another flag.

testing/citest/spinnaker_testing/spinnaker_test_scenario.py
```diff
@@ -35,12 +35,12 @@
...
-    appengine_scenario_support.AppEngineScenarioSupport,
-    kubernetes_scenario_support.KubernetesScenarioSupport,
-    openstack_scenario_support.OpenStackScenarioSupport,
+    # aws_scenario_support.AwsScenarioSupport,
+    # # appengine depends on google so order it after
+    # google_scenario_support.GoogleScenarioSupport,
+    # appengine_scenario_support.AppEngineScenarioSupport,
+    # kubernetes_scenario_support.KubernetesScenarioSupport,
+    # openstack_scenario_support.OpenStackScenarioSupport,
     azure_scenario_support.AzureScenarioSupport
```
ewiseblatt: are these a problem>
Why did you need to comment them out? Is there something I need to fix/change?
(and, obviously uncomment them)

rguthriemsft:
@EricJizbaMSFT @ewiseblatt PTAL, all test are working now.

testing/citest/spinnaker_testing/azure_scenario_support.py
ewiseblatt: I still think you should default the app key to an environment variable.
Passing the other parameters on the commandline is fine, but you wouldnt want to handle the credentials the same way so it just creates more work. You have a scheme where you get them all the same way from the deployment, but that wont work in general (e.g. when the deployment isnt azure, which wont always be the case). It isnt extra work because you dont have to use it, in which case you are back to where you already are here.

testing/citest/spinnaker_testing/azure_scenario_support.py
```diff
@@ -0,0 +1,88 @@
...
+        help='The service principal secret.')
+    builder.add_argument(
+        '--test_azure_rg_location', 
+        default = defaults.get('TEST_AZURE_RG_LOCATION', 'westus'),
+        help='The location of the azure resource group where test resources should be created.')
+    builder.add_argument(
+        '--azure_storage_account_name', 
+        dest='azure_storage_account_name',
+        help='The name of the Azure storage account used by Front50 in Spinnaker.')
+    builder.add_argument(
```
ewiseblatt: ditto for environment variable default value here too if this is a sensitive value.

testing/citest/spinnaker_testing/spinnaker.py
```diff
@@ -371,7 +371,7 @@ def new_native_instance(cls, name, status_factory, base_url):
       return None
 
     logger.info('%s is available at %s', name, base_url)
-    env_url = os.path.join(base_url, 'env')
+    env_url = os.path.join(base_url, 'resolvedEnv')
```
ewiseblatt: I made this fix this several days ago. Sync your repo.

testing/citest/tests/azure_smoke_test.py
```diff
@@ -0,0 +1,303 @@
...
+            description= 'Test - create application {app}'.format(app=self.bindings['TEST_APP']),
+            application=self.bindings['TEST_APP'])
+
+    builder = az.AzContractBuilder(self.az_observer)
+    (builder.new_clause_builder(
+        'Application Created', retryable_for_secs=30)
+      .collect_resources(
+          az_resource='storage',
+          command='blob',
+          args=['exists', '--container-name', 'front50',
```
ewiseblatt: What is "front50" here? Is this the name of an azure resource (e.g. are you assuming spinnaker is configured with a container named "front50" ? Is this something you hardcode into spinnaker, or was this a deployment choice? If a deployment choice, it should be parameterized (e.g. get it from the agent.deployed_config) so you can run the test against any deployment without having to overprescribe configuration.

testing/citest/tests/azure_smoke_test.py
```diff
@@ -0,0 +1,303 @@
...
+    (builder.new_clause_builder(
+        'Security Group Created', retryable_for_secs=30)
+    .collect_resources(
+         az_resource='network',
+         command='nsg',
+         args=['show', '--name', self.TEST_SECURITY_GROUP, '--resource-group', self.TEST_SECURITY_GROUP_RG],
+         no_resources_ok=True)
+    #sec grp name matches expected
+    .contains_match({'name': jp.STR_SUBSTR(self.TEST_SECURITY_GROUP)})
+    .contains_path_pred(
```
ewiseblatt: This looks older than our conversation earlier today.
I'm going to stop here and ping you because I think this is out of date.

ewiseblatt: My bad, my browser was clipping #207 and I didnt realize it. This looks current.

testing/citest/tests/azure_smoke_test.py
```diff
@@ -0,0 +1,303 @@
...
+    """
+    super(AzureSmokeTestScenario, self).__init__(bindings, agent)
+    bindings = self.bindings
+
+    # create default sec grp name  
+    # Spinnaker-azure appends region to actual RG used by an app to support multi-region
+    # pylint: disable=invalid-name
+    self.TEST_SECURITY_GROUP = 'sec_grp_' + self.bindings['TEST_APP']
+    self.TEST_SECURITY_GROUP_RULE_1=self.TEST_SECURITY_GROUP + '_RULE_1'
+    self.TEST_SECURITY_GROUP_RG = self.bindings['TEST_APP'] + '-' + self.bindings['TEST_AZURE_RG_LOCATION']
```
ewiseblatt: I assume the  GROUP_RULE_1 is an arbitrary name you are picking for this test
and he GROUP_RG is a specific name generated by your clouddriver heuristics.

If this is the case, I suggest you add a method to spinnaker_testing.azure_support that knows this policy and can return the names. it will serve as some documentation, add clarity to future tests, and help distinguish the arbitrary names from the intentional ones.

If you are following standard spinnaker frigga naming and I'm missing that mapping then you can use spinnaker_testing.frigga.Naming to generate the names.

testing/citest/tests/azure_smoke_test.py
rguthriemsft: its the name of the container that is set in spinnaker-local.yml.  we default it to front50 so its easy to understand what the storage container is for in the storage account.

testing/citest/spinnaker_testing/azure_scenario_support.py
rguthriemsft: I ave removed the subscription parameters.  I will look into the storage account key and putting into environment variable.  Can we push this and that is something I will look into immediately as I am unsure if we consider the storage account key sensitive info

testing/citest/spinnaker_testing/azure_scenario_support.py
rguthriemsft: app key is removed, we don't need it in the test as we assume you have already logged into the cli before running the test.

PR #1562, opened 2017-04-12
feat(nightly): Publish nightly test results to spinnaker.github.io.

jtk54:
@ttomsu Can you check that the post format is correct (front matter, etc)? I'll expand this in the future to push and update the changelog from the previous release tag.

### DIFF
--- a/dev/publish_test_results.py
+++ b/dev/publish_test_results.py
@@ -1,0 +1,127 @@
+#!/usr/bin/python
+#
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


dev/publish_test_results.py
```diff
@@ -0,0 +1,127 @@
...
+
+    A private key that has access to --githubio_repo needs added
+    to a running ssh-agent on the machine this script will run on:
+
+    > <copy or rsync the key to the vm>
+    > eval `ssh-agent`
+    > ssh-add ~/.ssh/<key with access to github repos>
+    """
+    self.__checkout_githubio_repo()
+    self.__read_test_results()
```
lwander: Should we publish something when we fail to run the tests?

jtk54: This implies either a build failure or a catastrophic test (failure to launch), either way these are problems before here. We can report that a different way, i.e. Jenkins build status, etc.

PR #1571, opened 2017-04-14
feat(build_image): Add script to build GCE image of one component.

jtk54:
I did a bit of a refactor of the build_*_image.sh functions so that things that are common are now in build_google_image_functions.sh. I added a second script to create an image per spinnaker component by using halyard, named with the component version derived from the toplevel version.

### DIFF
--- a/dev/halyard_install_component.sh
+++ b/dev/halyard_install_component.sh
@@ -1,0 +1,80 @@
+#!/bin/bash
+#
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Installs Halyard and installs a single component along with spinnaker-monitoring.
+
+set -e
+
+function show_usage() {
+    fix_defaults
+
+cat <<EOF
+Usage:  $0 [options]
+
+   --component COMPONENT
+       [$COMPONENT]
+       The name of the Spinnaker component to install. Does not contain the 'spinnaker-' prefix.
+
+   --version VERSION
+       [$VERSION]
+       The exact Spinnaker version we are baking images for.
+
+EOF
+}
+
+function process_args() {
+  while [[ $# > 0 ]]; do
+    local key="$1"
+    shift
+
+    case $key in
+        --help)
+            show_usage
+            exit
+            ;;
+
+        --component)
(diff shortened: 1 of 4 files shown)


dev/halyard_install_component.sh
```diff
@@ -0,0 +1,83 @@
...
+          >&2 echo "Unrecognized argument '$key'."
+          exit -1
+    esac
+  done
+}
+
+function main() {
+  echo "Installing git..."
+  sudo apt-get install -y git
+
```
ewiseblatt: where is this called from and where is it run?
Why does this install git and clone the repo every time?
What does _component mean in the script name?

If you are installing git and cloneing the repo to install halyard then that seems like a bad idea. Especially if this is outside a developer environment, which it appears.
I though halyard was in debian packages. If not you should be able to curl it from github and pipe that through bash.

jtk54:
When I was testing this I executed it from `/dev` with `./build_component_image.sh --version nightly --install_script halyard_install_component.sh --project <project images will be created in> --account <service account>`

dev/halyard_install_component.sh
jtk54: Previously I needed to copy a subdirectory of Halyard that contained configuration. I'll change this to a curl -> pipe.

dev/build_component_image.sh
```diff
@@ -0,0 +1,264 @@
...
+
+   --install_script INSTALL_SCRIPT
+       [$INSTALL_SCRIPT]
+       The path or URL to the install script to use.
+
+   --no_update_os
+       Do not force an upgrade-dist of the base OS.
+
+   --project PROJECT
+       [$PROJECT]
```
jtk54: OK

dev/build_component_image.sh
```diff
@@ -0,0 +1,264 @@
...
+  gcloud compute instances add-metadata $BUILD_INSTANCE \
+      --project $PROJECT \
+      --account $ACCOUNT \
+      --zone $ZONE \
+      --metadata-from-file ssh-keys=$HOME/.ssh/google_empty.pub
+
+  # This second instance will be used later to clean the image
+  # we dont need it yet, but will spin it up now to have it ready.
+  # Assumes a firewall rule allowing tcp:22 for this tag.
+  echo "`date` Warming up '$CLEANER_INSTANCE' for later"
```
jtk54: Should I make these changes or merge down and then let you refactor?

jtk54:
Essentially It just needs a service account configured with VM and image creation access in a project, the Spinnaker toplevel version for Halyard to unpack, and the install script to run on the VM.

jtk54:
@ewiseblatt updated this to include your comments minus further refactoring.

dev/halyard_install_component.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/bash
```
ewiseblatt: shoudl this be in spinnaker/install as opposed to spinnaker/dev?
The intent was that runtime scripts are in install, where developer only scripts (not part of an operational deployment) are in dev.

Otherwise LGTM

jtk54: This is meant to be used as a part of our build process, so I think it belongs in /dev.

PR #1573, opened 2017-04-15
fix(tests): Fixed native test startup / configuration

ewiseblatt:
Get configuration parameters from the "resolveEnv" endpoint.
Show exception information if agent fails
If not given a location, assume localhost

The net effect is that if you are on the same machine as spinnaker,
and the tests primary microservice is on its standard port, then
you can run a typical test without any command-line parameters provided
it can get whatever configuration it needs from the server.

@jtk54  
@rguthriemsft

### DIFF
--- a/testing/citest/spinnaker_testing/spinnaker.py
+++ b/testing/citest/spinnaker_testing/spinnaker.py
@@ -256,7 +256,10 @@ class SpinnakerAgent(service_testing.HttpAgent):
       elif bindings['NATIVE_HOSTNAME']:
         host_platform = 'native'
       else:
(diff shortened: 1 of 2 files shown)


testing/citest/spinnaker_testing/spinnaker.py
```diff
@@ -256,7 +256,11 @@ def __determine_host_platform(cls, bindings):
       elif bindings['NATIVE_HOSTNAME']:
         host_platform = 'native'
       else:
-        raise ValueError('No --native_hostname nor --gce_project')
+        # raise ValueError('No --native_hostname nor --gce_project')
```
jtk54: Remove commented line.

PR #1578, opened 2017-04-18
feat(changelog): script for publishing changelog

danielpeach:
@lwander or @ttomsu please review.


dev/publish_changelog.py
```diff
@@ -0,0 +1,119 @@
...
+class ChangelogPublisher():
+
+  def __init__(self, options):
+    self.__version = options.version
+    self.__githubio_repo_uri = options.githubio_repo_uri
+    self.__repo_name = ''
+    self.__changelog_gist_uri = options.changelog_gist_uri
+
+
+  def __checkout_githubio_repo(self):
```
lwander: It's a nitpick, but it feels like this function should return the repo path rather than store it as internal state

dev/publish_changelog.py
```diff
@@ -0,0 +1,119 @@
...
+    to origin/master.
+
+    A private key that has access to --githubio_repo needs to be added
+    to a running ssh-agent on the machine this script will run on:
+
+    > <copy or rsync the key to the vm>
+    > eval `ssh-agent`
+    > ssh-add ~/.ssh/<key with access to github repos>
+    
+    If you are running this script on Jenkins, you can configure Jenkins to handle SSH credentials.
```
lwander: This seems to be the only doc comment that exceeds 80 chars

dev/publish_changelog.py
```diff
@@ -0,0 +1,119 @@
...
+    day = '{:%Y-%m-%d}'.format(datetime.datetime.now())
+    post_name = '{day}-{version}-changelog.md'.format(day=day, version=self.__version)
+    post_path = os.path.join(self.__repo_name, POSTS_DIR, post_name)
+    # Path to post file relative to the git root.
+    post_rel_path = os.path.join(POSTS_DIR, post_name)
+    with open(post_path, 'w') as post_file:
+      post_file.write(post_content)
+
+    check_run_quick('git -C {0} add {1}'.format(self.__repo_name, post_rel_path))
+    message = 'Changelog for version {0}'.format(self.__version)
```
lwander: I'd add a comment that this commit was auto-generated by a script (you can get the name of this current script by using `__file__`, a global constant)

dev/publish_changelog.py
danielpeach: Good point. It feels like this function shouldn't really deal with the repo path at all.

Issue #1589, opened 2017-04-21
[GKE] [Kubernetes] deployment guide not working on GKE


lihan:
Cloud Provider:  **GKE**, Kubernetes

AWS, GCP, Kubernetes, Azure, Cloud Foundry, etc. Please assign a label from the right so your issue can be properly sorted.

*Environment*

I am trying to deploy Spinnaker using [this guide](https://github.com/spinnaker/spinnaker/tree/master/experimental/kubernetes/simple) onto GKE. ```

*Feature Area*

Install

*Description*

Following the guide will not create a working Spinnaker

*Steps to Reproduce*

Follow this [guide](https://github.com/spinnaker/spinnaker/tree/master/experimental/kubernetes/simple) and trying to create it on GKE 

*Additional Details*

At least 
1. Documentation should state that `services.desk.host` from `spinnaker-local.yml` is to change to `localhost`, or you will hit `access-control-allow-origin` issue, and after that you should restart `gate` service.
2. After solve above 1, hit [this issue](https://github.com/spinnaker/spinnaker/issues/1276)

lwander:
Deployment guide is deprecated in favor of [Halyard](https://www.spinnaker.io/setup/install/halyard/).

Issue #1276, opened 2016-10-30
Deck in Kubernetes always appends provider=aws to end of API queries


plombardi89:
An issue is not a place to ask questions. Please use [Slack](http://join.spinnaker.io) or [Stack Overflow](http://stackoverflow.com/questions/tagged/spinnaker). 

Before you open an issue, please check if a similar issue already exists or has been closed before.

Make sure you have followed the steps of the [Troubleshooting Guide](http://www.spinnaker.io/docs/troubleshooting-guide) to provide as much information as possible. 

Please provide:

_Deck in Kubernetes always appends provider=aws to end of API queries_

When you connect to the Deck UI after following the deploy instructions for Kubernetes (simple). You end up in a situation where much of the UI does not work because deck is making queries to gate with `?provider=aws` appended to each URL. This causes a 404.

_Cloud Provider_

Kubernetes

_Environment_

I followed the instructions in `experimental/spinnaker/simple`. 

_Feature Area_

Deck

_Description_

Spinning wheel of nothingness when trying to define a Cluster for Kubernetes. I popped open the Chrome Developer console and noticed 404 errors due to Spinnaker trying to hit URL's such as `http://localhost:8084/loadBalancers?provider=aws`

Returns:

``` json
{"error":"Not Found","failureCause":"retrofit.RetrofitError: 404 Not Found","message":"Not Found","status":404,"timestamp":1477788723158,"url":"http://spin-clouddriver.spinnaker.svc.cluster.local:7002/aws/loadBalancers"}
```

Hitting what should be the right URL: `http://localhost:8084/loadBalancers?provider=kubernetes` generates more sane output:

``` json
[]
```

_Steps to Reproduce_
1. Deploy Spinnaker via the instructions in `experimental/spinnaker/simple`
2. Connect to Deck
3. Create an Application (e.g. `test`)
4. Go to cluster page and click "Create Server Group"
5. Spinning wheel of nothing.

_Additional Details_

I don't see anything out the ordinary in the logs.

lwander:
@danielpeach when you get a chance can you take a look?

plombardi89:
If someone can point me at a good place to start debug this I might have some free cycles to mess around in the next day or two. My current working theory is that Deck requires `reconfigure_spinnaker.sh` to be run per the Troubleshooting Guide for a classic VM install.

tomaslin:
I would suggest troubleshooting this in the Slack channel if you want to get more interactive feedback instead of github issues.

[...]

PR #1598, opened 2017-04-24
docs(experimental/kubernetes): Added common pitfalls when setting up Spinnaker on Kubernetes

lihan:
docs(experimental/kubernetes): Added common falls when setting Spinnaker on kubernetes

This is to address https://github.com/spinnaker/spinnaker/issues/1589

### DIFF
--- a/experimental/kubernetes/simple/README.md
+++ b/experimental/kubernetes/simple/README.md
@@ -211,3 +211,14 @@ mechanism (Redis), run
 ```
 $ bash scripts/cleanup-spinnaker.sh # This deletes everything with application name spin
 ```
+
+
+
+## Common Pitfalls
+
+1. If you are running on GKE, and its UI isn't working. Check all logs from `clouddriver` Pod. If the errors are relating kubernates accounts, many relate to [this issue](http://www.spinnaker.io/docs/troubleshooting-guide#section-my-gke-google-container-engine-cluster-isn-t-showing-up-as-a-kubernetes-account-). After the change, make sure you re-fetch the credentials by `gcloud container clusters get-credentials $CLUSTER_NAME` again. 
+2. If Spinnaker's UI isn't working properly, and if your browser console has `Access-Control-Allow-Origin` errors, try including the following block in your spinnaker-local.yml. You need to run `scripts/update-config.sh` to update config and restart `gate` service to make this change to take effect
+```
+cors:
+  allowedOriginsPattern: http://localhost:9000
+``` 


experimental/kubernetes/simple/README.md
```diff
@@ -211,3 +211,14 @@ mechanism (Redis), run
 ```
 $ bash scripts/cleanup-spinnaker.sh # This deletes everything with application name spin
 ```
+
+
+
+## Common falls
```
danielpeach: How about `Common Pitfalls`?

experimental/kubernetes/simple/README.md
```diff
@@ -211,3 +211,14 @@ mechanism (Redis), run
...
+
+1. clouddriver not configured properly, check all the logs from `clouddriver` pod, many relate to [this link](http://www.spinnaker.io/docs/troubleshooting-guide#section-my-gke-google-container-engine-cluster-isn-t-showing-up-as-a-kubernetes-account-)
```
danielpeach: I wouldn't call this a Clouddriver configuration issue - and it might be helpful to provide more context here (e.g., if you're running on GKE, if your browser console is showing these errors, etc.). You could also provide a link to an issue (e.g., https://github.com/spinnaker/spinnaker/issues/1276), which has a lot of context.

experimental/kubernetes/simple/README.md
```diff
@@ -211,3 +211,14 @@ mechanism (Redis), run
...
+2. http://localhost:9000 hits `Access-Control-Allow-Origin` issue. Include following block to `spinnaker-local.yml`
```
danielpeach: Better descriptions of symptoms would also be helpful here - e.g., "If Spinnaker's UI isn't working properly, and if your browser console has `Access-Control-Allow-Origin` errors, try including the following block in your `spinnaker-local.yml`..."

lihan:
@danielpeach Thanks for the improvements, pushed my updates on the branch.

robotmay:
Any chance you could expand on the CORS issue a bit? I've actually just run into that problem but I don't know where in the `spinnaker-local.yaml` file that block should go, and it doesn't seem to have had an effect so far. I'm also fairly new to kubernetes, so I'm not sure if I'm missing anything, but when you say to restart the 'gate' service, is it sufficient to kill the existing pod, or is there another way of doing this?

lihan:
@robotmay Add CORS to the root of the `spinnaker-local.yaml`

With regards to restart `gate` service, you simply do `bash scripts/update-config.sh gate` command from `kubernetes/simple` directory.

robotmay:
Great, thanks @lihan. In the end I got it working by putting it in the `gate-local.yaml` file (as I think it's only needed there), ran the update script, then deleted the existing container and it finally took effect.

lwander:
Closing as the experimental install has been removed

PR #1599, opened 2017-04-24
feat(halyard_release): Implement support for Halyard release tracks.

jtk54:
This makes the change to publish Halyard to the 'trusty-nightly' distribution rather than just 'trusty'. We record the version built during the nightly release and a commit hash in a GCS file so we can promote Halyard by re-building it later from the same commit hash.

I'll implement the second part (promoting Halyard) next.

### DIFF
--- a/dev/annotate_source.py
+++ b/dev/annotate_source.py
@@ -241,11 +241,11 @@ class Annotator(object):
     This is so that gradle will use the latest resolved semantic version from
     our tag pattern when it builds the package.
     """
+    print ('Deleting {0} unwanted git tags locally from {1}'
+           .format(len(self.__tags_to_delete), self.path))
     for bad_hash_tag in self.__tags_to_delete:
-      # NOTE: The following command prints output to STDOUT, so we don't
-      # explicitly log anything.
       run_quick('git -C {path} tag -d {tag}'
-                .format(path=self.path, tag=bad_hash_tag.tag))
+                .format(path=self.path, tag=bad_hash_tag.tag), echo=False)
 
   def checkout_branch(self):
     """Checks out a branch.
@@ -253,16 +253,22 @@ class Annotator(object):
     run_quick('git -C {path} checkout {branch}'.format(path=self.path,
                                                        branch=self.branch))
 
+  def get_head_commit(self):
+    """Retrieves the head commit hash.
+    """
+    head_commit_res = run_quick('git -C {path} rev-parse HEAD'
+                                .format(path=self.path),
+                                echo=False)
+    return head_commit_res.stdout.strip()
+
+
   def __is_head_current(self):
     """Checks if the current version is at HEAD.
 
     Returns:
       [Boolean]: True if the current version tag is on HEAD, else False.
     """
-    head_commit_res = run_quick('git -C {path} rev-parse HEAD'
-                                    .format(path=self.path),
-                                echo=False)
-    head_commit = head_commit_res.stdout.strip()
+    head_commit = self.get_head_commit()
     return self.__current_version.hash == head_commit
 
   def __determine_current_version(self):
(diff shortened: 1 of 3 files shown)


dev/annotate_source.py
```diff
@@ -241,28 +241,34 @@ def delete_unwanted_tags(self):
...
     """Checks out a branch.
     """
     run_quick('git -C {path} checkout {branch}'.format(path=self.path,
                                                        branch=self.branch))
 
+  def get_head_commit(self):
+    """Retrieves the head commit hash.
+    """
+    head_commit_res = run_quick('git -C {path} rev-parse HEAD'
+                              .format(path=self.path),
```
lwander: This indentation seems a little funny

dev/build_prevalidation.py
```diff
@@ -23,16 +23,48 @@
...
+  bucket_uri = options.hal_nightly_bucket_uri
+  build_number = options.build_number or os.environ.get('BUILD_NUMBER', '')
+  local_bucket_name = os.path.basename(bucket_uri)
+  # Copy all the bucket contents to local (-r) and get rid of extra stuff (-d).
+  if not os.path.exists(local_bucket_name):
+    os.mkdir(local_bucket_name)
+  check_run_quick('gsutil rsync -r -d {remote_uri} {local_bucket}'
+                  .format(remote_uri=bucket_uri, local_bucket=local_bucket_name))
+  hal_version = version_bump.version_str.replace('version-', '')
+  new_hal_nightly_entry = ('{version}-{build}={commit}'
```
lwander: Why not record this in `yaml` format to make parsing this easier later?

dev/build_prevalidation.py
```diff
@@ -23,16 +23,48 @@
...
+                           .format(version=hal_version, build=build_number, commit=head_hash))
+  nightly_entry_file = '{0}/nightly-version-commits.yml'.format(local_bucket_name)
+  with open(nightly_entry_file, 'a') as nef:
+    nef.write('{0}\n'.format(new_hal_nightly_entry))
+  # Now sync the local dir with the bucket again after the update.
+  check_run_quick('gsutil rsync -r -d {local_bucket} {remote_uri}'
```
lwander: Does `rsync` do any sort of optimistic locking? If not, do we have to be careful not to run builds concurrently for other services?

dev/build_prevalidation.py
jtk54: Are you suggesting making this a nested list, or just keep the same structure but change `=` to `:`?

dev/build_prevalidation.py
jtk54: Reading through `gsutil rsync --help` - says it uses bucket listing which is eventually consistent, so we do need to be careful about running concurrent tasks trying to update the same bucket at the same time. That being said, there is only one nightly job touching the GCS file in question. I'll add a comment to the docstring about the concurrency of this function though.

dev/build_prevalidation.py
lwander: just `=` :arrow_forward: `: `

dev/build_prevalidation.py
```diff
@@ -23,16 +23,53 @@
...
+
+  Assumes that gsutil is installed on the machine this script is run from.
+
+  This function uses `gsutil rsync` to read the GCS file, changes it in-place,
+  and then uses `gsutil rsync` to write the file again. `rsync` is eventually
+  consistent, so running this script (or manually manipulating the GCS file)
+  concurrently could likely result in file corruption. Don't parallelize this.
+  """
+  bucket_uri = options.hal_nightly_bucket_uri
+  build_number = options.build_number or os.environ.get('BUILD_NUMBER', '')
```
jtk54: Turns out it already is, taking this out.

PR #1601, opened 2017-04-26
fix(bake): Fixes bake of consul & vault servers

lwander:
The problem was the we were baking images based on the `artifact` alone, when it should have been the `service`. However, the `artifact` needs to be known to name & version the baked image, so I put both `artifact`s & `service`s into an associative array.

### DIFF
--- a/dev/build_google_component_image.sh
+++ b/dev/build_google_component_image.sh
@@ -208,15 +208,16 @@ function create_component_prototype_disk() {
 
 
 function create_component_image() {
-  local comp=$1
-  COMP_VERSION="$(hal version bom $VERSION --artifact-name ${comp} --quiet --color false)"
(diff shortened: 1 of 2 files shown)


dev/build_google_component_image.sh
```diff
@@ -245,7 +246,19 @@ function fix_defaults() {
 
 process_args "$@"
 
-declare -a COMPONENTS=('clouddriver' 'deck' 'echo' 'fiat' 'front50' 'gate' 'igor' 'orca' 'rosco' 'consul' 'vault' 'redis')
+declare -A COMPONENTS=( ['clouddriver']='clouddriver' \
```
jtk54: Can you describe what's going on here in a comment, and point out which names are artifacts and which names are services?

PR #1609, opened 2017-04-27
feat(spinnaker_release): Add orchestration script for Spinnaker release.

jtk54:


dev/publish_bom.py
```diff
@@ -122,13 +121,14 @@ def publish_changelog_gist(self):
       # Also add some identifying information.
       raw_content_lines[0] = spinnaker_version
       timestamp = '{:%Y-%m-%d %H:%M:%S}'.format(datetime.datetime.now())
-      signature = '\n\nGenerated by {0} at {1}'.format(socket.gethostname(), timestamp)
+      signature = '\n\nGenerated by at {0}'.format(timestamp)
```
ewiseblatt: I think you should just replace socket.gethostname() with the logical user (e.g. spinnaker-release or organization).
If it is just a timestamp then s/by //. However I think there should be organizational traceability here.

dev/publish_bom.py
```diff
@@ -122,13 +121,14 @@ def publish_changelog_gist(self):
...
       raw_content_lines.append(signature)
       content = InputFileContent(''.join(raw_content_lines))
       filename = os.path.basename(self.__changelog_file)
       gist = self.__github.get_user().create_gist(True, {filename: content}, description=description)
-      print ('Wrote changelog to Gist at https://gist.github.com/{user}/{id}'
-             .format(user=self.__gist_user, id=gist.id))
+      gist_uri = 'https://gist.github.com/{user}/{id}'.format(user=self.__gist_user, id=gist.id)
+      print ('Wrote changelog to Gist at {0}.'.format(gist_uri))
+      return gist_uri
```
ewiseblatt: You didnt actually write it, you are just initializing the uri.
The write happens later -- and could be written elsewhere.

dev/publish_changelog.py
```diff
@@ -64,6 +64,11 @@ def __format_changelog_post(self):
       ''
     ]
 
+    # Since we include the changelog gist as a script, make sure we're
+    # referring to the gist javascript file.
+    if not self.__changelog_gist_uri.endswith('.js'):
```
ewiseblatt: if self.__changelog_gist_uri and not self.__changelog_gist_uri.endswith('.js')

dev/publish_bom.py
jtk54: We do actually write it in `create_gist` on L128.

dev/publish_bom.py
jtk54: I like using the github publisher, adding that in.

jtk54:
I'm about to do a bit of a refactor of this script to automate opening a PR with the changelog post, so I'll address the empty `githubio_uri` then.

PR #1618, opened 2017-05-02
fix(publish_bom): Set versions.yml entry link to changelog gist.

jtk54:


dev/publish_bom.py
```diff
@@ -126,9 +127,9 @@ def publish_changelog_gist(self):
       content = InputFileContent(''.join(raw_content_lines))
       filename = os.path.basename(self.__changelog_file)
       gist = self.__github.get_user().create_gist(True, {filename: content}, description=description)
-      gist_uri = 'https://gist.github.com/{user}/{id}'.format(user=self.__gist_user, id=gist.id)
-      print ('Wrote changelog to Gist at {0}.'.format(gist_uri))
-      return gist_uri
+      self.__gist_uri = 'https://gist.github.com/{user}/{id}'.format(user=self.__gist_user, id=gist.id)
```
lwander: Does ID include `.js`? Also https://github.com/spinnaker/deck/pull/3611#discussion_r114327299

jtk54: No it doesn't contain the `.js`.

PR #1626, opened 2017-05-05
fix(prevalidate): Support for building patch releases.

jtk54:


dev/generate_bom.py
```diff
@@ -146,11 +146,16 @@ def write_container_builder_gcr_config(self):
...
         gradle_version = self.__version_from_tag(comp)
+
+        # Gradle complains if the git tag version doesn't match the branch's version, i.e.
+        # in patch releases. As a workaround, we checkout the HEAD commit of the
+        # current branch in a 'detached HEAD' state so Gradle doesn't complain about
+        # our branch version and git tag version not matching.
         gradle_cmd = ''
         if comp == 'deck':
-          gradle_cmd = './gradlew build -PskipTests'
+          gradle_cmd = 'git rev-parse HEAD | xargs git checkout ; ./gradlew build -PskipTests'
```
ewiseblatt: why not add factor out this common command prefix since it is something you always do independent of the component?

jtk54: I'll initialize `gradle_cmd` with that, good call.

PR #1635, opened 2017-05-11
fix(component_image): Don't leave residual images in build project.

jtk54:
…ect. This also deletes images in the publish project if we are recreating them. Additionally, I changed how we wait for the subprocesses so that errors in any subprocess should be surfaced now.

### DIFF
--- a/dev/build_google_component_image.sh
+++ b/dev/build_google_component_image.sh
@@ -15,6 +15,7 @@
 # limitations under the License.
 
 set -e
+set -o pipefail
 
 # Import some functions from other scripts.
 source $(dirname $0)/build_google_image_functions.sh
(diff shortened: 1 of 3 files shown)


dev/build_google_component_image.sh
```diff
@@ -251,25 +263,37 @@ declare -A COMPONENTS=( ['clouddriver']='clouddriver' \
...
+for job in $(jobs -p); do
+  echo "Waiting for job $job"
+  wait $job || let "FAILED+=1"
+done
+
+if [ "$FAILED" == "0" ]; then
+  echo "All jobs succeeded."
+else
+  echo "Some jobs failed. Exiting..."
+  exit "$FAILED"
```
lwander: I wouldn't exit until you hit the log statement below

PR #1636, opened 2017-05-11
feat(testing): Script for deploying and validating a bom.

ewiseblatt:
Initial version only supports localdebian install on Google.
However the framework and process are set up to accomodate other
deployment types and platforms.

@lwander

### DIFF
--- a/dev/all_tests.yaml
+++ b/dev/all_tests.yaml
@@ -1,0 +1,131 @@
+# This file is used to specify the tests used by the validate_bom.sh script.
+# See validate_bom__test.py for more info.
+
+aliases:
+  standard_google_provider_params:
+    managed_gce_project: $google_account_project
+    spinnaker_google_account: $google_account_name
+    gce_credentials_path: $google_account_credentials
+
+  standard_aws_provider_params:
+    spinnaker_aws_account: $aws_account_name
+    aws_profile: citest
+
+
+tests:
+  bake_and_deploy_test:
+    requires:
+      configuration:
+        google_account_enabled: true
+        jenkins_master_enabled: true
+      services: [clouddriver, echo, front50, gate, igor, orca, rosco]
+    quota:
+      google_cpu: 1
+    api: gate
+    args:
+      alias: [standard_google_provider_params]
+      test_google: true
+      jenkins_token: TRIGGER_TOKEN
+      jenkins_job: TestTriggerProject
+      jenkins_url: $jenkins_master_address
+      jenkins_master: $jenkins_master_name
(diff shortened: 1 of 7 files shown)


dev/validate_bom__config.py
```diff
@@ -0,0 +1,439 @@
...
+    """Implements interface."""
+    if options.gcs_storage_credentials:
+      file_set.add(options.gcs_storage_credentials)
+
+  def add_config(self, options, script):
+    """Implements interface."""
+    project = options.gcs_storage_project or options.google_deploy_project
+    if options.spinnaker_storage == 'gcs':
+      hal = (
+          'hal --color=false config storage gcs edit --debug'
```
lwander: is `--debug` still needed?

dev/validate_bom__deploy.py
```diff
@@ -0,0 +1,392 @@
...
+    options = self.options
+    return [
+        'gcloud', 'alpha', 'compute', 'ssh',
+        '--quiet', '--verbosity=none',
+        options.google_deploy_instance,
+        '--ssh-key-file={ssh_key}'.format(ssh_key=self.EMPTY_SSH_KEY),
+        '--project={project}'.format(project=options.google_deploy_project),
+        '--zone={zone}'.format(zone=options.google_deploy_zone),
+        '--', '-L {local_port}:localhost:{remote_port}'.format(
+            local_port=local_port, remote_port=remote_port)]
```
lwander: Since you're using this just to open a connection, you can pass `-N` as well to avoid spawning a shell on the target machine.

dev/validate_bom__test.py
```diff
@@ -0,0 +1,757 @@
...
+    print 'TERMINATE: {0}'.format(test_name)
+    thread_pool.terminate()
+    return True
+
+  def add_extra_arguments(self, test_name, args, commandline):
+    """Add extra arguments to the commandline.
+
+    Args:
+      test_name: [string] Name of test specifying the options.
+      args: [dict] Specification of additioanl arguments to pass.
```
lwander: How do you pass valueless flags?

dev/all_tests.yaml
```diff
@@ -0,0 +1,131 @@
...
+    aws_profile: citest
+
+
+tests:
+  bake_and_deploy_test:
+    requires:
+      configuration:
+        google_account_enabled: true
+        jenkins_master_enabled: true
+      services: [clouddriver, echo, front50, gate, igor, orca, rosco]
```
lwander: Nice, I really like this.

PR #1637, opened 2017-05-11
fix(echo): Add Slack botName to echo configs

catherinetcai:
An extremely small fix. The Spinnaker [config docs] (http://www.spinnaker.io/docs/target-deployment-configuration) imply that all configs will get pulled over from the spinnaker.yml or spinnaker-local.yml. This is not the case with the Slack notifications. The botName configuration needs to be pulled into the echo.yml in order for notifications to actually trigger.


anotherchrisberry:
Nice, thanks for catching this!

PR #1641, opened 2017-05-13
feat(profiles): Publish a tar.gz for profile directories

lwander:
@jtk54 what's the best way for me to run/validate this?

### DIFF
--- a/dev/generate_bom.py
+++ b/dev/generate_bom.py
@@ -277,7 +277,7 @@ class BomGenerator(Annotator):
     result = run_quick('hal admin publish bom --color false --bom-path {0}'
                        .format(bom_path))
     if result.returncode != 0:
-      print "'hal admin publish bom' command failed with: \n{0}\n exiting...".format(result.stdout)
+      print "'hal admin publish bom' command failed with: \n{0}\nexiting...".format(result.stdout)
       exit(result.returncode)
 
   def __publish_config(self, component, profile_path):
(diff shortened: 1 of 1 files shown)


dev/generate_bom.py
```diff
@@ -277,7 +277,7 @@ def publish_bom(self, bom_path):
     result = run_quick('hal admin publish bom --color false --bom-path {0}'
                        .format(bom_path))
     if result.returncode != 0:
-      print "'hal admin publish bom' command failed with: \n{0}\n exiting...".format(result.stdout)
+      print "'hal admin publish bom' command failed with: \n{0}\nexiting...".format(result.stderr)
```
jtk54: @ewiseblatt correct me if I'm wrong, isn't `stderr` piped through `stdout` in `run_quick`?

dev/generate_bom.py
```diff
@@ -289,14 +289,25 @@ def __publish_config(self, component, profile_path):
     """
     for profile in os.listdir(profile_path):
       full_profile = os.path.join(profile_path, profile)
+      profile_path = full_profile
```
jtk54: I think `profile_path` and `full_profile` will always point to the individual profile files in `halconfig` the way this is invoked now. Seems like you can get rid of the loop and just tar up the `profile_path`, which is a directory.

jtk54: Talked offline, turns out there can be directories in the halconfig `profile_path`, which we'd want to tar up in the future, while treating the toplevel files the same.

dev/generate_bom.py
lwander: Yup looks like that's right

ewiseblatt: your right.
we can split them with a kwarg if you need.

PR #1643, opened 2017-05-14
feat(testing): ValidateBom with kubernetes deployment.

ewiseblatt:
@lwander 

Also fixes some aws configuration setup, but not enough to properly configure AWS.

On the k8s front, this basically adds a K8s and Docker configurator and changes the deployer into two roles -- hal and spinnaker. They are both the same implementation but the hal role is just for deploying and running hal, where the spinnaker role is for running spinnaker. Since hal deploys spinnaker, the main concern with the role here is to port forward into it to run the tests. So the port forwarding is on the spinnaker role, not the hal role.

The deployer class methods are such that they will mediate to the right role and it will bind the roles depending on the type of deployment you make.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -73,6 +73,10 @@ class StorageConfigurator(object):
     if options.spinnaker_storage not in ['gcs']:
       raise ValueError('Unknown --spinnaker_storage="{0}"'
                        .format(options.spinnaker_storage))
+    if options.spinnaker_storage == 'gcs':
+      if not options.gcs_storage_bucket:
+        raise ValueError('Specified --spinnaker_storage="gcs"'
+                         ' but not --gcs_storage_bucket')
 
   def add_files_to_upload(self, options, file_set):
     """Implements interface."""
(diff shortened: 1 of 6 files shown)


dev/validate_bom__config.py
```diff
@@ -250,6 +275,61 @@ def add_files_to_upload(self, options, file_set):
...
+    account_params = [options.docker_account_name,
+                      '--address', options.docker_account_address]
+    if options.docker_account_credentials:
+      cred_basename = os.path.basename(options.docker_account_credentials)
+      script.append('sudo chmod 600 {file}'.format(file=cred_basename))
+      account_params.extend(
+          ['--password-file', options.docker_account_credentials])
+    if options.docker_account_registry_username:
+      account_params.extend(
+          ['--password-file', options.docker_account_credentials])
```
lwander: This should probably be `--username`, no?

dev/validate_bom__deploy.py
```diff
@@ -124,48 +130,58 @@ def deploy(self, config_script, files_to_upload):
...
-    options = self.options
-    raise NotImplementedError(options.deploy_platform)
+    raise NotImplementedError(self.__class__.__name__)
 
-  def undeploy_helper(self):
+  def do_undeploy(self):
     """Hook for specialized platforms to implement the concrete undeploy()."""
-    options = self.options
-    raise NotImplementedError(options.deploy_platform)
+    raise NotImplementedError(self.__class__.__name__)
```
lwander: I think you can safely call `hal deploy clean` here

PR #1645, opened 2017-05-15
fix(generate_bom): Unconflate tarfile and profile paths.

jtk54:


dev/generate_bom.py
```diff
@@ -289,12 +289,12 @@ def __publish_config(self, component, profile_path):
...
     for profile in os.listdir(profile_path):
       full_profile = os.path.join(profile_path, profile)
-      profile_path = full_profile
       if os.path.isfile(full_profile):
+        # Don't zip with `tar` if the profile is a file.
         pass
       elif os.path.isdir(full_profile):
-        profile_path = '{0}.tar.gz'.format(full_profile)
-        result = run_quick('tar -cvf {0} {1}/*'.format(profile_path, full_profile))
+        tar_path = '{0}.tar.gz'.format(full_profile)
```
lwander: This guy was set to `profile_path` because that's what needs to be uploaded

PR #1660, opened 2017-05-18
fix(validate): Add --deploy_verison

ewiseblatt:
@jtk54

### DIFF
--- a/dev/validate_bom__deploy.py
+++ b/dev/validate_bom__deploy.py
@@ -236,6 +236,11 @@ class BaseValidateBomDeployer(object):
   def add_hal_deploy_script_statements(self, script):
     """Adds the hal deploy statements prior to "apply"."""
     options = self.options
+
+    script.append('hal --color=false config version edit'
+                  ' --version {version}'
+                  .format(version=options.deploy_version))
+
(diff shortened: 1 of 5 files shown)


dev/validate_bom__deploy.py
```diff
@@ -640,6 +649,10 @@ def make_bool_value(value):
            ' --deploy_spinnaker_type=distributed')
 
   parser.add_argument(
+      '--deploy_version', default='nightly',
```
jtk54: Default is now `<branch>-latest-unvalidated`.

dev/validate_bom__main.py
```diff
@@ -118,7 +118,8 @@ def main():
     test_controller.run_tests()
   finally:
     if sys.exc_info()[0] is not None:
-      logging.exception('Caught Exceptio')
+      logging.error('Caught Exception')
+      logging.exception('Caught Exception')
```
jtk54: Repeated this log statement.

ewiseblatt:
LGTM but fix the test faliure.

ewiseblatt:
Above comment was to wrong pr

PR #1664, opened 2017-05-19
fix(azure): Allow azure smoke test to run from anywhere.

ewiseblatt:
I took out the app tests that assumed azs storage.
These belong in an azure_front50_test.

I kept the old implementation around so you can migrate them there.

@rguthriemsft 

If you need this as is for the time being because of internal processes
you may have, I can add a flag here controlling the behavior. However,
ideally I'd like to have consistent expectations across platforms;
the suggestion here is the current thought.


ewiseblatt:
@dcaro

dcaro:
LGTM

PR #1665, opened 2017-05-19
chore(validate): Cleanup error handling and reporting.

ewiseblatt:
@lwander

### DIFF
--- a/dev/validate_bom__deploy.py
+++ b/dev/validate_bom__deploy.py
@@ -148,9 +148,16 @@ class BaseValidateBomDeployer(object):
     script = []
     self.add_install_hal_script_statements(script)
     self.add_platform_deploy_script_statements(script)
+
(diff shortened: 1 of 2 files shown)


dev/validate_bom__deploy.py
```diff
@@ -148,6 +148,13 @@ def deploy(self, config_script, files_to_upload):
     script = []
     self.add_install_hal_script_statements(script)
     self.add_platform_deploy_script_statements(script)
+
+    # Add the version first to avoid warnings or facilitate checks
+    # with the configuration commands
+    script.append('hal --color=false config version edit'
```
lwander: If this is a general reporting cleanup I would recommend replacing `--color=false` with `-q --log=all`

ewiseblatt: done

lwander:
LGTM

PR #1679, opened 2017-05-23
fix(publish_changelog): Don't use full path name for file.

jtk54:
Removes the unnecessary whole file path to the script generating the changelog.


jtk54:
i.e. in https://github.com/spinnaker/spinnaker.github.io/pull/110.

PR #1684, opened 2017-05-24
fix(aws/init): Adding missing '/' in meta-data AZ url

mmlac:
fix(aws/init): Adding missing '/' in meta-data AZ url

Availability zone meta-data endpoint will always return an empty response if the trailing / is missing.

### DIFF
--- a/packer/init.aws.region
+++ b/packer/init.aws.region
@@ -16,7 +16,7 @@ function get_aws_metadata_value() {
 
 
 function set_aws_defaults_from_environ() {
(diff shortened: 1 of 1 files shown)


mmlac:
This might have broken starting spinnaker in a private subnet on i.e. us-west-1 as clouddriver will default to trying to reach us-east-1, which is not allowed without a NAT. (not confirmed this fixes that issue)

https://spinnakerteam.slack.com/archives/C091CCWRJ/p1495655524866708

PR #1686, opened 2017-05-25
fix(bake): Allow auto-delete flag to propagate

lwander:
A race condition is my only guess as to why sometimes the `--no-auto-delete` behavior isn't preserved every now and then on one or two of our builds. PTAL @ewiseblatt

### DIFF
--- a/dev/build_google_component_image.sh
+++ b/dev/build_google_component_image.sh
@@ -173,13 +173,15 @@ function create_component_prototype_disk() {
       --command="sudo DEBIAN_FRONTEND=noninteractive apt-get -y dist-upgrade && sudo apt-get autoremove -y"
   fi
 
-  echo "`date`: Deleting '$BUILD_INSTANCE' but keeping disk"
+  echo "`date`: Setting auto-delete behavior '$BUILD_INSTANCE' disk to false"
   gcloud compute instances set-disk-auto-delete $BUILD_INSTANCE \
     --project $BUILD_PROJECT \
     --account $ACCOUNT \
     --zone $ZONE \
     --no-auto-delete \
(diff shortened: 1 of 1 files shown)


dev/build_google_component_image.sh
```diff
@@ -173,14 +173,16 @@ function create_component_prototype_disk() {
...
-  echo "`date`: Deleting '$BUILD_INSTANCE' but keeping disk"
+  echo "`date`: Setting auto-delete behavior '$BUILD_INSTANCE' disk to false"
   gcloud compute instances set-disk-auto-delete $BUILD_INSTANCE \
     --project $BUILD_PROJECT \
     --account $ACCOUNT \
     --zone $ZONE \
     --no-auto-delete \
     --disk $BUILD_INSTANCE
 
+  sleep 20
```
lwander: I'm hesitant to poll because if it's a consistency problem it won't guarantee the flag is set. Also, given that the builds take ~15 minutes now this is a negligible increase in wait time.

ewiseblatt: It's 15 minutes to succeed, not necessarily fail.
Another way would be to create the instance with --no-boot-disk-auto-delete and have an internal flag as to whether or not it needs to be deleted.

lwander: If the script made it this far, then the long parts (boot the VMs & successfully run the install scripts) must have passed. 

There is already a lot of stateful behavior with global flags for both the target VM & the `trap` statements in that script making it hard to reason about, adding another, especially one that implemented incorrectly will leave artifacts around causing future builds to fail, feels like more trouble than it is worht.

ewiseblatt:
ok. lgtm.

PR #1691, opened 2017-05-26
feat(validation): Monitor validation

ewiseblatt:
@lwander 

I'm hacking in apt-get install python-dev as a dependency; I thought you added this earlier in the week but am not seeing it.

I'd like to have a way of setting --client_host (in the prometheus install) to expose the daemon to the network to be polled by a remote prometheus.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -79,7 +79,7 @@ class AzsStorageConfiguratorHelper(object):
         'AZS_PASSWORD=$(cat {file})'
         .format(file=os.path.basename(options.storage_azs_credentials)))
     hal = (
-        'hal --color=false config storage azs edit'
+        'hal -q --log=info config storage azs edit'
         ' --storage-account-name {name}'
         ' --storage-account-key "$AZS_PASSWORD"'
         .format(name=options.storage_azs_account_name))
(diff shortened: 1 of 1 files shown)


lwander:
I'd added it to Halyard's installation of spinnaker-monitoring - this is the right place to do it though

lwander:
I'd added it to Halyard's installation of spinnaker-monitoring - this is the right place to do it though

lwander:
I'd added it to Halyard's installation of spinnaker-monitoring, maybe it's not up-to-date?

dev/validate_bom__config.py
```diff
@@ -818,15 +819,33 @@ class MonitoringConfigurator(object):
...
+      raise ValueError('gateway is only applicable to '
+                       ' --monitoring_install_which="prometheus"')
 
   def add_config(self, options, script):
     """Implements interface."""
-    pass
+    if not options.monitoring_install_which:
+      return
+
+    script.append('sudo apt-get install python-dev -y')
```
lwander: Shouldn't be necessary, I'm doing a github release for Halyard 0.26.0 that should cover this.

ewiseblatt:
I ran this using master-latest-unvalidated last night with halyard-nightly
and saw that error until I added the pip install ahead of it.

PR #1693, opened 2017-05-26
Config network protocol used

ewiseblatt:
@lwander


ewiseblatt:
no. i've done nothing security oriented. we can talk about how if you wnat.

PR #1700, opened 2017-05-31
fix(dev): various fixes to build scripts

ewiseblatt:
There are two or three categories here but I've only tested them together so am submitting them together.

   SSH refactoring
      * change the ssh key generation so the keys are user specific
      * moved the non-standard GCE pub key changes from the file
        to the use site so the files are compliant with other systems
        if they are needed.
      * Dont use gcloud alpha ssh --internal-ip because I cannot
        get it to reliably connect
      * test ssh with retries to ensure the keys propagated to the
        machine before attempting to scp/ssh with them.

   Build/Validate related
      * Added --build_debug to pass --debug to gradlew
      * Specify explicit context when interacting with kubernetes

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -199,6 +199,9 @@ class Builder(object):
       target = 'buildDeb'
       extra_args = []
 
+    if self.__options.debug_gradle:
+      extra_args.append('--debug')
+
     if name == 'deck' and not 'CHROME_BIN' in os.environ:
       extra_args.append('-PskipTests')
     elif name == 'halyard':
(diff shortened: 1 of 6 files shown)


ewiseblatt:
@jtk54

dev/build_release.py
```diff
@@ -199,6 +199,9 @@ def start_deb_build(self, name):
       target = 'buildDeb'
       extra_args = []
 
+    if self.__options.debug_gradle:
+      extra_args.append('--debug')
```
jtk54: I'm not sure how `--debug` interacts with `--stacktrace`, which is already passed. Just noting, we'll see how it goes.

PR #1715, opened 2017-06-05
feat(bom): Adds artifact sources so BOM is self-describing.

jtk54:


jtk54:
Added a `githubOrganization` that defaults to the spinnaker org.

dev/generate_bom.py
```diff
@@ -108,14 +117,18 @@ def init_argument_parser(cls, parser):
...
                         help="Type of builder to use. Currently, the supported options are {'gcb', 'docker'}.")
     parser.add_argument('--docker_registry', default='',
                         help="Docker registry to push the container images to.")
-    parser.add_argument('--changelog_output', default='',
-                        help="Output file to write the changelog to.")
-    parser.add_argument('--bom_alias', default='',
-                        help="Alias to rename the 'real' BOM as. This also sets the Spinnaker version as the alias.")
+    parser.add_argument('--google_image_project', default='marketplace-spinnaker-release',
+                        help="GCE project we publish HA Spinnaker component images to.")
+    parser.add_argument('--github_org', default='https://github.com/spinnaker',
```
lwander: Why only github?

jtk54: I think that's the only thing that will get use immediately (i.e. forks). We can add support later if users want to use something else.

PR #1721, opened 2017-06-06
feature(experimental/docker-compose): Add Docker Compose V2 Manifests

ethanfrogers:
Using Docker Compose Version 2 eliminates the need to use `$DOCKER_IP`
in the `docker-compose.yaml` file to set links to other services. An
entry in `/etc/hosts` is created for every container in the stack
meaning that we can just referenece the service by name!

I also took the opportunity to move Front 50 over to S3 using Minio
since Cassandra is on it's way out. Using version 2, we can get a full
Spinnaker stack using 1 command `docker-compose up -d` (without any
Cloud Provider configuration of course).

I added these in the `docker-compose/version-2` directory because
existing deployments of the Compose environments would cease to work.
This _should_ help with migration from V1 to V2.

@tomaslin I believe you responded to my post on Slack about this about a month ago.


tomaslin:
Very nice, we should just replace the original v1 files with these instead of having them in its own directory. LGTM

ethanfrogers:
@tomaslin I'm not against it. I'll add something in the README that existing deployments with V1 manifests will not work! Thanks for taking a look. Will ping when updated.

enkoder:
I fired up the docker compose stack and it worked well for me. I didn't get too deep in terms of config but its just what I was looking for because I am not on a Mac so I didn't want/need to use the DOCKER_IP envar.

experimental/docker-compose/README.md
```diff
@@ -46,7 +45,7 @@ Sometimes it is useful to use the docker compose configuration to test changes t
 * Open Kitematic and click on **'DOCKER CLI'** at the bottom.
 * Create a directory called build as a sibling to the spinnaker directory.
 * Run `../spinnaker/dev/refresh_source.sh --pull_origin --use_ssh --github_user default` to download Spinnaker source files.
-* Navigate back to the `docker-compose` directory.
+* Navigate back to the `docker-compose/version-2` directory.
```
tomaslin: should just be docker-compose

experimental/docker-compose/README.md
```diff
@@ -1,19 +1,18 @@
...
+This is an experimental integration of Spinnaker and Docker Compose using docker images for Spinnaker published on Quay.io. It requires no installation of any software aside from the Docker For Mac to test out Spinnaker.
 
-*Note: These instructions have been tested on Mac OS 10.11.1 and Docker Toolbox 1.9.0b, but might require slight tweaking if you are using a different platform.*
+*Note: These instructions have been tested on Mac OS 10.12.5 and Docker for Mac 17.03.1-ce-mac12, and might require slight tweaking if you are using a different platform.*
 
 ## Limitations
 
 * The docker compose project is intended as a 'try Spinnaker now' tool. It's not intended for production use.
-* Since we use an in-memory Cassandra and an in-memory Redis, changes to applications and pipelines will not be persisted across restarts.
+* Since we use an in-memory S3 and an in-memory Redis, changes to applications and pipelines will not be persisted across restarts.
```
tomaslin: would say in-memory S3 ( via Minio )

experimental/docker-compose/README.md
ethanfrogers: Good catch, thanks!

ethanfrogers:
@tomaslin updated with feedback and squashed.

ethanfrogers:
@tomaslin This is ready to go! Thanks for the review!

tomaslin:
Doesn't seem to work when trying simple things like creating an application. Seeing errors like `2017-06-12 18:49:38.560 DEBUG 1 --- [0.0-8084-exec-1] o.s.web.servlet.DispatcherServlet        : Could not complete request

retrofit.RetrofitError: Failed to connect to localhost/0:0:0:0:0:0:0:1:8083
` which would indicate connectivity errors.

ethanfrogers:
Crap, I could have sworn I fixed those. I'll circle back. Thanks!

ethanfrogers:
What are the odds that I left out the `orca` section in `spinnaker-local.yaml` when I updated the `host` and `baseUrl` of all the other services? @tomaslin we should be good to go now. I tore down my entire environment and brought it back up and could create applications.

tomaslin:
Now getting `      "error" : "Failed to connect to localhost/0:0:0:0:0:0:0:1:8080", ` when trying to create application. Seems like maybe front50 config is not right

ethanfrogers:
That's so odd. `curl localhost:8083/resolvedEnv`  shows `services.front50.baseUrl = http://front50:8080` on my environment after a full `docker-compose down`.

```
#config/spinnaker-local.yml
...
front50:
  host: front50
  port: 8080
  baseUrl: ${services.default.protocol}://${services.front50.host}:${services.front50.port}
...
```

and `config/orca.yml` only defaults to `localhost:8080` is that value isn't set.

PR #1731, opened 2017-06-08
fix(build): Write Halyard version to test into a global file.

jtk54:


dev/build_prevalidation.py
```diff
@@ -57,18 +57,24 @@ def __record_halyard_nightly_version(version_bump, options):
...
   # Now sync the local dir with the bucket again after the update.
   check_run_quick('gsutil rsync -r -d {local_bucket} {remote_uri}'
                   .format(remote_uri=bucket_uri, local_bucket=local_bucket_name))
 
+  with open(options.hal_version_under_test_file_path, 'w') as hal_version_file:
+    hal_version_file.write('{}'.format(full_hal_version))
+
 def init_argument_parser(parser):
   parser.add_argument('--hal_nightly_bucket_uri', default='',
                       help='The URI of the bucket to record the version and commit at which we built Halyard.')
```
ewiseblatt: ok, but could you give this a better name? It isnt the version under test, it is the version that was built.

dev/build_prevalidation.py
```diff
@@ -57,18 +57,24 @@ def __record_halyard_nightly_version(version_bump, options):
...
+  new_hal_nightly_entry = ('{full_hal_version}: {commit}'
+                           .format(full_hal_version=full_hal_version, commit=version_bump.commit_hash))
   nightly_entry_file = '{0}/nightly-version-commits.yml'.format(local_bucket_name)
   with open(nightly_entry_file, 'a') as nef:
     nef.write('{0}\n'.format(new_hal_nightly_entry))
   # Now sync the local dir with the bucket again after the update.
   check_run_quick('gsutil rsync -r -d {local_bucket} {remote_uri}'
                   .format(remote_uri=bucket_uri, local_bucket=local_bucket_name))
 
+  with open(options.hal_version_under_test_file_path, 'w') as hal_version_file:
```
ewiseblatt: check for it being non-empty. by default it does not write.

jtk54: 'w' should create a new [file each time](https://docs.python.org/2/tutorial/inputoutput.html#reading-and-writing-files).

dev/build_prevalidation.py
jtk54: Yep, `hal_build_version_file_path`?

ewiseblatt: what about --output_built_halyard_version=<path>?

jtk54: Sounds good.

PR #1733, opened 2017-06-08
fix(build): Revert change to build numbers.

jtk54:
@spinnaker/google-reviewers Reverting change to ISO instants. Our builds are broken until this is merged. We need to rethink what we want this to be, since apparently the Nebula Gods require more sacrifice before they are placated.

TLDR: The date-underscore format is not accepted by Nebula, which then defaults to 0.0.0 for all components.

### DIFF
--- a/dev/annotate_source.py
+++ b/dev/annotate_source.py
@@ -15,7 +15,6 @@
 # limitations under the License.
 
 import argparse
-import datetime
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


ewiseblatt:
Another GCB error, but different reason than the others.
http://localhost:4040/job/Build_BomFromGithub/20/artifact/logs/gcb/gate-gcb-build.log

PR #1735, opened 2017-06-09
fix(validate): Install halyard with --user

ewiseblatt:
@lwander


dev/validate_bom__deploy.py
```diff
@@ -540,7 +542,7 @@ def do_deploy(self, script, files_to_upload):
           ' -o StrictHostKeyChecking=no'
           ' -o UserKnownHostsFile=/dev/null'
           ' {user}@{ip}'
-          ' "sudo ./{script_name}"'
```
lwander: You probably still need sudo - unless the user running this is root.

ewiseblatt: I still sudo inside the script in places

lwander: SGTM

PR #1744, opened 2017-06-12
feat(udf): Adding citests for userdata functionality.

dbakshiyev:


testing/citest/tests/google_server_group_test.py
```diff
@@ -161,6 +161,7 @@ def create_server_group(self):
       'instanceMetadata': {
         'load-balancer-names': self.__lb_name
       },
+      'userData':'customUserData=testingUserData',
```
ewiseblatt: space after colon

testing/citest/tests/google_server_group_test.py
```diff
@@ -266,6 +275,13 @@ def clone_server_group(self):
      .list_resource(self.__mig_manager_name, **self.__mig_manager_kwargs)
      .contains_path_value('baseInstanceName', self.__cloned_server_group_name))
 
+    (builder.new_clause_builder('Instance Template Created', retryable_for_secs=150)
```
ewiseblatt: Is this clause something like "Instance template preserved"?
If the value here is tied to above, then use a variable or constant to capture it rather than duplicating it.

testing/citest/tests/google_server_group_test.py
dbakshiyev: done

testing/citest/tests/google_server_group_test.py
dbakshiyev: Done

testing/citest/tests/google_server_group_test.py
```diff
@@ -377,9 +396,9 @@ def delete_load_balancer(self):
     job = [{
       "loadBalancerName": self.__lb_name,
       "networkLoadBalancerName": self.__lb_name,
-      "region": self.TEST_REGION,
+      "region": "us-central1",
```
ewiseblatt: These should be self.TEST_REGION. Was there a reason you changed it?

testing/citest/tests/google_server_group_test.py
```diff
@@ -161,6 +164,7 @@ def create_server_group(self):
       'instanceMetadata': {
         'load-balancer-names': self.__lb_name
       },
+      'userData': 'customUserData=' + self.__custom_user_data,
```
ewiseblatt: this doesnt look right. If it is then 'customUserData' should be parameterized as well. I didnt realize this was. (__test_custom_user_data_key, __test_custom_user_data_value)

dbakshiyev: added a custom_user_data_key parameter

testing/citest/tests/google_server_group_test.py
dbakshiyev: Don't remember the reason for changing this--I changed it back.

testing/citest/tests/google_server_group_test.py
```diff
@@ -161,6 +165,7 @@ def create_server_group(self):
       'instanceMetadata': {
         'load-balancer-names': self.__lb_name
       },
+      'userData': '='.join([self.__custom_user_data_key, self.__custom_user_data_value]),
```
ewiseblatt: Though this produces the right output, I dont think it makes sense semantically because I'd read that as '=' over a list, but really it only makes sense on a 2-tuple (which it happens to be).  Instead I'd be more explicit with '{key}={value}'.format(key=..., value=...) (you dont necessarily need to name the parameters, {0} and {1} are fine)

FWIW, If it did need to be be more general ','.join(['{key}={value}'.format(key=key, value=value) for key,value in tupleList]) where tupleList is something like [('keyA', 'valueA'), ('keyB', 'valueB')]

dbakshiyev: Done

PR #1745, opened 2017-06-12
feat(bom): Adds utility to reconstruct source from BOM.

jtk54:


dev/reconstruct_source.py
```diff
@@ -0,0 +1,96 @@
...
+  'deck',
+  'echo',
+  'front50',
+  'gate',
+  'igor',
+  'orca',
+  'rosco',
+  'fiat',
+  'spinnaker-monitoring',
+  'spinnaker'
```
lwander: This doesn't have a bom entry, right?

jtk54: spinnaker does have a BOM entry, but we don't build a package from the source. It's just to track the tools we used to build it.

PR #1753, opened 2017-06-13
feat(build): Build and push rpms

gdanko:

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -23,8 +23,8 @@ The gradle script does not yet coordinate a complete build, so
 this script fills that gap for the time being. It triggers all
 the subsystem builds and then publishes the resulting artifacts.
 
-Publishing is typically to bintray for debian packages and a docker repository
-for containers.
+Publishing is typically to bintray for debian and redhat packages and
+a docker repository for containers.
 
 Usage:
   export BINTRAY_USER=
@@ -45,6 +45,7 @@ import argparse
 import base64
 import collections
 import datetime
+import fnmatch
 import glob
 import os
 import multiprocessing
@@ -62,10 +63,11 @@ import refresh_source
 from spinnaker.run import check_run_quick
 from spinnaker.run import run_quick
 
-
 SUBSYSTEM_LIST = ['clouddriver', 'orca', 'front50',
                   'echo', 'rosco', 'gate', 'igor', 'fiat', 'deck', 'spinnaker']
 ADDITIONAL_SUBSYSTEMS = ['spinnaker-monitoring', 'halyard']
+
+VALID_PLATFORMS = ['debian', 'redhat']
 
 
 class BackgroundProcess(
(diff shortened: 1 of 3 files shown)


gdanko:
@jtk54 @ewiseblatt

dev/build_release.py
```diff
@@ -231,6 +250,83 @@ def start_deb_build(self, name):
...
+    bintray_user = os.environ['BINTRAY_USER']
+
+    if self.__options.nebula:
+      target = 'buildRpm'
+      extra_args = [
+          '--stacktrace',
+          '-Prelease.useLastTag=true',
+          '-PbintrayPackageBuildNumber={number}'.format(
+              number=self.__build_number),
+          '-PbintrayOrg="{org}"'.format(org=org),
```
jtk54: I think the `-Pbintray*` properties aren't required here if we're not doing the upload via gradle.

dev/generate_bom.py
```diff
@@ -140,7 +140,7 @@ def __version_from_tag(self, comp):
     Returns:
       [string] Component version with build number and without 'version-'.
     """
-    version_bump = self.__component_versions[comp]
+    version_bump = dict(self.__component_versions.items() + self.__halyard_version.items())[comp]
```
ewiseblatt: @jtk54 is halyard versioning handled independent of spinnaker?

dev/build_release.py
jtk54: If we are managing the Bintray package constructs (not necessarily doing the RPM uploads) then this would be required.

dev/generate_bom.py
```diff
@@ -200,6 +200,19 @@ def write_docker_version_files(self):
       with open(config_file, 'w') as cfg:
         cfg.write(docker_tag)
 
+  def write_rpm_version_files(self):
+    """Write a file containing the version number for each microservice. Used
+    by build_release.py for rpm builds.
+    """
+    self.determine_and_tag_halyard()
+    for comp in dict(self.__component_versions.items() + self.__halyard_version.items()):
```
ewiseblatt: If you are just iterating over the keys, why build up the dictionaries as opposed to just the union of their keys?

dev/build_release.py
```diff
@@ -516,23 +665,35 @@ def build_packages(self):
...
+    elif self.__options.platform == 'redhat':
+      pids = self.start_copy_redhat_target(subsys)
 
     for p in pids:
       p.check_wait()
     print 'Finished copying {0}.'.format(subsys)
 
   @classmethod
-  def init_argument_parser(cls, parser):
+  def init_argument_parser(cls, parser, valid_platforms):
```
jtk54: I'd just refer to the `VALID_PLATFORMS` constant than pass those in as a parameter in this function.

dev/generate_bom.py
```diff
@@ -200,6 +200,19 @@ def write_docker_version_files(self):
...
+      if comp == 'spinnaker':
```
ewiseblatt: instead of duplicating this check, could you refactor it to something like
  if comp in EXCLUDE_VERSIONS_FOR_COMPONENTS
and define EXCLUDE_VERSIONS_FOR_COMPONENTS to ['spinnaker']?
(and update existing checks for this as well)

jtk54:
Requires https://github.com/spinnaker/spinnaker-gradle-project/pull/31 merged and each component's `build.gradle` updated to implement in the release process.

dev/generate_bom.py
jtk54: Yes.

jtk54:
@gdanko Looks like the last update is missing a bracket, can you fix that up and run through the RPM publish once more?

dev/build_release.py
```diff
@@ -516,23 +665,35 @@ def build_packages(self):
...
       refresh_source.Refresher.init_argument_parser(parser)
+      parser.add_argument('--platform', default='debian', action='store',
```
ewiseblatt: should this be a list so all the builds can be concurrent off the same process?

dev/build_release.py
```diff
@@ -633,9 +799,8 @@ def do_build(cls, options, build_number=None, container_builder=None):
   @classmethod
   def main(cls):
     parser = argparse.ArgumentParser()
-    cls.init_argument_parser(parser)
+    cls.init_argument_parser(parser, VALID_PLATFORMS)
```
ewiseblatt: I'd leave this global rather than adding it as an argument.
You are already using it global in do_build anyway.

dev/build_release.py
```diff
@@ -231,6 +250,83 @@ def start_deb_build(self, name):
...
+    assume that the source has been properly tagged to use the latest tag as the
+    package version for each component.
+
+    Args:
+      name [string]: Name of the subsystem repository.
+
+    Returns:
+      BackgroundProcess
+    """
+    jarRepo = self.__options.jar_repo
```
ewiseblatt: I didnt check but this looks like it is pretty similar to the debian build up front with a few differences at the end. If so, is it easy to consolidate them so they share the same logic for the nontrival common stuff? If it makes it more complicated then dont bother, but this duplication looks awkward to maintain.

dev/build_release.py
gdanko: This was not working at some time. Probably because of all my hacks in the previous script. Works now.. Done.

dev/build_release.py
jtk54: I think I'll take care of the concurrency and refactor if necessary when we integrate this into the release process.

jtk54:
Merging, thanks for your contribution @gdanko.

dev/build_release.py
gdanko: @ewiseblatt I thought about that up front but to make sure it worked absolutely I split them. If you like, I could work on consolidation unless @jtk54 wants to own it.

dev/build_release.py
gdanko: I think @jtk54 will be taking this one.

dev/build_release.py
gdanko: Done

PR #1762, opened 2017-06-16
fix(install): Supply --user to halyard install script

lwander:


lwander:
Assorted fixes from testing against C2D as well

PR #1769, opened 2017-06-19
fix(validate): Various validation fixes and enhancements

ewiseblatt:
* Add KeepAlive probe to appease kubectl port forwarding
 * Determine Azure IP address
 * Retry resolvedEnv scrape

 * Comment out Front50 None attributes to make S3 test happy.
 * Split kube namespaces and only use the first

@jtk54

### DIFF
--- a/dev/validate_bom__test.py
+++ b/dev/validate_bom__test.py
@@ -429,10 +429,27 @@ class ValidateBomTestController(object):
 
     # Redirect stdout to prevent buffer overflows (at least in k8s)
     # but keep errors for failures.
+    class KeepAlive(threading.Thread):
+      def run(self):
+        while True:
+          try:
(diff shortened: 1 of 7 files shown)


dev/validate_bom__test.py
```diff
@@ -500,22 +517,28 @@ def wait_on_service(self, service_name, port=None, timeout=120):
     # I've only seen this happen once.
     time.sleep(1)
 
+    threadid = hex(threading.current_thread().ident)
```
jtk54: Why do you need to `hex` this? Seems fine, just curious.

ewiseblatt: the id's are very large. Or at least mine are. hex provides a larger alphabet to make them differ from each other a little more noticably on inspection.

PR #1786, opened 2017-06-22
fix(c2d): Ownership of hal files should be ubuntu

lwander:


lwander:
Yup, during validation of the flow

PR #1788, opened 2017-06-22
perf(build): Enable using gradle cache on builds.

jtk54:
This is for building debian/rpm packages locally for now. Will add a PR that uses the gradle cache for the container builds soon.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -233,6 +233,9 @@ class Builder(object):
     if self.__options.debug_gradle:
       extra_args.append('--debug')
 
+    if self.__options.gradle_cache_path:
+      extra_args.append('--gradle_user_home={}'.format(self.__options.gradle_cache_path))
+
     if name == 'deck' and not 'CHROME_BIN' in os.environ:
       extra_args.append('-PskipTests')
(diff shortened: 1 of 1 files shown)


jtk54:
Yes, this is a result of that conversation -- it turns out we weren't referencing the cache during our builds. This will add the ability to use gradle's read-through caching in our package builds (not containers yet, coming soon).

dev/build_release.py
```diff
@@ -740,6 +746,9 @@ def init_argument_parser(cls, parser):
       parser.add_argument(
           '--gcb_service_account', default='',
           help='Google service account to invoke the gcp container builder with.')
+      parser.add_argument(
+          '--gradle_cache_path', default='',
+          help='Path to a gradle cache directory to use for the builds.')
```
ewiseblatt: Why would this not default to $HOME/.gradle or something within?
The cache is global not to the process.

jtk54: Good call, the behavior of the `--gradle-user-home` is to just create a cache if that directory doesn't exist. I'll default to `$HOME/.gradle`.

PR #1789, opened 2017-06-22
fix(c2d): Fix multi-provider deploy permission clobbering

lwander:


install/first_halyard_boot.sh
```diff
@@ -61,6 +61,8 @@ set -u
 
 METADATA_URL="http://metadata.google.internal/computeMetadata/v1"
 INSTANCE_METADATA_URL="$METADATA_URL/instance"
+HALYARD_USER=ubuntu
```
ewiseblatt: is there a relationship between this an /opt/spinnaker/config/halyard.user?
if so, should you either get this from there and if not should you update that?

LGTM otherwise.

lwander: Oh good idea

PR #1790, opened 2017-06-23
fix(validate): Removed a source of flakiness from aws smoke test.

ewiseblatt:
@jtk54

### DIFF
--- a/testing/citest/tests/aws_smoke_test.py
+++ b/testing/citest/tests/aws_smoke_test.py
@@ -62,6 +62,8 @@ import citest.base
 import spinnaker_testing as sk
 import spinnaker_testing.gate as gate
 
+from botocore.exceptions import (BotoCoreError, ClientError)
+
 
 class AwsSmokeTestScenario(sk.SpinnakerTestScenario):
   """Defines the scenario for the smoke test.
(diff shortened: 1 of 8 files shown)


testing/citest/tests/aws_smoke_test.py
```diff
@@ -245,17 +247,20 @@ def upsert_load_balancer(self, use_vpc):
...
-         {'HealthCheck': jp.DICT_MATCHES({
-             key: jp.EQUIVALENT(value) for key, value in health_check.items()}),
-          'AvailabilityZones': jp.LIST_SIMILAR(avail_zones),
-          'ListenerDescriptions/Listener':
-              jp.DICT_MATCHES(
-                  {key: jp.NUM_EQ(value)
-                   for key, value in listener['Listener'].items()})
-         })
-    )
+     .expect_value_list_contains(
```
jtk54: This is OK for now, but seems like we can provide a cleaner helper function when we're trying to match against a single payload object. I think the end goal for these should be to have them be expressive and clean to read and understand what's being validated.

ewiseblatt: Revisiting...

PR #1791, opened 2017-06-23
chore(validate): more monitoring, less logging

ewiseblatt:
Added/fixed monitoring into validation
Removed some logging chatter
@jtk54

### DIFF
--- a/dev/all_tests.yaml
+++ b/dev/all_tests.yaml
@@ -44,6 +44,21 @@ tests:
     requires:
       configuration:
         spinnaker_storage: gcs
+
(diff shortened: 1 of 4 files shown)


dev/all_tests.yaml
```diff
@@ -44,6 +44,20 @@ tests:
     requires:
       configuration:
         spinnaker_storage: gcs
+
+        # Localdebian because for some reason kubectl doesnt work
```
jtk54: Make this a TODO or NOTE since it's still an open question.

PR #1794, opened 2017-06-23
chore(c2d): Cleanup hal logs

lwander:


lwander:
I created that variable thinking you might like it

PR #1795, opened 2017-06-23
feat(launcher): enables necessary apis on halyard boot

skim1420:
@lwander @ewiseblatt @jtk54

### DIFF
--- a/install/first_halyard_boot.sh
+++ b/install/first_halyard_boot.sh
@@ -94,7 +94,6 @@ function replace_startup_script() {
 }
 
 function enable_apis() {
-  gcloud service-management enable compute.googleapis.com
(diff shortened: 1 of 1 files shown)


install/first_halyard_boot.sh
```diff
@@ -93,6 +93,21 @@ function replace_startup_script() {
   clear_instance_metadata "startup-script"
 }
 
+function enable_apis() {
```
ewiseblatt: what scopes are required for this call?

skim1420: Scopes are handled in the DM jinja file, depending on which providers are enabled. They are in alignment with the APIs that are enabled in this function.

ewiseblatt:
I need to know the scopes for automated testing outside DM.

PR #1797, opened 2017-06-26
Google tests

ewiseblatt:
@jtk54 

There's some extra reformatting in appengine test
but mostly this complements the PR to citest converting the google tests over to the new verifier API

### DIFF
--- a/testing/citest/tests/google_http_lb_upsert_scenario.py
+++ b/testing/citest/tests/google_http_lb_upsert_scenario.py
@@ -23,10 +23,13 @@ import time
 import citest.gcp_testing as gcp
 import citest.json_predicate as jp
 import citest.service_testing as st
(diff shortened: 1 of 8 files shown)


testing/citest/tests/google_http_lb_upsert_scenario.py
```diff
@@ -337,17 +345,20 @@ def delete_http_load_balancer(self):
     (contract_builder.new_clause_builder('Health Check Removed',
                                          retryable_for_secs=30)
      .list_resource('httpHealthChecks')
-     .excludes_path_value('name', self.__proto_hc['name'])
+     .EXPECT(ov_factory.value_list_path_excludes(
```
jtk54: I like these -- seems like with this structure we can iterate on the expressiveness.

PR #1804, opened 2017-06-26
docs(boot): documents necessary scopes

skim1420:


ewiseblatt:
ping

skim1420:
This is obsolete. We've given it cloud-platform scope for all cases, as required by `gcloud service-management enable..`

Issue #190, opened 2014-11-06
Invalid applications in application list


cfieber:
Seeing applications that were deleted

When clicked shows generic not found page

PR #1807, opened 2017-06-28
fix(generate_bom): Adds container builder base image as param.

jtk54:


dev/generate_bom.py
```diff
@@ -49,12 +49,13 @@
 GOOGLE_CONTAINER_BUILDER_SERVICE_BASE_CONFIG = {
   'steps': [
     {
-      'name': 'spinnakerrelease/gradle_cache',
-      'env': ['GRADLE_USER_HOME=/gradle_cache/.gradle'],
+      'name': '',
```
ewiseblatt: This seems like an awkward definition now.
It might be cleaner to instead make this a factory function that returns the configuration dictionary and takes an options parameter. Then instead of fixing it up later (~#190), call the factory function which outright constructs the dictionary with the values you are poking into it below. I'd consider the same for the other config too for consistency even if it isnt as parameterized.

dev/generate_bom.py
```diff
@@ -126,6 +129,10 @@ def init_argument_parser(cls, parser):
                         help="Output file to write the changelog to.")
     parser.add_argument('--container_builder', default='gcb',
                         help="Type of builder to use. Currently, the supported options are {'gcb', 'docker'}.")
+    parser.add_argument('--container_builder_base_image', default='spinnakerrelease/gradle_cache',
+                        help="Base image to start from in the container builds.")
+    parser.add_argument('--container_builder_env_vars', default='GRADLE_USER_HOME=/gradle_cache/.gradle',
+                        help="Comma-separated list of environment variables to set in the container builds.")
```
ewiseblatt: Is this the right default value?

jtk54: Yes https://github.com/spinnaker/spinnaker/blob/master/google/release/gradle_cache_base_image.docker#L7

jtk54:
Pushed a commit with a refactor of GCB config generation.

PR #1809, opened 2017-06-29
chore(tests): Drop the explicit executionEngine attribute on canned p…

duftler:
…ipelines.


duftler:
fyi @jtk54

jtk54:
Cool, this will get us some feedback in the integration tests.

PR #1827, opened 2017-07-10
chore(build): Add '--info' logging to builds.

jtk54:


jtk54:
@ewiseblatt FYI -- adds `--info` to the Debian builds.

ewiseblatt:
Thanks. I thought this would have been parameterized since ideally we
should not need it long term, but would want it available under times of
duress.

PR #1830, opened 2017-07-12
feat(build_release): Add GCB container builds via triggers.

jtk54:
Adds GCB triggered builds. Won't be back until Monday, but drop me some comments/concerns and I'll fix 'em up next week and then we can turn this stuff on for our builds.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -47,20 +47,24 @@ import collections
 import datetime
 import fnmatch
 import glob
+import json
 import os
 import multiprocessing
 import multiprocessing.pool
 import re
+import shlex
 import shutil
 import subprocess
 import sys
 import tempfile
+import threading
 import time
 import urllib2
 from urllib2 import HTTPError
 
 import refresh_source
-from spinnaker.run import check_run_quick
+
+from google.cloud import pubsub
 from spinnaker.run import run_quick
 
 SUBSYSTEM_LIST = ['clouddriver', 'orca', 'front50',
@@ -69,45 +73,8 @@ ADDITIONAL_SUBSYSTEMS = ['spinnaker-monitoring', 'halyard']
 
 VALID_PLATFORMS = ['debian', 'redhat']
 
+GCB_BUILD_STATUS_TIMEOUT = 1200
 
-class BackgroundProcess(
-    collections.namedtuple('BackgroundProcess', ['name', 'subprocess', 'log'])):
-  """Denotes a running background process.
-
-  Attributes:
-    name [string]: The visible name of the process for reporting.
-    subprocess [subprocess]: The subprocess instance.
-  """
-
-  @staticmethod
-  def spawn(name, args, logfile=None):
-    sp = None
-    log = None
-    if logfile:
-      log = open(logfile, 'w', 0) # Unbuffered
-      sp = subprocess.Popen(args, shell=True, close_fds=True,
-                            stdout=log, stderr=log)
-    else:
-      sp = subprocess.Popen(args, shell=True, close_fds=True,
-                            stdout=sys.stdout, stderr=subprocess.STDOUT)
-    return BackgroundProcess(name, sp, log)
-
-  def wait(self):
-    if not self.subprocess:
-      return None
-    return self.subprocess.wait()
-
-  def check_wait(self):
-    return_code = self.wait()
-    if return_code:
-      error = '{name} failed.'.format(name=self.name)
-      raise SystemError(error)
-
-    if not return_code is None and self.log:
-      self.log.close()
-
-
-NO_PROCESS = BackgroundProcess('nop', None, None)
 
 def determine_project_root():
   return os.path.abspath(os.path.dirname(__file__) + '/..')
@@ -146,6 +113,21 @@ def determine_package_version(platform, gradle_root):
     if re.match('-$', version): version = version + '0'
     return version
 
+def run_shell_and_log(cmd_list, logfile, cwd=None):
+  for cmd in cmd_list:
+    parsed = shlex.split(cmd)
+    log = None
+    if not os.path.exists(logfile):
+      log = open(logfile, 'w')
+    else:
+      log = open(logfile, 'a')
+    log.write('Executing command: {}\n---\n'.format(cmd))
+    subprocess.check_call(parsed, stdout=log, stderr=log, cwd=cwd)
+    log.write('\n---\nFinished executing command: {}'.format(cmd))
+    if log:
+      log.close()
+
+
 class Builder(object):
   """Knows how to coordinate a Spinnaker release."""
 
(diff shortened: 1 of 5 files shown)


dev/build_release.py
```diff
@@ -200,59 +179,57 @@ def start_deb_build(self, name):
...
-    return BackgroundProcess.spawn(
-      'Building and publishing Debian for {name}...'.format(name=name),
-      'cd "{gradle_root}"; ./gradlew {extra} {target}'.format(
-          gradle_root=gradle_root, extra=' '.join(extra_args), target=target),
-      logfile='{name}-debian-build.log'.format(name=name)
-    )
+    cmds = [
+      './gradlew {extra} {target}'.format(extra=' '.join(extra_args), target=target)
+    ]
+    logfile = '{name}-debian-build.log'.format(name=name)
```
ewiseblatt: do you want to delete the existing one since you append? Dont rely on the caller/jenkins to do that.

dev/build_release.py
```diff
@@ -280,54 +257,51 @@ def start_rpm_build(self, name):
 
     Args:
       name [string]: Name of the subsystem repository.
-
-    Returns:
-      BackgroundProcess
     """
-    jarRepo = self.__options.jar_repo
-    parts = self.__options.bintray_repo.split('/')
+    print 'Building and publishing Redhat for {name}...'.format(name=name),
```
ewiseblatt: print

dev/build_release.py
```diff
@@ -339,42 +313,118 @@ def start_container_build(self, name):
...
-          'Build/publishing container image for {name} with'
-          ' Google Container Builder...'.format(name=name),
-          'cd "{gradle_root}"'
-          '; gcloud container builds submit --account={account} --project={project} --config="../{name}-gcb.yml" .'
-        .format(gradle_root=gradle_root, name=name, account=self.__gcb_service_account,
-                project=self.__options.gcb_project),
-        logfile='{name}-gcb-build.log'.format(name=name)
-      )
+      print 'Build/publishing container image for {name} with GCB...'.format(name=name)
+      self.__gcb_build(name, gradle_root, self.__options.gcb_service_account, self.__options.gcb_project)
```
ewiseblatt: print here and others below

dev/build_release.py
```diff
@@ -339,42 +313,118 @@ def start_container_build(self, name):
...
+    cmds = [
+      ('gcloud container builds submit --account={account} --project={project} --config"../{name}-gcb.yml"'
+       .format(name=name, account=gcb_service_account, project=gcb_project))
+    ]
+    logfile = '{name}-gcb-build.log'.format(name=name)
+    run_shell_and_log(cmds, logfile, cwd=gradle_root)
+
+  @classmethod
+  def __gcb_trigger_build(cls, name, gradle_root, gcb_service_account, gcb_service_account_json, gcb_project, mirror_base_url):
+    tag = run_quick('cat {name}-gcb-trigger.yml'.format(name=name), echo=False).stdout.strip()
```
ewiseblatt: how big is tag? run_quick is subject to deadlock if it produces more output than the buffer.

dev/build_release.py
```diff
@@ -339,42 +313,118 @@ def start_container_build(self, name):
...
+    subscription = topic.subscription('cloud-builds-{name}'.format(name=name))
+    subscription.create()
+    if not subscription.exists():
+      raise LookupError('GCB pubsub subscription creation for subscription id {} failed.'.format(subscription.name))
+
+    def fail_build(name):
+      raise Exception('GCB triggered build for {} timed out'.format(name))
+
+    # Set an egg timer to fail.
+    timer = threading.Timer(1200, fail_build, (name))
```
ewiseblatt: where did 1200 come from?
This shoudl probably be a constant up top.

dev/build_release.py
```diff
@@ -339,42 +313,118 @@ def start_container_build(self, name):
...
+            subscription.acknowledge([ack_id])
+            status = payload['status']
+            print 'Received status: {} for building tag {} of {}'.format(status, tag_name, comp_name)
+            if status in ['SUCCESS', 'FAILURE']:
+              completed = True
+              build_id = payload['id']
+              print 'Retrieving logs for build_id: {}'.format(build_id)
+              get_log_cmd = ('gcloud container builds log --project {project} --account {account} {id}'
+                             .format(project=gcb_project, account=gcb_service_account, id=build_id))
+              build_log = run_quick(get_log_cmd, echo=False).stdout.strip()
```
ewiseblatt: Is this going to be too big for run_quick?

dev/build_release.py
```diff
@@ -339,42 +313,118 @@ def start_container_build(self, name):
...
+              with open(logfile, 'a') as log:
+                log.write(build_log)
```
ewiseblatt: When you append this log, do you need to indicate the command it came from?

dev/build_release.py
```diff
@@ -748,6 +798,12 @@ def init_argument_parser(cls, parser):
           '--gcb_service_account', default='',
           help='Google service account to invoke the gcp container builder with.')
       parser.add_argument(
+          '--gcb_service_account_json', default='',
+          help='Path to service account credentials to invoke the gcp container builder with.')
+      parser.add_argument(
```
ewiseblatt: I'm not a fan of an implicit default here at the moment.

dev/build_release.py
```diff
@@ -339,42 +313,118 @@ def start_container_build(self, name):
...
+    print 'Building and publishing container image via GCB for {name}...'.format(name=name)
+    cmds = [
+      ('gcloud container builds submit --account={account} --project={project} --config"../{name}-gcb.yml"'
+       .format(name=name, account=gcb_service_account, project=gcb_project))
+    ]
+    logfile = '{name}-gcb-build.log'.format(name=name)
+    run_shell_and_log(cmds, logfile, cwd=gradle_root)
+
+  @classmethod
+  def __gcb_trigger_build(cls, name, gradle_root, gcb_service_account, gcb_service_account_json, gcb_project, mirror_base_url):
```
ttomsu: This has a "long method" code smell - seems like you could break up the various functionality (establish git remote, pubsub setup, listen for pubsub messages) into separate methods.

dev/build_release.py
ttomsu: For the release process, this fork makes the most sense.

dev/build_release.py
```diff
@@ -146,6 +111,20 @@ def determine_package_version(platform, gradle_root):
     if re.match('-$', version): version = version + '0'
     return version
 
+def run_shell_and_log(cmd_list, logfile, cwd=None):
+  print 'Executing command list: {}'.format(cmd_list)
```
jtk54: My thought here was to provide some logging to the caller as well as in the individual log files, but I think there's enough information just in the log files to determine what happened. I'll add the command context and output in the log files only.

dev/build_release.py
jtk54: This tag is one line.

dev/build_release.py
jtk54: How big is the buffer? These are generally around 100 lines. I didn't notice any issues in my testing.

dev/build_release.py
jtk54: Refactored.

dev/build_release.py
jtk54: Makes sense.

jtk54:
Rebased and addressed comments, take another look.

jtk54:
Ping -- testing one more time and will merge and integrate into the nightly if there are no complaints.

ttomsu:
:+1:

PR #1835, opened 2017-07-14
fix(build): Dont buffer log output

ewiseblatt:
@ttomsu 
This is to ensure access to all the log information available in the event of a deadlock


ewiseblatt:
@lwander

PR #1840, opened 2017-07-17
fix(c2d): Fix bug when custom kubeconfig supplied

lwander:


lwander:
Correct param: 
param

lwander:
Merging & rebuilding the image since this is fairly urgent

PR #1853, opened 2017-07-21
feat(release): Publish Halyard JARS independently

lwander:


dev/build_prevalidation.py
```diff
@@ -92,6 +92,8 @@ def main():
 
   bom_generator = BomGenerator(options)
   bom_generator.determine_and_tag_versions()
+  bom_generator.determine_and_tag_halyard()
```
jtk54: Seems like adding this would cause problems since we already tag Halyard in https://github.com/lwander/spinnaker/blob/3dc2be0723766d69ce69a59820cc8009ad766d8a/dev/build_prevalidation.py#L91.

dev/build_release.py
```diff
@@ -250,6 +250,7 @@ def __debian_build(cls, name, options, build_number, gradle_root):
       os.remove(logfile)
     run_shell_and_log(cmds, logfile, cwd=gradle_root)
 
+
```
jtk54: Extra line.

dev/build_prevalidation.py
lwander: Last I tried it failed without to know halyard's version in the write_component_version w/o this line. But I might have had a conflicting change - will give it a shot

lwander:
Any concerns with merging @jtk54?

jtk54:
LGTM.

PR #1859, opened 2017-07-24
fix(release): Tag halyard once

lwander:
I forgot I had a staged change for testing on the jenkins dev vm - running this fix there now

### DIFF
--- a/dev/build_prevalidation.py
+++ b/dev/build_prevalidation.py
@@ -25,19 +25,6 @@ from refresh_source import Refresher
 
 from spinnaker.run import check_run_quick
 
-
-def __annotate_component(annotator, component):
-  """Annotate the component's source but don't include it in the BOM.
-
-  Returns:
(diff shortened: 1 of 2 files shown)


dev/build_prevalidation.py
```diff
@@ -87,12 +87,9 @@ def main():
   init_argument_parser(parser)
   options = parser.parse_args()
 
-  annotator = Annotator(options)
-  halyard_bump = __annotate_component(annotator, 'halyard')
-
   bom_generator = BomGenerator(options)
   bom_generator.determine_and_tag_versions()
-  # bom_generator.determine_and_tag_halyard()
+  halyard_bump = bom_generator.determine_and_tag_halyard()
```
jtk54: You can get rid of the `__annotate_component` function, I believe it's not used anywhere else.

lwander: Done

PR #1870, opened 2017-07-31
fix(dev): Separate stderr and stdout in run_quick

jtk54:
FYI @ewiseblatt didn't make any changes -- seems like a reasonable workaround to me.

### DIFF
--- a/pylib/spinnaker/run.py
+++ b/pylib/spinnaker/run.py
@@ -114,7 +114,7 @@ def run_and_monitor(command, echo=True, input=None,
   return RunResult(process.returncode, ''.join(out), ''.join(err))
 
 
-def run_quick(command, echo=True):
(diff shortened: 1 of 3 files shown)


pylib/spinnaker/run.py
```diff
@@ -161,12 +164,16 @@ def check_run_and_monitor(command, echo=True, input=None):
   return result
 
 
-def check_run_quick(command, echo=True):
+def check_run_quick(command, echo=True, dup_stderr_to_stdout=True):
```
lwander: Should the added parameter read `dump_stderr_to_stdout` instead of `dup_...`?

ewiseblatt:
I think Jake called it "dup" because it acts as a dup(2) on the stream.

PR #1877, opened 2017-08-03
feat(ha_images): Added HA image janitor.

jtk54:


jtk54:
On advice from @duftler, we should have a ramp-up period for this script since we're manipulating 'live' infrastructure customers may be using. I'll make two changes to implement this:

1. During a dry run, write out the `gcloud image delete` commands to a script so we can manually validate whether we're deleting the correct images.
2. Comment out/disable the actual `gcloud image delete` commands so we don't accidentally start deleting images if we forget a tag.

jtk54:
Updated the script to just output a shell script with several `gcloud compute image delete` commands, no dry run and no automatic image deletion.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def __resolve_images_in_bom(bom_version, contents_by_name):
+  bom_content_str = contents_by_name[bom_version]
+  bom_dict = yaml.load(bom_content_str)
+  service_entries = bom_dict['services']
+  return [__format_image_name(s, service_entries) for s in SERVICES]
+
+
+def __format_image_name(service_name, service_entries):
```
ewiseblatt: This could be local to __resolve_images_in_bom

duftler:
Would still like somebody more python-fluent to review this.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+  'echo',
+  'front50',
+  'gate',
+  'igor',
+  'orca',
+  'rosco',
+  'fiat'
+]
+
+RELEASED_TAG_KEY = 'released'
```
ewiseblatt: images are "published", not "released".
Could these tags have a value for the date they were published?
Eventually there could be a sunset or deprecated tag.
FWIW, GCE images use some standard tags which are used by pantheon for browsing so you dont see older images. We still want to keep them around but it might be worthwhile to use those tags too so that one could browse images in pantheon and see just the current.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+import json
+import os
+import re
+import sys
+import yaml
+
+from google.cloud import storage
+from spinnaker.run import check_run_quick, run_quick
+
+
```
jtk54: Added.

google/release/ha_image_janitor.py
jtk54: Switched nomenclature and added the value as the `creationTimestamp`.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+]
+
+RELEASED_TAG_KEY = 'released'
+
+
+def __partition_boms(gcs_client, bucket_name):
+  bucket = gcs_client.get_bucket(bucket_name)
+  # Get all BOM blobs but remove 'bom/' directory.
+  all_bom_blobs = [b for b in bucket.list_blobs(prefix='bom') if not b.name.endswith('/')]
+  bom_contents_by_name = {__bom_to_version(b): b.download_as_string() for b in all_bom_blobs}
```
jtk54: This is a feature of the storage API -- the `name` of a blob is actually a path relative to the root bucket, which is why we need to use `basename` here. I changed the filter to look for `.yml` extensions.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+
+  versions_to_tag = [__bom_to_version(bom) for bom in all_bom_blobs if __tagged_bom(bom)]
+  versions_to_delete = [__bom_to_version(bom) for bom in all_bom_blobs if not __tagged_bom(bom)]
+  return (versions_to_tag, versions_to_delete, bom_contents_by_name)
```
jtk54: Changed variable name.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+def __tagged_bom(bom_blob):
```
jtk54: Moved to local functions and changed the name.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+  name = os.path.basename(bom_blob.name)
+  return (RELEASED_VERSION_MATCHER.match(name)
+          or LATEST_UNVALIDATED_MATCHER.match(name))
+
+
+def __recent_image(image_json, days_before):
+  # HACK: Cut off the timezone because strptime can't deal with timezones.
```
jtk54: Refactored.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+  images_to_tag = set([])
+  for bom_version in versions_to_tag:
+    to_tag = [i for i in __resolve_images_in_bom(bom_version, bom_contents_by_name) if i in project_images]
+    images_to_tag.update(to_tag)
+  for image in images_to_tag:
+    result = run_quick('gcloud compute images describe --project={project} --account={account} --format=json {image}'
+                       .format(project=project, account=account, image=image), echo=False)
+    # Adding labels is idempotent, adding the same label again doesn't break anything.
+    if not result.returncode:
+      run_quick(
```
jtk54: Changed.

google/release/ha_image_janitor.py
```diff
@@ -0,0 +1,192 @@
...
+        if dry_run:
+          print('DRY_RUN: gcloud compute images delete --project={project} --account={account} {image}'
+                .format(project=project, account=account, image=image))
+        else:
+          check_run_quick('gcloud compute images delete --project={project} --account={account} {image}'
+                          .format(project=project, account=account, image=image))
+
+
+def __bom_to_version(bom_blob):
+  return os.path.basename(bom_blob.name).replace('.yml', '')
```
jtk54: Changed so the BOM blobs are filtered by having a `.yml` extension.

google/release/ha_image_janitor.py
jtk54: Moved.

jtk54:
Merging if no complaints.

PR #1899, opened 2017-08-14
feat(testing): add basic smoke tests for DC/OS

willgorman:
We still need to get a DC/OS cluster that can run these but I wanted to at least get this out there to see if there are any other changes we'd need to make.

Instructions for setting up the account in clouddriver, either with clouddriver-local.yml or halyard: https://github.com/spinnaker/spinnaker.github.io/pull/315

### DIFF
--- a/testing/citest/tests/dcos_smoke_test.py
+++ b/testing/citest/tests/dcos_smoke_test.py
@@ -1,0 +1,387 @@
+# Copyright 2017 Cerner Corporation All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Smoke test to see if Spinnaker can interoperate with DC/OS.
+
+See testable_service/integration_test.py and spinnaker_testing/spinnaker.py
+for more details.
+
+Sample Usage:
+  These tests expect the 'dcos' utility to be available and configured for the DC/OS cluster being tested against.
+
+  PYTHONPATH=$CITEST_ROOT:$CITEST_ROOT/spinnaker \
+    python $CITEST_ROOT/tests/dcos_smoke_test.py \
+    --native_hostname=$HOSTNAME \
+    --native_port=$PORT \
+    --spinnaker_dcos_account=$ACCOUNT \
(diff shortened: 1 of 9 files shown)


ewiseblatt:
looks good. Some comments. You can keep the no_results_ok as is, I'm just suggesting a new direction I'm trying to go in there. I'm open to debate/disagreement on it.

I'm trying to configure and environment to try this out...

ewiseblatt:
Sorry, I thought these comments were sent out earlier this week.

testing/citest/spinnaker_testing/dcoscli_agent.py
```diff
@@ -0,0 +1,35 @@
+# Standard python modules.
```
willgorman: We can definitely look into contributing that to citest.  For now I'll move it another package in this repo.  I'm not very experienced with Python though and I'm running into a problem.  Right now it looks like:

```
.
├── citest
│   ├── __init__.py
│   └── dcos_testing
│       ├── __init__.py
│       ├── dcos_contract.py
│       └── dcoscli_agent.py
├── spinnaker_testing
├── tests
│   ├── dcos_smoke_test.py
```

`dcos_smoke_test.py` has the following imports:
```python
import citest.json_contract as jc
import citest.json_predicate as jp
import citest.service_testing as st
import citest.dcos_testing as dcos
```

The package in this repo looks like it's  shadowing the citest library because it reports this error when run: `ImportError: No module named json_contract` Is there a way to prevent that from happening?

ewiseblatt: Looks like I gave you bad advice and packages are encapsulated. I guess that makes sense given the __init__.py and other metadata type files in a package directory. I'd still like to move it out. Maybe a spinnaker/testing/citest/citest_contrib directory and put dcos_testing/dcos_agent or whatever you called it under there? It would be imported as citest_contrib.dcos_testing because of the path, but eventually would like to move it into the google/citest repo if you dont mind. At least take it out of the spinnaker-specific packages.

ewiseblatt:
ping

testing/citest/spinnaker_testing/dcos_contract.py
```diff
@@ -0,0 +1,155 @@
...
+    Args:
+      no_resource_ok: Whether or not the resource is required.
+          If the resource is not required, "not found" is treated as a valid
+          check. Because resource deletion is asynchronous, there is no
+          explicit API here to confirm that a resource does not exist.
+    """
+    self.observer = self.__factory.new_get_marathon_resources(
+        type, action='list', extra_args=extra_args)
+
+    if no_resource_ok:
```
willgorman: There weren't any cases where we were actually setting `no_resource_ok` so I removed the parameter and left the rest of the test as is.

testing/citest/tests/dcos_smoke_test.py
```diff
@@ -0,0 +1,373 @@
+"""
```
ewiseblatt: copywrite

willgorman: Oops, missed that one.  It's added now.

PR #1907, opened 2017-08-18
feat(validate): Support to configure DC/OS accounts and run their cit…

ewiseblatt:
…ests.

@jtk54 
The actual test is introduced by a pending PR. This can be merged first since these changes require explicit configuration to enable.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -45,7 +45,27 @@ import os
 from validate_bom__deploy import write_data_to_secure_path
 
 
-class AzsStorageConfiguratorHelper(object):
+class Configurator(object):
+  """Interface used to control hal configuration of a particular feature set."""
+
+  def init_argument_parser(self, parser):
+    """Adds command-line arguments to control configuration."""
+    pass
+
+  def validate_options(self, options):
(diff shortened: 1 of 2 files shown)


dev/validate_bom__config.py
```diff
@@ -564,6 +564,79 @@ def add_files_to_upload(self, options, file_set):
...
+        help='The DC/OS account user name the service principal.'
+             'This is required if DC/OS is enabled.')
+    parser.add_argument(
+        '--dcos_account_credentials', default=None,
+        help='Path to DC/oS credentials file containing the password'
+             ' for the uid. This is required if DC/OS is enabled.')
+
+
+  def validate_options(self, options):
+    """Implements interface."""
```
jtk54: Which interface?

dev/validate_bom__config.py
```diff
@@ -957,6 +1030,7 @@ def add_files_to_upload(self, options, file_set):
     AppengineConfigurator(),
     AzureConfigurator(),
     DockerConfigurator(),
+    DcosConfigurator(),  # must be after docker
```
jtk54: Why does it need to be after Docker?

PR #1911, opened 2017-08-21
chore(springBoot): Config and script changes for Spring Boot upgrade

ttomsu:


dev/validate_bom__config.py
```diff
@@ -1003,7 +1003,31 @@ class SecurityConfigurator(Configurator):
...
+    pass
+
+  def validate_options(self, options):
+    """Implements interface."""
+    pass
+
+  def add_config(self, options, script):
+    """Implements interface."""
+    script.append('mkdir -p  ~/.hal/default/profiles')
+    script.append('echo "management.security.enabled: false"'
```
ewiseblatt: Guard this with a flag. Add the flag to #1011. You can default it off for now and I'll turn it on with my PR later.

ttomsu: Discussed offline - decided that a flag was not needed until a use case for it is found. Eric is working on an implementation that doesn't need to scrape `/resolvedEnv`, which is why we needed to disable management endpoint security to begin with.

PR #1925, opened 2017-08-23
fix(validate): Tolerate 404's to resolvedEnv, which is no longer visi…

ewiseblatt:
…ble by default.

@ttomsu

### DIFF
--- a/testing/citest/README.md
+++ b/testing/citest/README.md
@@ -144,10 +144,34 @@ it using `chmod 200`.
 
 # Running the tests
 
-The following parameters can be passed on the command-line using a flag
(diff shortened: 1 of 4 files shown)


testing/citest/README.md
```diff
@@ -143,11 +143,35 @@ it using `chmod 200`.
...
+
+In spinnaker-local.yml, add the following:
+```
+management:
+  security:
+    enabled: false
+
+endpoints:
+  resolvedEnv:
+    keys-to-sanitize: password
```
ttomsu: consider sanitizing some of the other more common options: `password,secret,key,token`

PR #1944, opened 2017-09-08
feat(release): Minimum halyard version

lwander:


lwander:
The default will quickly go out-of-date, this will just need to be something manual for now until we come up with a better system.

PR #1956, opened 2017-09-15
chore(dev): Accomodate mixed-branch checkouts

ewiseblatt:
@jtk54 
This is clumsy but lets you refresh_source using a branch without requiring the branch to exist in all the repos. The point of this is to be able to use your script to build a Bom to feed the test automation over a change in a microservice. The default behavior is as before (requires the branch to be uniform) however if you --allow_master then a branch failure will fallback on master to distinguish between the repository not existing (still required) vs the branch not being present.

### DIFF
--- a/dev/refresh_source.py
+++ b/dev/refresh_source.py
@@ -78,6 +78,10 @@ class Refresher(object):
   branch is 'master'. This is intended to perform complete updates of the
   local repositories.
 
+  If the branch is not found, then pull the --default_branch if one is set.
+  This allows builds of an individual feature branch affecting only a
+  subset of repositories.
+  
   --push_branch (or --push_master, implying 'master' branch) will push the
   local repository branch back to the origin, but only if the local repository
   is in the specified branch. This is for safety to prevent accidental pushes.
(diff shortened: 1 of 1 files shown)


dev/refresh_source.py
```diff
@@ -523,6 +539,10 @@ def init_argument_parser(cls, parser):
                                ' otherwise skip it.'
                                ' If cloning, then clone this branch.')
 
+      parser.add_argument('--allow_master', default=False, action='store_true',
```
jtk54: maybe `fallback_master` or `default_master`? `allow` seems like it would deal with permissions.

dev/refresh_source.py
```diff
@@ -226,6 +230,18 @@ def git_clone(self, repository, owner=None):
       shell_result = run_and_monitor(
           'git clone {url} -b {branch}'.format(url=origin_url, branch=branch),
           echo=False)
+      if (shell_result.returncode
+          and branch != "master"
+          and self.__options.allow_master):
+        if shell_result.stdout:
```
jtk54: Is there a specific pattern we'd want to look for in the output? This would default to master no matter the failure (e.g. network unavailability, whatever) which may cause some confusion. If we know exactly which message we need to retry on we can check for it here.

ewiseblatt:
@jtk54 ptal

PR #1978, opened 2017-09-20
refactor(release): Remove unused nebula code

lwander:
Removes all references to unused `nebula` & `nonebula` flags, as well as removes code to publish install script (which isn't referenced/validated against these releases anyhow).

This is done to make xenial support a bit easier in a coming PR.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -210,21 +210,17 @@ class Builder(object):
     bintray_key = os.environ['BINTRAY_KEY']
     bintray_user = os.environ['BINTRAY_USER']
 
-    if options.nebula:
-      target = 'candidate'
-      extra_args = [
-        '--stacktrace',
-        '-Prelease.useLastTag=true',
-        '-PbintrayPackageBuildNumber={number}'.format(number=build_number),
-        '-PbintrayOrg="{org}"'.format(org=org),
-        '-PbintrayPackageRepo="{repo}"'.format(repo=packageRepo),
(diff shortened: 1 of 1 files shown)


ewiseblatt:
I dont remember exactly offhand, but I recall wanting an option to bypass
nebula and it looks like this might be doing away with that. I dont really
do builds at the moment, but when I did, nebula is a pain to deal with as a
developer because it requires clean synced repos (i.e. you cannot build
changes you are working on because your repos arent properly tagged, etc).
If the change here does not prevent you from being able to just build your
current developer state then I cannot think of a problem with it.

Ideally nebula would be changed to be able to bypass some of its
restrictions, but that's a different issue.

lwander:
Do any of us actually use it? The code is immortalized in version control, so removing it from HEAD doesn't make it irrecoverable.

ewiseblatt:
Not at the moment, but seems fundamentally wrong to me to not be able to
perform informal builds.

lwander:
Well I can't seem to get it to work, there isn't any documentation/description of how to use it, and as far as we know there isn't anyone actively using it. I'd like to extend these scripts to support more flavors of ubuntu, and this code makes that very difficult.

When you say informal, do you mean "without publishing to our proper bintray repo"? or "without publishing to any repo"? or something else?

ewiseblatt:
Forget it. It seems there are more issues that crept in preventing
convenient informal builds.

PR #2045, opened 2017-10-19
feat(google): Configure test applications for  platform health checks

ewiseblatt:
@jtk54, @duftler 
Allow test applications platformHealthOnly and explicitly create an application for server group tests. It still isnt clear why this was suddenly a problem when run against old branches.


ewiseblatt:
@jtk54 ping

PR #2048, opened 2017-10-20
Patch release testing

ewiseblatt:
@jtk54 3 cherry picks for testing the patch release


ewiseblatt:
Good catch. I wonder why it is showing a diff. Master should already have all these.
I opened another PR against the release branch.

PR #2054, opened 2017-10-23
feat(bom): Set sane default maven lookup path.

jtk54:
Netflix publishes supporting packages (e.g. EmbeddedRedis) to `https://bintray.com/spinnaker/spinnaker`, and then requests a sync from that repository to a Maven repository located at `https://jcenter.bintray.com/com/netflix/spinnaker`. Occasionally the sync misbehaves and the two are out of sync, which results in our builds being broken.

This change allows our builds to point at any maven repository, defaulting to the Maven repo Netflix publishes packages to directly so we can avoid problems with the sync to the second repository.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -42,8 +42,6 @@ Usage:
 """
 
 import argparse
-import base64
-import collections
 import datetime
 import fnmatch
 import glob
@@ -58,8 +56,6 @@ import subprocess
 import sys
 import tempfile
 import time
-import urllib2
-from urllib2 import HTTPError
 
 import refresh_source
 
@@ -222,6 +218,8 @@ class Builder(object):
       '-PbintrayUser="{user}"'.format(user=bintray_user)
     ]
 
+    if options.maven_custom_init_file:
+      extra_args.append('-I {}'.format(options.maven_custom_init_file))
 
     if options.info_gradle:
       extra_args.append('--info')
@@ -648,6 +646,11 @@ class Builder(object):
       parser.add_argument(
           '--do_jar_build', type=bool, default=True,
           help='Build & publish jars independently to GCS.')
+      parser.add_argument('--maven_custom_init_file', default=os.path.join(os.path.dirname(__file__), 'maven-init.gradle'),
+          help='Path to a gradle init file to add to the debian builds.'
+               'Used to specify any custom behavior in the gradle builds.'
+               'Argument is a file path relative to the directory this script is executed in.'
+               'The default value assumes we run this script from the parent directory of spinnaker/spinnaker.')
 
   def __verify_bintray(self):
     if not os.environ.get('BINTRAY_KEY', None):
--- a/dev/maven-init.gradle
+++ b/dev/maven-init.gradle
@@ -1,0 +1,13 @@
+// By default, the Spinnaker microservices use only jcenter.
+// For standard Spinnaker builds, we want to pick fetch JARs from
+// the standard Spinnaker releases repository, then Netflix's repository,
+// then JCenter. The JCenter repository may sometimes lag behind Netflix
+// for newly built components (Netflix pushes from their repository to JCenter).
+
+allprojects {
+  repositories {
+    maven { url 'https://spinnaker-releases.bintray.com/jars'  }
+    maven { url 'https://spinnaker.bintray.com/spinnaker'  }
+    maven { url 'https://jcenter.bintray.com'  }
+  }
+}


dev/build_release.py
```diff
@@ -222,6 +222,9 @@ def __debian_build(cls, name, options, build_number, gradle_root):
       '-PbintrayUser="{user}"'.format(user=bintray_user)
     ]
 
+    if os.path.exists('maven-override.gradle'):
```
ewiseblatt: I think this should be controlled by a flag. Otherwise the override is sticky (earlier runs with the override will inadvertantly affect the later runs). The flag can be the name of the file. Setting it enables the override.

It's not clear to me how this is used and whether this is a standard name or not.

dev/generate_bom.py
```diff
@@ -110,6 +110,8 @@ def init_argument_parser(cls, parser):
                         help="GCE project we publish HA Spinnaker component images to.")
     parser.add_argument('--git_prefix', default='https://github.com/spinnaker',
                         help="Prefix to the component source URIs.")
+    parser.add_argument('--maven_override', default='http://spinnaker.bintray.com/spinnaker',
```
ewiseblatt: This is a confusing name. You have an override file, but htis is specifying a repository. I'd call this something like -maven_override_repository. However it isnt clear if this is the default repo or not, in which case it isnt --maven_override_repository, rather --maven_repository. However  it is only spinnaker, so maybe this is the --spinnaker_maven_repository.

It isnt clear how this goes with the maven override. It seems simpler if you always have the override file (in which case it is no longer an override file) but that requires scripting. So this is looking too messy and complicated. I dont understand it well enough to provide an alternative. We can talk offline.

jtk54:
Ping, refactored this a bit and addressed offline comments. Should look/feel a bit better now.

dev/build_release.py
```diff
@@ -222,6 +222,9 @@ def __debian_build(cls, name, options, build_number, gradle_root):
       '-PbintrayUser="{user}"'.format(user=bintray_user)
     ]
 
+    if options.debian_gradle_init_file:
```
ewiseblatt: This still looks awkward because you initialize this in two different places and the parameter is specified in build_release but is based on a file in a different module entirely.

Why not instead have a parameter for the "maven_custom_initialization_file", which is empty by default. Then here check if there was one supplied or not. If so then use it. Otherwise generate a temporary file containing a static string (the body of the init script below) and use that instead. Create a local variable containing the name of the file to delete (empty by default) then wrap the command execution in a try/finally where you check and delete in the finally block.

There is no change to generate_bom or build_prevalidate. This assumes that there is only one context for executing maven. You encapsulate the maven configuration here in the same module that runs maven, and you localize the use of the configuration right where its used.

jtk54: SGTM.

jtk54:
Test finally finished successfully, PTAL @ewiseblatt.

dev/maven-init.gradle
```diff
@@ -0,0 +1,7 @@
+allprojects {
```
ewiseblatt: If this file accepts comments, you might want to explain it

dev/build_release.py
```diff
@@ -648,6 +647,11 @@ def init_argument_parser(cls, parser):
       parser.add_argument(
           '--do_jar_build', type=bool, default=True,
           help='Build & publish jars independently to GCS.')
+      parser.add_argument('--maven_custom_init_file', default='spinnaker/dev/maven-init.gradle',
```
ewiseblatt: The default here is assuming that you are running from build (or the parent). Is this really the case? I think it would be safer to say
   os.path(os.path.dirname(__file__), 'maven-init.gradle')

dev/build_release.py
```diff
@@ -222,6 +218,9 @@ def __debian_build(cls, name, options, build_number, gradle_root):
       '-PbintrayUser="{user}"'.format(user=bintray_user)
     ]
 
+    if options.maven_custom_init_file:
+      init_file_path = os.path.abspath(options.maven_custom_init_file)
```
ewiseblatt: why do you need the abspath here?

dev/maven-init.gradle
jtk54: Done.

dev/build_release.py
jtk54: Done.

dev/build_release.py
jtk54: Jenkins's workspace directory symlinking throws off the path building.

dev/maven-init.gradle
```diff
@@ -0,0 +1,12 @@
+// This block configures several source JAR repos for the gradle build
+// to search for dependencies in, in descending order of priority.
+// By default, the components have only the third configured, which is not
```
ewiseblatt: Clarify this. Something like the following

By default, the spinnaker microservices only use jcenter.
For standard spinnaker builds, we want to pick up from the standard spinnaker releases repository first, then from netflix's repository. The netflix repository is included here because the jcenter repository may sometimes lag behind netflix for newly built components (netflix pushes from their repository to jcenter).

dev/build_release.py
ewiseblatt: Is this really still a problem after the adding the path below to the default value?

jtk54: We'll see, I'll try it.

jtk54: Seems like it works.

PR #2061, opened 2017-10-26
feat(testcase) : Adding Statefulset & Daemonset creation Testcase

prashantvicky:
@ewiseblatt Please review the Daemonset & Statefulsets creation testcase in smoke test.
Regards
Prashant 

### DIFF
--- a/testing/citest/tests/kube_smoke_test.py
+++ b/testing/citest/tests/kube_smoke_test.py
@@ -247,6 +247,11 @@ class KubeSmokeTestScenario(sk.SpinnakerTestScenario):
             bindings['SPINNAKER_KUBERNETES_ACCOUNT']),
         application=self.TEST_APP)
     builder = kube.KubeContractBuilder(self.kube_observer)
+    (builder.new_clause_builder('StatefulSet Added', retryable_for_secs=60)
+     .get_resources(
+         'statefulsets', extra_args=[self.TEST_APP,'--namespace', self.TEST_NAMESPACE])
+     .contains_path_value('kind', "StatefulSet"))
+
     return st.OperationContract(
         self.new_post_operation(
             title='create_statefulset', data=payload, path='pipelines/'+self.TEST_APP+'/statefulset-pipeline',
@@ -254,50 +259,32 @@ class KubeSmokeTestScenario(sk.SpinnakerTestScenario):
         contract=builder.build())
 
   def save_statefulset_pipeline(self):
-      bindings = self.bindings
-      job={
-	                "keepWaitingPipelines": "false",
-                        "application": self.TEST_APP,
-                        "name": "statefulset-pipeline",
-	                "lastModifiedBy": "anonymous",
-	                "limitConcurrent": "true",
-	                "parallel": "true",
-	                "stages": [{
-		                "clusters": [{
-			                "account": bindings["SPINNAKER_KUBERNETES_ACCOUNT"],
-			                "application": self.TEST_APP,
-			                "cloudProvider": "kubernetes",
-			                "containers": [{
-				                    "imageDescription": {
-					                "repository": "nginx"
-				                        },
-				            "imagePullPolicy": "ALWAYS",
-				            "limits": {
-					            "memory": "200Mi"
-				                    },
-				            "name": "nginx",
-				            "requests": {
-					            "cpu": "100m",
-					            "memory": "200Mi"
-				                }
-			                }],
-			        "kind": "StatefulSet",
-			        "namespace": "default",
-			        "provider": "kubernetes",
-			        "region": "default",
-			        "targetSize": 0,
-			        "terminationGracePeriodSeconds": 30
-		                }],
-		                "name": "Deploy",
-		                "refId": "1",
+      deploy_stage = self.make_deploy_stage_statefulset()
+      job = dict(
+          keepWaitingPipelines= 'false',
+                    application= self.TEST_APP,
+                    name= 'statefulset-pipeline',
+	                lastModifiedBy= 'anonymous',
+	                limitConcurrent= 'true',
+	                parallel= 'true',
+	                stages= [deploy_stage]
+      )
+      payload = self.agent.make_json_payload_from_kwargs(**job)
+      expect_match = {key: jp.EQUIVALENT(value)
+                    for key, value in job.items()}
+      expect_match['stages'] = jp.LIST_MATCHES(
+      [
+       jp.DICT_MATCHES({key: jp.EQUIVALENT(value)
+                        for key, value in deploy_stage.items()})])
 
-		            "type": "deploy"
-	                }],
-	                "triggers": [{"enabled": "true"}]
-                }
-      val=json.dumps(job)
-      builder = kube.KubeContractBuilder(self.kube_observer)
-      return st.OperationContract(self.new_post_operation(title='create_statefulset', data=val, path='pipelines',status_class=st.SynchronousHttpOperationStatus),contract=builder.build())
+      builder = st.HttpContractBuilder(self.agent)
+      (builder.new_clause_builder('Has Pipeline',
+                                  retryable_for_secs=15)
+       .get_url_path(
+          'applications/{app}/pipelineConfigs'.format(app=self.TEST_APP))
+      .contains_match(expect_match))
+
+      return st.OperationContract(self.new_post_operation(title='create_statefulset', data=payload, path='pipelines',status_class=st.SynchronousHttpOperationStatus),contract=builder.build())
 
   def create_daemonset_pipeline(self):
     bindings = self.bindings
@@ -311,6 +298,11 @@ class KubeSmokeTestScenario(sk.SpinnakerTestScenario):
             bindings['SPINNAKER_KUBERNETES_ACCOUNT']),
         application=self.TEST_APP)
     builder = kube.KubeContractBuilder(self.kube_observer)
+    (builder.new_clause_builder('Daemonset Added', retryable_for_secs=60)
+     .get_resources(
+         'daemonsets', extra_args=[self.TEST_APP,'--namespace', self.TEST_NAMESPACE])
+     .contains_path_value('kind', "DaemonSet"))
+
     return st.OperationContract(
         self.new_post_operation(
             title='Create_daemonset', data=payload, path='pipelines/'+self.TEST_APP+'/daemonset-pipeline',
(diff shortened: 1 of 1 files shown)


testing/citest/tests/kube_smoke_test.py
```diff
@@ -235,6 +234,159 @@ def delete_load_balancer(self):
...
+        application=self.TEST_APP)
+    builder = kube.KubeContractBuilder(self.kube_observer)
+    (builder.new_clause_builder('Run pipeline', retryable_for_secs=15)
+     .get_resources(
+         'svc',
+         extra_args=['--namespace', self.TEST_NAMESPACE],
+         no_resource_ok=True))
+    return st.OperationContract(
+        self.new_post_operation(
+            title='create_state_ful_sets', data=payload, path='pipelines/'+self.TEST_APP+'/statefulset-pipeline',
```
ewiseblatt: is this intentinoally "state_ful_" as opposed to "stateful_" ?

testing/citest/tests/kube_smoke_test.py
```diff
@@ -235,6 +234,159 @@ def delete_load_balancer(self):
...
+		                'refId': '1',
+
+		            'type': 'deploy'
+	                }],
+	                'triggers': [{'enabled': 'true'}]
+                }],
+                    description='Save Daemonset Pipenine:' ,application=self.TEST_APP)
+      builder = kube.KubeContractBuilder(self.kube_observer)
+      builder.new_clause_builder('Run pipeline', retryable_for_secs=15).get_resources('svc',extra_args=['--namespace', self.TEST_NAMESPACE],no_resource_ok=True)
+      my_dict = eval(payload)
```
ewiseblatt: I am confused. You are never using the json payload as json.
Instead "my_dict" is just the dictionary you built up to pass into the make_json_payload_from_kwargs, so just build up the dictionary.
Furthermore, you are just extracting 'job' out, so just set a job variable here to the dictionary you want. But then you convert it to a string and change the quoting syntax then pass it as the data to the post operation. So I assume it is in fact supposed to be a JSON encoding.

So really I think the "payload" above should just be the job dictionary, then return here with data=payload. Or call json.JSONEncoder().encode directly with the job dictionary if you'd prefer to specify the payload as a dictionary given its complexity.

testing/citest/tests/kube_smoke_test.py
```diff
@@ -518,6 +669,18 @@ def test_a_create_app(self):
 
   def test_b_upsert_load_balancer(self):
     self.run_test_case(self.scenario.upsert_load_balancer())
+  def save_statefulset_pipeline(self):
```
ewiseblatt: I'm pretty sure that tests have to start with 'test_' in order for the test runner to pick them up as a test case rather than a helper method. So I do not think these tests are actually being run.

Furthermore, there's a caveat in the test running in that it runs tests alphabetically rather than the order of declaration. That's something I never figured out how to fix easily and intended on changing anyway (though never got around to it).

That's why there's the test_a_... and so forth. So you need to sequence these. at least relative to each other if they have dependencies on one another or other tests. For example, you might need these to run between the test_a_create_app and test_z_delete_app. If so, you can treat a letter decorator as a phase, and add additional numbers or some other scheme after that. So for example:
test_c_1_create_statefulset_pipeline
test_c_2_...

You'll probably need to rename some of the existing tests. You can add a gap.
The create_server_group is independent of these so you can stick it in the same phase and it's order doesnt matter so name could remain the same, but htat might feel weird and unintentionally imply that it is part of that phase when really it doesnt matter. So at least a comment

testing/citest/tests/kube_smoke_test.py
```diff
@@ -527,6 +690,22 @@ def test_c_create_server_group(self):
   def test_d_create_find_image_pipeline(self):
     self.run_test_case(self.scenario.create_find_image_pipeline())
 
+  def test_cd_save__statefulsets_pipeline(self):
```
ewiseblatt: As noted above, this will run after c but before d (not because of the d in the name, but because "c" < "cd" < "d". 

Also, do you intend the double _ here and below with the daemonset?

testing/citest/tests/kube_smoke_test.py
```diff
@@ -527,6 +690,22 @@ def test_c_create_server_group(self):
...
+    self.run_test_case(self.scenario.save_statefulset_pipeline())
+
+
+  def test_d_create_statefulsets_pipeline(self):
+    self.run_test_case(self.scenario.create_statefulset_pipeline())
+
+
+  def test_cd_save__daemonset_pipeline(self):
```
ewiseblatt: This is going to execute before the statefulsets because of the lexical ordering.
That's confusing so I'd rename and/or reorder the tests so that they are in the order you intend to execute as well as lexographic.

The long term intention is to specify the dependencies of the tests and the runner will perform the partial ordering to satisfy the constraints. However I never got around to implementing that and the tests have generally been small and simple enough that the lexographic ordering was good enough.

testing/citest/tests/kube_smoke_test.py
```diff
@@ -518,6 +669,18 @@ def test_a_create_app(self):
 
   def test_b_upsert_load_balancer(self):
     self.run_test_case(self.scenario.upsert_load_balancer())
+  def save_statefulset_pipeline(self):
+    self.run_test_case(self.scenario.save_statefulset_pipeline(), max_retries=1,
+                       timeout_ok=True)
```
ewiseblatt: I'm ok with timeout_ok being true. However if this is supposed to execute fairly quickly (relative to say a deploy) then having it fail on a timeout might be desirable depending on the significance of it timing out (e.g. it times out for now, but at some point a later test is going to complain if it still hasnt completed). Again, that's only if you care.

testing/citest/tests/kube_smoke_test.py
```diff
@@ -235,6 +234,159 @@ def delete_load_balancer(self):
...
+		                'refId': '1',
+
+		            'type': 'deploy'
+	                }],
+	                'triggers': [{'enabled': 'true'}]
+                }],
+                    description='Save Statefulsets:' ,application=self.TEST_APP)
+      builder = kube.KubeContractBuilder(self.kube_observer)
+      builder.new_clause_builder('Run pipeline', retryable_for_secs=15).get_resources('svc',extra_args=['--namespace', self.TEST_NAMESPACE],no_resource_ok=True)
+      my_dict = eval(payload)
```
ewiseblatt: See comment later at #384 about this eval(payload).

testing/citest/tests/kube_smoke_test.py
```diff
@@ -235,6 +234,159 @@ def delete_load_balancer(self):
...
+		                'name': 'Deploy',
+		                'refId': '1',
+
+		            'type': 'deploy'
+	                }],
+	                'triggers': [{'enabled': 'true'}]
+                }],
+                    description='Save Statefulsets:' ,application=self.TEST_APP)
+      builder = kube.KubeContractBuilder(self.kube_observer)
+      builder.new_clause_builder('Run pipeline', retryable_for_secs=15).get_resources('svc',extra_args=['--namespace', self.TEST_NAMESPACE],no_resource_ok=True)
```
ewiseblatt: I'm not sure what this pipeline is doing because I'm not familiar with the operations. It looks like you dont have a contract here. You are getting resources but then not doing anything with it. You say no_resource_ok but not that you expect no resources. So basically it doesnt matter what happens here, so you havent actually tested anything other than the operation completing. But even then you permit a timeout when you call it.

There's a newer way to test for errors without using no_resource_ok. I havent updated the kube_agent for it (to discourage no_resources_ok and perhaps add explicit support for error checking as I did in gcp and aws agents). I think the most readable way is:

from citest.json_contract import ObservationPredicateFactory
ov_factory = ObservationPredicateFactory()

Then, assuming you are expecting an error:
(builder.new_clause_builder(...)
 .get_resouces('svc', extra_arguments=...)
 .EXPECT(ov_factory.error_list_matches(...))   # or ov_factory.error_list_contains([]))
)

Or if you are expecting it either way, you can add another OR clause with what you would expect.

I think this makes the code and intent clearer as well as clear up the reporting.

testing/citest/tests/kube_smoke_test.py
```diff
@@ -235,6 +234,159 @@ def delete_load_balancer(self):
...
+        description='Create Daemonset'.format(
+            self.__lb_name,
+            bindings['SPINNAKER_KUBERNETES_ACCOUNT']),
+        application=self.TEST_APP)
+    builder = kube.KubeContractBuilder(self.kube_observer)
+    (builder.new_clause_builder('Run pipeline', retryable_for_secs=15)
+     .get_resources(
+         'svc',
+         extra_args=['--namespace', self.TEST_NAMESPACE],
+         no_resource_ok=True))
```
ewiseblatt: similar to above, I dont think you are checking for anything here. You allow it to be empty but do not require it to be empty.

testing/citest/tests/kube_smoke_test.py
```diff
@@ -235,6 +234,159 @@ def delete_load_balancer(self):
...
+		                'name': 'Deploy',
+		                'refId': '1',
+
+		            'type': 'deploy'
+	                }],
+	                'triggers': [{'enabled': 'true'}]
+                }],
+                    description='Save Daemonset Pipenine:' ,application=self.TEST_APP)
+      builder = kube.KubeContractBuilder(self.kube_observer)
+      builder.new_clause_builder('Run pipeline', retryable_for_secs=15).get_resources('svc',extra_args=['--namespace', self.TEST_NAMESPACE],no_resource_ok=True)
```
ewiseblatt: not enforcing anything here either (same as above).

prashantvicky:
Changes 
1) Name miss spelled rectified ("state_ful_" as opposed to "stateful_" )
2) Json payload comment incorporated
3) Test cases starts with "test_" (Though it works perfectly if they does not starts with tests)
4) lexical ordering issue resolved
5) resource been called but not been used (resolved)
@ewiseblatt please review the code.
Thanks
Prashant

ewiseblatt:
Sorry for the individual responses above. I didnt realize I was doing that.

With regard to your #3 Test cases starts with "test_" (Though it works perfectly if they does not starts with tests), are you sure it ran those tests? I dont think the test runner adds methods that dont start with 'test_'. I suspect the fixture passed, but if you look at the test cases run, it didnt have those methods. If it did then I think something is broken in my code...

prashantvicky:
Added all the comments As well as contract. @ewiseblatt please review.

ewiseblatt:
Replaced with #2122

PR #2073, opened 2017-10-31
fix(http_lb): Update upsert description for UHCs.

jtk54:


jtk54:
The unified health checks have a `healthCheckType`, which the L7 LB upsert chokes on if it's missing.

PR #2129, opened 2017-11-28
feat(build): checkout source from bom

ewiseblatt:
@jtk54


jtk54:
I think you can import https://github.com/spinnaker/spinnaker/blob/master/dev/reconstruct_source.py#L37 to use here.

PR #2141, opened 2017-12-01
chore(spinnaker): break out gcs interaction in gcb build

skim1420:
@ewiseblatt @ttomsu 
Your call on if/whether/when you merge and try this.

### DIFF
--- a/dev/build_release.py
+++ b/dev/build_release.py
@@ -371,11 +371,24 @@ class Builder(object):
     if os.path.isdir(gradle_cache):
       # Tell rmtree to delete the directory even if it's non-empty.
       shutil.rmtree(gradle_cache)
-    cmds = [
-      ('gcloud container builds submit --log-http'
-       ' --account={account} --project={project} --config="../{name}-gcb.yml" .'
-       .format(name=name, account=gcb_service_account, project=gcb_project))
-    ]
+    if self.__options.gcb_externalize_gcs_upload:
+      tar_filename = name + '-{:%Y%m%d%H%M%S}'.format(datetime.datetime.utcnow())
+      gcs_staging_bucket = self.__options.gcb_gcs_staging_bucket
+      cmds = [
+        ('tar -cf {tar_filename}.tar .'
+          .format(tar_filename=tar_filename)),
+        ('gsutil cp {tar_filename}.tar {gcs_staging_bucket}'
+          .format(tar_filename=tar_filename, gcs_staging_bucket=gcs_staging_bucket)),
+        ('gcloud container builds submit --log-http'
+         ' --account={account} --project={project} --config="../{name}-gcb.yml" {gcs_staging_bucket}/{tar_filename}'
+         .format(name=name, account=gcb_service_account, project=gcb_project, gcs_source=gcs_source_uri, tar_filename=tar_filename))
+      ]
+    else:
+      cmds = [
+        ('gcloud container builds submit --log-http'
+         ' --account={account} --project={project} --config="../{name}-gcb.yml" .'
+         .format(name=name, account=gcb_service_account, project=gcb_project))
+      ]
     logfile = '{name}-gcb-build.log'.format(name=name)
     if os.path.exists(logfile):
       os.remove(logfile)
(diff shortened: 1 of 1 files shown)


dev/build_release.py
```diff
@@ -371,10 +371,16 @@ def __gcb_build(cls, name, gradle_root, gcb_service_account, gcb_project):
     if os.path.isdir(gradle_cache):
       # Tell rmtree to delete the directory even if it's non-empty.
       shutil.rmtree(gradle_cache)
+    tar_filename = name + '-{:%Y%m%d%H%M%S}'.format(datetime.datetime.utcnow())
```
ewiseblatt: ok, but could you feature flag this? You can default the feature to true and not provide a CLI option for it. That will make it easier to remove when this is no longer needed.

Also bucket name looks like s/gbc/gcb and would need to be parameterized to accomodate non-official builds.

skim1420: Done.

dev/build_release.py
```diff
@@ -371,11 +371,24 @@ def __gcb_build(cls, name, gradle_root, gcb_service_account, gcb_project):
...
       # Tell rmtree to delete the directory even if it's non-empty.
       shutil.rmtree(gradle_cache)
-    cmds = [
-      ('gcloud container builds submit --log-http'
-       ' --account={account} --project={project} --config="../{name}-gcb.yml" .'
-       .format(name=name, account=gcb_service_account, project=gcb_project))
-    ]
+    if __self.options.externalize_gcs_upload:
+      tar_filename = name + '-{:%Y%m%d%H%M%S}'.format(datetime.datetime.utcnow())
+      gcs_staging_bucket = __self.options.gcs_staging_bucket or 'gs://spinnaker-build-gcb-staging'
```
ewiseblatt: give the option a default and scrap the or

ewiseblatt: dont you need to add --gce_staging_bucket and --externalize_gcs_upload  to the argparser?

skim1420: Done.

dev/build_release.py
```diff
@@ -371,11 +371,24 @@ def __gcb_build(cls, name, gradle_root, gcb_service_account, gcb_project):
...
-       .format(name=name, account=gcb_service_account, project=gcb_project))
-    ]
+    if self.__options.externalize_gcs_upload:
+      tar_filename = name + '-{:%Y%m%d%H%M%S}'.format(datetime.datetime.utcnow())
+      gcs_staging_bucket = self.__options.gcs_staging_bucket
+      cmds = [
+        ('tar -cf {tar_filename}.tar .'
+          .format(tar_filename=tar_filename)),
+        ('gsutil cp {tar_filename}.tar {gcs_staging_bucket}'
+          .format(tar_filename=tar_filename, gcs_staging_bucket=gcs_staging_bucket)),
```
ewiseblatt: You can have an account conflict here between gsutil and gcloud.

dev/build_release.py
```diff
@@ -654,6 +667,12 @@ def init_argument_parser(cls, parser):
                'Used to specify any custom behavior in the gradle builds.'
                'Argument is a file path relative to the directory this script is executed in.'
                'The default value assumes we run this script from the parent directory of spinnaker/spinnaker.')
+      parser.add_argument(
```
ewiseblatt: ok, though ideally these should be named gcb_... to scope them to the gcb build module.

dev/build_release.py
ewiseblatt: gsutil doesnt let you set the account, so you'll need to gcloud set account before running this feature (which will have the side effect of changing the account for gcloud after the command is run).

dev/build_release.py
skim1420: Done

dev/build_release.py
skim1420: I added the `gcloud config set account ...` call. I don't see a way out of explicitly doing this, and all subsequent gcloud calls are specifying the --account parameter. If you have alternative methods, please suggest.

dev/build_release.py
```diff
@@ -371,11 +371,26 @@ def __gcb_build(cls, name, gradle_root, gcb_service_account, gcb_project):
...
-       ' --account={account} --project={project} --config="../{name}-gcb.yml" .'
-       .format(name=name, account=gcb_service_account, project=gcb_project))
-    ]
+    if self.__options.gcb_externalize_gcs_upload:
+      tar_filename = name + '-{:%Y%m%d%H%M%S}'.format(datetime.datetime.utcnow())
+      gcs_staging_bucket = self.__options.gcb_gcs_staging_bucket
+      cmds = [
+        ('tar -cf {tar_filename}.tar .'
+          .format(tar_filename=tar_filename)),
+        ('gcloud config set account {account}'
```
ewiseblatt: I dont think you should do this. It's up to the caller (e.g. jenkins script)
This is a global system wide side effect.
Youd definitely dont want to do that -- it woudl affect other independent jobs running.

I was just pointing out the potential vulnerability in the script.
I dont know how to fix it. It is a gsutil limitation.

skim1420: Is it avoidable? How else will gsutil get credentials to do the upload?

skim1420: Done, removed the set account call.

dev/build_release.py
ewiseblatt: as discussed offline, call is put in the caller of this script rather than hidden in this script to avoid subtle permanent side effects leading to confusion and possible bugs. the change is still made, but by a different role (that already manages accounts) than this script so should be more clear that it is happening.

ewiseblatt:
Can you close this out? It is obsolete.

PR #2156, opened 2017-12-08
fix(annotate_source): Remove errant re-tagging if no changes.

jtk54:


dev/annotate_source.py
```diff
@@ -220,8 +220,6 @@ def tag_head(self):
       # We didn't manually specify, but want to force a rebuild of the old tag.
       elif self.__force_rebuild:
```
lwander: This only happens when head and the last tag are equal?

jtk54: Yes, otherwise it's a new semver because there are changes since the last tag.

PR #2182, opened 2017-12-18
Buildtool.scripts

ewiseblatt:
@jtk54,
@brandonnelson3 

I'm probably going to send these out as individual reviews for easier feedback/discussion/focus.
However this is a preview of where it is going and how it all comes together.

I have not done the publish_ scripts for release yet, however this is capable of supporting the pre-release build and validation. I've verified this on VMs but not yet on containers (though can build the containers).

### DIFF
--- a/dev/buildtool/flow_generate.sh
+++ b/dev/buildtool/flow_generate.sh
@@ -1,0 +1,128 @@
+#!/bin/bash
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Usage:  spinnaker/dev/buildtool/flow_generate.sh <defaults-file>
+#   e.g.  spinnaker/dev/buildtool/flow_generate.sh ga-defaults.yml
+#
+# Note: This can be run from any directory, however root_dir path
+#       in the <defaults-file> will be relative to the CWD so it
+#       likely makes sense to run from the project parent directory.
+
+set -e
+
+# Import support functions
+source $(dirname $0)/support_functions.sh
+
+
+########################################################
+# Fetches the main sources (wait for completion)
+#
+#   Some specialized source repositories might not be
+#   included here, particularly if they are only used
+#   by individual command modules.
+########################################################
+function run_fetch_flow() {
+  start_command_unless NO_FETCH "fetch_source"
+
+  # Synchronize here so we have source before we continue.
+  wait_for_commands_or_die "Fetch"
+}
+
+
+########################################################
+# Generate all the release artifacts (wait for completion)
+#
+#    This flow generates all the artifacts so they can
+#    be verified and validated. It does not publish any
+#    of them beyond the needs for validation processes.
+########################################################
+function run_generate_flow() {
+  start_command_unless NO_HALYARD "build_halyard"
+  start_command_unless NO_CONTAINERS "build_bom_containers"
+  start_command_unless NO_CHANGELOG "generate_changelog"
+  start_command_unless NO_BOM "generate_bom"
+  start_command_unless NO_APIDOCS "generate_api_docs"
+
+  # Avoid a race condition where different jobs are git cloning into the
+  # scratch directory
+  echo "$(date): PAUSING flow 15s to avoid a potential git race condition"
+  sleep 15
+  echo "$(date): RESUMING"
+
+  start_command_unless NO_DEBIANS "build_debians" "--max_local_builds=6"
+
+  # Synchronize here so we have all the artifacts generated before we continue.
+  wait_for_commands_or_die "Build"
+
+  # Only publish the bom for this build if the build succeeded.
+  #
+  # NOTE(ewiseblatt): 20171220
+  # We probably want to publish if at least one implementation succeeded
+  # (e.g. if we have VMs but not containers or vice-versa)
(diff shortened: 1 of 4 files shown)


dev/buildtool/source_code_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    base_url = bom['artifactSources']['gitPrefix']
+    origin_url = spec.get(
+        'sourceRepository',
+        '{base}/{name}'.format(base=base_url, name=repository_name))
+    branch = spec.get(
+        'sourceBranch',
+        bom['artifactSources'].get('gitBranch', '(unknown)'))
+
+    self.__git.clone_repository_to_path(origin_url, git_dir, commit=commit)
+    self.__git.reinit_local_repository_with_tag(
```
jtk54: Looks like we are re-initializing with the git repo with a `version-{}` tag before we run the gradle builds. What are the are the resulting versions for the debians and container images - do they contain `version-`?

ewiseblatt: If you are looking at this, I assume you are choosing the review it all at once option.
I thought it might be easier to digest in pieces, but your call (sorry brandon -- we can do a separate walk through)

This part is a little weird because of nebula. I thought that the repo needs to be tagged before gradle because gradle uses the tag. It is tagged with the version.

> git -C scratch/build/clouddriver show-ref --tags
d6887d526968ef544ca4c205fa9e716607fa1a5a refs/tags/2.0.0-20171218175040

However I also record what I want the "version" tag to be for later
> cat scratch/clouddriver/clouddriver-summary.yml
commit_id: 623e2457708e9f2b86dfe45811c22016abb572f5
tag: version-2.0.0
version: 2.0.0

The artifact gradle publishes to bintray is this:
About This Version
Website
https://home/ewiseblatt/build/newtest/clouddriver
Issue Tracker
https://home/ewiseblatt/build/newtest/clouddriver
VCS
https://home/ewiseblatt/build/newtest/clouddriver.git
Licenses
Apache-2.0
VCS Tag
v2.0.0-20171218175040
Debian Details
spinnaker-clouddriver_2.0.0-20171218175040_all.deb
Name	spinnaker-clouddriver
Version	2.0.0-20171218175040
Indexed Architecture	i386, amd64
Indexed Component	spinnaker
Indexed Distribution	xenial, trusty ControlFileContent:	
Package	spinnaker-clouddriver
Source	spinnaker-clouddriver
Version	2.0.0-20171218175040
Section	java
Priority	optional
Architecture	all
Installed-Size	180375
Maintainer	ewiseblatt
Description	spinnaker-clouddriver clouddriver-web


The sequencing is a bit different from the original scripts, but I dont have to re-pull anything or modify the local repository associated with the upstream/origin that we'll be pushing tags to later. When I had a "version-" tag in the gradle repo, I was getting something weird that caused parsing errors in gradle if I recall.

The scratch/<component> contains the intermediate state information passed among commands, which are now executing in different processes so dont have shared memory. But before were being regenerated anyway, so was confusing and not as clearly connected to where and when the version information was being computed.

The scratch/build/<component> is the munged directory to satisfy nebula. If/when nebula is fixed then this can go away and the abstraction maintaining it can be changed to refer to the <root_dir>/<component> repository.

BTW, the example above is version 2.0 because of this:
D 17:50:41.950 [Thread-5.23936] Commit is considered "MAJOR" because it says "('BREAKING CHANGE: Stops support of CF until maintainer is found',)"

jtk54: Just had a question on the usage of this specific function, would still like to review this in pieces :)

Thanks for the explanation.

ewiseblatt:
Ok, this PR is now all that's left so is ready to be reviewed.

[I still need to write the publishing commands after this]

dev/buildtool/flow_generate.sh
```diff
@@ -0,0 +1,131 @@
...
+#    This flow generates all the artifacts so they can
+#    be verified and validated. It does not publish any
+#    of them beyond the needs for validation processes.
+########################################################
+function run_generate_flow() {
+  # Not sure if this is needed or if it is covered by build_halyard
+  # start_command_unless NO_HALYARD "build_halyard_containers"
+
+  start_command_unless NO_HALYARD "build_halyard"
+  start_command_unless NO_CONTAINERS "build_bom_containers"
```
jtk54: Which builds the debians?

dev/buildtool/flow_generate.sh
```diff
@@ -0,0 +1,131 @@
...
+
+########################################################
+# Generate all the release artifacts (wait for completion)
+#
+#    This flow generates all the artifacts so they can
+#    be verified and validated. It does not publish any
+#    of them beyond the needs for validation processes.
+########################################################
+function run_generate_flow() {
+  # Not sure if this is needed or if it is covered by build_halyard
```
jtk54: If `build_halyard` runs the Halyard jar build, I believe the containers are covered. @lwander please confirm.

dev/buildtool/support_functions.sh
```diff
@@ -0,0 +1,194 @@
+#!/bin/bash
```
jtk54: Any particular reason for orchestrating this in bash over python?

dev/buildtool/flow_generate.sh
ewiseblatt: build_halyard seems to build everything (I say seems because it wasnt clear how it is supposed to work because of the release/ scripts and I havent run halyard from a container yet, only on a VM).

dev/buildtool/support_functions.sh
ewiseblatt: It's growing organically. Maybe eventually python. It was pretty simplistic except for the process array. It might tip the complexity scale when I write the publishing script. But for now the flow_*.sh looks a lot easier to read and understand than I imagine python would. 

I can switch now if you think I should.

jtk54: Just pointing it out so you're aware. I think this time around things are logically separated and will do enough reporting so that we may not have the issues we've had with orchestration in bash in the past.

Fine for now, we can port it if we need to.

dev/buildtool/flow_generate.sh
lwander: Yes that's right

PR #2183, opened 2017-12-18
feat(buildtool): buildtool.source_code_manager module

ewiseblatt:
@jtk54, @brandonnelson3 

The second PR in the sequence adds source code management.
This is mostly to define the different spinnaker repositories and manage concurrency operating across them.

The weird nebula mirrored source code management is actually in the build module because it feels like a hack that I'm trying to hide where it is needed rather than institutionalize it here. That might prove to be a mistake later, but is how I feel at the moment.

### DIFF
--- a/dev/buildtool/source_code_manager.py
+++ b/dev/buildtool/source_code_manager.py
@@ -1,0 +1,294 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Responsible for managing the source code that we are building.
+
+This includes fetching the source to be built as well as manipulating
+the git repositories for tagging and annotations.
+"""
+
+# Ideally something like a Pool is better because it will run in
+# different processes to bypass some of the global interpreter lock.
+# However because we often run commands in these, which forks another process,
+# we seem to deadlock sometimes (always when running gradle). I think this
+# could be because this pool fork has some locks grabbed which leads to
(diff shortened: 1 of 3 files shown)


dev/buildtool/source_code_manager.py
```diff
@@ -0,0 +1,294 @@
...
+
+from multiprocessing.pool import (
+    Pool,
+    ThreadPool)
+
+import logging
+import os
+import yaml
+
+# pylint: disable=relative-import
```
jtk54: If pylint doesn't like relative imports, what is the proper 'pythonic' way to handle separate modules? Different published libraries? Just curious, I'm fine with the relative imports we have.

dev/buildtool/source_code_manager.py
```diff
@@ -0,0 +1,294 @@
...
+    pull_version = 1 if bom_version else 0
+    pull_path = 1 if bom_path else 0
+    pull_branch = 1 if git_branch else 0
+    have = pull_version + pull_path + pull_branch
+    if have == 0:
+      logging.debug('No source pull requests to process.')
+      return
+
+    if have > 1:
+      raise ValueError('Ambguous source code requests.')
```
jtk54: s/Ambguous/Ambiguous/

dev/buildtool/source_code_manager.py
```diff
@@ -0,0 +1,294 @@
...
+      else:
+        pool = Pool(num_threads, maxtasksperchild=1)
+      logging.info('Mapping with %s %d/%s',
+                   pool.__class__,
+                   len(all_repos.keys()), all_repos.keys())
+      try:
+        raw_list = pool.map(worker, all_repos.values())
+        result = {name: value for name, value in raw_list}
+      except Exception:
+        logging.error('Map caught exception')
```
jtk54: Will we have more contextual information about the failure other than this, or can we dump the exception or exception, task pair for more traceability?

dev/buildtool/source_code_manager.py
ewiseblatt: It wants a leading "."
Or maybe I could just change my pylint.rc file.

dev/buildtool/source_code_manager.py
ewiseblatt: In practice the exeption was already logged when it was thrown so logging the details here is redundant. This message here is confirmation that we unwound this far afterward.

jtk54: Ok, makes sense.

PR #2185, opened 2017-12-18
feat(buildtool): Added command processor

ewiseblatt:
@jtk54, @brandonnelson3 
This is the mechanism for adding process steps into the tool.

### DIFF
--- a/dev/buildtool/command.py
+++ b/dev/buildtool/command.py
@@ -1,0 +1,267 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


dev/buildtool/command.py
```diff
@@ -0,0 +1,259 @@
...
+        *self.__factory_method_pos_args, **self.__factory_method_kwargs)
+
+  def add_argparser(self, subparsers, defaults):
+    """Add specialized argparser for this command.
+
+    Args:
+      subparsers: The subparsers from the argparser parser.
+
+    Returns:
+      A new parser for to add additional custom parameters for this command.
```
jtk54: s/for//

dev/buildtool/command.py
```diff
@@ -0,0 +1,259 @@
...
+            self.__class__.__name__))
+
+  def _do_command(self):
+    self._do_preprocess()
+    result_dict = self.source_code_manager.foreach_source_repository(
+        _do_call_do_repository, self,
+        use_threadpool=self.__use_threadpool)
+    return self._do_postprocess(result_dict)
+
+  def _do_preprocess(self):
```
jtk54: To be extended, right?

ewiseblatt: yea, I clarified instructions in _do_command.
This is a hook so derived classes dont need to implement _do_command rather one of the new _do_ methods depending on where in the processing they want to add behaviors.

PR #2190, opened 2017-12-19
fix(buildtool): detect variants of the same github repository url

ewiseblatt:
@jtk54

### DIFF
--- a/dev/buildtool/git.py
+++ b/dev/buildtool/git.py
@@ -461,6 +461,41 @@ class RepositorySummary(collections.namedtuple(
 class GitRunner(object):
   """Helper class for interacting with Git"""
 
+  @staticmethod
+  def __normalize_repo_url(url):
+    """Normalize a repo url for purposes of checking equality.
+
+    Returns:
+      Either a tuple (HOST, OWNER, PATH) if url is a github-like URL
+         assumed to be in the form <PROTOCOL://HOST/OWNER/PATH> where
+         or ssh@<HOST>:<USER><REPO>
+         in these cases, a '.git' REPO postfix is considered superfluous.
+
+      Otherwise a string assuming the url is a local path
+         where the string will be the absolute path.
+    """
+    dot_git = '.git'
+    gitless_url = (url[:-len(dot_git)]
+                   if url.endswith(dot_git)
+                   else url)
+
+    # e.g. http://github.com/USER/REPO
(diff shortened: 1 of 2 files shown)


dev/buildtool/git.py
```diff
@@ -461,6 +461,41 @@ def to_yaml(self, with_commit_messages=True):
...
+  def __normalize_repo_url(url):
+    """Normalize a repo url for purposes of checking equality.
+
+    Returns:
+      Either a tuple (HOST, OWNER, PATH) if url is a github-like URL
+         assumed to be in the form <PROTOCOL://HOST/OWNER/PATH> where
+         or ssh@<HOST>:<USER><REPO>
+         in these cases, a '.git' REPO postfix is considered superfluous.
+
+      Otherwise a string assuming the url is a local path
```
jtk54: How does this behave if you try to compare a local path and remote? Does this occur in practice? Seems like those two would be incomparable using this helper.

unittest/buildtool/git_test.py
```diff
@@ -110,6 +110,31 @@ def test_determine_tag_at_head(self):
...
         self.assertEquals(version, tag)
 
+  def test_same_repo(self):
+    variants = [
+        'http://github.com/user/spinnaker',
+        'http://github.com/user/spinnaker.git',
+        'https://github.com/user/spinnaker',
+        'https://github.com/user/spinnaker.git',
+        'git@github.com:user/spinnaker.git',
+        'git@github.com:user/spinnaker.git'
```
jtk54: Local path is missing for `True` - not supported?

dev/buildtool/git.py
ewiseblatt: They are different repos unless you were somehow looking at a local path that the "remote" was serving from. A clone of a repo isnt the same as the repo it cloned for this purpose.

This function is a safety check to minimize risk of pushing upstream when the upstream is the origin but with a different url (e.g. ssh vs https or https vs http). In the case of local vs remote, safety is already assured.

FWIW, the GitRemoteRepository class has the origin and the upstream so tracks the relationship you are alluding to. There is a function in this module that can derive one of those objects from an existing repository if the fetch stage were in a separate process so that relationship isnt locally known.

unittest/buildtool/git_test.py
ewiseblatt: not applicable

PR #2193, opened 2017-12-19
chore(buildtool): Asynchronous subprocesses and more logging.

ewiseblatt:
@jtk54

### DIFF
--- a/dev/buildtool/util.py
+++ b/dev/buildtool/util.py
@@ -60,19 +60,20 @@ def maybe_log_exception(where, ex, action_msg='propagating exception'):
   logging.error('"%s" %s', where, action_msg)
 
 
-def run_subprocess(cmd, stream=None, echo=False, **kwargs):
-  """Returns retcode, stdout."""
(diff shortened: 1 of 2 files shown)


dev/buildtool/util.py
```diff
@@ -81,17 +82,35 @@ def run_subprocess(cmd, stream=None, echo=False, **kwargs):
       stdout=subprocess.PIPE,
       stderr=subprocess.STDOUT,
       **kwargs)
+  logging.log(log_level, 'Running %s as pid %s', split_cmd[0], process.pid)
+  process.start_date = start_date
+
   time.sleep(0) # yield this thread
+  return process
```
jtk54: Is the `time.sleep(0)` still needed here? I thought this didn't produce the yield as we expected.

ewiseblatt: It wont hurt. It didnt go to the extent I thought, but might still do something, especially in other use cases that arent under as much load.

PR #2194, opened 2017-12-19
feat(buildtool): Added build_* commands

ewiseblatt:
@jtk54 @brandonnelson3 

This PR is where it gets interesting.

This is just build, not publish/release.

I was able to deploy and validate the boms created by this (bom creation is a future PR) last weekend on VMs. I think I hit an issue on containers that was halyard related and submitted changes there but havent retried yet because I need the reviews to catch up so I dont risk an accident. The GCB containers build without complaint, but I havent deployed them.

The halyard build is sketchy to me because of the release scripts it is using under its control so isnt as clear. I've made some modifications there to be able to inject some control over it but I still am not sure of the different variations and how I can confirm them. Also, the build jars step which also seems to publish some things but not the debian packages. I tried to consolidate here but it might need more refinement or maybe changing the gradle tasks (down the road).

My goal in this PR are to introduce control over building the various artifacts needed to run through a validation, including both halyard and spinnaker runtime.

This build is much faster than the current one. In part because I build the containers and debians in parallel[*] (the containers are external on GCB so dont consume local machine resources so it's cheap to do). Even so, I think it is about 1/3 the time so there's some other savings I'm not entirely sure how I got other than perhaps tuning the "foreach repository" mechanism in the base command processor to accomodate the GIL. I havent paid much attention to performance difference yet. It seems to be building the same artifacts. I add the bom and changelog in separate PRs but those are negligible in time.

[*] This PR is just commands, not flow. So commands operate on repositories in parallel, but the commands themselves are independent from one another and there is no coordination in this PR. When I say they run in parallel, it is because an outside caller is running multiple buildtool commands at the same time. So dont look for that concurrency here. It isnt.

### DIFF
--- a/dev/buildtool/build_commands.py
+++ b/dev/buildtool/build_commands.py
@@ -1,0 +1,731 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Implements build command for buildtool."""
+
+import datetime
+import logging
+import os
+import shlex
+import shutil
+import subprocess
+import yaml
+
+from buildtool.command import (
+    RepositoryCommandFactory,
+    RepositoryCommandProcessor)
+
+from buildtool.source_code_manager import (
+    SPINNAKER_BOM_REPOSITORIES,
+    SPINNAKER_HALYARD_REPOSITORIES)
+
+from buildtool.util import (
+    check_subprocess,
+    timedelta_string,
+    timestring,
+    check_subprocesses_to_logfile,
+    write_to_path)
+
+
+class GradleCommandFactory(RepositoryCommandFactory):
+  """Base class for build commands using Gradle."""
+
+  @staticmethod
+  def add_bom_parser_args(parser, defaults):
+    """Adds publishing arguments of interest to the BOM commands as well."""
+    if hasattr(parser, 'added_gradle'):
+      return
+    parser.added_gradle = True
+
+    GradleCommandFactory.add_argument(
+        parser, 'build_bintray_repository', defaults, None,
+        help='bintray repository to publish debians into.')
+
+  def _do_init_argparser(self, parser, defaults):
+    """Adds command-specific arguments."""
+    super(GradleCommandFactory, self)._do_init_argparser(parser, defaults)
+    self.add_bom_parser_args(parser, defaults)
+
+    # The gradle jobs in the foreach_repository seem to hang when using
+    # pools but not threadpools. I think this is a deadlock related to
+    # job forks. It only hangs after all the jobs completed (cycling through
+    # the reuse of workers). Forcing the pool to not reuse works seems to
+    # have fixed the problem. However I'm leaving this flag in as a
+    # break glass if needed should this crop up again.
+    #
+    # In general using Processes are more resource-efficient than threads
+    # because of the python GIL. The gradle jobs are so CPU heavy that
+    # they cannot run that concurrently anyway so this might not matter
+    # much where builds are local.
+    self.add_argument(
+        parser, 'use_threadpool', defaults, False, action='store_true',
+        help='Use ThreadPool rather than Pool as a workaround for'
+        ' gradle hanging.')
+    self.add_argument(
+        parser, 'max_local_builds', defaults, None, type=int,
+        help='Maximum build concurrency.')
+    self.add_argument(
+        parser, 'run_unit_tests', defaults, False, action='store_true',
+        help='Run unit tests during build for all components other than Deck.')
+    self.add_argument(
+        parser, 'build_jar_repository', defaults, None,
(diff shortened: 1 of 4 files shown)


dev/buildtool/build_commands.py
```diff
@@ -0,0 +1,732 @@
...
+    check_subprocesses_to_logfile,
+    write_to_path)
+
+
+class GradleCommandFactory(RepositoryCommandFactory):
+  """Base class for build commands using Gradle."""
+
+  @staticmethod
+  def add_bom_parser_args(parser, defaults):
+    """Adds publishing arguments of intrest to the BOM commands as well."""
```
jtk54: s/intrest/interest

dev/buildtool/build_commands.py
```diff
@@ -0,0 +1,732 @@
...
+        parser, 'run_unit_tests', defaults, False, action='store_true',
+        help='Run unit tests during build for all components other than Deck.')
+    self.add_argument(
+        parser, 'build_jar_repository', defaults, None,
+        help='bintray repository to publish jar files into.')
+    self.add_argument(
+        parser, 'maven_custom_init_file', defaults,
+        os.path.join(os.path.dirname(__file__), '..', 'maven-init.gradle'),
+        help='Path to a gradle init file to add to the debian builds.'
+        'Used to specify any custom behavior in the gradle builds.'
```
jtk54: Space at the beginning of this string to add a space between the sentences.

dev/buildtool/build_commands.py
```diff
@@ -0,0 +1,732 @@
...
+        'Argument is a file path relative to the directory this script is'
```
jtk54: Same here.

dev/buildtool/build_commands.py
```diff
@@ -0,0 +1,732 @@
...
+        logging.debug('Forcing a repository update because there is no %s',
+                      summary_path)
+        self._get_nebula_repository_path(name, force_new=True)
+
+      with open(summary_path, 'r') as stream:
+        data = yaml.load(stream)
+        self.__nebula_repository_version[name] = data['version']
+    return self.__nebula_repository_version[name]
+
+  def _do_determine_source_repositories(self):
```
jtk54: Generally single and double underscore functions are meant to be private, seems a little strange to be overriding. I missed this in the interface definition. I think the underscore is signalling this is used in another public function in the superclass, which is OK.

jtk54: Relevant SO: https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-a-single-and-a-double-underscore-before-an-object-name - I think this is fine but just be aware of the limitations, e.g. `from M import *` won't import single underscore functions.

dev/buildtool/build_commands.py
```diff
@@ -0,0 +1,732 @@
...
+        'Snapshot from {path} to satisfy nebula.\n\n'
+        'commit: {commit}\ntag: {tag}\nversion: {version}'
+        .format(path=original_repo_path,
+                commit=summary.commit_id,
+                tag=summary.tag,
+                version=summary.version))
+    self.__nebula_repository_version[name] = summary.version
+    summary_path = os.path.join(
+        options.scratch_dir, name, '{name}-summary.yml'.format(name=name))
+    write_to_path(summary.to_yaml(with_commit_messages=False), summary_path)
```
jtk54: What's the intended use of this summary file?

dev/buildtool/build_commands.py
```diff
@@ -0,0 +1,732 @@
...
+
+    # Use an absoluate path here because we're going to
+    # pass this to the gcloud command, which will be running
+    # in a different directory so relative paths wont hold.
+    config_path = os.path.abspath(os.path.join(
+        name_scratch_dir, '{name}-gcb.yml'.format(name=name)))
+    write_to_path(gcb_config, config_path)
+
+    # NOTE(ewiseblatt): 20171213
+    # This is an old comment. Not sure if it is still relevant or not.
```
jtk54: This is still relevant for docker-based builds e.g. GCB or local.

dev/buildtool/build_commands.py
```diff
@@ -0,0 +1,732 @@
...
+                   config_path=config_path))
+
+    logfile = os.path.join(
+        name_scratch_dir, '{name}-gcb-build.log'.format(name=name))
+    check_subprocesses_to_logfile(
+        '{name} container build'.format(name=name), logfile,
+        [cmd], cwd=nebula_dir)
+
+  def __make_gradle_gcb_step(self, name, env_vars_list):
+    # Nebula complains if the git tag version doesn't match the branch's
```
jtk54: This comment is describing a `git checkout` but it doesn't appear in this block of code.

dev/buildtool/build_commands.py
ewiseblatt: The single underscores are intended to be protected. The methods starting with _do_ are meant to be hook methods that derived classes can specialize. I'm using inheritance here so always super. This usage is entirely consistent with that thread.

These are generally methods, not functions. Top level things within a module that have a leading _ are intended to be private with regard to anything other than a unit test.

dev/buildtool/build_commands.py
ewiseblatt: The summary file is persisting state about the repository (GitSummary) so it is available to other processes because the commands are run one at a time in individual processes. It is generally cheap to compute so I dont actually need it, except the nebula repository are created fresh so dont have the commit hashes or version information, never mind any tags. So the summary files provide a cache.

There is a potential vulnerability here with the different repositories in they can get out of sync. It shouldnt happen, especially if there are higher level scripts that manage it (create a new root_dir/scratch_dir or wipe the existing one)

dev/buildtool/build_commands.py
ewiseblatt: oops

PR #2201, opened 2017-12-20
feat(buildtool): Added ApiDocs commands.

ewiseblatt:
@jtk54 @brandonnelson3

### DIFF
--- a/dev/buildtool/apidocs_commands.py
+++ b/dev/buildtool/apidocs_commands.py
@@ -1,0 +1,230 @@
+# Copyright 2017 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


dev/buildtool/apidocs_commands.py
```diff
@@ -0,0 +1,230 @@
...
+          default_branch='master')
+    self.__templates_directory = os.path.join(git_dir, '_api_templates')
+
+  def _do_command(self):
+    """Implements CommandProcessor interface."""
+    self._ensure_templates_directory()
+    redis_service = 'redis-server'
+    start_redis = run_subprocess('sudo service {redis} start'.format(
+        redis=redis_service))[0] != 0
+    logging.debug('redis-server %s running.',
```
jtk54: We log first then try to run redis. Maybe this goes in the `finally`.

PR #2204, opened 2017-12-22
feat(buildtool): Base command for initiating pull requests

ewiseblatt:
This also refactors GitRunner to accomodate authentication
and to consolidate standard commandline parameters for various
use cases involving git.

@jtk54 @brandonnelson3

### DIFF
--- a/dev/buildtool/git.py
+++ b/dev/buildtool/git.py
@@ -23,6 +23,7 @@ import logging
 import os
 import re
 import shutil
+import tempfile
 import time
 
 from subprocess import CalledProcessError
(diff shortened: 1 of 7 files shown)


dev/buildtool/git.py
```diff
@@ -490,12 +551,68 @@ def __normalize_repo_url(url):
...
 
+  @property
+  def options(self):
+    """Return bound options."""
+    return self.__options
+
+  def __init__(self, options):
+    self.__options = options
+    self.__auth_env = {}
+    if self.__options.github_token_path:
```
jtk54: I know we've done it this way in the past, but storing tokens in plaintext files is not a good pattern. I'm not sure about the other ramifications of opting for an env var over a file (like implicitly providing subprocesses with the git credentials per your comment), but we may want to rethink using files for this.

ewiseblatt: Ok. I agree. I have a plan B. Hold on...

ewiseblatt: ok. have a look. I can disregard the whole special casing if not.

PR #2218, opened 2018-01-03
feat(buildtool): Instrument build scripts.

ewiseblatt:
Adds instrumentation into the build scripts.
There are three different implementations for the time being,
each with its own limitations while I'm experimenting.
I think the file based one is where I'll end up but am holding
on to the alternatives for now.

@jtk54 @brandonnelson3


dev/buildtool/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+        help='Frequency at which to push metrics in seconds.')
+    add_parser_argument(
+        parser, 'monitoring_system', defaults, 'file',
+        choices=['file', 'prometheus', 'stackdriver'],
+        help='Where to store metrics.')
+
+  @staticmethod
+  def startup_metrics(options):
+    """Startup metrics module with concrete system."""
+    monitoring_systems = {
```
jtk54: datadog eventually? Which are we planning to use for our nightly builds?

PR #2226, opened 2018-01-05
feat(buildtool): Build GCE component images.

ewiseblatt:
This command uses the old bash scripts for the actual work.
However it modifies thos scripts to remove some built in assumptions
that we are using published artifacts.

@jtk54 @brandonnelson3

### DIFF
--- a/dev/buildtool/ga-defaults.yml
+++ b/dev/buildtool/ga-defaults.yml
@@ -35,6 +35,7 @@
 # Commented values surrounded by '[]' are code defaults.
 # Not all code defaults are shown, only those for commenting purposes.
 
+# root_path:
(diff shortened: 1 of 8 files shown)


dev/buildtool/ga-defaults.yml
```diff
@@ -35,6 +35,7 @@
 # Commented values surrounded by '[]' are code defaults.
 # Not all code defaults are shown, only those for commenting purposes.
 
+# root_path:
```
jtk54: Extraneous, or just there for documentation?

ewiseblatt: docs

PR #2236, opened 2018-01-09
chore(buildtool): implementation cleanup

ewiseblatt:
Changed strategy for dealing with nebula friction by using a decoy origin
repository rather than removing all the confusing tags and versions.
Using the nebula release.verison switch eliminates the need to remove tags,
however it insists on pushing a new tag (and the wrong pattern at that).
The decoy repository tricks nebula into pushing the tag to somewhere irrelvant.

Cleaned up some of the gradle build command classes, adding build_rpms

@jtk54 
This contains the other changes I was working on before fixing nebula.
I combined them together because it was easier for me and doesnt affect anyone else yet.

### DIFF
--- a/dev/buildtool/build_commands.py
+++ b/dev/buildtool/build_commands.py
@@ -17,6 +17,7 @@
 import datetime
 import logging
 import os
+import re
 import shlex
 import shutil
 import subprocess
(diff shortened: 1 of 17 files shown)


dev/buildtool/build_commands.py
```diff
@@ -32,15 +33,101 @@
...
+
+    summary_match = re.match(
+        r"[> ]*(.*?)'(.+?)': HTTP/[0-9\.]+ ([0-9]{3}) ([\w]+)(.*)",
+        summary_line)
+    failed_by = 'unknown'
+    failed_reason = 'unknown'
+    if not summary_match:
+      logging.error('Cannot interpret cause of failure:\n%s', summary_line)
+    else:
+      # abstract = summary_match.group(1)
```
jtk54: Will be used in the future?

dev/buildtool/build_commands.py
```diff
@@ -234,272 +306,127 @@ def __is_newer_repo(self, test_path, ref_path, log):
...
-        options.scratch_dir, name, '{name}-summary.yml'.format(name=name))
-    write_to_path(summary.to_yaml(with_commit_messages=False), summary_path)
-    return nebula_repo_path
+    jar_repo = options.build_jar_repository
+    if not jar_repo or len(jar_repo.split('/')) != 1:
+      raise ValueError('Expected --build_jar_repository in the form <repo>'
+                       ' using the implied owner from'
+                       ' --build_bintray_repository')
+
+    parts = (options.build_bintray_repository or '').split('/')
```
jtk54: It always seemed weird that we were explicitly passing a single jar repo string, but the org and deb package repo as `{org}/{deb_repo}`.

ewiseblatt: I changed the config options. have a look.

dev/buildtool/build_commands.py
ewiseblatt: i dont know. it's there for documentation. As I encounter and handle more error cases I might change the label schema to better accomodate them. For now we just want to be able to characterize and quantify impediments to building artifacts.

PR #2246, opened 2018-01-12
Cherry pick commit into release-1.5.x: Xenial builds (#2116)

sbwsg:
Construct xenial debs when building from the `release-1.5.x` branch.  This is a cherry-pick from master, commit d2dc70fb6a7876d1cdad189d76ff4e67cb433fa0

* feat(build_release): Add xenial builds for debian package.

PR #2273, opened 2018-01-25
feature(Testcase):Adding Daemonset & Statefulsets creation testcase

prashantvicky:
@lwander These test case is dependent on PR https://github.com/spinnaker/clouddriver/pull/2246
FYI @chlung 

### DIFF
--- a/testing/citest/tests/kube_smoke_test.py
+++ b/testing/citest/tests/kube_smoke_test.py
@@ -473,6 +473,214 @@ class KubeSmokeTestScenario(sk.SpinnakerTestScenario):
             title='delete_server_group', data=payload, path='tasks'),
         contract=builder.build())
 
+  def create_daemonset_pipeline(self):
+      bindings = self.bindings
+      payload = self.agent.make_json_payload_from_kwargs(
+          job=[{
+              'type': 'manual',
+              'user': '[anonymous]'
+          }],
+          description='Create Daemonset in Application' + self.TEST_APP,
+          application=self.TEST_APP)
+      builder = kube.KubeContractBuilder(self.kube_observer)
+      (builder.new_clause_builder('Daemonset Added', retryable_for_secs=15)
+       .get_resources(
+          'daemonsets', extra_args=[self.TEST_APP+'-tst-v001', '--namespace', self.TEST_NAMESPACE])
+       .contains_path_value('kind', 'DaemonSet'))
+      return st.OperationContract(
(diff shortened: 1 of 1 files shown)


testing/citest/tests/kube_smoke_test.py
```diff
@@ -530,6 +738,18 @@ def test_d_create_find_image_pipeline(self):
...
+
+  def test_e_2_create_statefulset_pipeline(self):
+    self.run_test_case(self.scenario.create_statefulset_pipeline())
+
+  def test_f_1_save_daemonset_pipeline(self):
+    self.run_test_case(self.scenario.save_daemonset_pipeline())
+
+  def test_f_2_create_daemonset_pipeline(self):
+    self.run_test_case(self.scenario.create_daemonset_pipeline())
+
```
lwander: Doesn't this need to delete the daemonset and stateful set? This is why this test couldn't pass before.

prashantvicky: No It was suppose to be there issue was with clouddriver we need to provide the functionality to delete daemonset/statefulset , which is the part of PR  #2246 which chasling is working on it.
We have tested with her changes all the test cases are passing with the flying colors :)

chlung: From my understanding, the framework will delete every controller with the same name not individual call.  This is why we need to change the logic to allow that behavior.

lwander: Again, this feels like a bad practice (clicking "delete" on one server group will possibly delete others). Also if any of the names in the test change independently, it will start to break because it's no longer relying on this strange delete behavior?

chlung: I will agree with you.  In https://github.com/chlung/clouddriver/blob/2671bd95a6d063ab445fd17a1537b4bfa2460575/clouddriver-kubernetes/src/main/groovy/com/netflix/spinnaker/clouddriver/kubernetes/v1/deploy/ops/servergroup/DestroyKubernetesAtomicOperation.groovy#L45, it is not if-else for replicaset and replicationcontroller.  It goes through that each type of controller.

lwander: Can you feature-flag this behavior in clouddriver, and make sure that this test doesn't depend on it?

chlung: with `@test`

PR #2282, opened 2018-01-30
fix(test): Update validate_bom__config to support artifact storage

sbwsg:


dev/validate_bom__config.py
```diff
@@ -191,6 +191,47 @@ def add_config(cls, options, script):
...
+  def add_files_to_upload(cls, options, file_set):
+    """Implements interface."""
+    if options.artifact_gcs_credentials:
+      file_set.add(options.artifact_gcs_credentials)
+
+  @classmethod
+  def add_config(cls, options, script):
+    """Implements interface."""
+    script.append('hal -q --log=info config artifact gcs enable')
+    project = options.artifact_gcs_project or options.deploy_google_project
```
ewiseblatt: did artifact_gcce_project already exist? It isnt defined in this PR.

dev/validate_bom__config.py
```diff
@@ -257,6 +298,34 @@ def add_config(cls, options, script):
...
+  HELPERS = {
+    'gcs': GcsArtifactStorageConfiguratorHelper,
+  }
+
+  def init_argument_parser(self, parser):
+    """Implements interface."""
+    for helper in self.HELPERS.values():
+      helper.init_argument_parser(parser)
+
+  def validate_options(self, options):
```
ewiseblatt: HELPERS is a map but you are treating it as a list.
I presume it is a map for the same reason the storage service is a map.
However even so you are not using it that way. Other than init, you'll never iterate over all the values. It only works now because there is one element in the map -- adding another will break all the code here.

I'd propose either adding the configuration option, of which there is only one choice (and having the remaining code read out the option) or using a class constant as a place holder for that option, whose value is 'gcs' and using that symbol for the HELPERS lookup key. Otherwise get rid of the map and replace it with a variable or use the raw value and refactor later.

sbwsg: Ah I see.  I believe the intention is that multiple artifact configurators could all be enabled at once.  A test might need artifacts from multiple storage providers for a single scenario.  With that in mind I've converted the HELPERS map to a list and removed the `raise ValueError` from `GcsArtifactStorageConfiguratorHelper.validate_options()`, which should prevent an explosion from happening during configuration.

dev/validate_bom__config.py
```diff
@@ -257,6 +296,34 @@ def add_config(cls, options, script):
     script.append(hal)
 
 
+class ArtifactConfigurator(Configurator):
+  """Controls hal config artifact for Spinnaker artifact ."""
+
+  HELPERS = [
```
ewiseblatt: I guess this is ok since you have guards in the helpers themselves.

PR #2309, opened 2018-02-06
feat(dev): control halyard configuration when building images.

ewiseblatt:
@jtk54

### DIFF
--- a/dev/halyard_install_component.sh
+++ b/dev/halyard_install_component.sh
@@ -19,6 +19,7 @@
 set -e
 set -x
 
+RELEASE_TRACK=nightly
 HALYARD_INSTALL_PARAMS=
 
 function show_usage() {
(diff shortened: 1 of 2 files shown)


dev/halyard_install_component.sh
```diff
@@ -72,11 +88,15 @@ function process_args() {
             shift
             ;;
 
-        --halyard_repository)
-            HALYARD_INSTALL_PARAMS="$HALYARD_INSTALL_PARAMS --halyard-repository $1"
+        --spinnaker_registry)
```
jtk54: Only two of the new flags added to the `--help` seem to be implemented.

jtk54: Or do these get delegated to the python script?

ewiseblatt: They are passed to the install script at line 119

jtk54: Got it.

PR #2311, opened 2018-02-06
chore(buildtool): Refactored buildtool monitoring

ewiseblatt:
@jtk54 #1 of 10


dev/buildtool/base_metrics.py
```diff
@@ -235,128 +236,112 @@ def __init__(self, options):
...
+    gauge.set(value)
+    return gauge
 
-  def track(self, name, labels, func, *pos_args, **kwargs):
+  def track_call(self, name, labels, description, func, *pos_args, **kwargs):
     """Track number of active calls to the given function."""
-    self.gauge(name, labels).track(func, *pos_args, **kwargs)
+    family = self.__ensure_family(
+        MetricFamily.GAUGE, name, labels, description)
+    gauge = family.get(labels)
```
jtk54: These two `family` and `gauge|counter|timer` calls appear several places, maybe a helper function? Maybe one just for `gauge`?

dev/buildtool/base_metrics.py
```diff
@@ -435,15 +416,15 @@ def flush_final_metrics(self):
...
     if not self.options.monitoring_enabled:
       logging.warning('Monitoring disabled -- dont push incremental metrics.')
       return
 
-    with self.__mutex:
-      self._do_flush_updated_metrics(self.__updated_metrics)
+    with self.__update_mutex:
+      updated_metrics = self.__updated_metrics
       self.__updated_metrics = set([])
+    self._do_flush_updated_metrics(updated_metrics)
```
jtk54: Should this be inside the `__update_mutex`? Had it inside before.

dev/buildtool/errors.py
```diff
@@ -0,0 +1,105 @@
...
+    super(ResponseError, self).__init__(message, self.RUNTIME, cause=server)
+
+class UnexpectedError(BuildtoolError):
+  def __init__(self, message, cause=None):
+    super(UnexpectedError, self).__init__(message, self.INTERNAL,
+                                          cause=cause or 'unexpected')
+
+
+def maybe_log_exception(where, ex, action_msg='propagating exception'):
+  """Log the exception and stackdrace if it hasnt been logged already."""
```
jtk54: s/stackdrace/stacktrace

dev/buildtool/errors.py
```diff
@@ -0,0 +1,105 @@
...
+  if not hasattr(ex, 'loggedit'):
+    text = traceback.format_exc()
+    logging.error('"%s" caught exception\n%s', where, text)
+    ex.loggedit = True
+
+  logging.error('"%s" %s', where, action_msg)
```
jtk54: Does `logging` not support the bracket notation for string formatting? It's kinda funny - java does it the exact opposite way, regular string formatting requires the `%s`, but `java.logging` supports brackets...

ewiseblatt: python brackets in strings are resolved by ".format" method on the string. logging takes a string parameter, so I dont think you can use brackets. you could use brackets to pass the string parameter in so logging sees a single string arg. Style guides discourage that because it forces strings to be resovlved eagerly rather than lazily so you cant avoid it if you disable logging. That's not a concern here, but feels simpler to do that way.

jtk54: Makes sense just curious.

PR #2312, opened 2018-02-06
Adds graph of how the new build system works

ttomsu:


jtk54:
Sweet.

ewiseblatt:
This should clarify that this is the "build" phase of the process. The "publish" at the end is misleading because in the process publish happens before validation. The "publish" command in the tool is a publish, but in this diagram the "publish" needs to be qualified that this is the "unvalidated" bom.

In the "publish" phase of the process there will be another bom publish, which is the actual release (in halyard terms this is a halyard bom release vs halyard bom publish).

Other than that this looks good.

dev/build_and_publish.dot
```diff
@@ -0,0 +1,135 @@
...
+
+		publish_nightly_bom -> git_clone_from_BOM_hash_60;
+
+        subgraph cluster_61 {git_clone_from_BOM_hash_60
+            label = "For Each Service..."
+	        color=aquamarine3
+
+	        git_clone_from_BOM_hash_60 -> get_hal_configs;
+	        get_hal_configs -> hal_admin_publish_profile;
+	        hal_admin_publish_profile -> hal_admin_publish_profile_alias_copy;
```
ewiseblatt: I'm taking this copy out. Hal profiles are based on the component version, not the bom version. So this is in fact superfluous.

PR #2314, opened 2018-02-06
chore(buildtool): Refactored git support

ewiseblatt:
@jtk54 #3 of 10

### DIFF
--- a/dev/buildtool/git_support.py
+++ b/dev/buildtool/git_support.py
@@ -25,59 +25,105 @@ import re
 import tempfile
 import time
 
-from subprocess import CalledProcessError
-
 # pylint: disable=no-name-in-module
 # pylint: disable=import-error
 from distutils.version import LooseVersion
 import yaml
 
-from buildtool.util import (
+from buildtool import (
+    add_parser_argument,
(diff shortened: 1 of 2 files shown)


dev/buildtool/git_support.py
```diff
@@ -262,6 +309,9 @@ def make_list_from_result(response_text):
   def make(entry):
     """Create a new CommitMessage from an individual entry"""
     match = CommitMessage._MEDIUM_PRETTY_COMMIT_MATCHER.match(entry)
+    if match is None:
```
jtk54: We currently don't block merges with incorrectly formatted commit messages. Will this failure surface if someone doesn't follow the conventions?

ewiseblatt: No. This will fail if the structured format from git is different or the regex is wrong.
This parses out the author/data/content structure. The issue you are talking about is inside the content part of the structure, which is parsed elsewhere in determine_semver_implications (#437), which does a compliance check to see if it can interpret the message.

jtk54: SGTM.

Issue #2043, opened 2017-10-18
Issue in creating server groups in spinnaker


ramimmadi:
I am trying to do the codlabs given in spinnaker. following the steps given in _https://www.spinnaker.io/guides/tutorials/codelabs/gce-source-to-prod/_ but I am unabe to create the server group(create without template). It is getting loaded for infinite time. I could say I got stuck at that point. I did that multiple times to check if I am missig any step but I think I am not missing anything.

One more thing is that I am able to do ssh tunneling using Azure VMs and Local machines but not with Google cloud VMs. It is not opening beyond "Hello" page If I use Google cloud VMs to do "SSH tunneling". Please help me out to solve this. 

  Thank you

ezimanyi:
I've upgraded the version of spinnaker in the codelab image to spinnaker 1.5.4, which addresses the bug noted here.  Following the instructions in the codelab now works as expected.

PR #2343, opened 2018-02-13
chore(google): Fix gce codelab

ezimanyi:
-Update the codelab to use a more recent base image of spinnaker
-Change post-install configuration of Spinnaker to update new
post-Halyard config file locations
-Bump the versions of Jenkins and aptly we install
-Replace some instances of the deprecated gcloud command 'copy-files'
with 'scp'
-Improve cleanup of image data


ezimanyi:
@lwander

ezimanyi:
Mind taking a look, @ewiseblatt?

google/codelab/first_codelab_boot.sh
```diff
@@ -14,15 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-PYTHONPATH=/opt/spinnaker/pylib python /opt/spinnaker/install/codelab_config.py
-echo "debianRepository: http://$(hostname):9999/ trusty main" \
-    > /opt/spinnaker/config/rosco-local.yml
-chown spinnaker:spinnaker /opt/spinnaker/config/rosco-local.yml
+hal config ci jenkins enable
+echo admin | hal config ci jenkins master add CodelabJenkins --address http://localhost:5656 --username admin --password
```
ewiseblatt: I'm not sure about this. Might it be possible to use a unique password? The pattern in GCE is (or at least used to be) that passwords are written to the VM metadata as a label. I actually dont like that practice, but that seems to be it and is better than a hardcoded default. Since you are configuring hal here with the password, you can also give hal that password so spinnaker knows. The user would just need to check for the password in the metadata. Or to the file where the password is. Jenkins has a file where it puts the password. I forget what that is offhand. You could do that (which I'd recommend for starters). IN summary:
   1) Put a unique password in the place jenkins wants it if not already there.
       be sure it is chmod 400 and owned by the jenkins user.
   2) feed the contents of that file to hal
   3) Document a reminder to the user where they can learn what the password is. They will need root access.

google/codelab/first_codelab_boot.sh
```diff
@@ -14,15 +14,15 @@
...
+    > ~ubuntu/.hal/default/profiles/rosco-local.yml
+
+sudo -u ubuntu cat <<EOF > ~ubuntu/.hal/default/profiles/clouddriver-local.yml
 credentials:
   challengeDestructiveActionsEnvironments:
 EOF
-chown spinnaker:spinnaker /opt/spinnaker/config/clouddriver-local.yml
 
-/opt/spinnaker/install/first_google_boot.sh
+/var/spinnaker/startup/first_halyard_boot.sh
```
ewiseblatt: I'm not familiar with this. Where did /var come from?

google/codelab/provision_spinnaker_codelab_image.sh
```diff
@@ -20,18 +20,18 @@
 # specifically for creating a new (running) codelab image out of a fresh
 # spinnaker image.
 
+if ! hal --version; then
+  echo "Spinnaker is not installed. Look at http://www.spinnaker.io for installation instructions."
```
ewiseblatt: Do you mean "Spinnaker" or "halyard"?
Also s/Look at/Refer to/

google/codelab/provision_spinnaker_codelab_image.sh
```diff
@@ -52,22 +52,27 @@ apt-mark unhold $SPINNAKER_SUBSYSTEMS
 
 # acquire and configure jenkins
 apt-get install -y git
-wget http://pkg.jenkins-ci.org/debian/binary/jenkins_2.1_all.deb
+wget https://pkg.jenkins.io/debian-stable/binary/jenkins_2.89.3_all.deb
```
ewiseblatt: call this version out into a constant at top so it is more easily maintainable

google/codelab/create_codelab_image_wrapper.sh
```diff
@@ -0,0 +1,118 @@
+#!/bin/bash
```
ewiseblatt: Does this really need to have so much logic in it?
I thought the helper library was structured to make this easier. This looks like it is all boilerplate except you are uploading a couple custom files and executing a custom script against it.

Maybe because this is bash it isnt structured properly to let you inject that in. If you see a simple refactoring in the helper libraries that would let . you do that, then most of this logic can be eliminated which will be safer and easier to maintain.

google/codelab/first_codelab_boot.sh
ezimanyi: Post-halyard, the VM startup script is in `/var/spinnaker/startup/` rather than `/opt/spinnaker/install`; see #1486 for the PR that added the halyard startup script.  I'm just calling it from the pre-existing location here.

google/codelab/provision_spinnaker_codelab_image.sh
ezimanyi: Fixed

google/codelab/create_codelab_image_wrapper.sh
ezimanyi: I'll take a look to see if there's any way I can reduce the logic here.  For context, all this logic was directly in the Jenkins job prior to my change---I'm just checking in what was on the Jenkins job (with a few small changes) so that it's managed by source control.  I'll take a look to see if there's any reasonable way to clean this up.

google/codelab/provision_spinnaker_codelab_image.sh
ezimanyi: Done

google/codelab/first_codelab_boot.sh
ezimanyi: @ewiseblatt : Is the suggestion here that we generate a unique password each time that a user provisions an instance based on the image, or that we simply pick something better than 'admin' but that will be the same each time an instance is provisioned?  Given that is is a codelab image and not intended for production use, I want to make sure that we're not making it overly complex for a user to follow the instructions.

google/codelab/create_codelab_image_wrapper.sh
ezimanyi: I was able to clean up this logic somewhat by making use of some of the existing functions in the helper library (and pushing some others into that library).

ezimanyi:
@ewiseblat: I've requested some clarification on the first comment (regarding the password) and addressed all other passwords.  Mind taking a look again?  Thanks!

google/codelab/first_codelab_boot.sh
ezimanyi: I looked a bit more into options here, and it feels like this might be more effort that it's worth (though certainly not impossible).  (It's also worth noting that the admin/admin username/password is pre-existing in the codelab, I've just changed the code to set it up to use Halyard.)

On initial install Jenkins can generate a random password, but as currently set up we install Jenkins before we take an image, so it would be the same random password for all installs.

We could ask the user to supply a Jenkins password in instance metadata, or just generate a random one when spinning up the codelab image; there does not seem to be an easy way to change the password on a Jenkins instance via script (beyond generated the salted hash of the chosen password and inserting it into the Jenkins user's XML file).

I'm leaning towards deferring making this change in the current PR, which was intended just to unbreak the codelab (which is currently broken) but not make any further functional changes---but happy to discuss further.

ewiseblatt: Isnt the first_boot script run when the instance is first booted? That's where you'd generate a password and tell halyard about it.

google/codelab/first_codelab_boot.sh
ewiseblatt: ok

ewiseblatt:
also, do a manual inspection on the instance that comes out the other end and make sure it doesnt have superfluous artifacts left behind from the install.

google/codelab/first_codelab_boot.sh
ezimanyi: `first_codelab_boot.sh` is run when the instance is started, but by that time Jenkins has already been installed and configured with a password (before we generated the image).  So we'd need to splice a new password's hash into the existing Jenkins config file.

ezimanyi:
I ran Build_GceImageWithHalyard pointing at this branch. The Jenkins job completed without errors, and I verified that:
-The output image spinnaker-unlicensed-1-5-4 was properly created
-Starting a VM pointing at that image led to a working installation of spinnaker
-The contents of the disk on that VM were as expected (ie, properly cleaned of the data we're trying to clean)
-No superflous disks, images, or instances were left in the project; the expected output image was the only artifact that remained from the pipeline.

ezimanyi:
I've built a new codelab image with these changes and pushed it to the marketplace.  I am now able to follow the codelab instructions without any issues.  This now closes #2343 and #2043.

PR #2347, opened 2018-02-14
fix(google): Remove needless userdel from cleanup script

ezimanyi:
The disk we are cleaning is attached to a cleaner instance; running
userdel will only delete users on the cleaner instance not the users
on the mounted (but not running) disk we're trying to clean.


ezimanyi:
@lwander : Rolling back half of yesterday's change #2344---I had made this change to make the cleanup script more robust, but it's actually incorrect so rolling it back.  (It will in general be a no-op but could lead to strange behavior if we delete the active user from the cleaner instance.)

The main goal of the prior change (removing ssh keys from all accounts) is still fine; I'm not rolling that back here.

PR #2349, opened 2018-02-15
fix(c2d): rely on randomly generated bucket name

lwander:


skim1420:
@ewiseblatt ptal

lwander:
@skim1420 has the full context, but we ran into a case where a project had a period in it causing bucket creation to fail. I don't have a strong preference towards either a random bucket or a statically chosen one.

ewiseblatt:
If it is just the illegal name, we could fix the transformation. The
comment had explicit mentioned that '.' was valid so if it was wrong, that
was intentional based on misinformation.

My concerns are that the opaque buckets we originally used were hard to
trace back to spinnaker. This was a real problem for us in our testing when
we wanted to leave buckets around. Most of our testing no longer goes
through this script so that isnt a concern for us anymore. However
customers may have other sources of random bucket names leading to the same
effect. I'm also concerned that it might be hard to find the bucket if they
need to reassociate it (e.g. migrating from the C2D project to a production
one, or some other reason to migrate projects). The project name as we have
it does have a risk of collision with existing bucket, but that seems
unlikely given we do add "spinnaker" into the name and I dont think we've
encountered that problem in practice to date.

I trust your judgement since you both have more customer facing
interactions with a more relevant perspective. As long as you are aware of
these concerns, I'm fine with whatever tradeoffs you prioritize.

skim1420:
Good points. I'd prefer to converge the bucket naming/validation thing to fewer places. I also felt not re-using the bucket names better supports the notion of either having multiple spinnaker instances in a GCP project, or when spinnaker instances are deleted then recreated (i.e. bucket should also be a new creation), as a default.

I think the use cases you bring are valid, but feel they should be deliberately handled (documentation, tooling..) rather than bleeding into our provisioning paths. I'll merge for now, thanks.

ewiseblatt:
Normally you would specify the bucket and have all that control. This is
specific to C2D. So it does make sense to have multiple spinnaker instances
(perhaps more so than a "normal" use case). However I still think the names
should be tied. We could satisfy both by adding a unique phrase into the
name (e.g. project name + timestamp). Then every instance would be unique,
but you'd still be able to trace it back to spinnaker (though not whose or
which). If the timestamp isnt hashed then a sort would tell you which are
old vs which are new in terms of creation (not use).

PR #2352, opened 2018-02-16
feat(buildtool): artifact janitor

ewiseblatt:
@jtk54 
Here's the bom anaysis underlying the janitor.

It's notably more complex than the image janitor in that it also verifies the integrity of the BOMs as well as uncovers redundant builds. The latter is so I can see the impact of the build streamlining changes and audit that the processes are working  as expected. The bits that look for unused artifacts is of similar complexity to the existing image janitor.

This PR just emits the analysis and lists of things to delete. There is no logic that consumes that to delete anything. That will follow once this goes in. in the meantime, those files could feed simple shell scripts

### DIFF
--- a/dev/buildtool/inspection_commands.py
+++ b/dev/buildtool/inspection_commands.py
@@ -12,100 +12,1189 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Implements commands for buildtool."""
+"""Implements inspection commands for buildtool.
 
+   1) buildtool.sh collect_bom_versions
+   2) buildtool.sh collect_artifact_versions
+   3) buildtool.sh audit_artifact_versions
+
+   This will produce files of things to prune.
+   The should be reviewed. Those remove those that you wish to keep.
+   Then to remove each of the artifacts:
+
+      for url in $(cat prune_jars.txt); do
+         curl -s -u$BINTRAY_USER:$BINTRAY_KEY -X DELETE $url &
+      done
+      wait
+
+      for url in $(cat prune_debians.txt); do
+         curl -s -u$BINTRAY_USER:$BINTRAY_KEY -X DELETE $url &
+      done
+      wait
+
+      for url in $(cat prune_containers.txt); do
+         gcloud -q container images delete $url --force-delete-tags &
+      done
+      wait
+
+      for image_name in $(cat prune_images.txt); do
+         gcloud -q compute images --project $PROJECT delete $image_name &
+      done
+
+      for url in $(cat prune_boms.txt); do
+         gsutil rm $url
+      done
+"""
+
+from threading import current_thread
+from multiprocessing.pool import ThreadPool
+
+import base64
+import json
 import logging
-import time
+import os
+import re
 import urllib2
+import yaml
 
 from buildtool import (
-    RepositoryCommandProcessor,
-    RepositoryCommandFactory,
-    BomSourceCodeManager,
-    GitRepositorySpec,
+    CommandFactory,
+    CommandProcessor,
+    SemanticVersion,
+    check_options_set,
+    check_path_exists,
+    check_subprocess,
+    maybe_log_exception,
     raise_and_log_error,
+    write_to_path,
+    ConfigError,
+    UnexpectedError,
     ResponseError)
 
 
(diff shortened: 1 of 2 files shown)


dev/buildtool/inspection_commands.py
```diff
@@ -12,100 +12,1154 @@
...
+          continue
+      missing = self.difference(versions, jar_map.get(key))
+      if missing:
+        missing_jars[key] = missing
+
+    return missing_jars
+
+  def find_missing_debian_versions(self, jar_map, debian_map):
+    missing_debians = {}
+    for package, versions in jar_map.items():
```
jtk54: Seems like we may be interested in debians listed in BOMs that are missing as well.

dev/buildtool/inspection_commands.py
```diff
@@ -12,100 +12,1154 @@
...
+        bad_list.append(text)
+        self.__invalid_versions[name] = bad_list
+        logging.error('Ignoring invalid %s version "%s": %s', name, text, ex)
+    return sorted(sem_vers, cmp=SemanticVersion.compare)[-1].to_version()
+
+  def determine_prunings(self):
+    def filter_not_version(version, version_list):
+      return [value for value in version_list
+              if not value.startswith(version)
+              or value < version + '-20180201']
```
jtk54: Datestring should be parameterized here, right?

dev/buildtool/inspection_commands.py
```diff
@@ -12,100 +12,1154 @@
...
+          (self.__unused_gce_images, self.__prune_gce_images),
+          (self.__unused_containers, self.__prune_containers)]:
+        unused_list = unused_map.get(name, None)
+        if unused_list is None:
+          unused_list = unused_map.get('spinnaker-' + name, [])
+
+        newest_version = self.most_recent_version(name, unused_list)
+        candidates = filter_not_version(newest_version, unused_list)
+
+        # We're going to keep malformed versions. These are rare so
```
jtk54: :+1:

dev/buildtool/inspection_commands.py
```diff
@@ -12,100 +12,1154 @@
...
+    add_if_nonstandard('debianRepository', self.__expect_debian_repository)
+    if len(info) > 2:
+      self.__non_standard_boms.append(info)
+
+    return info
+
+  def analyze_bom(self, bom):
+    """Analyzes one bom and breaks it down into this threads result_map.
+
+    Boms are processed within a single thread, but multiple boms can be
```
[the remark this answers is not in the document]
ewiseblatt: I do check the file name matches the version and report that among the "invalid boms". This does not seem to happen in practice because halyard writes it. It could happen if you deposit the boms directly in the bucket. But again, it would get picked up elsewhere.

dev/buildtool/inspection_commands.py
```diff
@@ -12,100 +12,1154 @@
...
+    return (package_name, content['versions'])
+
+  def difference(self, versions, target):
+    missing = []
+    for version in versions:
+      if not version in target:
+        missing.append(version)
+    return missing
+
+  def find_missing_jar_versions(self, jar_map, debian_map):
```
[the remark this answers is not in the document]
ewiseblatt: We dont execute jars. We dont really need to know the jar is missing, but I thought it would be informative in determining integrity and deciding what can be deleted. If some services "expect" jars and they are not present, then those services might be desired to discard. The jars arent part of the runtime, but could be part of client/interoperational compatability where they are used in other service builds, meaning you would not be able to build against that version.

dev/buildtool/inspection_commands.py
ewiseblatt: Yea, but not yet. Actually this wasnt supposed to survive. It was more explicit. There are some "18" boms during a transition period, so < 2018 is going to prune those. None of those were released to my knowledge so we can prune them out after a bit more time and add the option.

dev/buildtool/inspection_commands.py
ewiseblatt: This is just an artifact collection, independent of boms.
The two get tied together in the AuditArtifactVersions which this feeds.
The audit will feed all the boms into it. This command will produce an additional table saying "by the way, that debian was missing a  corresponding jar".

PR #2353, opened 2018-02-18
chore(buildtool): Be efficient at doing nothing

ewiseblatt:
@jtk54 
There are three optimizations here.

First is a change to build_bom that, when refreshing an existing bom,
will only refresh the bom if the repository commits have changed.
This makes it transparent to maintain a bom in conjunction with the git
repos (the primitive command requires you explicitly set a parameter to
do this, however the flow script sets the parameter by default unless you
tell it not to, such as for a new branch).

Second, given that boms are easily maintained with source changes, the
"flow_build" script starts a build using the bom from the previous build.
This also adds an interim local bom for "unbuilt" artifacts so that failed
builds can continue where they left off. The next build will still pick up
on any repository changes (e.g. a fix to the build failure).

Third, given that this makes it common for most repositories to offer no
work, the RepositoryCommands check the remote repo for changes before even
cloning it. While the clone isnt that expensive, this also cuts down on the
semver analysis and all the debugging logging noise.

### DIFF
--- a/dev/buildtool/halyard_commands.py
+++ b/dev/buildtool/halyard_commands.py
@@ -162,7 +162,11 @@ class BuildHalyardCommand(GradleCommandProcessor):
 
   def find_commit_version_entry(self, repository):
     logging.debug('Looking for existing halyard version for this commit.')
-    commit_id = self.git.query_local_repository_commit_id(repository.git_dir)
+    if os.path.exists(repository.git_dir):
+      commit_id = self.git.query_local_repository_commit_id(repository.git_dir)
+    else:
+      commit_id = self.git.query_remote_repository_commit_id(
+          repository.origin, self.options.git_branch)
     commits = self.load_halyard_version_commits().split('\n')
     commits.reverse()
     postfix = ' ' + commit_id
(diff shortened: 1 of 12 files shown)


dev/buildtool/git_support.py
```diff
@@ -152,6 +152,22 @@ def make(tag):
     return SemanticVersion(match.group(1),
                            *[int(num) for num in match.groups()[1:]])
 
+  @staticmethod
+  def compare(semver, other):
```
jtk54: Another candidate for de-duping once the dust has settled.

dev/buildtool/halyard_commands.py
```diff
@@ -181,8 +184,15 @@ def _do_repository(self, repository):
...
         self.__emit_last_commit_entry(entry)
-        return
+        return True
+    return False
 
+  def _do_repository(self, repository):
+    """Implements RepositoryCommandProcessor interface."""
+    # We need the version number here, which means we need the source
+    # so this is not in the can_skip method. However if we dont know
+    # about the commit, we probbaly dont have the debian anyway. And
```
jtk54: s/probbaly/probably

PR #2458, opened 2018-02-23
feat(citest/pubsub): Adds GCS -> pubsub -> GAE citest.

jtk54:
Adds the citest and an agent for interacting with GCS with
pub/sub notifications configured, and observing triggered
pipelines that execute from the GCS events.

This assumes quite a lot of external bucket, topic, and Echo config
that isn't in place yet. The test parameters are all there to work
when supplied, however, and I have tested this quite a lot given
my GCS -> pub/sub configuration.

The point of this WIP is to see where I went off the reservation 
and bent the normal citest workflows for my use cases, and
how we can improve either citest or this implementation to
achieve the same goal.

### DIFF
--- a/dev/all_tests.yaml
+++ b/dev/all_tests.yaml
@@ -218,6 +218,22 @@ tests:
       test_gcs_bucket: $storage_gcs_bucket
       branch: master
 
+  # gcs_pubsub_gae_test:
+  #   requires:
+  #     configuration:
+  #       appengine_account_enabled: true
+  #     services: [clouddriver, front50, gate, orca, echo]
+  #   quota:
+  #     # TODO: Loadbalancer?
+  #     appengine_deployment: 1
+  #   path: citest/tests/gcs_pubsub_gae_test
+  #   api: gate
+  #   args:
+  #     alias: [standard_appengine_provider_params]
+  #     git_repo_url: https://github.com/GoogleCloudPlatform/python-docs-samples.git
+  #     app_directory_root: appengine/standard/hello_world
+  #     test_gcs_bucket: $storage_gcs_bucket
+  #     branch: master
 
   dcos_smoke_test:
     requires:
--- a/testing/citest/requirements.txt
+++ b/testing/citest/requirements.txt
@@ -15,3 +15,5 @@ pyopenssl
 
 # OpenStack client needed for openstack_smoke_test
 python-openstackclient
+
+google-cloud-storage>=1.7
--- a/testing/citest/tests/gcs_pubsub_gae_test.py
+++ b/testing/citest/tests/gcs_pubsub_gae_test.py
@@ -1,0 +1,465 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# This test currently expects some preconfiguration to set up a GCS bucket,
+# a Google pub/sub topic and subscription, and notifications enabled on the
+# GCS bucket pushing to the pub/sub topic.
+#
+# The full setup and config instructions are listed in the following sections:
+# * https://www.spinnaker.io/guides/tutorials/codelabs/pubsub-to-appengine/#create-a-gcs-bucket-to-store-artifacts
+# * https://www.spinnaker.io/guides/tutorials/codelabs/pubsub-to-appengine/#set-up-google-cloud-pubsub-to-listen-to-bucket-object-changes
+#
+# Summarized here:
+# * Create bucket: gsutil mb -p <project> <bucket_name>
+# * Create topic and notification channel: gsutil notification create -t <topic> -f json <bucket_name>
+# * Create subscription: gcloud beta pubsub subscriptions create <subscription> --topic <topic>
+#
+# In addition to the external config, this test expects a pub/sub subscription and artifacts configured
+# in Spinnaker.
+#
+# Invoke this test by executing this python command:
+#
+# PYTHONPATH=$CITEST_ROOT \
+# python spinnaker/testing/citest/tests/gcs_pubsub_gae_test.py
+# --native_hostname
+# --native_port
+# --git_repo_url <app to deploy>
+# --app_directory_root <root of app>
+# --test_storage_account_name <configured artifact storage account>
+# --appengine_primary_managed_project_id <managed gce project>
+# --appengine_credentials_path <path to service account credentials>
+# --test_subscription_name <configured subscription name>
+# --test_gcs_bucket <bucket notifying pub/sub topic>
+# --spinnaker_appengine_account
+# --test_app
+# --test_stack
+#
+# pylint: disable=bad-continuation
+# pylint: disable=invalid-name
+# pylint: disable=missing-docstring
+
+# Standard python modules.
+import json
+import logging
+import os
+import subprocess
+import shutil
+import sys
+import tempfile
+import time
+
+# citest modules.
+import citest.base
+import citest.gcp_testing as gcp
+import citest.json_contract as jc
+import citest.json_predicate as jp
+import citest.service_testing as st
+
+# Spinnaker modules.
+import spinnaker_testing as sk
+import spinnaker_testing.gate as gate
+import spinnaker_testing.frigga as frigga
+
+ov_factory = jc.ObservationPredicateFactory()
+
+class GcsPubsubGaeTestScenario(sk.SpinnakerTestScenario):
+  """
+  Scenario for testing GAE deploys of GCS artifacts via pub/sub triggers.
+  """
+  @classmethod
+  def new_agent(cls, bindings):
+    return gate.new_agent(bindings)
+
+  @classmethod
+  def initArgumentParser(cls, parser, defaults=None):
+    """Initialize command line argument parser.
+
+    Args:
+    parser: argparse.ArgumentParser
+    """
+    super(GcsPubsubGaeTestScenario, cls).initArgumentParser(
+      parser, defaults=defaults)
+
+    defaults = defaults or {}
+    parser.add_argument(
+      '--app_directory_root', default=None,
+      help='Path from the root of source code repository to the application directory.')
+    parser.add_argument(
+      '--branch', default='master',
+      help='Git branch to be used when deploying from source code repository.')
+    parser.add_argument(
+      '--git_repo_url', default=None,
+      help='URL of a git source code repository used by Spinnaker to deploy to App Engine.')
+    parser.add_argument(
+      '--test_gcs_bucket', default=None,
+      help='GCS bucket to upload GAE app source code to.')
+    parser.add_argument(
+      '--test_storage_account_name', default=None,
+      help='Storage account when testing GCS buckets.'
+      ' If not specified, use the application default credentials.')
+    parser.add_argument(
+      '--test_subscription_name', default=None,
+      help='Google pub/sub subscription name configured in Echo.')
+
+  def __init__(self, bindings, agent=None):
+    super(GcsPubsubGaeTestScenario, self).__init__(bindings, agent)
+    self.logger = logging.getLogger(__name__)
+
+    bindings = self.bindings
+
+    if not bindings['GIT_REPO_URL']:
+      raise ValueError('Must supply value for --git_repo_url')
+
+    if not bindings['APP_DIRECTORY_ROOT']:
+      raise ValueError('Must supply value for --app_directory_root')
+
+    # We'll call out the app name because it is widely used
+    # because it scopes the context of our activities.
+    self.TEST_APP = bindings['TEST_APP']
+    self.TEST_STACK = bindings['TEST_STACK']
+    self.__EXPECTED_ARTIFACT_ID = 'deployable-gae-app-artifact'
+
+    self.__gcp_project = bindings['APPENGINE_PRIMARY_MANAGED_PROJECT_ID']
+    self.__cluster_name = frigga.Naming.cluster(self.TEST_APP, self.TEST_STACK)
+    self.__server_group_name = frigga.Naming.server_group(self.TEST_APP, self.TEST_STACK)
+    self.__lb_name = self.__cluster_name
+
+    self.__subscription_name = bindings['TEST_SUBSCRIPTION_NAME']
+    self.__gcs_pubsub_agent = sk.GcsFileUploadAgent(bindings['APPENGINE_CREDENTIALS_PATH'])
+
+    # Python is clearly hard-coded as the runtime here, but we're just asking App Engine to be a static file server.
+    self.__app_yaml = ('\n'.join(['runtime: python27',
+                                  'api_version: 1',
+                                  'threadsafe: true',
+                                  'service: {service}',
+                                  'handlers:',
+                                  ' - url: /.*',
+                                  '   static_dir: .']).format(service=self.__lb_name))
+
+    self.__app_directory_root = bindings['APP_DIRECTORY_ROOT']
+    self.__branch = bindings['BRANCH']
+
+    self.pipeline_id = None
+
+    self.bucket = bindings['TEST_GCS_BUCKET']
+    self.__test_repository_url = 'gs://' + self.bucket
+
+
+  def create_app(self):
+    # Not testing create_app, since the operation is well tested elsewhere.
+    # Retryable to handle platform flakiness.
+    contract = jc.Contract()
+    return st.OperationContract(
+      self.agent.make_create_app_operation(
+        bindings=self.bindings,
+        application=self.TEST_APP,
+        account_name=self.bindings['SPINNAKER_APPENGINE_ACCOUNT']),
+      contract=contract)
+
+  def delete_app(self):
+    # Not testing delete_app, since the operation is well tested elsewhere.
+    # Retryable to handle platform flakiness.
+    contract = jc.Contract()
+    return st.OperationContract(
+      self.agent.make_delete_app_operation(
+        application=self.TEST_APP,
+        account_name=self.bindings['SPINNAKER_APPENGINE_ACCOUNT']),
+      contract=contract)
+
+  def make_deploy_stage(self):
+    return {
+      'clusters': [
+        {
+          'account': self.bindings['SPINNAKER_APPENGINE_ACCOUNT'],
+          'application': self.TEST_APP,
+          'cloudProvider': 'appengine',
+          'configFilepaths': [],
+          'configFiles': [self.__app_yaml],
+          'expectedArtifactId': self.__EXPECTED_ARTIFACT_ID,
+          'fromArtifact': True,
+          'gitCredentialType': 'NONE',
+          'interestingHealthProviderNames': [
+            'App Engine Service'
+          ],
+          'provider': 'appengine',
+          'region': self.bindings['TEST_GCE_REGION'],
+          'sourceType': 'gcs',
+          'stack': self.TEST_STACK,
+          'storageAccountName': self.bindings.get('TEST_STORAGE_ACCOUNT_NAME')
+        }
+      ],
+      'name': 'Deploy',
+      'refId': 'DEPLOY',
+      'requisiteStageRefIds': [],
+      'type': 'deploy'
+    }
+
+  def make_pubsub_trigger(self):
+    return {
+      'attributeConstraints': {'eventType': 'OBJECT_FINALIZE'},
+      'constraints': {},
+      'enabled': True,
+      'expectedArtifactIds': [
+        self.__EXPECTED_ARTIFACT_ID
+      ],
+      'payloadConstraints': {},
+      'pubsubSystem': 'google',
+      'subscriptionName': self.__subscription_name, # Logical name assigned in Echo.
+      'type': 'pubsub'
+    }
+
+  def make_expected_artifact(self):
+    return {
+      'defaultArtifact': {
+        'kind': 'custom'
+      },
+      'id': self.__EXPECTED_ARTIFACT_ID,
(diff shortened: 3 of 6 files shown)


jtk54:
FYI we caught a bug in 1.6.0 developing this :)

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+    builder = st.HttpContractBuilder(self.agent)
+    (builder.new_clause_builder('Has Pipeline',
+                                retryable_for_secs=5)
+     .get_url_path(
+       'applications/{app}/pipelineConfigs'.format(app=self.TEST_APP))
+     .excludes_path_value('name', pipeline_id))
+
+    return st.OperationContract(
+      self.new_delete_operation(
+        title='delete_bake_pipeline', data=payload, path=path,
```
ewiseblatt: Is this really synchronous?
Also you call it "bake_pipeline" but it is a deploy pipeline.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+     )
+    )
+
+    executions_path = 'executions?pipelineConfigIds={}&limit=1&statuses=SUCCEEDED'.format(self.__pipeline_id)
+    return st.OperationContract(
+      self.__gcs_pubsub_agent.new_gcs_pubsub_trigger_operation(
+        gate_agent=self.agent,
+        title='monitor_gcs_pubsub_pipeline',
+        bucket_name=self.bucket,
+        upload_path='{name}'.format(name=name),
```
ewiseblatt: why is this quoted but not others?

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+  def test_a_create_app(self):
+    self.run_test_case(self.scenario.create_app(),
+                       retry_interval_secs=8, max_retries=8)
+
+  def test_b_create_pipeline(self):
+    self.run_test_case(self.scenario.create_deploy_pipeline(),
+                       retry_interval_secs=8, max_retries=8)
+
+  def test_d_run_pipeline(self):
+    time.sleep(60)
```
ewiseblatt: I'm not comfortable with this. At least log before and after the sleep so the time delay in the logs is accounted for.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+  def test_x_delete_load_balancer(self):
+    self.run_test_case(self.scenario.delete_load_balancer(),
+                       retry_interval_secs=8, max_retries=8)
+
+  def test_y_delete_pipeline(self):
+    self.run_test_case(
+      self.scenario.delete_deploy_pipeline(self.scenario.pipeline_id))
+
+  def test_z_delete_app(self):
+    time.sleep(60) # Wait for cache update.
```
ewiseblatt: we never do this in any other test and they all delete apps except for the kato tests. What's different here?

dev/all_tests.yaml
```diff
@@ -218,6 +218,21 @@ tests:
       test_gcs_bucket: $storage_gcs_bucket
       branch: master
 
+  gcs_pubsub_gae_test:
+    requires:
+      configuration:
+        appengine_account_enabled: true
+      services: [clouddriver, front50, gate, orca, echo]
+    quota:
```
ewiseblatt: i think you need to account for the LBs here too.
Note that this is only used from the validation controller scripts, not the test itself, so you might want the test to verify the quota is available if this is sensitive.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+    logging.info('Fetching %s', git_repo)
+    subprocess.Popen(command, stderr=sys.stderr, shell=True).wait()
+
+
+  @classmethod
+  def tearDownClass(cls):
+    runner = citest.base.TestRunner.global_runner()
+    scenario = runner.get_shared_data(GcsPubsubGaeTestScenario)
+    bindings = scenario.bindings
+    shutil.rmtree(scenario.temp)
```
ewiseblatt: where did scenario.temp come from?

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+    branch = bindings['BRANCH']
+    git_repo = bindings['GIT_REPO_URL']
+
+    gcs_path = 'gs://{bucket}'.format(bucket=scenario.bucket)
+    topic = '{}-topic'.format(bindings['TEST_APP'])
+    subscription = '{}-subscription'.format(bindings['TEST_APP'])
+
+    # App to tar and upload to GCS.
+    command = 'git clone {repo} -b {branch} {dir}'.format(
+      repo=git_repo, branch=branch, dir=scenario.temp)
```
ewiseblatt: what is scenario.temp?

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+      raise ValueError('Must supply value for --git_repo_url')
+
+    if not bindings['APP_DIRECTORY_ROOT']:
+      raise ValueError('Must supply value for --app_directory_root')
+
+    # We'll call out the app name because it is widely used
+    # because it scopes the context of our activities.
+    self.TEST_APP = bindings['TEST_APP']
+    self.TEST_STACK = bindings['TEST_STACK']
+    self.EXPECTED_ARTIFACT_ID = 'deployable-gae-app-artifact'
```
ewiseblatt: Is this arbitrary? If so you might want to tie it to this app so it is easier to trace, can have mulitple tests running, and arent affected by leaks in failed or broken tests.

testing/citest/requirements.txt
```diff
@@ -15,3 +15,5 @@ pyopenssl
 
 # OpenStack client needed for openstack_smoke_test
 python-openstackclient
+
+google-cloud-storage==1.7.0
```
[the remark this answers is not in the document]
jtk54: Moving to be min version.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -0,0 +1,153 @@
...
+      self.__client = storage.Client.from_service_account_json(credentials_path)
+    else:
+      self.__client = storage.Client()
+
+    # Allow up to 13 minutes to wait on operations.
+    # 13 minutes is arbitrary. The current test takes around 6-7 minutes
+    # end-to-end. Other use cases might make it more clear what this should be.
+    self.default_max_wait_secs = 780
+
+  def upload_file(self, bucket_name, upload_path, local_filename):
```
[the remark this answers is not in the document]
jtk54: This class interfaces with GCS, which then indirectly interacts with pub/sub via a preconfigured notification listener on the bucket. There wasn't support in the citest gcs agent for uploading objects, and the discovery doc only specified the name of the `Object` class for an `/insert` without any attributes for us to roll our own.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -0,0 +1,153 @@
...
+    super(BaseGcsPubsubOperation, self).__init__(title, gcs_pubsub_agent, max_wait_secs=max_wait_secs)
+    if not gcs_pubsub_agent or not isinstance(gcs_pubsub_agent, GcsPubsubAgent):
+      raise TypeError('agent is not a GcsPubsubAgent: ' + gcs_pubsub_agent.__class__.__name__)
+
+  def export_to_json_snapshot(self, snapshot, entity):
+    snapshot.edge_builder.make_mechanism(entity, 'Gcs Pubsub Agent', self.agent)
+    super(BaseGcsPubsubOperation, self).export_to_json_snapshot(snapshot, entity)
+
+  def execute(self, agent=None, trace=True):
+    if not self.agent:
```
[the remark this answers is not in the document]
jtk54: Removing, this seems unnecessary.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -0,0 +1,153 @@
...
+  def finished(self):
+    resp = self.__trigger_response
+    executions = json.JSONDecoder().decode(resp.output) # Executions list
+    print executions
+    return resp and resp.http_code == 200 and executions and len(executions) == 1
+
+  @property
+  def finished_ok(self):
+    resp = self.__trigger_response
+    return resp and resp.http_code
```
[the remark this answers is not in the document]
jtk54: That would be helpful.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -0,0 +1,153 @@
...
+    return self.__class__.__name__
+
+  @property
+  def detail(self):
+    return 'agent {} doing a GET at path {} with response {}'.format(self.__gate_agent, self.__status_path, self.__trigger_response)
+
+  @property
+  def timed_out(self):
+    resp = self.__trigger_response
+    return resp and resp.http_code == 504
```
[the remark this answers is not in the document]
jtk54: Placeholder, I need to figure out what the timeout response looks like.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    self.__gate_agent = gate_agent
+    operation.bind_agent(gate_agent)
+    self.__status_class = status_class
+    self.__status_path = status_path
+    super(GcsPubsubTriggerOperationStatus, self).__init__(operation)
+    self.__trigger_response = self.__gate_agent.get(self.__status_path)
+
+  def export_to_json_snapshot(self, snapshot, entity):
+    super(GcsPubsubTriggerOperationStatus, self).export_to_json_snapshot(
```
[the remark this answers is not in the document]
jtk54: Done.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -0,0 +1,153 @@
...
+    resp = self.__trigger_response
+    return resp and resp.http_code
+
+  @property
+  def id(self):
+    return self.__class__.__name__
+
+  @property
+  def detail(self):
+    return 'agent {} doing a GET at path {} with response {}'.format(self.__gate_agent, self.__status_path, self.__trigger_response)
```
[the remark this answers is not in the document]
jtk54: Done.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+
+    defaults = defaults or {}
+    parser.add_argument(
+      '--app_directory_root', default=None,
+      help='Path from the root of source code repository to the application directory.')
+    parser.add_argument(
+      '--branch', default='master',
+      help='Git branch to be used when deploying from source code repository.')
+    parser.add_argument(
+      '--git_repo_url', default=None,
```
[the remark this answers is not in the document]
jtk54: I followed the pattern of `appengine_smoke_test` and used the same app there. Should I add that app as a default to this arg or hard code it in the test?

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+          'configFilepaths': [],
+          'configFiles': [self.__app_yaml],
+          'expectedArtifactId': self.EXPECTED_ARTIFACT_ID,
+          'fromArtifact': True,
+          'gitCredentialType': 'NONE',
+          'interestingHealthProviderNames': [
+            'App Engine Service'
+          ],
+          'provider': 'appengine',
+          'region': 'us-central',
```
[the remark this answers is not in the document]
jtk54: Done.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+
+    # Need to query Gate for the id of the pipeline we just created...
+    def create_pipeline_id_extractor(_ignored, context):
+      pipeline_config_resp = self.agent.get(pipeline_config_path)
+      pipeline_config_list = json.JSONDecoder().decode(pipeline_config_resp.output)
+      found = next(x for x in pipeline_config_list if x['name'] == self.pipeline_id)
+      context['pipelineId'] = found['id'] # I don't know how to reference this later, so I'm saving it in self for now.
+      self.__pipeline_id = found['id'] # I don't know how to reference this later, so I'm saving it in self for now.
+      logging.info('Created pipeline config with id: %s', context['pipelineId'])
+      print('Created pipeline config with id: {}'.format(context['pipelineId']))
```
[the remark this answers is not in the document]
jtk54: Done.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+     .EXPECT(
+       jp.LIST_MATCHES([self.make_dict_matcher(pipeline_spec)])))
+
+    # Need to query Gate for the id of the pipeline we just created...
+    def create_pipeline_id_extractor(_ignored, context):
+      pipeline_config_resp = self.agent.get(pipeline_config_path)
+      pipeline_config_list = json.JSONDecoder().decode(pipeline_config_resp.output)
+      found = next(x for x in pipeline_config_list if x['name'] == self.pipeline_id)
+      context['pipelineId'] = found['id'] # I don't know how to reference this later, so I'm saving it in self for now.
+      self.__pipeline_id = found['id'] # I don't know how to reference this later, so I'm saving it in self for now.
```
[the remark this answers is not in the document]
jtk54: I don't know how to reference the `pipelineId` key in the context after I use this hook to populate it. I get around that by just setting an attribute on `self`. Which pattern is preferable?

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+      logging.info('Created pipeline config with id: %s', context['pipelineId'])
+      print('Created pipeline config with id: {}'.format(context['pipelineId']))
+
+    return st.OperationContract(
+      self.new_post_operation(
+        title='create_gcs_gae_pubsub_pipeline', data=payload, path='pipelines',
+        status_class=st.SynchronousHttpOperationStatus),
```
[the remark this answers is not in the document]
jtk54: Yes.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+        version='v000')
+
+    # Triggered pipeline does a deploy, check for that server group.
+    server_group_path = 'applications/{app}/serverGroups'.format(app=self.TEST_APP)
+    builder = st.HttpContractBuilder(self.agent)
+    (builder.new_clause_builder('GAE Deploy Pipeline Succeeded',
+                                retryable_for_secs=300)
+     .get_url_path(server_group_path)
+     .EXPECT(
+       ov_factory.value_list_path_contains(
```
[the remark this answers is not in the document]
jtk54: I'll try your suggestion.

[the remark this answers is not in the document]
jtk54: Worked great.

testing/citest/tests/gcs_pubsub_gae_test.py
jtk54: Yes AFAIK.

jtk54: Fixed naming.

testing/citest/tests/gcs_pubsub_gae_test.py
jtk54: Made consistent.

testing/citest/tests/gcs_pubsub_gae_test.py
jtk54: I agree, this needs solved more generally -- we need to be able to check that Echo's pipeline cache is synchronized to the pipeline writes, either via metrics inspection or exposing an endpoint. Right now this is a hack because we know roughly what the cache interval is.

testing/citest/tests/gcs_pubsub_gae_test.py
jtk54: Removed.

dev/all_tests.yaml
jtk54: I don't know what quota needs accounted for, I'll add a TODO and ask Dan.

testing/citest/tests/gcs_pubsub_gae_test.py
jtk54: It's a temp directory to clone the test appengine app -- we use it in multiple steps of the test so it needs to live the duration of the scenario.

testing/citest/tests/gcs_pubsub_gae_test.py
jtk54: It's arbitrary, but baked into the pipeline only which is scoped to the app. I can parameterize it if you want.

testing/citest/tests/gcs_pubsub_gae_test.py
ewiseblatt: Why is it removed here? Ownership of the directory seems off. What if multiple fixtures are accessing it?

dev/all_tests.yaml
ewiseblatt: The project has quota. The test controller is allocated a certain amount of quota to work with and manages that quota at a test script level using this config. So if the test is creating load balancers, and there are project quotas, then this test needs to specify how many to allocate it when managing script concurrency.

testing/citest/spinnaker_testing/gcs_agent.py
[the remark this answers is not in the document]
ewiseblatt: could you add the IO support into GcsAgent then use it from there?

dev/all_tests.yaml
ewiseblatt: Is pub sub always available or does it require explicit configuration? If it is opt in then you should have something in the configuration section that tests for it so that the test will be skipped if the deployment configuration is not applicable. Your test itself should test for it if it can.

This config file is tied to the validate_bom script and the parameters here in the configuration are the configuration parameters to the deploy step. The test deployed the servers so it knows how they were configured. This file ties the deployment configuration parameters into the testing configuration. So you can test for the configuration here to specify what's needed.

testing/citest/spinnaker_testing/gcs_agent.py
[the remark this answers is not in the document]
jtk54: Discussed out of band - leaving this functionality out of the citest GcsAgent for now but will rename to something more specific.

testing/citest/tests/gcs_pubsub_gae_test.py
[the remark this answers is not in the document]
jtk54: Discussed out of band - I need to reference the `pipelineId` in a different test case, so I need to save it in the `scenario` not the `context`, which is scoped only to a single test case.

testing/citest/tests/gcs_pubsub_gae_test.py
jtk54: The temp directory is created and managed in this fixture only.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -111,21 +105,20 @@ class GcsPubsubTriggerOperationStatus(base_agent.AgentOperationStatus):
   def finished(self):
     resp = self.__trigger_response
     executions = json.JSONDecoder().decode(resp.output) # Executions list
-    print executions
-    return resp and resp.http_code == 200 and executions and len(executions) == 1
+    return resp and resp.http_code and resp.http_code == 200 and executions and len(executions) == 1
```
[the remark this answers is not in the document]
jtk54: Done.

testing/citest/spinnaker_testing/gcs_agent.py
```diff
@@ -111,21 +105,20 @@ class GcsPubsubTriggerOperationStatus(base_agent.AgentOperationStatus):
...
 
   @property
   def finished_ok(self):
     resp = self.__trigger_response
-    return resp and resp.http_code
+    return resp and resp.http_code and resp.http_code == 200
```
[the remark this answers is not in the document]
jtk54: Done.

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,453 @@
...
+                                  'api_version: 1',
+                                  'threadsafe: true',
+                                  'service: {service}',
+                                  'handlers:',
+                                  ' - url: /.*',
+                                  '   static_dir: .']).format(service=self.__lb_name))
+
+    self.__app_directory_root = bindings['APP_DIRECTORY_ROOT']
+    self.__branch = bindings['BRANCH']
+    self.temp = tempfile.mkdtemp()
```
ewiseblatt: You create this directory here in the scenario constructor, but then delete it in the fixture tear down. That's violating ownership. The scenario can span fixtures. It doesnt now, but doing this makes it less maintainable.

Either you should create the directory in the fixture setUpClass method and inject it into the scenario, or you should register some kind of global shutdown function that cleans it up (e.g. atexit)

jtk54:
Agreed, I moved it into the setupClass method and injected it.

jtk54:
I think I addressed everything we discussed @ewiseblatt, can you take one more pass and see if I missed anything?

testing/citest/tests/gcs_pubsub_gae_test.py
```diff
@@ -0,0 +1,449 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Invoke this test by executing this python command:
+#
+# PYTHONPATH=$CITEST_ROOT \
+# python spinnaker/testing/citest/tests/gcs_pubsub_gae_test.py
+# --native_hostname
+# --native_port
```
ewiseblatt: Add preconfig steps and remove parameters that arent defaults, adding default values where possible if they arent there already. Dont assume external values (e.g. not the name of the subscription, but teh app/stack). Ideally there is a public repo you can run this on, such as the one we use for appengine smoke test if this can be the same.

jtk54: Added.

PR #2542, opened 2018-03-15
fix(c2d): Install kubectl if it's not already installed

dparrish:
The ubuntu 14.04 image that the click-to-deploy Spinnaker uses doesn't seem to have `kubectl` installed by default, causing the first boot to fail if Kubernetes is enabled. This change installs the latest stable `kubectl` binary if none is found locally, before attempting to do any Kubernetes configuration.


skim1420:
kubectl not required with v1 provider. Revisiting click-to-deploy startup scripts compatibility with RBAC. Closing this in the meanwhile.

PR #2547, opened 2018-03-16
fix(validation): Validation more resilient to bintray unreliability

ewiseblatt:
This also removes the hardcoded ubuntu 14.04 assupmtions.
@jtk54

### DIFF
--- a/dev/validate_bom__deploy.py
+++ b/dev/validate_bom__deploy.py
@@ -129,6 +129,11 @@ class BaseValidateBomDeployer(object):
     """The metrics regisry bound at construction."""
     return self.__metrics
 
+  @property
+  def hal_user(self):
+    """Returns the Halyard User within the deployment VM."""
+    return self.__hal_user
+
   def __init__(self, options, metrics, runtime_class=None):
     if runtime_class:
       self.__spinnaker_deployer = runtime_class(options, metrics)
@@ -136,6 +141,8 @@ class BaseValidateBomDeployer(object):
       self.__spinnaker_deployer = self
     self.__options = options
     self.__metrics = metrics
+    self.__hal_user = options.deploy_hal_user
+    logging.info('hal_user="%s"', self.__hal_user)
 
   def make_port_forward_command(self, service, local_port, remote_port):
     """Return the command used to forward ports to the given service.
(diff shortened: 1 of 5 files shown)


ewiseblatt:
@lwander

dev/validate_bom__deploy.py
```diff
@@ -1230,17 +1278,17 @@ def init_argument_parser(parser, defaults):
   # pylint: disable=line-too-long
   add_parser_argument(
       parser, 'halyard_install_script', defaults,
-      'https://raw.githubusercontent.com/spinnaker/halyard/master/install/nightly/InstallHalyard.sh',
+      'https://raw.githubusercontent.com/spinnaker/halyard/master/install/debian/InstallHalyard.sh',
```
lwander: This script defaults to "stable", since it is the replacement for the stable install script. Maybe the `halyard_version` parameter above should do the same now?

ewiseblatt: Not sure I follow. The halyard_version is left empty by default and only passed in if set. So by default it wont specify anything, which will pick up the default in the script, which is 'stable'.
The intent here is to use the default, whatever it may happen to be, unless explicitly specified.

lwander: I'm just pointing out this is a change in behavior -- the default (using the script that was deleted in this diff) was 'nightly'. Now it's 'stable'.

ewiseblatt: ok. that's actually good. in practice we were injecting stable by default in the processes anyway.

dev/validate_bom__deploy.py
```diff
@@ -499,19 +508,12 @@ def ssh_key_path(self, path):
...
-
   def __init__(self, options, metrics, **kwargs):
     super(GenericVmValidateBomDeployer, self).__init__(
         options, metrics, **kwargs)
     self.__instance_ip = None
-    self.__hal_user = options.deploy_hal_user
-    logging.info('hal_user="%s"', self.__hal_user)
     self.__ssh_key_path = os.path.join(os.environ['HOME'], '.ssh',
-                                       '{0}_empty_key'.format(self.__hal_user))
+                                       '{0}_empty_key'.format(self.hal_user))
```
jtk54: Nit: I don't think you need the positional zero.

ewiseblatt:
I broke this out into two separate prs

PR #2557, opened 2018-03-21
fix(validate): Tolerate some bintray unreliability.

ewiseblatt:
@jtk54

### DIFF
--- a/dev/validate_bom__deploy.py
+++ b/dev/validate_bom__deploy.py
@@ -22,6 +22,7 @@ from multiprocessing.pool import ThreadPool
 import json
 import logging
 import os
+import shutil
 import stat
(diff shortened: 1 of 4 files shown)


dev/validate_bom__deploy.py
```diff
@@ -600,30 +634,30 @@ def do_deploy(self, script, files_to_upload):
           ExecutionError('Caught %s provisioning vm' % ex.message,
                          program='provisionVm'))
     finally:
+      import shutil
```
jtk54: Any reason import is here instead of the top?

ewiseblatt:
One last look after fixing a bug with the recovery loop.  Also, I added instrumentation for common install errors.

PR #2563, opened 2018-03-21
chore(citest/pubsub): Add validation configurators for pub/sub test.

jtk54:
Also some small changes to the test and agent, nothing that changes behavior drastically.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -43,10 +43,16 @@ interface, which is provided via free functions.
 """
 
 
+import logging
 import os
-from buildtool import add_parser_argument
+
+from buildtool import add_parser_argument, check_options_set, check_path_exists, raise_and_log_error, ConfigError
 
 from validate_bom__deploy import write_data_to_secure_path
+
+from google.cloud import pubsub
+from google.cloud import storage
+from google.oauth2 import service_account
 
 
 class Configurator(object):
@@ -70,6 +76,14 @@ class Configurator(object):
 
   def add_config(self, options, script):
     """Writes bash commands (e.g. hal) to configure feature set."""
+    pass
+
+  def setup_environment(self, options):
+    """Configures external platform infrastructure."""
+    pass
+
+  def teardown_environment(self, options):
+    """Tears down external platform infrastructure created in setup_environment()."""
     pass
 
   def add_files_to_upload(self, options, file_set):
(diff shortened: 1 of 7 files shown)


dev/validate_bom__config.py
```diff
@@ -899,6 +913,179 @@ def add_files_to_upload(self, options, file_set):
...
+    add_parser_argument(
+        parser, 'pubsub_google_name', defaults, None,
+        help='The logical name of the configured subscription in Spinnaker.')
+    add_parser_argument(
+        parser, 'pubsub_google_template_path', defaults, None,
+        help='Path to the Jinja message translation template.')
+
+  def validate_options(self, options):
+    """Implements interface."""
+    if (options.pubsub_google_project is None
```
ewiseblatt: Rather than is None check not to fail empty values as well.

dev/validate_bom__config.py
```diff
@@ -899,6 +913,179 @@ def add_files_to_upload(self, options, file_set):
...
+    subscription = options.pubsub_google_subscription_name or None
+    template_path = os.path.basename(options.pubsub_google_template_path) or None
+
+    if name is None:
+      return
+
+    script.append('hal -q --log=info config pubsub google enable')
+    script.append('hal -q --log=info config pubsub google subscription'
+                  ' add {name}'
+                  ' --project {project}'
```
ewiseblatt: I dont think the None values will work here -- will they?
I think instead you want to add the parameters to the call if they are specified.

dev/validate_bom__config.py
```diff
@@ -899,6 +913,179 @@ def add_files_to_upload(self, options, file_set):
...
+        parser, 'gcs_pubsub_topic', defaults, None,
+        help='The name of the topic to create.')
+    add_parser_argument(
+        parser, 'gcs_pubsub_subscription', defaults, None,
+        help='The name of the subscription to create.')
+
+  def validate_options(self, options):
+    """Implements interface."""
+    if (options.gcs_pubsub_bucket is None
+        or options.gcs_pubsub_topic is None
```
ewiseblatt: use the check_options_set

dev/validate_bom__config.py
```diff
@@ -899,6 +913,179 @@ def add_files_to_upload(self, options, file_set):
...
+                       ' --gcs_pubsub_topic="{}"'
+                       ' --gcs_pubsub_project="{}"'
+                       ' --gcs_pubsub_credentials_path="{}"'
+                       ' --gcs_pubsub_subscription="{}"'
+                       .format(options.gcs_pubsub_bucket,
+                               options.gcs_pubsub_topic,
+                               options.gcs_pubsub_project,
+                               options.gcs_pubsub_credentials_path,
+                               options.gcs_pubsub_subscription))
+    if options.gcs_pubsub_subscription != options.pubsub_google_subscription_name:
```
ewiseblatt: instead of raise ValueError call
raise_and_log_error(ConfigError(...))

The raise_and_log_error (from buildtool) is instrumented. The ConfigError will better categorize the source of problem to be configuration related ( a form of user error) as opposed to something else (a testing bug uncovered)

If I missed other errors thrown, this raise_and_log_error woudl apply there too. See the errors in buildtool/errors.py as the preferred choices. You can still use ValueError but categorizing them for trend analysis is more useful than precision since the errors are not actually handled.

dev/validate_bom__config.py
```diff
@@ -899,6 +913,179 @@ def add_files_to_upload(self, options, file_set):
...
+    notification = bucket.notification(options.gcs_pubsub_topic, topic_project=options.gcs_pubsub_project,
+                                       payload_format=storage.notification.JSON_API_V1_PAYLOAD_FORMAT)
+    notification.create()
+    logging.debug('Created bucket notification %s', notification)
+
+  def teardown_environment(self, options):
+    """Implements interface."""
+    publisher_client, subscriber_client, storage_client = self.__instantiate_clients(options)
+
+    build_number = os.environ['BUILD_NUMBER']
```
ewiseblatt: why BUILD_NUMBER? I dont see this used elsewhere. I'd prefer it be a parameter if you need one rather than environment variable because this should be easily run by people in shells, not just a CI server.

dev/validate_bom__config.py
```diff
@@ -1194,6 +1383,20 @@ def make_scripts(options):
   return init_script, config_script
 
 
+def setup_environments(options):
```
ewiseblatt: should this be singular "setup_environment"?
same for teardown below

dev/validate_bom__main.py
```diff
@@ -163,6 +165,7 @@ def main(options, metrics):
       deployer.collect_logs()
     if options.deploy_undeploy:
       deployer.undeploy()
+      validate_bom__config.teardown_environments(options)
```
ewiseblatt: should there be a try/finally so the teardown happens even if undeploy fails? I'm not sure, but suspect it should. Manually undeploying isnt so bad but manually tearing down might be, so better to make a stronger effort.

testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
```diff
@@ -121,8 +121,14 @@ def detail(self):
 
   @property
   def timed_out(self):
+    ping = datetime.datetime.utcnow()
+    diff = ping - self.__start
+    return diff > datetime.timedelta(minutes=1)
```
ewiseblatt: Please make this a constant up in the class.
Better yet would be a property on the agent that you coudl initialize as 1, even if you dont provide a setter at this time.

testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
```diff
@@ -121,8 +121,14 @@ def detail(self):
...
+
+  def __check_finished_ok(self):
     resp = self.__trigger_response
-    return resp and resp.http_code == 504
+    executions = json.JSONDecoder().decode(resp.output) # Executions list
+    return resp.ok() and executions and len(executions) == 1
```
ewiseblatt: executions wont be None (right?) so len(executions) == 1 is sufficient

dev/validate_bom__config.py
jtk54: Done.

dev/validate_bom__config.py
jtk54: Done.

dev/validate_bom__config.py
jtk54: Done.

dev/validate_bom__config.py
jtk54: Done.

dev/validate_bom__config.py
jtk54: Unused, forgot to remove in a refactor.

dev/validate_bom__config.py
jtk54: Done.

dev/validate_bom__main.py
jtk54: Done.

testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
jtk54: Done.

testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
jtk54: Done.

jtk54:
Re-running once more to verify I didn't break anything responding to comments.

dev/validate_bom__config.py
```diff
@@ -899,6 +913,149 @@ def add_files_to_upload(self, options, file_set):
...
+  def validate_options(self, options):
+    """Implements interface."""
+    check_options_set(options, ['pubsub_google_project', 'pubsub_google_credentials_path', 'pubsub_google_subscription_name', 'pubsub_google_template_path'])
+    check_path_exists(options.pubsub_google_credentials_path, "pubsub_credentials_path")
+    check_path_exists(options.pubsub_google_template_path, "pubsub_template_path")
+    options.pubsub_google_enabled = options.pubsub_google_subscription_name is not None
+
+  def add_config(self, options, script):
+    """Implements interface."""
+    name = options.pubsub_google_name or None
```
ewiseblatt: These variables arent helping you.

if not options.pubsub_google_name:
   return

subscription_cmd.append('add ' + options.pubsub_google_name)
if options.pubsub_google_project:
 subscription_cmd.append('--project ' + options.pubsub_google_project)
...
script.append('hal -q ..')
script.append(' '.join(subscription_cmd))

testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
```diff
@@ -121,8 +121,22 @@ def detail(self):
...
+    return self.__timeout
+
+  @timeout.setter
+  def timeout(self, val):
+    self.__timeout = val
+
+  def __check_finished_ok(self):
     resp = self.__trigger_response
-    return resp and resp.http_code == 504
+    executions = json.JSONDecoder().decode(resp.output) # Executions list
```
ewiseblatt: If resp is not ok, is there resp.output?
If not then checks need to be separate.

testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
```diff
@@ -121,8 +121,22 @@ def detail(self):
...
+    ping = datetime.datetime.utcnow()
+    diff = ping - self.__start
+    return diff > self.timeout
+
+  @property
+  def timeout(self):
+    return self.__timeout
+
+  @timeout.setter
+  def timeout(self, val):
```
ewiseblatt: it isnt clear this is a timedelta. I'd normally expect seconds.

PR #2564, opened 2018-03-22
fix(validate): only configure/run pub sub tests when pubsub is enabled

ewiseblatt:
@jtk54 

There are also a bunch of formating changes here to bring it mostly to 80 columns and use 4 space indents for continuing lines (which is the style used throughout).

The main gist here is to not require pubsub be configured and gate the pubsub test on pubsub being configured as is the pattern for all the other configurators.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -46,7 +46,12 @@ interface, which is provided via free functions.
 import logging
 import os
 
-from buildtool import add_parser_argument, check_options_set, check_path_exists, raise_and_log_error, ConfigError
+from buildtool import (
+    add_parser_argument,
+    check_options_set,
+    check_path_exists,
+    raise_and_log_error,
+    ConfigError)
 
 from validate_bom__deploy import write_data_to_secure_path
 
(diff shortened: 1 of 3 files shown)


ewiseblatt:
@jtk54 also, the pubsub setup doenst work in my account.
I think because the validation script runs with a different service account that does not have pubsub credentials in it. The job only seems to account for giving spinnaker a service account with pubsub credentials. However the test controller runs from a different service account, it is needs the credentials to setup pub sub. This PR doesnt fix that issue.

dev/validate_bom__config.py
```diff
@@ -936,38 +941,54 @@ def init_argument_parser(self, parser, defaults):
...
+    if not options.pubsub_google_enabled:
       return
 
     script.append('hal -q --log=info config pubsub google enable')
     subscription_cmd = ['hal -q --log=info config pubsub google subscription']
     if options.pubsub_google_name:
       subscription_cmd.append('add {}'.format(options.pubsub_google_name))
     if options.pubsub_google_project:
-      subscription_cmd.append('--project {}'.format(options.pubsub_google_project))
+      subscription_cmd.append('--project ' + options.pubsub_google_project)
```
jtk54: On a second look at this, I think all of these are required, so if any are missing any then we should fail. Does `check_options_set` fail if any of the arguments are `None`? If so I think we can assume without guarding each of these with an `if`.

jtk54: Discussed out of band, LGTM.

PR #2575, opened 2018-03-24
feat(buildtool): Write changelog gists

ewiseblatt:
This also changes publishing spinnaker to use changelog gists.
There is a process change here where the publish directly writes
the changelog entry to spinnaker.github.io, but takes the existing
gist URI as a parameter as discussed offline. This gist is assumed
to already be curated so no need for additional review step.

The push_changelog_to_gist introduced here is a new process step
where the changelog is uploaded to a separated volatile gist
(control outside this script) which is used offline while curating
the official changelog gist.

@jtk54 
@lwander

### DIFF
--- a/dev/buildtool/base_metrics.py
+++ b/dev/buildtool/base_metrics.py
@@ -353,6 +353,7 @@ class BaseMetricsRegistry(object):
       except Exception as ex:
         logging.exception('label_func failed with %s', ex.message)
         raise ex
+      raise
     finally:
       timer = self.get_metric(MetricFamily.TIMER, name, outcome_labels)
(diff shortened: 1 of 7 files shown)


dev/buildtool/base_metrics.py
```diff
@@ -353,6 +353,7 @@ def time_call(self, name, labels, label_func,
       except Exception as ex:
         logging.exception('label_func failed with %s', ex.message)
         raise ex
+      raise
```
jtk54: Intentional?

ewiseblatt: yea, this was eating exceptions before

PR #2653, opened 2018-04-12
chore(validate): Deploy kayenta

ewiseblatt:
@duftler 
First pass is only deploying on GCE for google with stackdriver.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -918,6 +918,77 @@ class DockerConfigurator(Configurator):
       file_set.add(options.docker_account_credentials)
 
 
+class CanaryConfigurator(Configurator):
+  """Controls hal config canary configuration."""
+
+  def init_argument_parser(self, parser, defaults):
+    """Implements interface."""
+    add_parser_argument(
+        parser, 'canary_aws', defaults, False, type=bool,
+        help='Whether or not to canary aws account deployments.'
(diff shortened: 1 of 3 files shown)


dev/validate_bom__config.py
```diff
@@ -918,6 +918,59 @@ def add_files_to_upload(self, options, file_set):
...
+             ' If so, this will use the aws account, project, credentials.')
+
+  def validate_options(self, options):
+    """Implements interface."""
+    options.canary_enabled = options.canary_google or options.canary_aws
+
+    if not options.canary_enabled:
+      return
+
+    #check_options_set( options, [])
```
jtk54: Seems like this should be implemented.

dev/validate_bom__config.py
```diff
@@ -918,6 +918,59 @@ def add_files_to_upload(self, options, file_set):
...
+      if options.deploy_hal_platform == 'gce':
+        script.append('hal -q --log=info config canary google edit'
+                      ' --stackdriver-enabled true')
+      script.append('hal -q --log=info config canary google enable')
+      
+    script.append('hal -q --log=info config canary enable')
+
+  def add_files_to_upload(self, options, file_set):
+    """Implements interface."""
+    if not options.canary_enabled:
```
jtk54: Is this a stub for later or it superfluous?

PR #2677, opened 2018-04-19
chore(validate): Modernized halyard validation restart

ewiseblatt:
Halyard no longer uses upstart.

@lwander 
I dont like any of this PR from an API perspective.

I think hal should have startup/shutdown commands for the server.
But the critical omission is being able to supply credentials in the config file.
That should be there along with a complement to the --config-bucket parameter that I'm using to change the bucket. If you can change the bucket then it stands to reason you might need to change [provide] credentials so the two should be managed together. If I could set the credentials to begin with, then I wouldnt need to stop and start the server again for this use case (nor would I need to hack to startup script to inject them as I am doing here).

This should work in the short term if it is easier to add the credential scopes. However if you can add explicit credential configuration support then I can replace most of the changes in this PR to use that mechanism instead.

### DIFF
--- a/dev/validate_bom__deploy.py
+++ b/dev/validate_bom__deploy.py
@@ -181,13 +181,12 @@ class BaseValidateBomDeployer(object):
     logging.info('Deploying with hal on %s...', platform)
     script = list(init_script)
     self.add_install_hal_script_statements(script)
-    self.add_platform_deploy_script_statements(script)
     if self.options.halyard_config_bucket_credentials:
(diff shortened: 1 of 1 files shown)


lwander:
When you say "supply credentials in the config file", which config file are you referring to, and which credentials?

ewiseblatt:
I mean a json path in the halyard.yml file.

ewiseblatt:
ping

PR #2802, opened 2018-05-18
 feat(spinbot): label issues when requested

lwander:
Depends on prior spinbot PR

### DIFF
--- a/spinbot/event/label_issue_comment_event_handler.py
+++ b/spinbot/event/label_issue_comment_event_handler.py
@@ -1,0 +1,26 @@
+from gh import AddLabel
+from .handler import Handler
+from .command import GetCommands
+from .issue_event import GetIssue
+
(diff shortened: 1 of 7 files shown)


jtk54:
The bot has spoken.

spinbot/event/label_issue_comment_event_handler.py
```diff
@@ -0,0 +1,26 @@
+from gh import AddLabel
+from .handler import Handler
+from .command import GetCommands
+from .issue_event import GetIssue
+
+class LabelIssueCommentEventHandler(Handler):
+    def __init__(self):
+        super().__init__()
```
jtk54: Is `super()` python3 style? I'm not familiar with that that does.

lwander: Yup, it just calls the parent class's implementation of the requested method.

PR #2804, opened 2018-05-18
fix(spinbot): read commit messages, not PR titles

lwander:


spinbot/event/release_branch_pull_request_handler.py
```diff
@@ -17,17 +17,22 @@ def handles(self, event):
...
+        if pull_request is None:
+            self.logging.warn('Unable to determine PR that created {}'.format(event))
+            return
+
+        commits = pull_request.get_commits()
+        bad_commits = []
+
+        for commit in commits:
+            message = ParseCommitMessage(commit.commit.message)
+            if message is None or message.get('type') != 'fix':
```
ewiseblatt: Is this saying release branches can only have fixes?
Is there an issue during pre-release when cherry-picking in new features, chores, etc?

lwander: Even during pre-release we say "fixes" only. Maybe we can include chore? What's the set of types that doesn't bump the minor version number?

jtk54: Let's look back and expand the set of acceptable tags to what we've cherry picked in the past.

ewiseblatt: the semver rules are here https://github.com/spinnaker/spinnaker/blob/master/dev/buildtool/git_support.py#L290

The available types were based on repository usage in practice. I think it is either subject to change depending on what people need to express, or you might enforce the types here. Ideally via a yaml file that the above can import so that the list is maintained in only one place. If you do that, I'd restrict down the scope of types. There's redundancy in that list because that's what was being used.

ewiseblatt: LGTM as it stands now, especially since this is a soft warning. If it were enforced then you'd need an escape hatch for cherry picking anything into pre-release branches.

lwander: Ah good point -- I will add a config option to this event handler to omit repos

lwander: Added config option to set the allowed types as well as the repos to scan in this PR

lwander: Repos to avoid scanning * 

The running bot now skips PRs against spinnaker/spinnaker in release branches only.

PR #2816, opened 2018-05-22
feat(spinbot): collect repo-level stats

lwander:
image


spinbot/gh/client.py
```diff
@@ -40,8 +40,14 @@ def get_label(self, repo, name, create=True):
 
         return label
 
+    def repos(self):
```
jtk54: `@property`?

lwander: I've avoided that for now... will do a pass to introduce where necessary

PR #2902, opened 2018-06-13
test(kube_v2_smoke): Tests basic deploy & delete manifest

lwander:
More will be added to this test, this just starts it off


testing/citest/tests/kube_v2_smoke_test.py
```diff
@@ -0,0 +1,267 @@
+# Copyright 2016 Google Inc. All Rights Reserved.
+#
```
ewiseblatt: 2018

testing/citest/spinnaker_testing/kubernetes_v2_scenario_support.py
```diff
@@ -0,0 +1,59 @@
...
+      defaults: [dict] Default binding value overrides.
+         This is used to initialize the default commandline parameters.
+    """
+
+    #
+    # Operation Parameters
+    #
+    builder.add_argument(
+        '--spinnaker_kubernetes_v2_account',
+        default=defaults.get('SPINNAKER_KUBERNETES_V2_ACCOUNT', None),
```
ewiseblatt: Do you want this to be v2 account, especially long term?
In practice these are run in separate processes so names wouldnt conflict with other tests.

testing/citest/tests/kube_v2_smoke_test.py
```diff
@@ -0,0 +1,267 @@
...
+    self.TEST_APP = bindings['TEST_APP']
+
+    # Take just the first if there are multiple
+    # because some uses below assume just one.
+    self.TEST_NAMESPACE = bindings['TEST_NAMESPACE'].split(',')[0]
+    self.pipeline_id = None
+
+    # We will deploy two images. One with a tag that we want to find,
+    # and another that we don't want to find.
+    self.__image_registry = 'gcr.io'
```
ewiseblatt: I dont think these should be hardcoded. Hardcoding parameter defaults is ok.
One should be able to run this test from custom build configurations.

testing/citest/tests/kube_v2_smoke_test.py
```diff
@@ -0,0 +1,267 @@
...
+        description='Deploy manifest',
+        application=self.TEST_APP)
+
+    builder = kube.KubeContractBuilder(self.kube_v2_observer)
+    (builder.new_clause_builder('Deployment created',
+                                retryable_for_secs=15)
+     .get_resources(
+         'deploy',
+         extra_args=[name, '--namespace', self.TEST_NAMESPACE])
+     .contains_path_eq('spec/replicas', 1))
```
ewiseblatt: ok, but you could EXPECT a match if you want to look at more stuff.
See EXPECT(ov_factory.value_list_contains(jp.DICT_MATCHES in other tests.

testing/citest/tests/kube_v2_smoke_test.py
```diff
@@ -0,0 +1,267 @@
...
+        application=self.TEST_APP,
+        description='Destroy Manifest')
+
+    builder = kube.KubeContractBuilder(self.kube_v2_observer)
+    (builder.new_clause_builder('Replica Set Removed')
+     .get_resources(
+         'deployment',
+         extra_args=[name, '--namespace', self.TEST_NAMESPACE],
+         no_resource_ok=True)
+     .contains_path_eq('targetSize', 0))
```
ewiseblatt: Should this accomodate 404 errors as well?
See the error handling in the kube smoke test.

ewiseblatt: I just noticed the no_resource_ok=True -- that is deprecated and I'm hoping to remove it in favor of the more generalized and standardizable response handling via EXPECT and ov_factory.error_list predicates.

lwander: Can you share an example usage?

testing/citest/spinnaker_testing/kubernetes_v2_scenario_support.py
lwander: This is testing v2 accounts only, so I do expect it to be a v2 account long term unless I'm misunderstanding.

lwander:
Yeah I'm holding off on adding this to the test runner until I make sure we've provisioned a v2 account.

testing/citest/tests/kube_v2_smoke_test.py
ewiseblatt: from the kube_smoke_test.py

      (builder.new_clause_builder('Daemonset Deleted', retryable_for_secs=15)
       .get_resources(
          'daemonsets', extra_args=[self.TEST_APP + self.POSTFIX_STATEFUL_DAEMONSET,
                        '--namespace', self.TEST_NAMESPACE])
       .EXPECT(ov_factory.value_list_path_contains('targetSize', jp.NUM_EQ(0)))
       .OR(self.__not_found_observation_predicate()))

where
ov_factory = jc.ObservationPredicateFactory()

and

  def __not_found_observation_predicate(self, title='Not Found Permitted'):
    return ov_factory.error_list_contains(
        st.CliAgentRunErrorPredicate(
            title=title, error_regex='.* not found'))

What this is doing is using explicit predicates on the observation (the ov_factory helper) rather than predicates on the observation value. The ov_factory helper provides convienence to say whether you mean on the observation "value list" to look for good results, or on the "error list" to look for bad results. Being explicit this way, lets you say you are expecting to observe a 404 error (though the agent doesnt give 404's so we regex the output text) or a result with a 0 size. The "no_resources_ok" is simpler syntax but more wishy washy, harder to implement internally, and doesnt generalize to other types of errors.

I'd like a simpler syntax but havent realized one yet. So for the meantime I've been using that pattern so if I do find a simpler syntax, it will be straightforward to refactor to it and for the time being, the consistency and familiarity should make it a little easier to read.

testing/citest/tests/kube_v2_smoke_test.py
ewiseblatt: Here's an example from google_smoke_test
https://github.com/spinnaker/spinnaker/blob/master/testing/citest/tests/google_smoke_test.py#L193

There it is taking the configuration it passed in, makes a tweak to it to add the name it expects it to create, then passes that as [a subset of] the expected observation in order to verify that the configuration was plumbed through spinnaker and into the actual platform resources created.

I think this is particularly important because mocked unit tests arent going to tell you that the actual API isnt behaving the way you thought and your resources are not configured the way you expected.

This test is sufficient to verify that it got created which is probably normally ok if you can take the config for granted. But since nothing else is testing the config, this would be a good place for it.

testing/citest/tests/kube_v2_smoke_test.py
lwander: Oh nice this seems like exactly what I needed -- thanks!

testing/citest/tests/kube_v2_smoke_test.py
lwander: This has nothing to do with a build environment, we are just picking an arbitrary image public image to deploy -- the actual choice has no bearing on the test

lwander:
Switched to use EXPECT style predicates

testing/citest/tests/kube_v2_smoke_test.py
```diff
@@ -0,0 +1,273 @@
...
+        description='Deploy manifest',
+        application=self.TEST_APP)
+
+    builder = kube.KubeContractBuilder(self.kube_v2_observer)
+    (builder.new_clause_builder('Deployment created',
+                                retryable_for_secs=15)
+     .get_resources(
+         'deploy',
+         extra_args=[name, '--namespace', self.TEST_NAMESPACE])
+     .EXPECT(ov_factory.value_list_contains(jp.DICT_MATCHES(
```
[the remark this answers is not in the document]
lwander: I'll add more here in the future, I plan to deploy/test a few more resources as well. This also removes all the deprecation warnings from the logs.

testing/citest/tests/kube_v2_smoke_test.py
```diff
@@ -0,0 +1,273 @@
...
+        }],
+        application=self.TEST_APP,
+        description='Destroy Manifest')
+
+    builder = kube.KubeContractBuilder(self.kube_v2_observer)
+    (builder.new_clause_builder('Replica Set Removed')
+     .get_resources(
+         'deployment',
+         extra_args=[name, '--namespace', self.TEST_NAMESPACE])
+     .EXPECT(self.__not_found_observation_predicate()))
```
ewiseblatt: I thought before you had a size of 0 here (or not found).
The two are different. I think you want both.

The difference is that you were using 0 size as confirmation that eventually the right thing would happen, willing to trust kubernetes from that point. Here you will actually wait for it to finish deleting (which might have flaky timeouts). It's a bit more rigorous but can take longer and doesnt give much value since we're willing to trust k8s. However other tests that run after this which want the delete to be finished can be assured that their preconditions are met and avoid retries on those operations.

As a rule of thumb (not necessarily right, but at this time I think is fine) I settle for the minimum conditions to give confidence that spinnaker is ok. Thus permit the 0 size. You still need the 404 because you might not see the 0 size transition and 404 is the final steady state. The 0 size is just an acknowledgement that eventually you can expect a 404, which is good enough.

lwander: The test of size 0 was incorrect, it was only passing because we had stated that a 404 was also OK. The resource never has a desired spec size of 0 after a delete is called.

lwander: To elaborate, in V1, where your example came from we did hit size 0 at some point during the delete process -- this doesn't happen for the deployment object.

lwander: I did, but I removed it since it was copied from the v1 test and not appropriate here. The size of 0 will never be reflected in that call, so it's not testing for anything valid. Kubernetes will not render that resource with a desired size of 0, so there is no transition we can even catch in some cases.

PR #2923, opened 2018-06-18
test(provider/kubernetes): more v2 smoke tests

lwander:


lwander:
Now the test:

creates a manifest
updates it
rolls it back
scales it
deletes it

PR #2954, opened 2018-06-26
feat(codelabs): add spin install

lwander:


codelabs/base/install/setup.sh
```diff
@@ -39,6 +39,18 @@ bold "Using bucket $BUCKET_URI..."
 
 gsutil mb $BUCKET_URI
 
+bold "Installing the Spinnaker CLI..."
```
jtk54: Spinnaker or Spin?

lwander: Good call, will say "spin cli"

PR #2956, opened 2018-06-26
test(provider/kubernetes): v2 artifact tests

lwander:
This is just the start, will test a few more flows in subsequent PRs

### DIFF
--- a/testing/citest/tests/kube_v2_artifact_test.py
+++ b/testing/citest/tests/kube_v2_artifact_test.py
@@ -1,0 +1,290 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


testing/citest/tests/kube_smoke_test.py
```diff
@@ -42,6 +42,7 @@
 
 # Standard python modules.
 import sys
+import random
```
jtk54: Unused?

testing/citest/tests/kube_v2_artifact_test.py
```diff
@@ -0,0 +1,290 @@
...
+    """
+    super(KubeV2ArtifactTestScenario, self).__init__(bindings, agent)
+    bindings = self.bindings
+
+    # We'll call out the app name because it is widely used
+    # because it scopes the context of our activities.
+    # pylint: disable=invalid-name
+    self.TEST_APP = bindings['TEST_APP']
+
+    # Take just the first if there are multiple
```
jtk54: Does it matter which namespace this deploys to? Are there problems when multiple tests are targeting the same namespace?

testing/citest/tests/kube_smoke_test.py
lwander: Whoops, yes

testing/citest/tests/kube_v2_artifact_test.py
lwander: Doesn't matter, we make sure all the names are unique enough

PR #2967, opened 2018-06-27
test(provider/kubernetes): test unversioned configmap

lwander:


testing/citest/tests/kube_v2_artifact_test.py
```diff
@@ -112,6 +112,45 @@ def __docker_image_artifact(self, name, image):
       'uuid': id_
     }
 
+  def deploy_unversioned_config_map(self, value):
+    """Creates OperationContract for deploying an unversioned configmap
+
+    To verify the operation, we just check that the configmap was created with
+    the correct 'value'.
+    """
+    bindings = self.bindings
```
sbwsg: minor nit: `bindings` is only used once in the code below so this appears unnecessary

lwander: :+1: will fix

PR #2985, opened 2018-06-29
fix(citest): Add queueing to gate Hystrix config

ezimanyi:
A number of the intermittent failures observed in running our CI tests are due to Gate returing a 500 on /tasks/task_id because Hystrix did not have an available thread to use to query Orca.  As this is primarily a problem with requests being bursty rather than an overall throughput problem, adding a modest queue to the Hystrix config so that Gate can handle and recover from short bursts of queries.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -45,7 +45,6 @@ interface, which is provided via free functions.
 
 import logging
 import os
-import textwrap
 
 from buildtool import (
     add_parser_argument,
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -1445,13 +1445,13 @@ class SpinnakerConfigurator(Configurator):
     script.append('mkdir -p  ~/.hal/default/profiles')
     script.append('echo "management.security.enabled: false"'
                   ' > ~/.hal/default/profiles/spinnaker-local.yml')
-    hystrix_config = textwrap.dedent('''\
-      hystrix:
-        threadpool:
-          default:
-            maxQueueSize: 20
-            queueSizeRejectionThreshold: 20
-    ''')
+    hystrix_config = '''\
+hystrix:
+  threadpool:
+    default:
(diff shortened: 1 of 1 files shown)


ezimanyi:
Testing now in Jenkins; will merge if things look good.

ajordens:
fwiw this is what we've been running with

```
threadpool:
    applications:
      coreSize: 25
      maxQueueSize: 100
    tasks:
      coreSize: 20
      maxQueueSize: 100
    builds:
      coreSize: 15
      maxQueueSize: 25
    serverGroups:
      coreSize: 75
      maxQueueSize: 100
    clusters:
      coreSize: 20
      maxQueueSize: 50
    loadBalancers:
      coreSize: 15
      maxQueueSize: 10
```

It hasn't been touched in awhile.

dev/validate_bom__config.py
```diff
@@ -1444,6 +1445,14 @@ def add_config(self, options, script):
     script.append('mkdir -p  ~/.hal/default/profiles')
     script.append('echo "management.security.enabled: false"'
                   ' > ~/.hal/default/profiles/spinnaker-local.yml')
+    hystrix_config = textwrap.dedent('''\
```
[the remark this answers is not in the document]
ezimanyi: Didn't realize this was possible in python; fixed.

dev/validate_bom__config.py
```diff
@@ -45,6 +45,7 @@
 
 import logging
 import os
+import textwrap
```
jtk54: Can remove this import.

ezimanyi: I mentally removed it; I'll remove it from the actual code now

ezimanyi:
@ajordens : Thanks!  Looks like I may have been even a bit too modest in bumping the parameters, but I think (based on my very basic load testing) this should be sufficient for our CI test.  But if we still run into issues, I'll bump up the thread pool and queue even more.

PR #2988, opened 2018-06-30
fix(citest): Fix two intermittently failing tests

ezimanyi:


ezimanyi:
* The `bake_and_deploy` test has been intermittently timing out; the logs show the pipeline is progressing, but just needs more time, so increase the timeout from 13 to 20 minutes.
* The `gcs_pubsub_gae` test throws a Python error when the first poll for the generated pipeline doesn't return anything.  This is because the `status_extractor` runs before we check the contract.  The fix is to have `status_extractor` silently do nothing if it doesn't find the expected pipeline.  (The contract check will then run and fail, causing another polling cycle or failing if the timeout is reached.)

testing/citest/spinnaker_testing/jenkins_agent.py
```diff
@@ -93,15 +93,15 @@ def export_to_json_snapshot(self, snapshot, entity):
 
 class JenkinsAgent(base_agent.BaseAgent):
   """A specialization of BaseAgent for interacting with Jenkins."""
-  def __init__(self, baseUrl, auth_path, owner_agent, logger=None):
+  def __init__(self, baseUrl, auth_path, owner_agent, logger=None, max_wait_secs=780):
     super(JenkinsAgent, self).__init__(logger=logger)
     self.__http_agent = http_agent.HttpAgent(baseUrl)
     self.__owner_agent = owner_agent
 
     # Allow up to 13 minutes to wait on operations.
```
lwander: Should the comment be moved to the parameter above?

ezimanyi: Good point; I tried moving it up and it looked a bit overly-prominent when above the whole function.  I ended up leaving it where it is, but re-worded it in a new commit so it's more clear that the default is set above.

PR #3000, opened 2018-07-04
fix(testing): Restructure BakeAndDeploy test reduces timeout flakiness.

ewiseblatt:
@jtk54


testing/citest/tests/bake_and_deploy_test.py
```diff
@@ -491,19 +522,44 @@ def create_bake_and_deploy_aws_pipeline(self):
     # FIXME(jacobkiefer): this is creating a gce deploy stage in an aws
     # pipeline. Not good.
     deploy_stage = self.make_deploy_google_stage(requisiteStages=['BAKE'])
+
```
jtk54: I don't know if my `FIXME` is actually relevant.

PR #3013, opened 2018-07-08
feat(buildtool): Can produce changelog relative to bom

ewiseblatt:
@jtk54

### DIFF
--- a/dev/buildtool/changelog_commands.py
+++ b/dev/buildtool/changelog_commands.py
@@ -23,6 +23,7 @@ import re
 import shutil
 import textwrap
 import urllib2
+import yaml
 
 
 from buildtool import (
@@ -40,6 +41,7 @@ from buildtool import (
     UnexpectedError,
     CommitMessage,
     GitRunner,
+    HalRunner,
 
     check_kwargs_empty,
     check_options_set,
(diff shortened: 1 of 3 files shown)


dev/buildtool/changelog_commands.py
```diff
@@ -321,12 +344,25 @@ def init_argparser(self, parser, defaults):
...
+
     self.add_argument(
         parser, 'include_changelog_details', defaults, False,
         action='store_true',
         help='Include a "details" section with the full commit messages'
              ' in time sequence in the changelog.')
 
+    HalRunner.add_parser_args(parser, defaults)
+    self.add_argument(
+        parser, 'relative_to_bom_path', defaults, None,
```
jtk54: This is a local file, correct? Can you specify that in the help?

dev/buildtool/changelog_commands.py
```diff
@@ -321,12 +344,25 @@ def init_argparser(self, parser, defaults):
...
+        help='If specified then produce the changelog relative to the'
+             ' commits found in the specified bom rather than the previous'
+             ' repository tag.')
+    self.add_argument(
+        parser, 'relative_to_bom_version', defaults, None,
+        help='If specified then produce the changelog relative to the'
```
jtk54: Same here but for the Halyard fetch.

dev/buildtool/git_support.py
```diff
@@ -815,20 +819,20 @@ def find_newest_tag_and_common_commit_from_id(
...
     return start_tag, start_commit
 
-
   def query_local_repository_commits_to_existing_tag_from_id(
-      self, git_dir, commit_id, commit_tags):
+      self, git_dir, commit_id, commit_tags, base_commit_id=None):
     """Returns the list of commit messages to the local repository."""
     # pylint: disable=invalid-name
 
     tag, found_commit = self.find_newest_tag_and_common_commit_from_id(
```
jtk54: Is this something expensive we can avoid if `base_commit_id` is specified?

ewiseblatt: I wanted to keep it simple and not break anything. The problem is that this method is also used to get the tag to use. The tag needs to analyze the absolute changelog even if the caller is only interested in the messages for the relative changelog.

This is offline processing and changelogs are probably small so performance isnt a concern here. It would be nice if things where different but isnt obvious how to restructure it and not worth the effort to pursue further at this time. So yes, this is messy and adds some complexity being overloaded. It might not add runtime cost because the work might be necessary regardless. It would still be desirable to reduce the maintainence / complexity costs but payoff doesnt seem worth it right now.

jtk54: SGTM.

ewiseblatt:
Oops, I missed the arlier comments somehow. I'll followup with another PR to address those requests.

PR #3045, opened 2018-07-13
test(provider/kubernetes): test manifest cache endpoint

lwander:
contains commits for https://github.com/spinnaker/spinnaker/pull/3044 (will rebase once the other test changes are merged).

### DIFF
--- a/testing/citest/tests/kube_v2_cache_test.py
+++ b/testing/citest/tests/kube_v2_cache_test.py
@@ -1,0 +1,255 @@
+# Copyright 2018 Google Inc. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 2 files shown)


testing/citest/tests/kube_v2_cache_test.py
```diff
@@ -0,0 +1,255 @@
...
+import citest.json_predicate as jp
+import citest.service_testing as st
+
+# Spinnaker modules.
+import spinnaker_testing as sk
+import spinnaker_testing.gate as gate
+import spinnaker_testing.frigga as frigga
+import citest.base
+
+ov_factory = jc.ObservationPredicateFactory()
```
jtk54: Not really a comment specific to this PR, but I see a lot of tests including this. We may want to just add this as a property to the `SpinnakerTestScenario` rather than have everyone declare it as a constant.

lwander: Good call, will start to look at wider scale cleanup as I rename some tests later.

PR #3047, opened 2018-07-13
chore(citest/appengine): Remove deployment quota for appengine tests.

jtk54:
According to the appengine docs, we can do 10k deployments
a day to the same project. As long as the deployments are
namespaced, I think we're in the clear.

### DIFF
--- a/dev/all_tests.yaml
+++ b/dev/all_tests.yaml
@@ -235,8 +235,6 @@ tests:
       configuration:
         appengine_account_enabled: true
       services: [clouddriver, front50, gate, orca]
-    quota:
-      appengine_deployment: 1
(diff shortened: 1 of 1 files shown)


ewiseblatt:
We need to make sure this is tested first under normal conditions. There might be other implications as well to guard against concurrent tests running within the same project if artifact names are fixed and not a function of the deconflicted TEST_STACK or CITEST_ID decorators.

I cannot look into it now, but can soon.

jtk54:
I'll run a test validation from this branch to verify.

jtk54:
Nice. Seems like this change is rolled into the larger quota change in https://github.com/spinnaker/spinnaker/pull/3219. Closing.

PR #3050, opened 2018-07-13
fix(test/appengine): remove old test path

lwander:


dev/all_tests.yaml
```diff
@@ -270,7 +270,6 @@ tests:
       services: [clouddriver, front50, gate, orca, echo]
     quota:
       appengine_deployment: 1
-    path: citest/tests/gcs_pubsub_gae_test.py
```
jtk54: Does this need replaced with something?

lwander: Nope, since the test name & entry above match

PR #3068, opened 2018-07-20
fix(validate): Log ntp time synchronization of instances

ezimanyi:
Some test failures, and the associated log timestamps, indicate that the machines running our tests may have a nontrivial time drift (on the order of hundreds of milliseconds).  To gather data on this, log the output of 'ntpq -p' before installing Spinnaker.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -1451,6 +1451,7 @@ class MonitoringConfigurator(Configurator):
 
   def add_init(self, options, script):
     """Implements interface."""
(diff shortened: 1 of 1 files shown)


ezimanyi:
Running citest pointed at this branch now; will check that the run looks good before merging.

ezimanyi:
My test run got `ntpq: read: Connection refused`.  I wonder if we tried to run the command to soon after the instance booted; I'll try moving it later.

ezimanyi:
Moved the `ntpq` command a bit later, and it worked.

PR #3077, opened 2018-07-23
test(provider/kubernetes): server group cache tests

lwander:


testing/citest/tests/kube_v2_cache_test.py
```diff
@@ -227,7 +281,16 @@ def test_b1_deploy(self):
...
     self.run_test_case(self.scenario.check_manifest_endpoint_exists('deployment'))
 
-  def test_b3_delete_deployment(self):
+  def test_b3_check_applications_endpoint(self):
+    self.run_test_case(self.scenario.check_applications_endpoint())
+
+  def test_b4_check_clusters_endpoint(self):
+    self.run_test_case(self.scenario.check_clusters_endpoint('deployment'))
+
+  def test_b4_check_server_groups_endpoint(self):
```
jtk54: This will just lex sort if they have the same number. Did you mean `b5`?

lwander: It doesn't matter what order they run, I'll actually make all these checks the same #

jtk54: Ok.

PR #3086, opened 2018-07-25
feat(buildtool): Adds spin CLI builds into buildtool.

jtk54:


dev/buildtool/spin_commands.py
```diff
@@ -0,0 +1,133 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Implements SpinCLI support commands for buildtool."""
+
+import logging, os
```
ewiseblatt: could you one per line?

dev/buildtool/spin_commands.py
```diff
@@ -0,0 +1,133 @@
...
+
+from collections import namedtuple
+
+from buildtool import (
+    BomSourceCodeManager,
+    RepositoryCommandFactory,
+    RepositoryCommandProcessor,
+
```
ewiseblatt: blank?

dev/buildtool/spin_commands.py
```diff
@@ -0,0 +1,133 @@
...
+class DistArch(namedtuple('DistArch', ['dist', 'arch'])):
+  """Describes an distribution/architecture pair.
+  """
+  def __repl__(self):
+    return '{},{}'.format(self.dist, self.arch)
+
+
+DIST_ARCH_LIST = [
+  DistArch('darwin', 'amd64'),
```
ewiseblatt: wrapped lines are 4 space indent

dev/buildtool/spin_commands.py
```diff
@@ -0,0 +1,133 @@
...
+]
+
+
+class BuildSpinCommand(RepositoryCommandProcessor):
+  def __init__(self, factory, options, **kwargs):
+    super(BuildSpinCommand, self).__init__(factory, options, **kwargs)
+    self.__gcs_uploader = SpinGcsUploader(options)
+
+  def _do_can_skip_repository(self, repository):
+    return repository.name != 'spin'
```
ewiseblatt: This is where you would check gcs to see if the build is already there

dev/buildtool/spin_commands.py
```diff
@@ -0,0 +1,133 @@
...
+
+  def _do_repository(self, repository):
+    """Implements RepositoryCommandProcessor interface."""
+    name = repository.name
+    build_version = self.scm.get_repository_service_build_version(repository)
+
+    if name == 'spin':
```
ewiseblatt: If you pass source_repository_names=['spin'] to the command constructor (or factory constructor to pass here) then the command will filter for you so this will only be called for 'spin'.

dev/buildtool/spin_commands.py
```diff
@@ -0,0 +1,133 @@
...
+                         cwd=config_root)
+
+        spin_path = '{}/spin'.format(config_root)
+        self.__gcs_uploader.upload_from_filename(
+          version_bin_path, spin_path)
+        self.__gcs_uploader.upload_from_filename(
+          nightly_bin_path, spin_path)
+        os.remove(spin_path)
+
+      with open('latest', 'w') as latest_file:
```
ewiseblatt: lastest_path = os.path.join(self.get_output_dir(), 'latest')

then just leave it in the output directory.

PR #3090, opened 2018-07-25
chore(buildtool): Update base config/template with spin build.

jtk54:


dev/buildtool/ga-defaults.yml
```diff
@@ -32,9 +32,15 @@ docs_repo_owner: spinnaker
 halyard_bom_bucket: halconfig
 
 
+################################
+# Spin CLI Commands
+################################
+spin_build_bucket: spinnaker-artifacts
+spin_build_credentials_path: key.json
```
ewiseblatt: leave the credentials out.
This is meant to document the external public topology.

PR #3091, opened 2018-07-26
fix(buildtool): Invoke build_spin with proper args.

jtk54:


jtk54:
One more fix required, build is dying on uploading bom since there's no `spin/halconfig`.

PR #3100, opened 2018-07-27
test(provider/kubernetes): test bake manifest

lwander:


testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
```diff
@@ -51,7 +51,6 @@ def upload_file(self, bucket_name, upload_path, local_filename):
     """Uploads a local file to a bucket at a relative upload path.
     """
     logging.info('Uploading local file %s to bucket %s at path %s', local_filename, bucket_name, upload_path)
-    print 'Uploading local file {} to bucket {} at path {}'.format(local_filename, bucket_name, upload_path)
```
jtk54: Nice.

testing/citest/tests/kube_v2_helm_test.py
```diff
@@ -0,0 +1,333 @@
...
+    return st.OperationContract(
+        self.agent.make_delete_app_operation(
+            application=self.TEST_APP,
+            account_name=self.bindings['SPINNAKER_KUBERNETES_V2_ACCOUNT']),
+        contract=contract)
+
+  def __create_helm_chart(self):
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    command = 'helm package {}/kube_v2_data/helm-test'.format(dir_path)
+    subprocess.Popen(command, stderr=sys.stderr, shell=True).wait()
```
jtk54: There should probably be a helper for running commands like this so you don't have to manage streams and waiting and such. We have a nice set of helper functions in `buildtool`, don't know if we want to include that lib in `testing/citest` though.

lwander: What would the helper do beyond calling subprocess.Popen(..).wait()? It feels pretty encapsulated already

jtk54: Cleans up the interface and provides logging, etc. [What I was referring to](https://github.com/spinnaker/spinnaker/blob/ae4006af07c08fc9259dd7c071bc7a39571cc6f9/dev/buildtool/subprocess_support.py#L130). May be overkill for one case but if this is more systematic then may be worth it.

PR #3096, opened 2018-07-26
chore(buildtool) Port buildtool to python3

ewiseblatt:
@jtk54

PR #3103, opened 2018-07-29
fix(build): fix stdout encoding with change to python 3

lwander:
Seems the encoding has changed to ascii with the upgrade to python3 on the system we're running on


lwander:
Related to https://github.com/spinnaker/spinnaker/pull/3096

PR #3107, opened 2018-07-30
fix(dev):  Fixed create_google_dev_vm script

ewiseblatt:
@jtk54 
Removed deprecated dependencies and modernized

### DIFF
--- a/dev/bootstrap_dev.sh
+++ b/dev/bootstrap_dev.sh
@@ -151,7 +151,7 @@ EOF
 # Start script
 
 # Install node
-NODE_VERSION=7.0.0
+NODE_VERSION=8.9.0
 . /etc/profile.d/nvm.sh
 nvm install $NODE_VERSION
 nvm alias default $NODE_VERSION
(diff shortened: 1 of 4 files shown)


dev/bootstrap_dev.sh
```diff
@@ -240,42 +249,40 @@ if [[ "$CONFIRMED_GITHUB_REPOSITORY_OWNER" != "none" ]]; then
...
-    # The default spinnaker-local will be configured for this machine as
-    # a better starting point than otherwise.
-    ./spinnaker/dev/stop_dev.sh bogus >& /dev/null || true
-fi
-
-# Some dependencies of Deck rely on Bower to manage their dependencies. Bower
-# annoyingly prompts the user to collect some stats, so this disables that.
-echo "{\"interactive\":false}" > ~/.bowerrc
+  echo "Setting up buildtool virtual environment in $PWD/venv"
+  virtualenv venv
```
jtk54: This is assuming `virtualenv` and `pip`, would we want to ensure these here since this is a bootstrap script?

ewiseblatt: This runs after the install_development script which installs them.

jtk54: Can you add that assumption in a comment here?

PR #3125, opened 2018-08-03
chore(build): Update gcloud build commands

ezimanyi:
The 'gcloud container builds' command group has been renamed to 'gcloud builds' as of Google Cloud SDK 210.0.0. The old commands are deprecated and will be removed in a future release; update
the calls in Spinnaker to account for this change.

### DIFF
--- a/codelabs/cicd-k8s-best-practice/app/scripts/build-image.sh
+++ b/codelabs/cicd-k8s-best-practice/app/scripts/build-image.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-gcloud container builds submit -q --tag gcr.io/spinnaker-playground/demo src/ --project spinnaker-playground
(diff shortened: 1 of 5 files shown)


ezimanyi:
The changed commands won't be compatible with gcloud <=209, but it looks like they are all run on systems that should be fully up to date:
* codelabs: cloud shell, which is always up to date
* builds: jenkins instance, which I'm updating now

@jtk54 : Does the change to `spinbot` look good?

ezimanyi:
Actually, @jtk54 , you can ignore my last question...confused `spinbot` and `spin-cli`

ezimanyi:
@lwander : Mind publishing the codelabs, or giving me write permission to the `gke-spinnaker-codelab` bucket?  (I got permission denied errors when trying to publish.)

PR #3136, opened 2018-08-07
fix(buildtool/spin): Remove spin from the BOM.

jtk54:
@stewchen FYI.


jtk54:
Builds will break without `--no-spin` at this point. Will fix that shortly. This is to unblock the release process.

PR #3099, opened 2018-07-27
chore(build): Removes extraneous heading from generated changelog

dorbin:
Also changes title to major.minor only.

---
### Instructions (that you should delete before submitting):

1. We prefer small, well tested pull requests.

2. Please refer to [Contributing to Spinnaker](https://spinnaker.io/community/contributing/).

3. When filling out a pull request, please consider the following:

    * Follow the commit message conventions [found here](https://www.spinnaker.io/community/contributing/submitting/).
    * Provide a descriptive summary for your changes.
    * If it fixes a bug or resolves a feature request, be sure to link to that issue.
    * Add inline code comments to changes that might not be obvious.
    * Squash your commits as you keep adding changes.
    * Add a comment to @spinnaker/reviewers for review if your issue has been outstanding for more than 3 days.

4. We are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them.

PR #3138, opened 2018-08-07
chore(build): Restores fullly-qualified version to "Latest Stable" list

dorbin:
#3099 removed an extraneous h1, and also changed the version displayed at the top of the [changelog page](https://www.spinnaker.io/community/releases/versions/1-8-5-changelog) so that it only shows major.minor, because the page actually contains all the point releases under that major minor.

A side effect was that the version listing under [Latest stable](https://www.spinnaker.io/community/releases/versions/#latest-stable) also now only showed major.minor, in contrast with earlier releases also shown on page. 

This should fix it.


ttomsu:
@ewiseblatt - [this](https://github.com/spinnaker/spinnaker.github.io/pull/953) change was made to support Dave's change here.

PR #3142, opened 2018-08-08
feat(buildtool/spin) Adds spin publish command.

jtk54:
In the process of testing, don't merge yet. Still PTAL.

### DIFF
--- a/dev/buildtool/spin_commands.py
+++ b/dev/buildtool/spin_commands.py
@@ -12,15 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Implements SpinCLI support commands for buildtool."""
+"""Implements Spin CLI support commands for buildtool."""
 
+import copy
 import logging
 import os
 
 from collections import namedtuple
 
 from buildtool import (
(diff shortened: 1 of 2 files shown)


dev/buildtool/spin_commands.py
```diff
@@ -133,6 +146,110 @@ def check_file_exists(self, path):
...
+    self.__scm = BranchSourceCodeManager(options_copy, self.get_input_dir())
+    # TODO(jacobkiefer): Add spin CLI autogenerated docs.
+
+    dash = gate_version.find('-')
+    semver_str = gate_version[0:dash]
+    semver_parts = semver_str.split('.')
+    if len(semver_parts) != 3:
+      raise_and_log_error(
+          ConfigError('Expected gate_version in the form X.Y.Z-N'))
+    self.__release_branch = 'release-{maj}.{min}.x'.format(
```
ewiseblatt: The release- version here is the gate version. Everywhere else release- version is the spinnaker version. I think that is going to cause confusion. Either this should be a spinnaker version or use a different string prefix than "release-".

jtk54: Good point.

PR #3146, opened 2018-08-08
fix(buildtool/spin): Fixes spin publishing.

jtk54:


dev/buildtool/spin_commands.py
```diff
@@ -192,17 +194,15 @@ def __init__(self, factory, options, **kwargs):
...
     if len(semver_parts) != 3:
       raise_and_log_error(
           ConfigError('Expected spinnaker version in the form X.Y.Z-N'))
     self.__release_branch = 'release-{maj}.{min}.x'.format(
         maj=semver_parts[0], min=semver_parts[1])
-    self.__release_tag = 'version-' + semver_str
+    self.__release_tag = 'version-' + spinnaker_version
 
-    self.__release_version = semver_str
+    self.__release_version = spinnaker_version
```
ewiseblatt: So you cannot patch spin without a full spinnaker release?

jtk54: That was unused. Removed.

jtk54: But this still raises a good point about fixes in spin.

ewiseblatt:
I'm not done, but this keeps changing and invalidating me so I'll come back to it later.
I wanted to look at "promote" more. It looked like it was finding the version, but it should already know the version if you are publishing it. Otherwise you dont know what you are publishing which means you cant be sure you have validated it. That might not be what is happening but I havent been able to look too closely yet.

jtk54:
Sorry, I found another modification I wanted to make. I stopped changing it.

dev/buildtool/spin_commands.py
```diff
@@ -180,72 +176,117 @@ def __init__(self, factory, options, **kwargs):
...
 
+    match = re.match(r'(\d+)\.(\d+)\.(\d+)-\d+', self.__gate_version)
+    if match is None:
+      raise_and_log_error(
+          ConfigError('gate version {version} is not X.Y.Z-<buildnum>'
+                      .format(version=self.__gate_version)))
+    semver_parts = self.__spinnaker_version.split('.')
+    if len(semver_parts) != 3:
+      raise_and_log_error(
+          ConfigError('Expected spinnaker version in the form X.Y.Z-N'))
```
lwander: I would append a `, got: <self.__spinnaker_version>` here for debugability.

dev/buildtool/spin_commands.py
```diff
@@ -180,72 +176,117 @@ def __init__(self, factory, options, **kwargs):
...
+
+    release_branch = 'release-{maj}.{min}.x'.format(
+        maj=semver_parts[0], min=semver_parts[1])
+    release_tag = 'version-' + self.__stable_version
     release_url = repository.origin
```
ewiseblatt: why the alias?

dev/buildtool/spin_commands.py
```diff
@@ -180,72 +176,117 @@ def __init__(self, factory, options, **kwargs):
...
-                      self.__release_tag, want_commit)
-        return
-
+                 release_branch, release_tag, release_url)
     git.check_run_sequence(
         git_dir,
         [
-            'checkout -b ' + self.__release_branch,
+            'checkout -b ' + release_branch,
             'remote add release ' + release_url,
```
ewiseblatt: release is a non-standard remote.

dev/buildtool/spin_commands.py
```diff
@@ -283,6 +324,12 @@ def __init__(self):
   def init_argparser(self, parser, defaults):
     super(PublishSpinCommandFactory, self).init_argparser(
         parser, defaults)
+    self.add_argument(
```
ewiseblatt: why "build" and not "publish"?

dev/buildtool/spin_commands.py
```diff
@@ -180,72 +176,117 @@ def __init__(self, factory, options, **kwargs):
...
+    # and minor versions so fixes (thus patch version) are decoupled between
+    # the two.
+    patch = '0' # Patch is reset on a new Gate major or minor.
+    tag_matcher = re.compile(r'version-{maj}.{min}.(\d+)'
+                             .format(maj=gate_major, min=gate_min))
+    tags = git.fetch_tags(git_dir)
+    tag_matches = [tag_matcher.match(t) for t in tags if tag_matcher.match(t)]
+    if tag_matches:
+      patch_versions = [m.group(1) for m in tag_matches]
+      max_patch = max(patch_versions)
```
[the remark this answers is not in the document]
jtk54: Good catch.

dev/buildtool/spin_commands.py
jtk54: I think it's superfluous.

dev/buildtool/spin_commands.py
jtk54: :+1:

dev/buildtool/spin_commands.py
jtk54: It's the same bucket as build. I can rename it to `spin_bucket` or something similar.

ewiseblatt: I didnt see that. It might be ok. I think I've done that in places too. Though I do like spin_bucket better since it is context-free.

dev/buildtool/spin_commands.py
```diff
@@ -180,72 +176,116 @@ def __init__(self, factory, options, **kwargs):
...
+                 release_branch, release_tag, repository.origin)
     git.check_run_sequence(
         git_dir,
         [
-            'checkout -b ' + self.__release_branch,
-            'remote add release ' + release_url,
-            'push release ' + self.__release_branch,
-            'tag ' + self.__release_tag,
-            'push release ' + self.__release_tag
+            'checkout -b ' + release_branch,
```
ewiseblatt: I was confused earlier in this git UI that doesnt show much context and assumed this was the publisher class. Why does a build create a branch and tag? Builds should not have side effects into the repo.  I dont know why push and promote are in this class. Those are publisher activities.

jtk54: This is in `class PublishSpinCommand`.

dev/buildtool/spin_commands.py
jtk54: Changed.

jtk54:
Any other concerns with this? I think everything is addressed.

ewiseblatt:
looks ok. I'm locked out of github so I cannot reply there.

PR #3151, opened 2018-08-09
feat(spinbot): check for new groovy files

lwander:
As discussed in https://github.com/spinnaker/clouddriver/pull/2838

### DIFF
--- a/spinbot/event/filetype_check_pull_request_handler.py
+++ b/spinbot/event/filetype_check_pull_request_handler.py
@@ -1,0 +1,47 @@
+from .handler import Handler
+from .pull_request_event import GetBaseBranch, GetPullRequest, GetTitle, GetRepo
+from gh import ReleaseBranchFor, ParseCommitMessage
+
+format_message = ('We prefer that non-test backend code be written in Java or Kotlin, rather ' +
+        'than Groovy. The following files have been added and written in Groovy:\n\n' +
+        '{}\n\n' + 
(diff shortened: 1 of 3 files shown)


spinbot/event/filetype_check_pull_request_handler.py
```diff
@@ -0,0 +1,47 @@
+from .handler import Handler
+from .pull_request_event import GetBaseBranch, GetPullRequest, GetTitle, GetRepo
+from gh import ReleaseBranchFor, ParseCommitMessage
+
+format_message = ('We prefer that non-test backend-code be written in Java or Kotlin, rather ' +
```
sbwsg: 'backend code' doesn't need the hyphen here i don't think. Also, is this a preference or is it a rule? I'd argue it's better to be a little more direct and include a call-to-action; something like

```python
"New classes must be written in Java or Kotlin. " +
"Please replace Groovy in the following new files:\n\n" + 
"{}\n\n" +
"See our server-side [commit conventions here](https://spinnaker.io/community/contributing/back-end-code/#choice-of-language)."
```

It seems a little less polite but it's also a clearer instruction of what to do next.

jtk54: +1

lwander: I'm hesitant to make it a hard rule, the enforcement is something we can discuss with Netflix when the time comes.

PR #3152, opened 2018-08-09
fix(citest): Add queueing to front50 Hystrix config

ezimanyi:
A number of the intermittent failures observed in running our CI tests are due to Front50 hitting a Hystrix circuit breaker when trying to fetch newly-created applications from cloud storage. We fixed a similar
issue in gate by adding queueing to the Hystrix config; do the same for front50, and also increase the queue size to give us a bit more buffer (as we send ~20 requests to create an application at the start of the tests). As this is primarily an issue with burstiness and not throughput, leaving the thread pool size at the default for now.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -1525,10 +1525,11 @@ class SpinnakerConfigurator(Configurator):
 hystrix:
   threadpool:
     default:
-      maxQueueSize: 20
-      queueSizeRejectionThreshold: 20
+      maxQueueSize: 100
+      queueSizeRejectionThreshold: 100
 '''
     script.append('echo "{}" > ~/.hal/default/profiles/gate-local.yml'.format(hystrix_config))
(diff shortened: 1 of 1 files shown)


ezimanyi:
@ewiseblatt : I don't think this would mask other issues---we're effectively adding resiliency to burstiness to the tests.  The same logic will execute, but will queue up of there are too many requests to GCS at the same time (as opposed to falling back to using the cache, as was happening before).

Staggering the startup times is another option, but I think that with the change here we're effectively increasing the ability of Spinnaker to deal with concurrent requests, so might not be as necessary if we merge this.  In some sense, we're forcing the tests to stagger themselves by adding queueing within the critical path that was saturating before.

ewiseblatt:
ok. lgtm.

By masking other issues, I meant "other issues" elsewhere where hystrix might have kicked in where now it wont.

PR #3171, opened 2018-08-10
fix(citest): Increase timeout for resize server group

ezimanyi:
When resizing a server group, Spinnaker reports success when the new instance is observed. Our test cases are actually testing that the instance group is reporting the new size, which can lag the instance actually appearing. Increase the amount of time we wait for the instance group to report its new size to reduce intermittent failures.


ezimanyi:
@ewiseblatt : I think this should be enough time (as the couple of cases I dug into only missed by ~10-20 s), but I guess if it still fails intermittently we can further increase the wait time.

The only issue I'm aware of with this test is the timing issue (not waiting long enough for the server group to report its new size after the instance shows up), but it's possible that there's some other issue I'm not aware of---every failure of this test I've dug into has been this timing issue.

PR #3179, opened 2018-08-13
fix(citest): Further increase timeout for resize server group

ezimanyi:
Looks like the timeout added in PR #3171 was not enough; let's extend to 10 minutes to reduce test flakiness.


ewiseblatt:
LGTM

PR #3182, opened 2018-08-13
fix(buildtool/spin): Set spin CLI version in binary.

jtk54:
Depends on: https://github.com/spinnaker/spin/pull/54/files.

### DIFF
--- a/dev/buildtool/spin_commands.py
+++ b/dev/buildtool/spin_commands.py
@@ -59,6 +59,12 @@ class BuildSpinCommand(RepositoryCommandProcessor):
       factory, options, source_repository_names=['spin'], **kwargs)
     self.__gcs_uploader = SpinGcsUploader(options)
     self.__build_version = None  # recorded after build
+    bom_contents = BomSourceCodeManager.load_bom(options)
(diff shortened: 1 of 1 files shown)


jtk54:
Dependent PR merged, PTAL.

dev/buildtool/spin_commands.py
```diff
@@ -101,10 +107,21 @@ def build_all_distributions(self, repository):
...
                   'GOOS': dist_arch.dist,
                   'GOARCH': dist_arch.arch})
+
+      # Note: spin CLI is coupled to the Gate major and minor version.
+      # Gate is a routing server, so features and breaking changes in Gate
+      # must be reflected in spin since it is a client.
+      dash = self.__gate_version.find('-')
+      gate_semver = self.__gate_version[:dash]
+      # Unset ReleasePhase tag for proper versions.
+      prefix = 'github.com/spinnaker/spin/version'
```
ewiseblatt: can you use repository.origin instead of the github.com/spinnaker/spin? That way you arent assuming the repo this is building.

jtk54: Yes, but it needs formatted as a go package for `ldflags`. Done.

PR #3199, opened 2018-08-15
fix(buildtool): Always compute version numbers when creating bom.

ewiseblatt:
Otherwise we dont handle patches on earlier releases because
we're re-using the earlier patched version (assuming master branch doesnt change
so continues to carry forward unchanged commits whose version was computed before
the change in earlier branch).

@stewchen 
@jtk54

### DIFF
--- a/dev/buildtool/bom_commands.py
+++ b/dev/buildtool/bom_commands.py
@@ -64,14 +64,14 @@ class BomBuilder(object):
   """Helper class for BuildBomCommand that constructs the bom specification."""
 
   @staticmethod
-  def new_from_bom(options, scm, bom):
-    return BomBuilder(options, scm, base_bom=bom)
+  def new_from_bom(options, scm, metrics, bom):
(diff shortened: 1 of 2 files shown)


dev/buildtool/bom_commands.py
```diff
@@ -182,11 +183,29 @@ def build(self):
...
+      index = build_version.find('-')
+      return build_version[:index] if index >= 0 else build_version
+
     for name, info in self.__services.items():
+      labels = {'repostiory': name, 'branch': options.git_branch, 'updated': True}
       if info['commit'] == services.get(name, {}).get('commit', None):
-        logging.debug('%s commit hasnt changed -- keeping existing %s',
-                      name, info)
-        continue
+        if to_semver(info['version']) == to_semver(services.get(name, {}).get('version', '')):
```
ewiseblatt: This check here is the actual fix.
The other changes are mostly preserving monitoring, though I'm not sure if the metrics are still of use.

PR #3205, opened 2018-08-16
Pipelines and application used in upcoming Kayenta tutorial

MrTrustor:
I will soon publish a Kayenta tutorial on cloud.google.com.
Those are the assets needed for this tutorial.

I am still polishing the CI for this solution. It will come as a second PR.

cc @duftler @lwander @demobox


codelabs/solutions/kayenta/pipelines/automated-canary.json
```diff
@@ -0,0 +1,431 @@
...
+                        "name": "SUCCESS_RATE",
+                        "valueFrom": {
+                          "configMapKeyRef": {
+                            "key": "SUCCESS_RATE",
+                            "name": "sampleapp-config-prod"
+                          }
+                        }
+                      }
+                    ],
+                    "image": "gcr.io/spinnaker-marketplace/sampleapp:latest",
```
lwander: How does someone w/o access to spinnaker marketplace build this image/run this pipeline?

codelabs/solutions/kayenta/README.md
```diff
@@ -0,0 +1,9 @@
+# Automated canary analysis on Kubernetes Engine with Spinnaker
```
lwander: Any reason these are in the solutions/ subdirectory?

codelabs/solutions/kayenta/pipelines/automated-canary.json
MrTrustor: `gcr.io/spinnaker-marketplace` is public. Not sure why someone wouldn't have access to this image?

The code for this image is in the app directory, in the same PR.

codelabs/solutions/kayenta/README.md
MrTrustor: It was suggested by @duftler. I can move it if needed.
It's not a codelab though.

duftler: After talking about it with @lwander a bit, I think best approach is to have `solutions` as a sibling to `codelabs`.

MrTrustor: Moved everything to the root of the repo.

codelabs/solutions/kayenta/pipelines/automated-canary.json
lwander: Well they can't push to the repo if they want to make changes or run this themselves -- if that's the intent then this LGTM

lwander:
One last question (above) about deploying from a repo users can't push to

codelabs/solutions/kayenta/pipelines/automated-canary.json
MrTrustor: That's the intent. They always run the same image in the tutorial. It's configurable with an env variable to change the error rate and simulate different behaviours for Kayenta.

MrTrustor:
Actually, just wait another hour please. I'll be able to push the CI scripts today in this PR.

MrTrustor:
Ok, just added the `solutions/kayenta/ci` directory. This uses the Concourse instance of my team and will run through the entire tutorial once a day. We receive the potential notifications and will handle any updates to the code base and/or tutorial.

solutions/kayenta/ci/scripts/automated-canary.sh
```diff
@@ -0,0 +1,159 @@
...
+DECK_POD=$(kubectl -n spinnaker get pods -l \
+    cluster=spin-deck,app=spin \
+    -o=jsonpath='{.items[0].metadata.name}')
+kubectl -n spinnaker port-forward $DECK_POD 8080:9000 \
+    >/dev/null &
+sleep 5
+
+curl -XPOST -d '{"job":[{"type":"updateApplication","application":{"name":"sampleapp","dataSources":{"enabled":["canaryConfigs"],"disabled":[]}},"user":"[anonymous]"}],"application":"sampleapp","description":"Update Application: sampleapp"}' \
+    --fail -sS -H "Content-Type: application/json" -H "Accept: */*" \
+    http://localhost:8080/gate/applications/sampleapp/tasks >/dev/null
```
lwander: Any reason not to use the `spin` cli for the application & pipeline tasks?

lwander:
That's really neat -- this will catch codelab breakages?

solutions/kayenta/ci/scripts/automated-canary.sh
lwander: If this is only internal it doesn't really matter -- if we plan to make this accessible or share this as part of the solution we should add a TODO to use the recommended tooling.

MrTrustor: It's only internal. But I used spin where I could, which is almost nowhere:
* This particular line is enabling the canary option for the `sampleapp` app. Not possible with spin.
* I need to pass a parameter when launching a pipeline, spin doesn't support that yet.
* 2 of the pipelines I create have dependencies on other things that I need to dynamically resolve. AFAIK, spin doesn't support that yet.

MrTrustor:
> That's really neat -- this will catch codelab breakages?
Hopefully :-) (not a codelab, but close enough! and could be turned into one, maybe)

solutions/kayenta/ci/scripts/automated-canary.sh
lwander: @jtk54 ^

jtk54: @MrTrustor What are the 'other things' that need dynamically resolved? Created tracking bugs for the first two things, thanks for the feedback.

solutions/kayenta/ci/scripts/automated-canary.sh
```diff
@@ -0,0 +1,159 @@
...
+sleep 5
+
+## Create automated canary pipeline ##
+export PIPELINE_ID=$(curl localhost:8080/gate/applications/sampleapp/pipelineConfigs/Simple%20deploy \
+    | jq -r '.id')
+export CANARY_CONFIG_ID=$(curl localhost:8080/gate/v2/canaryConfig | jq -r '.[0].id')
+jq '(.stages[] | select(.refId == "9") | .pipeline) |= env.PIPELINE_ID | (.stages[] | select(.refId == "8") | .pipeline) |= env.PIPELINE_ID | (.stages[] | select(.refId == "11") | .canaryConfig.canaryConfigId) |= env.CANARY_CONFIG_ID' spinnaker-git/solutions/kayenta/pipelines/automated-canary.json | \
+    curl -d@- -X POST \
+    -H "Content-Type: application/json" -H "Accept: */*" \
+    http://localhost:8080/gate/pipelines
```
MrTrustor: @jtk54 You have both examples on this one.

1. The pipeline that is created here depends on another pipeline (it fetches an artifact and also can trigger it). I need to inject the ID of that other pipeline in the definition of this new pipeline.
2. This new pipeline uses a Canary Analysis stage, and that takes a canary config as parameter. I need to inject the ID of this canary config in the definition of this new pipeline.

lwander: What's preventing you from generating the desired pipeline spec given that you know the pipeline & canary config IDs before submitting it with spin?

MrTrustor: Nothing, now that you mention it 😄However, I would expect a tool like spin to completely replace my need for curl (which I need to get the IDs). When writing the tutorial, I ruled against it for this reason: mixing curl and spin was super confusing. I just didn't think too much about it when writing the tests.

jtk54: The curl that was necessary was specifically for canary configs, right? That seems like a reasonable set of commands to add into `spin` going forward.

duftler: Seems like we could have a `spin pipeline get-id` command as a companion to this one: `spin pipeline get --name my-pipeline --application my-app`.
And a similar thing for the as-yet-to-exist `canary-config` command group.

duftler: Or just args to tweak the response content/format.

jtk54: What is wrong with using `jq` to parse the output coming from `spin` in the case where you want the pipeline id? Agreed that pipeline templates need a surface, but we need to think very carefully what that looks like given we are starting the design of managed pipeline templates v2.

demobox: > What is wrong with using jq to parse the output coming from spin in the case where you want the pipeline id

I can see that a dependency on `jq` is not ideal, but on the other hand providing _some_ amount of response parsing/control in `spin` itself risks getting into a "feature race" with tools like `jq`, which is not the intention of `spin`.

If sufficiently common use cases show up that merit out-of-the-box support I'd be up for @duftler's suggestion of adding commands like `get-id`, but I think we'd need to wait a bit before figuring out what those use cases might be?

MrTrustor: I'm not sure I'm following the conversation between here and spinnaker/spin#59, but for what it's worth: I agree that templating is probably something that should be considered very carefully.
However, regarding the output of spin, it seems to me that a `--json-path` flag is not that complicated to implement using existing libs (as it seems it has been done on spinnaker/spin#59). I don't have any hard data, but I see a lot of people using this option with kubectl. As long as you don't use it to _change_ the content of response, just its formatting, I don't think there's a risk of "feature race" with jq. And if people want to use jq, they still can.
To me it looks like a quick win that improves the usability of spin.

PR #3206, opened 2018-08-16
fix(spinbot): unchecked-in code, & abuse protection mechanism

lwander:


spinbot/policy/stale_issue_policy.py
```diff
@@ -8,7 +8,7 @@ def __init__(self):
         self.stale_days = self.config.get('stale_days')
         self.count = 0
         if not self.stale_days:
-            self.stale_days = 90
+            self.stale_days = 45
```
lwander: Forgot to check this in last time

PR #3233, opened 2018-08-23
fix(citest): Limit polling to US regions

ezimanyi:
We use a lot of GCE operation quota when running multiple tests concurrently.  As our tests only deploy to US regions, limit the configured Google account to only poll US regions to reduce our quota usage.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -822,8 +822,7 @@ class GoogleConfigurator(Configurator):
         parser, 'google_account_name', defaults, 'my-google-account',
         help='The name of the primary google account to configure.')
     add_parser_argument(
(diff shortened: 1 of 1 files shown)


dev/validate_bom__config.py
```diff
@@ -821,6 +821,10 @@ def init_argument_parser(self, parser, defaults):
     add_parser_argument(
         parser, 'google_account_name', defaults, 'my-google-account',
         help='The name of the primary google account to configure.')
+    add_parser_argument(
```
ewiseblatt: Can you make this default to None and only pass it through if it is set?
I think the default behavior for deploying a validation should be what customers see.

ezimanyi: Done

PR #3236, opened 2018-08-23
fix(citest): Limit polling to US regions (#3233)

ezimanyi:
We use a lot of GCE operation quota when running multiple tests concurrently.  As our tests only deploy to US regions, limit the configured Google account to only poll US regions to reduce our quota usage.


ezimanyi:
Cherry-pick #3233 to 1.9 to reduce hitting quota running tests against that branch.

PR #3247, opened 2018-08-24
fix(build): Publish debians to Bintray synchronously

ezimanyi:
We get intermittent failures in our build scripts because we publish debians to Bintray asynchronously, but later build steps expect these debians to be present. Change the publish call to be synchronous.

### DIFF
--- a/dev/buildtool/ga-defaults.yml
+++ b/dev/buildtool/ga-defaults.yml
@@ -21,6 +21,7 @@
 bintray_jar_repository: jars
 bintray_org: spinnaker-releases
 bintray_debian_repository: debians
+bintray_publish_wait_secs: -1
 
 
 ################################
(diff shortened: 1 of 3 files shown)


dev/buildtool/ga-defaults.yml
```diff
@@ -21,6 +21,7 @@
 bintray_jar_repository: jars
 bintray_org: spinnaker-releases
 bintray_debian_repository: debians
+bintray_publish_wait_secs: -1
```
ewiseblatt: This is correct, but this ga-defaults.yml file is not actually used anywhere.
It is just here for documentation purposes.
You will need to add this to the "buildtool.yml" file on the build server.

ezimanyi: Yup, that confused me at first, but I figured it out and have already updated that file.

PR #3261, opened 2018-08-28
fix(buildtool): Use Dockerfile.slim if present.

ewiseblatt:
@ttomsu


ewiseblatt:
halyard doesnt go through this flow because [for better or worse] it has
its own build scripts.
Before it was always using the Dockerfile.slim for everything other than
echo and spinnaker-monitoring.
So this is no worse than before with regard to halyard.

PR #3286, opened 2018-09-04
fix(build/containers): Exclude echo from slim docker builds.

jtk54:
Regression in https://github.com/spinnaker/spinnaker/pull/3261. @joonlim PTAL also.


joonlim:
LGTM

jtk54:
It was poorly documented before, so should be clear now.

PR #3292, opened 2018-09-05
feat(ha): Add HA configurator.

joonlim:


joonlim:
@ewiseblatt

dev/validate_bom__config.py
```diff
@@ -1543,6 +1543,64 @@ def add_files_to_upload(self, options, file_set):
...
 
 
+class HaConfigurator(Configurator):
+  """Controls hal config deploy ha (only supported by Kubernetes v2 accounts)."""
+
+  def init_argument_parser(self, parser, defaults):
+    """Implements interface."""
+    add_parser_argument(
+        parser, 'ha_clouddriver_enabled', defaults, False, type=bool,
+        help='Whether or not to deploy clouddriver in HA mode.'
```
ewiseblatt: I think the details of what "ha" means is outside the scope.
The responsibility here is to say you want HA and control halyard to enable HA in clouddriver (and echo). What specifically that means is a detail to halyard, not here. If halyard were to change what these are called, or the topology or other details, they would not have any effect here. The responsibility of these Configurators is to control halyard configuration and to inform the test controller about configuration details that it might need to adapt the tests to the details of the specific deployment's configuration (and those details do not [as of this time] require knowledge of HA topology).

dev/validate_bom__config.py
```diff
@@ -1543,6 +1543,64 @@ def add_files_to_upload(self, options, file_set):
     pass
 
 
+class HaConfigurator(Configurator):
+  """Controls hal config deploy ha (only supported by Kubernetes v2 accounts)."""
```
ewiseblatt: Not sure what this means. Is an account a Spinnaker concept? And the HA configuration is a halyard concept?

joonlim:
@ewiseblatt Is this good to merge?

PR #3298, opened 2018-09-06
fix(citest): Increase operation timeouts for bake and deploy.

jtk54:
Also initialize class variable for pub/sub test to avoid NPE.


testing/citest/tests/bake_and_deploy_test.py
```diff
@@ -204,10 +204,9 @@ def __init__(self, bindings, agent=None):
     self.docker_pipeline_id = None
     self.test_google = bindings['TEST_GOOGLE']
     self.test_aws = bindings['TEST_AWS']
-    # This test has been exceeding the default timeout of 13 minutes for the Jenkins agent,
```
ezimanyi: Did you mean to delete the comment here?

PR #3302, opened 2018-09-07
fix(codelab): use default cluster version

lwander:


codelabs/gke-base/install/setup.sh
```diff
@@ -92,8 +92,7 @@ gcloud beta pubsub subscriptions create $GCR_SUB --topic $GCR_TOPIC
 bold "Creating your cluster $GKE_CLUSTER..."
 
 gcloud container clusters create $GKE_CLUSTER --zone $ZONE \
-  --service-account $SA_EMAIL \
-  --username admin --cluster-version 1.8.8-gke.0 \
```
jtk54: What changed that we can omit this now? Update in gcloud behavior?

lwander: It was possible before, but for the codelab this was copied from (the IO one) we needed a higher versioned cluster than the default.

PR #3330, opened 2018-09-13
fix(validate): Revert bind appengine_ smoke-test to TEST_GCE_REGION

ezimanyi:
This reverts commit f896f0f1a0dbe5353b0e078811d0f876b14d1640.


ezimanyi:
The parent commit was causing appengine tests to fail as they were looking in the wrong region for deployed server groups.

Issue #3349, opened 2018-09-18
gcloud not in path by default on GCE Ubuntu 18.04


ezimanyi:
The appengine CI tests for Ubuntu 18.04 have been failing, as clouddriver is unable to find `gcloud` and `gcloud` is necessary for registering an appengine account.

It looks like the Ubuntu 18.04 image doesn't contain appengine, but that it is installed on first boot.  It seems like there is a race condition where the executable is not installed in time for clouddriver.  (Given how long it takes spinnaker to boot up, this may actually be a case where the process is installed but the environment for Spinnaker does not contain its path, or something similar.)

ezimanyi:
The above changes still haven't fully fixed the issue; we need clouddriver to know where the `gcloud` executable is, but it is started by systemd which doesn't start services in a login shell. One potential solution here is to add a configurable parameter "path to gcloud" to the halyard config and have clouddriver use that full path rather than rely on resolution based on the PATH environment.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

spinnakerbot:
This issue is tagged as 'stale' and hasn't been updated  in 45 days, so we are tagging it as 'to-be-closed'. It will be closed in 45 days unless updates are made. If you want to remove this label, comment:

> @spinnakerbot remove-label to-be-closed

spinnakerbot:
This issue is tagged as 'to-be-closed' and hasn't been updated  in 45 days, so we are closing it. You can always reopen this issue if needed.

PR #3353, opened 2018-09-18
fix(citest): Run hal deploy apply in a login shell

ezimanyi:
Ubuntu 18.04 on GCE installs gcloud using snap, so it ends up in the /snap/bin directory. This directory is added to the path for login shells, but not for non-login shells. Let's run hal deploy apply in a login shell to prevent issues like this.


ezimanyi:
Addresses https://github.com/spinnaker/spinnaker/issues/3349

ezimanyi:
@ewiseblatt : This fixes 18.04 validation, which is currently broken for the reasons in the linked issue

PR #3393, opened 2018-09-27
chore(config): delete outdated config dir

lwander:
The referenced config is roughly a year out-of-date, and we have no plans to maintain it.


ady624:
This folder is apparently required by the experimental spinnaker Docker compose...

PR #3399, opened 2018-09-28
feat(codelab/monitoring): Enable Stackdriver GKE Monitoring for codel…

duftler:
…ab base cluster.


codelabs/gke-base/install/setup.sh
```diff
@@ -97,10 +97,10 @@ gcloud beta pubsub subscriptions create $GCR_SUB --topic $GCR_TOPIC
 
 bold "Creating your cluster $GKE_CLUSTER..."
 
-gcloud container clusters create $GKE_CLUSTER --zone $ZONE \
+gcloud beta container clusters create $GKE_CLUSTER --zone $ZONE \
   --service-account $SA_EMAIL --username admin \
   --machine-type n1-standard-4 --image-type COS --disk-size 100 \
-  --num-nodes 3 --enable-cloud-logging --enable-cloud-monitoring
+  --num-nodes 3 --cluster-version 1.10.6 --enable-stackdriver-kubernetes
```
lwander: Can we parameterize the version? And have an option to leave it blank? I'm asking because once this version is removed from gcloud's list the command will fail for all codelabs based on this setup

duftler: Done.

PR #3403, opened 2018-10-01
fix(citest/gae/pubsub): Increase timeouts for pubsub -> gae test.

jtk54:


jtk54:
@danielpeach FYI.

ewiseblatt:
LGTM if this helps, but still need to investigate why

jtk54:
I'm not sure this is actually root cause, test run still failed in roughly the same way. This points to an issue in the pub/sub logic in Spinnaker. Will look today.

jtk54:
Closing, I'm fairly certain this is not needed. If it is, I'll add in a separate change that addresses the real problems with this test (when we determine them).

PR #3414, opened 2018-10-03
fix(citest/gae): Fix gae pubsub trigger agent.

jtk54:


ewiseblatt:
please patch this into 1.10

PR #3422, opened 2018-10-04
chore(buildtool/validate): Adds debug logging for bucket creation.

jtk54:
Allows us to inspect the storage class etc of buckets we are creating for validation purposes.


dev/validate_bom__config.py
```diff
@@ -279,7 +279,8 @@ def setup_environment(cls, options):
 
     storage_client = GcsArtifactStorageConfiguratorHelper.__instantiate_client(options)
     if storage_client.lookup_bucket(options.artifact_gcs_bucket) is None:
-      storage_client.create_bucket(options.artifact_gcs_bucket)
+      created_bucket = storage_client.create_bucket(options.artifact_gcs_bucket)
+      logging.debug('Created bucket for Artifact storage: %s', created_bucket)
```
ewiseblatt: Also log the deletes

PR #3434, opened 2018-10-05
feat(codelab/kayenta-workship): Add services and supporting artifacts.

duftler:
Will resolve the duplication with gke-source-to-prod later. Have not customized the README either.

### DIFF
--- a/codelabs/gke-base/install/enable_apis.sh
+++ b/codelabs/gke-base/install/enable_apis.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+bold() {
+  echo ". $(tput bold)" "$*" "$(tput sgr0)";
+}
+
+# TODO: Move this list of APIs to properties if we end up needing to enable other APIs.
+
+bold "Enabling Pub/Sub API..."
+
+gcloud services --project $PROJECT_ID enable pubsub.googleapis.com
+
+
+bold "Enabling Cloud Build API..."
+
+gcloud services --project $PROJECT_ID enable cloudbuild.googleapis.com
+
+
+bold "Enabling Kubernetes Engine API..."
+
+gcloud services --project $PROJECT_ID enable container.googleapis.com
--- a/codelabs/gke-kayenta-workshop/services/backend/main.go
+++ b/codelabs/gke-kayenta-workshop/services/backend/main.go
@@ -1,0 +1,138 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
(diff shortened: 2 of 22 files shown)


duftler:
@lwander 1st pass is ready for a review.

codelabs/gke-base/install/manifests.yml
```diff
@@ -42,7 +42,7 @@ spec:
     spec:
       containers:
       - name: halyard-daemon
-        image: gcr.io/spinnaker-marketplace/halyard:stable
+        image: gcr.io/spinnaker-marketplace/halyard@sha256:5c34d37b7e6fbcc1776d2cd9385ebaf063a7b6d670abacb0fa6c2b5886e5d47f
```
lwander: Intentional to pin this?

codelabs/gke-base/install/enable_apis.sh
```diff
@@ -0,0 +1,19 @@
...
+
+
+bold "Enabling Cloud Build API..."
+
+gcloud services --project $PROJECT_ID enable cloudbuild.googleapis.com
+
+
+bold "Enabling Kubernetes Engine API..."
+
+gcloud services --project $PROJECT_ID enable container.googleapis.com
```
lwander: Do you think we'll need more APIs in other codelabs? If so, maybe we can drop a note to move this list of apis to the `properties` file in a future change.

codelabs/gke-kayenta-workshop/services/backend/main.go
```diff
@@ -0,0 +1,137 @@
...
+
+	"google.golang.org/api/monitoring/v3"
+)
+
+var count = 1
+var lastTime = time.Now()
+
+// Started with: https://github.com/GoogleCloudPlatform/golang-samples/blob/master/monitoring/custommetric/custommetric.go
+func writeTimeSeriesValue(now string, s *monitoring.Service, metricType string, metricValue int64) error {
+	project_id, err := gce.ProjectID()
```
lwander: Worth checking these errors?

codelabs/gke-base/install/manifests.yml
duftler: At the time, it was the version that had the fix we needed. Looks like that version is now also marked `stable`. Fixed.

codelabs/gke-base/install/enable_apis.sh
duftler: Done.

codelabs/gke-kayenta-workshop/services/backend/main.go
duftler: Probably. There are a bunch of things I'd like to change in this driver. I dropped a TODO.

PR #3461, opened 2018-10-16
chore(*) add @maggieneterval to list of approvers

spinnaker-release:
Previously had approver access prior to governance; meets criteria.


danielpeach:
Whoops...

PR #3469, opened 2018-10-17
fix(builds): Adds composite build flag to buildtool to include nebula

ttomsu:
Should fix our nightly build failing to find the `candidate` task


sbwsg:
I'm merging this to try it out in our nightly

Issue #3456, opened 2018-10-16
feat(sponnet): Extend JSONNET library with additional pipeline options


kskewes:
### Issue Summary:

[Sponnet demo](https://github.com/spinnaker/spinnaker/tree/master/sponnet/demo) has initial support for creating pipelines with `Wait` and  `Deploy Manifest` stages.
This is a great start.
It would be nice to extend this with additional pipeline stages, configuration options, execution options, etc.

I have found using MPT? YAML pipelines with partials and multiple levels of inheritance to be fragile and difficult to manage compared with what JSONNET provides.

### Cloud Provider(s):

Kubernetes for now.

### Environment:

1.9.5 on Kubernetes using V2 provider for applications.

### Feature Area: 

sponnet: pipeline templates with JSONNET

### Description:

Expect to be able to create json pipeline definitions with all features available in Deck UI.
Currently can only create json pipeline definitions with `Wait` and `Deploy Manifest (text)` stages.

### Steps to Reproduce:

Try to create a pipeline.json file for demo app with triggers and stages using artifacts.

### Additional Details:

Have started extending this in [my fork of spinnaker repo](https://github.com/kskewes/spinnaker/commit/c3b8f4b72ab14b3369d626f121fb2ce5f1a23be3#diff-16290c192c8881156d617597a2ef886b) and looking for guidance as to whether this would be acceptable for a PR and additional stylistic, structural and general guidelines/improvements.

I understand pipeline templates methodology has been under discussion for some time and that maintaining 'sponnet' would be somewhat duplicating API in areas.

I'm happy to maintain my own library and/or submit PR's and assist with any related issues. I don't have access to the full gamut of providers that Spinnaker support though.

kskewes:
In addition to above I plan to use jsonnet to create and manage Application connfigs and Project configs.

lwander:
Awesome! Can you open a PR? Let's collaborate to build the sponnet library together so you don't have to worry about throwing away what's built already.

kskewes:
Sure, thanks, added.

PR #3472, opened 2018-10-17
feat(sponnet): extend pipeline library

kskewes:
RE: #3456 

Add jenkins stage
Add artifact option for deploy manifest
Add notifications
Add stage timeout
Add moniker
Add expectedArtifacts
Add triggers (docker and git only) constraints TODO

Notes:

The `jsonnet fmt` function is responsible for a most of the changes to pre-existing lines.
Inheritance using a sleek per app `appname.jsonnet` is not catered to here and subject to outcomes from the design doc on MPT v2, future PR's, etc.

Questions:

1. camelCase for variables and functions is common in other projects like Prometheus, Databricks. I'm tempted to change `deploy_manifest` etc. Thoughts?

### DIFF
--- a/sponnet/pipeline.libsonnet
+++ b/sponnet/pipeline.libsonnet
@@ -38,6 +38,9 @@
     gcsObject():: artifact('gcs/object'),
   },
 
+  // expected artifacts
+  // TODO: This section may need splitting out by artifact type due to differing field requirements.
+
   expectedArtifact(id):: {
     id: id,
     // Defaults to kind: custom
@@ -45,6 +48,7 @@
     withMatchArtifact(matchArtifact):: self + {
       matchArtifact+: {
         kind: std.splitLimit(matchArtifact.type, '/', 1)[0],
+        // TODO: For Docker, the name field should be registry and repository.
         name: matchArtifact.name,
         type: matchArtifact.type,
       },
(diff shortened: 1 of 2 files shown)


sponnet/pipeline.libsonnet
```diff
@@ -2,52 +2,118 @@
...
     id: id,
-    withMatchArtifact(matchArtifact):: self + {matchArtifact: matchArtifact},
-    withDefaultArtifact(defaultArtifact):: self + {defaultArtifact: defaultArtifact},
-    withUsePriorArtifact(usePriorArtifact):: self + {usePriorArtifact: usePriorArtifact},
-    withUseDefaultArtifact(useDefaultArtifact):: self + {useDefaultArtifact: useDefaultArtifact},
+    // Defaults to kind: custom
+    defaultArtifact: { kind: 'custom' },
+    withMatchArtifact(matchArtifact):: self + {
+      matchArtifact+: {
+        kind: std.splitLimit(matchArtifact.type, '/', 1)[0],
```
lwander: Kind is just for the UI, and doesn't (as far as I know) always follow this pattern. I'd drop this.

sponnet/pipeline.libsonnet
```diff
@@ -2,52 +2,118 @@
...
-    githubFile():: artifact("github/file"),
-    gcsObject():: artifact("gcs/object"),
+    dockerImage():: artifact('docker/image'),
+    githubFile():: artifact('github/file'),
+    gitlabFile():: artifact('gitlab/file'),
+    gcsObject():: artifact('gcs/object'),
   },
 
+  // expected artifacts
+  // TODO: This section may need splitting out by artifact type due to differing field requirements.
```
lwander: :+1:

sponnet/pipeline.libsonnet
kskewes: Dropped.

kskewes: https://github.com/spinnaker/spinnaker/issues/3491

lwander:
Nice work!

lwander:
Also, :+1: to changing from `deploy_manifest` to `deployManifest`, given that we are camel case in most of the sponnet library anyhow. Let's do that in a separate PR

PR #3499, opened 2018-10-24
feat(sponnet): camelCase, additional artifacts and stages

kskewes:
- Style: camelCase stage types and demo.jsonnet variables.
- Add additional artifact types.
- Add Kubernetes FindArtifactsFromResource stage
- fix Kubernetes DeployManifest stage correctly sets source artifact or text
  based on whether artifact supplied or not.
- Extend Jenkins stage - add properties file and configuration options
- Extend Wait stage - add custom skip message
- Add pipeline stage
- Add Wercker stage, untested as don't have access to a Wercker install.

Picked up some missing configuration options from initial stage add.

Moving on now to putting into use internally. Will come back to add more missing stages and options 
 in a future PR later.

### DIFF
--- a/sponnet/pipeline.libsonnet
+++ b/sponnet/pipeline.libsonnet
@@ -72,7 +72,7 @@
 
   notification:: {
     withAddress(address):: self + { address: address },
-    withCC(address):: self + { address: address },
+    withCC(cc):: self + { cc: cc },
     withLevel(level):: self + { level: level },
     // Custom notification messages are optional
     withWhen(when, message=false):: self + {
--- a/sponnet/pipeline.libsonnet
+++ b/sponnet/pipeline.libsonnet
@@ -49,11 +49,8 @@
 
   expectedArtifact(id):: {
     id: id,
-    // Defaults to kind: custom
-    defaultArtifact: { kind: 'custom' },
     withMatchArtifact(matchArtifact):: self + {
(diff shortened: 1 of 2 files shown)


kskewes:
Forgot need to refactor notifications before we can put to use here. Almost done will send commit tomorrow.

kskewes:
@lwander - Would you mind having a look when free?
Sorry the first commit message was ambiguous. Will be a few weeks/months+ before can look at additional stages and things.

sponnet/pipeline.libsonnet
```diff
@@ -63,26 +69,23 @@
   },
 
   // notifications
-  // TODO: refactor level to include stage level.
 
-  notification(address, type, conditions, when, message):: {
+  notification:: {
     withAddress(address):: self + { address: address },
-    level: 'pipeline',
+    withCC(address):: self + { address: address },
```
lwander: I address used by all notification types, not just email?

kskewes: Yes. `address` field is used for hipchat room, slack room, sms phone number, email address, etc.
Only email has an additional field available to it (`cc`) and have just pushed the naming fix for that now.
I hadn't bothered to split out `notification` into different types but can do. Might be more useful when the Github stuff comes along.

PR #3582, opened 2018-11-09
chore(validate): Pass metric filters to halyard to install.

ewiseblatt:
@jtk54  -- if you are around.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -1480,9 +1480,9 @@ class MonitoringConfigurator(Configurator):
   def add_files_to_upload(self, options, file_set):
     if options.monitoring_filter_dir:
       logging.info('Taring %s', options.monitoring_filter_dir)
-      os.system('tar cf monitoring_filter.tz -C %s .'
+      os.system('tar cf monitoring_filters.tar -C %s .'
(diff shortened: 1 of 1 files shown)


dev/validate_bom__config.py
```diff
@@ -1501,9 +1501,12 @@ def add_config(self, options, script):
                     ' --push-gateway {gateway}'
                     .format(gateway=options.monitoring_prometheus_gateway))
     if options.monitoring_filter_dir:
-      script.append('sudo mkdir -p /opt/spinnaker-monitoring/filters')
-      script.append('cat monitoring_filter.tz'
-                    ' | (cd /opt/spinnaker-monitoring/filters; sudo tar xf -)')
+      # Unpack the tar file into halyard's config directory.
+      script.extend([
+          'mkdir -p ~/.hal/default/profiles/monitoring-daemon/filters',
```
jtk54: There was a `sudo` dropped here, still needed?

ewiseblatt: Before it was dropping directly into the install directory under /opt.
Now it is dropping into halyards config directory, under $HOME, where halyard can pick it up.
Recent versions of halyard know about this.

PR #3614, opened 2018-11-16
feat(spinbot): automated cherrypicks

lwander:


spinbot/event/pull_request_cherry_pick_event_handler.py
```diff
@@ -0,0 +1,53 @@
+from gh import AddLabel, RemoveLabel, ParseCommitMessage
+from .handler import Handler
+from .command import GetCommands
+from .pull_request_event import GetPullRequest, GetRepo
+
+invalid_command_format = ("You must specify exactly 1 release to cherry-pick " +
```
jtk54: Would it make sense to extend this to a list? We sometimes pick changes into several versions.

lwander: My thinking was you just write:
> @spinnakerbot cherry-pick 1.10
> @spinnakerbot cherry-pick 1.9
> ...
WDYT?

jtk54: Seems OK. If this gets annoying we can change later.

lwander: btw that can all be done in 1 comment

PR #3619, opened 2018-11-19
feat(validate): Added --halyard_profile_dir

ewiseblatt:
@jtk54


dev/validate_bom__config.py
```diff
@@ -1524,11 +1524,18 @@ class SpinnakerConfigurator(Configurator):
 
   def init_argument_parser(self, parser, defaults):
     """Implements interface."""
-    pass
+    add_parser_argument(
+        parser, 'halyard_profile_dir', defaults, None,
+        help='Directory for custom service setting files'
```
jtk54: Maybe call out here that this is an 'input' directory and not where the directories will be written to.

PR #3652, opened 2018-11-27
fix(validate): Confirm AWS tags when looking for public ip

ewiseblatt:
@duftler


dev/validate_bom__deploy.py
```diff
@@ -1004,14 +1019,17 @@ def do_determine_instance_ip(self):
               server='ec2'))
 
     try:
-      ip = found[0]['Instances'][0]['PublicIpAddress']
+      # Although we filtered, sometimes was CLI returns others.
```
duftler: s/was/aws

ewiseblatt: heh

PR #3688, opened 2018-12-05
feat(deck): first draft at a functional test suite

sbwsg:
This is a first pass at deck functional tests. See `README.md` for the overall approach taken and see `wdio.conf.js` for the test runner's setup and teardown code.  The tests themselves are under the `deck-functional-tests/src` directory.  Just two at the moment - one smoke test to access the website and one to verify that Cores and Memory are listed correctly in the GCE clone modal.

Sorry for the size of the PR. The thousands of lines come from network captures that get used as test fixtures to mock out Gate's responses. These may end up going into a bucket in future so as to not balloon repository size further.


deck-functional-tests/docker/context/run-tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+# Update deck to latest master branch, build production assets & execute functional tests
+
+# This file is intended to be run inside a built deck-functional-tests container
+
+set -x
```
ewiseblatt: these are redundant

deck-functional-tests/docker/context/run-tests.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+# Update deck to latest master branch, build production assets & execute functional tests
```
ewiseblatt: why master?

deck-functional-tests/docker/context/update-everything.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# This file is intended to be run inside a built deck-functional-tests container
+
+cd /
+
+rm -rf /deck
+rm -rf /deck-functional-tests
+rm -rf /spinnaker
+
+git clone --depth 1 --branch master https://github.com/spinnaker/deck /deck
```
ewiseblatt: shoudl be able to test release branches, not just master.

deck-functional-tests/docker/Dockerfile.local
```diff
@@ -0,0 +1,7 @@
+FROM node:8-stretch
+LABEL last_update="2018-10-01 13:31"
```
ewiseblatt: what's this date?

deck-functional-tests/docker/context/install.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+# Install dependencies required to run Deck's functional tests
+
+# This file is intended to be run inside a debian stretch docker container
+
+echo 'deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main' | tee /etc/apt/sources.list.d/google-chrome.list
```
ewiseblatt: is this arch appropriate? Probably ok. just asking.

deck-functional-tests/docker/context/update-everything.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
+
+set -x
+set -e
+
+# Update deck to latest master branch and build production assets
```
ewiseblatt: why master?

deck-functional-tests/docker/publish-ci.sh
```diff
@@ -0,0 +1 @@
+gcloud docker -- push gcr.io/hebridean-sun/deck-functional-tests:ci
```
ewiseblatt: Do we need to tie these to GCP?
That means OSS users on other platforms cant fully benefit.
At least structure this to provide guidance for other platform providers to accomodate themselves and let them do the work. But provide the direction.

Also should there be a #!/usr/bin/bash env here?

deck-functional-tests/kubernetes/execute-tests-cron-job.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: batch/v1beta1
```
ewiseblatt: maybe but not sure if this policy belongs here or not.

deck-functional-tests/docker/build-ci.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+
+# Call this script from the root of the deck-functional-tests directory
+
+BUILD_TAG="deck-functional-tests:ci"
+PUBLISH_TAG="gcr.io/hebridean-sun/deck-functional-tests:ci"
```
ewiseblatt: What is this URL? What is "hebridean-sun"?

deck-functional-tests/package.json
```diff
@@ -0,0 +1,31 @@
+{
```
ewiseblatt: Does this need to be so precise and decoupled from deck?
Is there a way to tie deck together?

deck-functional-tests/docker/Dockerfile.local
sbwsg: I've removed this.  I added it originally as a quick way to invalidate the Docker build cache. When the date is updated all subsequent steps are re-run even if the contents of those commands are exactly the same.  No longer necessary I don't think since there's a `docker build --no-cache` flag that I wasn't originally aware of.

deck-functional-tests/docker/build-ci.sh
sbwsg: It's the GCP project I test work in. Once this process is solidified the CI image would ideally be available publicly for non-Google users to use.  This should be parameterized.

deck-functional-tests/docker/context/run-tests.sh
sbwsg: I'll parameterize this.

deck-functional-tests/package.json
ewiseblatt: Offline I propose perhaps this should be in deck repo so versions align. Can have a separate feature branch for now if it isnt "stable enough" to be in master itself. I dont know that it's a good idea, but seems like a better long term fit given its coupling and scoping to deck.

sbwsg:
@ewiseblatt made the excellent point that we want the tests to be versioned alongside Deck so that, for example, a 1.12 release is only tested using a 1.12-era suite. So this almost certainly needs to go in to the Deck repo to avoid maintenance headaches.

deck-functional-tests/docker/build-ci.sh
ewiseblatt: I think this needs to accomodate OSS users creating their own stuff. I'm not sure what's in the image, but if I have a custom deck extension I shoudl be able to test it using this framework in which case I might have things that arent going to be in the public image.

Also there are versioning concerns here. We would want to test 1.11 and master (and eventually 1.12) independent of one another.

Issue #3740, opened 2018-12-17
Expected Artifacts aren't displayed when pipline is created by using pipeline.libsonnet


Mahito:
### Issue Summary:

Expected Artifacts aren't displayed when pipline is created by using pipeline.libsonnet

### Cloud Provider(s):

kubernetes

### Environment:
I am running Spinnaker v1.11.0 on GKE

### Feature Area: 

pipeline.libsonnet

### Description:

I expected that the pipeline was made correctly, but expected Artifacts aren't displayed when pipline is created by using pipeline.libsonnet.

### Steps to Reproduce:

See Sponnet demo 
https://github.com/spinnaker/spinnaker/blob/master/sponnet/demo/README.md

1. `git clone https://github.com/spinnaker/spinnaker.git`
2. `cd spinnaker/sponnet/demo`
3. To install jsonnet
3. To build the demo application
`jsonnet demo-app.jsonnet > demo-app.json && spin application save --file demo-app.json`
4. To build the demo pipeline
`jsonnet demo-pipeline.jsonnet > demo-pipeline.json && spin pipeline save --file demo-pipeline.json`
5. Go to pipeline page on Deck

### Additional Details:

<img width="595" alt="2018-12-17 17 24 08" src="https://user-images.githubusercontent.com/210692/50075066-35205b00-0221-11e9-824a-9fa47b686581.png">

kskewes:
The `kind` attribute was removed from sponnet here: https://github.com/spinnaker/spinnaker/pull/3472/commits/73e3a6e6683fe129261c854c2c42cf6208f703b7

`kind` is not required to save pipelines or for pipelines to correctly operate so I think the idea is that Deck should infer kind and present appropriately (see referenced Deck PR in above sponnet PR).
For now, as you discovered Deck does still need `kind` for presenting the artifacts correctly.
So I agree that `kind` is useful and have been maintaining the kind patch from my PR above in our local version of sponnet. 

I wonder if I we should add it back in, either with your PR or mine. Or there is something else we can do with Deck to complete the empty boxes.
@lwander thoughts?

lwander:
I think when discussed, we figured this is best handled in Deck, rather than require all other clients to meet Deck's requirements. @sbwsg how feasible is it to remove the requirement to have (or infer) `kind`?

sbwsg:
[...]

PR #3741, opened 2018-12-17
fix(sponnet) Add kind attribute to Artifact

Mahito:
Artifacts are not displayed properly on Deck because pipeline.libsonnet doesn't have 'kind' attribute in Artifact.
This PR adds 'kind' attribute to Artifact in pipeline.libsonnet.

close: #3740

### DIFF
--- a/sponnet/pipeline.libsonnet
+++ b/sponnet/pipeline.libsonnet
@@ -25,8 +25,9 @@
 
   // artifacts
 
-  artifact(type):: {
+  artifact(type, kind):: {
     type: type,
+    kind: kind,
     withArtifactAccount(artifactAccount):: self + { artifactAccount: artifactAccount },
     withLocation(location):: self + { location: location },
     withName(name):: self + { name: name },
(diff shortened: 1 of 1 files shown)


sponnet/pipeline.libsonnet
```diff
@@ -36,15 +37,15 @@
...
-    embeddedBase64():: artifact('embedded/base64'),
-    gcsObject():: artifact('gcs/object'),
-    githubFile():: artifact('github/file'),
-    gitlabFile():: artifact('gitlab/file'),
-    httpFile():: artifact('http/file'),
+    bitbucketFile():: artifact('bitbucket/file', 'custom'),
+    dockerImage():: artifact('docker/image', 'docker'),
+    embeddedBase64():: artifact('embedded/base64', 'base64'),
+    gcsObject():: artifact('gcs/object', 'custom'),
+    githubFile():: artifact('github/file', 'custom'),
```
lwander: I think for most of these (check the source here) you can be more specific about the kind: https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/src/pipeline/config/triggers/artifacts/github/github.artifact.ts#L12

Mahito: Thank you for your comment ! It looks like a good change.
Should I change the other 'custom' to key too ?

lwander: Yes I would change all the kinds that have some sort of adapter here: https://github.com/spinnaker/deck/tree/master/app/scripts/modules/core/src/pipeline/config/triggers/artifacts

Mahito: I changed the kinds.
I didn't change the kind of `kubernetesObject`, because I couldn't find the key.
https://github.com/spinnaker/spinnaker/pull/3741/files#diff-98560c97d4b23f0cf12af197f63a07b2R48

lwander:
Thanks @Mahito!

@ezimanyi will this work with the changes you've made in deck & echo? This clearly is beneficial for older installs of Spinnaker, but we should be sure this works for future releases 1.12+ with your fixes as well

ezimanyi:
After my changes, the artifact model is as follows:
* There is no `kind` field; the UI rendering is entirely inferred from the existing `type` field
* There is a new field `customKind` that indicates whether the artifact is custom-configured (ie, by selecting 'custom' in the drop-down) and should be displayed in the UI using the full custom-configuring editor.  The field is optional, and defaults to `false`.

So for Spinnaker 1.12+ I'd recommend not sending the `kind` at all, and only sending `customKind` if you have an artifact that is one of the standard types but you've configured it in a non-standard way and want the custom editor to show up.  (I think the use case for `customKind` is pretty narrow, but it's there to account for that particular case.)  You can choose whether or not to support `customKind` in sponnet, but it's not necessary except for this particular use case; pipelines will work if it's absent.

The change is backwards compatible, so you can still send the `kind` going forward, though of course it would be nice to stop including this in pipelines generated by sponnet once users have upgraded to 1.12+.  On Spinnaker 1.12+, upon editing an artifact in the UI, the `kind` will be removed and (if applicable) the `customKind` will be added.

lwander:
Ok, in that case I will merge now, we can revisit (and potentially version) sponnet going forward.

Issue #3744, opened 2018-12-17
Click-to-deploy broken?


dorbin:
### Issue Summary:
[GCE Source to Prod codelab](https://www.spinnaker.io/guides/tutorials/codelabs/gce-source-to-prod/) broken; the [click-to-deploy image](https://console.google.com/marketplace/browse?q=spinnaker&project=spinnaker-beginnaker-test5&organizationId=433637338589) might be broken too.

### Cloud Provider(s):
GCE

### Description:
Going through the [GCE Source to Prod codelab](https://www.spinnaker.io/guides/tutorials/codelabs/gce-source-to-prod/), when I get to the point where I'm configuring the trigger, the Jenkins master does not show up in the UI. And Jenkins is not even enabled. 

I was able to manuall `hal config ci jenkins enable`, and add the master too. But there seems to be something wrong in the script(s) that breaks the codelab.

### Additional Details:

---

duftler:
Have reproduced the issue. The masters pulldown is not populated.

duftler:
The configuration scripts look correct, but I see this early on in `/var/log/syslog`:

```
Dec 18 15:48:21 gce-codelab-1 ec2: -----END SSH HOST KEY FINGERPRINTS-----
Dec 18 15:48:21 gce-codelab-1 ec2: #############################################################
Dec 18 15:48:25 gce-codelab-1 startup-script: INFO startup-script: #033[31m#033[1m! ERROR #033[0mFailed to connect to localhost/127.0.0.1:8064#033[0m#033[0m
Dec 18 15:48:25 gce-codelab-1 startup-script: INFO startup-script: #033[35m#033[1m? #033[0mIs your daemon running?#033[0m#033[0m
Dec 18 15:48:29 gce-codelab-1 startup-script: INFO startup-script: The password of the jenkins user to authenticate as.: #033[31m#033[1m! ERROR #033[0mFailed to connect to localhost/127.0.0.1:8064#033[0m#033[0m
Dec 18 15:48:29 gce-codelab-1 startup-script: INFO startup-script: #033[35m#033[1m? #033[0mIs your daemon running?#033[0m#033[0m
Dec 18 15:48:31 gce-codelab-1 startup-script: INFO startup-script: Waiting for halyard to start running...
[...]

PR #3756, opened 2018-12-18
fix(codelab/gce): Wait for halyard before issuing hal command.

duftler:


duftler:
Fixes https://github.com/spinnaker/spinnaker/issues/3744

PR #3762, opened 2018-12-19
feat(spinbot): no-lifecycle label support

lwander:
Allows spinbot to ignore stale/to-be-closed behavior on issues with this label.


lwander:
@ethanfrogers FYI this is live now

PR #3765, opened 2018-12-19
fix(buildtool/spin): Accomodates golang build assumptions.

jtk54:
Golang makes assumptions about where the source repo lives.
The previous build implementation was incorrectly building stale
spin binaries since we didn't know/follow go's assumptions.


dev/buildtool/spin_commands.py
```diff
@@ -84,15 +83,33 @@ def _do_can_skip_repository(self, repository):
...
 
-    check_subprocess('go get -d -v', cwd=config_root)
+    # NOTE: go build is opinionated about where the source repository lives --
+    # it expects the source repository to be under $GOPATH/src/<package path>,
+    # and will use that the source code in that directory *even if you clone
+    # your own copy of the source repo*.
+    #
+    # To facilitate go's pattern, we operate on the source under
+    # $GOPATH/src/<path to spin> to build spin.
+    gopath = check_subprocess('go env GOPATH')
```
ewiseblatt: can you set the GOPATH to be the output directory for this command?
Pass env={'GOPATH': path} to the subprocess commands?
That way the command side effects are still encapsulated within this filesystem.

jtk54: I'll give it a try. I thought the first time I tried that I ran into trouble dynamically setting `GOPATH`, but I just walked through it again now successfully. Making changes and re-testing.

jtk54: Getting a lot of this during testing: 

```
package _/var/lib/jenkins/workspace/Build_PrimaryArtifacts/build_output/build_spin: unrecognized import path "_/var/lib/jenkins/workspace/Build_PrimaryArtifacts/build_output/build_spin" (import path does not begin with hostname)
```

Seems that go's assumptions pretty severely breaks the buildtool's `build_input -> build_output` pattern due to the package naming resolution.

jtk54: I'm going to revert this back to operating on go's opinions of where the source repositories live reading from `go env`.

ewiseblatt: can I see what it looked like ? I dont see how that would matter.

You need input_path for the source analysis stuff (changelog, etc).
Since go is opinionated about where it goes, i guess you can make the GOPATH the commands build_output. Then when you run the command it will checkout again and build in the output dir. It breaks assumptions but is technically ok. The downside is an extra checkout and when we archive jenkins builds, we'll be archiving that source tree as well. So perhaps the GOPATH could be a different input directory (a subdir within the commands input dir).

The output directory is only used to direct log files and such. I dont know why your commands would even know about it. Perhaps you are executing in that directory. You can pass a different cwd to the input directory, but if you are setting GOPATH I'm not sure why go would be confused.

jtk54: I didn't see this before I merged so I assumed everything was OK. I'll follow up with another PR if I can clean any of this up. Sorry.

PR #3885, opened 2019-01-26
fix(approvers): update danielpeach  username

ethanfrogers:
dpeach is not @danielpeach username. also, since daniel isn't working on this team anymore, do we want to continue to allow approvals?

### DIFF
--- a/approvers.md
+++ b/approvers.md
@@ -13,7 +13,7 @@
 * csanden
 * dibyom
 * dorbin
-* dpeach
+* danielpeach
 * dreynaud
 * duftler
 * edwinavalos


ethanfrogers:
@lwander @skim1420

skim1420:
This should go up to the TOC.

lwander:
We haven't removed approvers/reviewers yet (AFAIK) for leaving a Spinnaker team. Let's add Dan's correct username for now, and decide on a more general policy later.

lwander:
Oops, didn't refresh before posting that. Will take it up with the TOC

ttomsu:
TOC approved.

PR #3912, opened 2019-01-30
feat(buildtool/spinnaker): Cuts branches for spin cli repo.

jtk54:


dev/buildtool/__init__.py
```diff
@@ -12,6 +12,9 @@
 # For building and validating a release
 SPINNAKER_PROCESS_REPOSITORY_NAMES = ['spinnaker']
 
+# Additional tools that accompany a release
+SPINNAKER_ADDITIONAL_REPOSITORY_NAMES = ['spin']
```
ewiseblatt: I dont like this name because it is only used by the spinnaker commands.
Build debians, etc does not operate on it. And for that matter, the spin commands
dont either.

I would this something like
SPIN_REPOSITORY_NAME (or list of NAMES) and use this in spin_commands as well to
tie the two references together.

jtk54: Done, PTAL.

Issue #175, opened 2014-11-03
Adding detach/attach as possible instance actions


aglover:
**Submitted by:**
mhuang@netflix.com

**From page:**
http://deck.prod.netflix.net/#/applications/api/clusters/prod/api-prod-c0us.ca/instanceDetails?instanceId=i-e533cd04

**Description:**
I don't think I have access to the Spinnaker repository, so I'm not sure if this is already a feature in the pipeline, but it will be extremely useful to be able to detach an existing instance from an ASG so that the ASG no longer manages its lifecycle.  This is handy when attempting to triage a problematic instance without worrying that the ASG will terminate it.  Similarly, it will be useful as well to expose the attach functionality to attach an orphaned instance back to an ASG.  Here are the AWS CLI commands for this functionality:

http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/detach-instance-asg.html http://docs.aws.amazon.com/AutoScaling/latest/DeveloperGuide/attach-instance-asg.html

PR #3958, opened 2019-02-11
fix(buildtool/spin): Set internal patch version properly.

jtk54:
Also fixes tag pushing on publish.

### DIFF
--- a/dev/buildtool/spin_commands.py
+++ b/dev/buildtool/spin_commands.py
@@ -138,17 +138,10 @@ class BuildSpinCommand(RepositoryCommandProcessor):
                   'GOOS': dist_arch.dist,
                   'GOARCH': dist_arch.arch})
 
-      # Note: spin CLI is coupled to the Gate major and minor version.
-      # Gate is a routing server, so features and breaking changes in Gate
-      # must be reflected in spin since it is a client.
-      dash = self.__gate_version.find('-')
-      gate_semver = self.__gate_version[:dash]
-
       version_package_prefix = os.path.join(spin_package_path, 'version')
-
       # Unset ReleasePhase tag for proper versions.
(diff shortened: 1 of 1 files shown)


dev/buildtool/spin_commands.py
```diff
@@ -171,6 +165,34 @@ def _do_repository(self, repository):
...
+  def __determine_internal_version(self, repository):
+    # Note: spin CLI is coupled to the Gate major and minor version.
+    # Gate is a routing server, so features and breaking changes in Gate
+    # must be reflected in spin since it is a client.
+    git_dir = repository.git_dir
+    git = self.__scm.git
+    match = re.match(r'(\d+)\.(\d+)\.(\d+)-\d+', self.__gate_version)
+    gate_major = match.group(1)
+    gate_min = match.group(2)
+    if match is None:
```
ewiseblatt: Move check up before #175

dev/buildtool/spin_commands.py
```diff
@@ -171,6 +165,34 @@ def _do_repository(self, repository):
...
+
+    tag_matcher = re.compile(r'version-{maj}.{min}.(\d+)'
+                             .format(maj=gate_major, min=gate_min))
+    tags = git.fetch_tags(git_dir)
+    tag_matches = [tag_matcher.match(t) for t in tags if tag_matcher.match(t)]
+    if tag_matches:
+      patch_versions = [int(m.group(1)) for m in tag_matches]
+      max_patch = max(patch_versions)
+      patch = str(max_patch + 1)
+      self.__internal_version = '{major}.{minor}.{patch}'.format(
```
ewiseblatt: why set this rather than return it?
You only use it in the one place calling this function.

dev/buildtool/spin_commands.py
jtk54: Done.

dev/buildtool/spin_commands.py
jtk54: Done.

PR #836, opened 2016-04-01
Replace distro -ne check with -lt.

duftler:
Closes #835.

PR #4051, opened 2019-02-27
feat(outbound): Add configs to make outbound calls to an endpoint rather than port-forwarding.

joonlim:


dev/validate_bom__test.py
```diff
@@ -735,6 +818,13 @@ def validate_test_requirements(self, test_name, spec, metric_labels):
       raise_and_log_error(
           UnexpectedError('Test "{name}" is missing an "api" spec.'.format(
               name=test_name)))
+
+    if spec['api'] in self.__service_endpoints:
```
ewiseblatt: I think this is premature. You still want to check the requirements are satisfied. I think this comes into play within wait_on_service() ~#578

dev/validate_bom__test.py
```diff
@@ -593,6 +630,52 @@ def wait_on_service(self, service_name, port=None, timeout=None):
...
 
+  def __wait_on_service_using_endpoint(self, service_name, timeout=None):
+    endpoint = self.__service_endpoints[service_name]
+
+    timeout = timeout or self.options.test_service_startup_timeout
+    end_time = time.time() + timeout
+    logging.info('Waiting on "%s..."', service_name)
+
+    threadid = hex(threading.current_thread().ident)
+    logging.info('WaitOn polling %s from thread %s', service_name, threadid)
```
ewiseblatt: You can use %x instead of hex()

dev/validate_bom__test.py
```diff
@@ -593,6 +630,52 @@ def wait_on_service(self, service_name, port=None, timeout=None):
...
+    except Exception as error:
+      raise_and_log_error(
+          ResponseError('{0} service endpoint got {1}'.format(service_name, error),
+                        server=endpoint))
+    except URLError as error:
+      if time.time() >= end_time:
+        logging.error(
+            'Timing out waiting for %s | %s', service_name, threadid)
+        raise_and_log_error(TimeoutError(service_name, cause=service_name))
+      time.sleep(2.0)
```
ewiseblatt: what is this sleep for?

testing/citest/spinnaker_testing/spinnaker_test_scenario.py
```diff
@@ -166,6 +166,14 @@ def _init_spinnaker_bindings_builder(cls, builder, defaults):
         '--test_app', default=defaults.get('TEST_APP', cls.__name__.lower()),
         help='Default Spinnaker application name to use with test.')
 
+    builder.add_argument(
+        '--outbound_hostname', default=defaults.get('OUTBOUND_HOSTNAME', None),
+        help='') # TODO(joonlim)
```
ewiseblatt: No time like the present to explain these.

testing/citest/spinnaker_testing/spinnaker_test_scenario.py
```diff
@@ -136,7 +136,7 @@ def _init_spinnaker_bindings_builder(cls, builder, defaults):
     # It probably isnt useful anymore.
     builder.add_argument(
         '--host_platform', default=defaults.get('HOST_PLATFORM', None),
-        help='Platform running spinnaker (gce, native).'
+        help='Platform running spinnaker (gce, native, outbound).'
```
ewiseblatt: can you explain "outbound" somewhere? Is this "proxied"?

dev/iap_generate_auth_token.py
```diff
@@ -0,0 +1,73 @@
...
+  """
+  bootstrap_credentials = google.oauth2.service_account.Credentials.from_service_account_file(service_account_file)
+  if isinstance(bootstrap_credentials,
+                google.oauth2.credentials.Credentials):
+    raise Exception('iap_generate_auth_token is only supported for service '
+                    'accounts.')
+  elif isinstance(bootstrap_credentials,
+                  google.auth.app_engine.Credentials):
+    requests_toolbelt.adapters.appengine.monkeypatch()
+
```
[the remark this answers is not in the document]
joonlim: No else needed here for our case, just continue.

dev/validate_bom__test.py
```diff
@@ -735,6 +818,13 @@ def validate_test_requirements(self, test_name, spec, metric_labels):
       raise_and_log_error(
           UnexpectedError('Test "{name}" is missing an "api" spec.'.format(
               name=test_name)))
+
+    if spec['api'] in self.__service_endpoints:
+      # No need to wait on service if we are using the service endpoint.
```
ewiseblatt: just because the api is for consideration does not mean that an endpoint is defined. You either need to check here, or remove from teh map the unconfigured endpoints.

dev/validate_bom__test.py
```diff
@@ -420,6 +430,35 @@ def __init__(self, deployer):
...
+        gate_service_config['service_account_email'] = get_service_account_email(self.options.test_gate_iap_credentials)
+      outbound_service_configs['gate'] = gate_service_config
+
+    self.__outbound_service_configs = outbound_service_configs
+
+  def __gate_service_bearer_auth_token_or_none(self):
+    iap_credentials_file = self.options.test_gate_iap_credentials
+    iap_client_id = self.options.test_gate_iap_client_id
+
+    if iap_credentials_file and not iap_client_id:
```
ewiseblatt:
```
def __bearer_auth_token_or_none(self, credentials_path, client_id):
if (credentials_path or '' != '') != (client_id or '' != ''):
   raise_and_log_error(ConfigError('Either both test_{service}_iap_credentials and test_{service}_iap_client_id must be set or neither.'.format(service=service)))
if credentials_path:
  return generate_auth_token(credentials_path, client_id)
return None
```

dev/validate_bom__test.py
```diff
@@ -420,6 +430,35 @@ def __init__(self, deployer):
         'echo-worker': 8089
     }
 
+    # Map of services with provided hosts and credentials.
+    outbound_service_configs = {}
+    if self.options.test_gate_service_host:
+      gate_service_config = {
+        'host': self.options.test_gate_service_host,
+      }
+      bearer_auth_token = self.__gate_service_bearer_auth_token_or_none()
```
ewiseblatt:
```
gate_credentials_path = self.options.test_gate_iap_credentials
if gate_credentials_path:
  gate_service_config['service_account_email'] = get_service_account_email(gate_credentials_path)
  client_id = self.options.test_gate_iap_client_id
  auth_token = self.__bearer_auth_token_or_none('gate', gate_credentials_path, client_id)
  if auth_token:
    gate_service_config['bearer_auth_token'] = auth_token
```
And you can wrap 435..443 into a add_gate_service_config(configs) function that
```
if not self.options.test_gate_service_host:
   return
436...444 but can remove "gate_" from local variables.
```

Then replace 435...444 with self.add_gate_service_config(outbond_service_configs)

dev/validate_bom__test.py
```diff
@@ -752,8 +830,15 @@ def validate_test_requirements(self, test_name, spec, metric_labels):
                                   'IncompatableConfig', metric_labels)
         return False
 
+    if spec['api'] in self.__outbound_service_configs:
+      # No need to wait on service if we are using the outbound service host.
+      service_name = spec['api']
+      self.__wait_on_service_using_host(service_name)
+      return True
```
ewiseblatt: still feels like this is premature. It skips the checks at 845 and beyond.
Coudl wait_on_service() figure out if it should use #836?

testing/citest/spinnaker_testing/scrape_spring_config.py
```diff
@@ -81,7 +81,7 @@ def infer(json):
   return expr_dict
 
 
-def scrape_spring_config(url, timeout=60, empty_if_404=True):
+def scrape_spring_config(url, timeout=60, empty_if_404=True, bearer_auth_token=None):
```
ewiseblatt: Seems like bearer_auth_token=None should instead be a list of headers.
The caller with a bearer auth token would pass in a list containing the authorization header.
If that's hard it could be a refactorization for later if other headers are needed.
But if it is all the same, it feels like this should be generalized.

testing/citest/spinnaker_testing/spinnaker.py
```diff
@@ -88,6 +88,14 @@ def name_value_to_dict(content):
     result[match.group(1)] = match.group(2).strip()
   return result
 
+def construct_base_url(host, port, protocol=None):
+  """Constructs a base_url from host, port, and an optional protocol.
+  """
+  return '{protocol}{host}:{port}'.format(
```
ewiseblatt: should be something like "http" or "https", so you should be adding in the "//" separator between the URL's protocol and host.

testing/citest/spinnaker_testing/spinnaker.py
```diff
@@ -358,13 +366,13 @@ def new_instance_from_bindings(cls, name, status_factory, bindings, port):
     """
     host_platform = cls.__determine_host_platform(bindings)
     if host_platform == 'native':
-      base_url = (None if not bindings['NATIVE_HOSTNAME']
-                  else 'http://{host}:{port}'.format(
-                      host=bindings['NATIVE_HOSTNAME'],
-                      port=bindings['NATIVE_PORT'] or port))
-
+      base_url = None
+      if bindings['NATIVE_HOSTNAME']:
```
ewiseblatt: why would the hostname include the protocol? That isnt a hostname. Is that a real thing?
Why not add a NATIVE_PROTOCOL binding to indicate http vs https?

dev/validate_bom__test.py
```diff
@@ -861,9 +946,35 @@ def make_test_command_or_none(self, test_name, spec, metric_labels):
...
-        '--native_host', 'localhost',
-        '--native_port', str(self.__forwarded_ports[microservice_api].port)
     ]
+    if microservice_api in self.__outbound_service_configs:
+      service_config = self.__outbound_service_configs[microservice_api]
+      host = service_config['host']
+      command.extend([
+          '--native_host', host
+      ])
+      if host.startswith('http://'):
```
ewiseblatt: This is odd.
Why not either have a service_config['host'], service_config['port'], service_config['protocol'] or a service_config['base_url']? If the service_config['port'] is not defined then you coudl default it based on the protocol.

dev/validate_bom__test.py
```diff
@@ -861,9 +946,35 @@ def make_test_command_or_none(self, test_name, spec, metric_labels):
...
+        command.extend([
+            '--native_port', '80'
+        ])
+      elif host.startswith('https://'):
+        command.extend([
+            '--native_port', '443'
+        ])
+      if 'bearer_auth_token' in service_config:
+        command.extend([
```
ewiseblatt: It is not ideal to be passing this on the command line from a security perspective, though would be ok especially since the token has a relatively short lifetime.

I dont think citest supports options from a file. If not then I'll add it later. But since you are adding this parameter in now, how about support passing it through an environment variable?

dev/validate_bom__test.py
```diff
@@ -593,6 +632,44 @@ def wait_on_service(self, service_name, port=None, timeout=None):
...
+  def __wait_on_service_using_host(self, service_name, timeout=None):
+    service_config = self.__outbound_service_configs[service_name]
+    host = service_config['host']
+
+    timeout = timeout or self.options.test_service_startup_timeout
+    end_time = time.time() + timeout
+    logging.info('Waiting on "%s..."', service_name)
+
+    threadid = threading.current_thread().ident
+    logging.info('WaitOn polling %s from thread %x', service_name, threadid)
```
ewiseblatt: does logging include the thread already? It's ok to leave given the concurrency alignment going through here.

dev/validate_bom__test.py
```diff
@@ -593,6 +632,44 @@ def wait_on_service(self, service_name, port=None, timeout=None):
...
+    try:
+      if 'bearer_auth_token' in service_config:
+        _make_get_request_with_bearer_auth_token('{host}/health'
+                                            .format(host=host),
+                                            service_config['bearer_auth_token'])
+      # Add more authentication cases here.
+      else:
+        urlopen('{host}/health'
+                .format(host=host))
+      logging.info('"%s" is ready on service host %s | %s',
```
ewiseblatt: you used %x as a thread id before if you are trying to align them.
rather than " | %s " you could "threadid=%x" here and " from thread %x" with " threadid=%x" above.

dev/validate_bom__test.py
```diff
@@ -593,6 +632,44 @@ def wait_on_service(self, service_name, port=None, timeout=None):
...
+          ResponseError('{0} service host got {1}'.format(service_name, error),
+                        server=host))
+    except Exception as error:
+      raise_and_log_error(
+          ResponseError('{0} service host got {1}'.format(service_name, error),
+                        server=host))
+    except URLError as error:
+      if time.time() >= end_time:
+        logging.error(
+            'Timing out waiting for %s | %s', service_name, threadid)
```
ewiseblatt: threadid=%x

It seems like you call this one time and not i a loop. So if you have another URL error in here, you are going to eat it whereas in a polling loop you'd retry and perhaps eventually timeout and throw then.

dev/iap_generate_google_auth_token.py
```diff
@@ -75,6 +69,18 @@ def generate_auth_token(service_account_file, client_id):
...
+
+def generate_auth_token(service_account_file, client_id):
+  """Generates an auth token to make requests to an IAP-protected endpoint.
+
+  Args:
+    service_account_file: [string] The path to a credentials file for a service
+       account that can be used to generate the token.
+    client_id: The client ID used by IAP to generate the token.
+  """
+  service_account_credentials =_make_service_account_credentials(service_account_file, client_id)
```
ewiseblatt: s/=_make/= _make/

dev/validate_bom__test.py
```diff
@@ -430,33 +430,30 @@ def __init__(self, deployer):
...
-    if iap_credentials_file and iap_client_id:
-      return generate_auth_token(iap_credentials_file, iap_client_id)
+    # Map of services with provided endpoints and credentials.
+    self.__outbound_service_configs = {}
+    self.__add_gate_service_config(self.__outbound_service_configs)
+
+  def __add_gate_service_config(self, configs):
+    if not self.options.test_gate_service_endpoint:
+      return
+    gate_service_config = {
```
ewiseblatt: all the local variable gate_ prefix here are superfluous because the method itself is scoped to gate.
Removing them will make it easier to read.

dev/validate_bom__test.py
```diff
@@ -578,6 +575,13 @@ def wait_on_service(self, service_name, port=None, timeout=None):
     Returns:
       The ForwardedPort entry for this service.
     """
+    if len(self.__outbound_service_configs) > 0 and service_name not in self.__outbound_service_configs:
```
ewiseblatt: This is equivalent to:
```
if self.__outbound_service_configs:
  if service_name in self.__outbound_service_configs:
     self.__wait_on_service_using_endpoint(service_name)
  return True

try:
 ...
```

But that doesnt look right, does it?
If it is in the outbound_service_configs then use the endpoint, otherwise dont.
Or do you really mean to not wait on the other services? That's ok if intentional,
in which case the simpler above makes it more clear. Throw a comment in as to
why you always return True if a non-empty outbound_service_cofnigs.

Also, didnt we change "outbound" to something else?

dev/validate_bom__test.py
```diff
@@ -1140,8 +1131,9 @@ def init_argument_parser(parser, defaults):
       help='The Jenkins job name to use in tests.')
 
   add_parser_argument(
-      parser, 'test_gate_service_host', defaults, None,
-      help='Gate host to use rather than port-forwarding.')
+      parser, 'test_gate_service_endpoint', defaults, None,
```
ewiseblatt: if this is a url then the name should say url, like "test_gate_service_url".  Then it is clear it is specified as a url.

testing/citest/spinnaker_testing/spinnaker.py
```diff
@@ -367,10 +359,13 @@ def new_instance_from_bindings(cls, name, status_factory, bindings, port):
...
-      if bindings['NATIVE_HOSTNAME']:
-        host = bindings['NATIVE_HOSTNAME']
-        protocol = None if host.startswith('http://') or host.startswith('https://') else 'http://'
-        base_url = construct_base_url(host, bindings['NATIVE_PORT'] or port, protocol)
+      if bindings['NATIVE_BASE_URL']:
+        base_url = bindings['NATIVE_BASE_URL']
+      elif bindings['NATIVE_HOSTNAME']:
+        base_url = 'http://{host}:{port}'.format(
+                   host=bindings['NATIVE_HOSTNAME'],
+                   port=bindings['NATIVE_PORT'] or port)
```
ewiseblatt: If port is a parameter, you cannot give bindings precedence and disregard the parameter.
If you dont always need port then it should be a kwarg "default_port" instead of positional.

joonlim: I believe that the port from bindings should take precedence, as the port parameter is the hardcoded default. I took the logic from the existing code: https://github.com/spinnaker/spinnaker/blob/master/testing/citest/spinnaker_testing/spinnaker.py#L364

dev/validate_bom__test.py
joonlim: I will add a comment indicating that we do not (actually cannot) wait for the other services, since we are expected to communicate with services via a url (instead of port forwarding).

Would it be clearer if we take an explicit --no-port-forward=true or --external-services=true in order to achieve our flow?

dev/validate_bom__test.py
joonlim: Agreed. Would you be opposed to "test_gate_service_base_url"?

ewiseblatt: SGTM

testing/citest/spinnaker_testing/spinnaker.py
ewiseblatt: I agree that's what you want. I was saying that the API is misleading. You have the write policy, but the wrong syntax. At least call it a "default port" and not The Port.

dev/validate_bom__test.py
ewiseblatt: I assume this is all or none. I think maybe the options specify whether the services are on public addresses or private. Private addresses require port forwarding. Or more precisely you need to know the network the services are on and whether they are on your network or not. A private network that you are on would not require port forwarding either.

If you go this general route, then the controller would need to figure out if it is on the same network or not. I think that's kind of what it does now to port forward. It looks up the metadata for the machine it wants to connect to then tries the private network, then the public network. At least for VMs. Kubernetes might be one way. I wonder if you can tell if you are on the same network by looking at the IP address you want to connect to and determining if it is public or private and, if private, see if you have an IP address on that same private network.

Is that too much? For now you can add some kind of option indicating whether to port forward or not. The default for port forward is true so make no-port-forward set port_forward to false.

Actually, lookup up an example for you I came across this:
    parser.add_argument(
        '--deploy_google_use_external_ip',
        dest='deploy_google_use_internal_ip', action='store_false',
        help='DEPRECATED: Use --deploy_google_use_internal_ip=false')

Maybe you can just use that flag. Set --deploy_google_use_internal_ip=false and then use that in the test module to connect with the external address. Does that satisfy your need?

dev/validate_bom__test.py
```diff
@@ -1104,13 +1100,18 @@ def init_argument_parser(parser, defaults):
 
   add_parser_argument(
       parser, 'testing_enabled', defaults, True, type=bool,
-      help='If false then dont run the testing phase.')
+      help='If false then don\'t run the testing phase.')
```
ewiseblatt: s/don\\`t/do not/

testing/citest/spinnaker_testing/spinnaker.py
```diff
@@ -341,6 +341,32 @@ def __determine_host_platform(cls, bindings):
...
+  @classmethod
+  def __determine_native_base_url(cls, bindings, default_port):
+    """Helper function to determine a native host platform's base URL.
+
+    The returned base URL may be None if bindings['NATIVE_BASE_URL'] and
+    bindings['NATIVE_HOSTNAME'] are both missing.
+
+    Args:
+      bindings: [dict] List of bindings to configure the endpoint
+          NATIVE_BASE_URL: The base URL to use, if given.
```
ewiseblatt: Mention that if base_url is not provided then the URL will be constructed using NATIVE_HOSTNAME and NATIVE_PORT and assuming http (no need to parameterize the protocol)

PR #4067, opened 2019-03-04
feat(sponnet): jsonnet libs for v2 MPTs.

jtk54:


jtk54:
@kskewes PTAL.

sponnet/demo/demo-v2-config.jsonnet
```diff
@@ -0,0 +1,16 @@
+local pipelines = import '../pipeline.libsonnet';
+
+local artifact = pipelines.artifact('front50/pipelineTemplate', '')
```
kskewes: The other artifact types (docker/gitlab/etc) are instantiated with `artifacts.type()`. 
I don't know if we would want to have `pipelineTemplate` as a separate `type()` or just leverage the underlying `artifact()` function directly as you have done which has all the required fields. We don't assert input correctness anyway.
I think you have a much better understanding of current and future plans so defer to you. :)

jtk54: That does make sense, I'll add in an explicit `front50/pipelineTemplate` type.

jtk54: Done.

PR #4092, opened 2019-03-09
fix(testing): Fixed gcs pubsub operation

ewiseblatt:
@jtk54 
Seems to work, but I'm not exactly sure there arent other issues (possibly introduced).
Take a critical look.

A few changes here
First, rather than rebinding the operation midstream from the gcs pubsub agent
to the gate agent, construct the operation with both agents and use the pubsub
agent to handle the pubsub operations, but bind the gate agent to the operation.
This is because the gate agent is used to check the status, which is what the
bound agent is for.

The PubsubTriggerOperationStatus seemed to have too much state.
It mostly delegates to the __trigger_status instead. Its only real added
state is whether or not it timed out.

Clarify between the trigger status and the trigger response to avoid confusing
the two.

### DIFF
--- a/testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
+++ b/testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
@@ -13,15 +13,14 @@
 # limitations under the License.
 
 
-import json
 import logging
-import os
+import time
 
-import datetime
-
-from citest.service_testing import (base_agent, http_agent)
+from citest.service_testing import base_agent
 
 from google.cloud import storage
+
+from spinnaker_testing.gate import GateAgent
 
 
 class GcsFileUploadAgent(base_agent.BaseAgent):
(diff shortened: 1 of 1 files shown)


testing/citest/spinnaker_testing/gcs_pubsub_trigger_agent.py
```diff
@@ -71,131 +71,128 @@ def new_gcs_pubsub_trigger_operation(
...
   def export_to_json_snapshot(self, snapshot, entity):
     snapshot.edge_builder.make_mechanism(
-        entity, 'Gcs Pubsub Agent', self.agent)
+        entity, 'Gcs Pubsub Agent', self.__pubsub_agent)
     super(BaseGcsPubsubTriggerOperation, self).export_to_json_snapshot(
         snapshot, entity)
 
   def execute(self, agent=None):
-    status = self._do_execute(self.agent)
+    status = self._do_execute(self.__pubsub_agent)
```
jtk54: The log message following this line uses `self.agent`, but we changed the reference here.

ewiseblatt:
@spinnakerbot cherry-pick 1.13

PR #4162, opened 2019-03-20
fix(appengine): Fix AppEngine git repo path reference.

joonlim:


joonlim:
Breaking commit for reference: https://github.com/spinnaker/spinnaker/commit/d789dceaff848d3f877b9f105f257357b12f66a0

PR #4186, opened 2019-03-24
fix(audit): Dont require jars for non-java repositories

ewiseblatt:
@jtk54

### DIFF
--- a/dev/buildtool/inspection_commands.py
+++ b/dev/buildtool/inspection_commands.py
@@ -531,8 +531,8 @@ class CollectArtifactVersions(CommandProcessor):
         key = package[len(prefix):]
       if not key in jar_map:
         key = package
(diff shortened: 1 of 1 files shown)


dev/buildtool/inspection_commands.py
```diff
@@ -1284,7 +1284,11 @@ def audit_service(service, versions):
         for _, buildnums in commits.items():
           for buildnum, info_list in buildnums.items():
             version_buildnum = '%s-%s' % (version, buildnum)
-            jar_ok = self.audit_jar(service, version_buildnum, info_list)
+            if service in ['monitoring-daemon', 'monitoring-third-party']:
+              # Uses debians, but not jars
```
jtk54: I'd add a "so jars_ok is trivially true" to explicitly describe what we're doing here.

PR #4198, opened 2019-03-26
feat(sponnet): Add skip eval SPeL to Deploy Manifest

kskewes:
Per: https://github.com/spinnaker/orca/pull/2761

Notes:

1.  Haven't added `evaluateOverrideExpressions` option to Bake Manifest stage. See Helm PR here: https://github.com/spinnaker/spinnaker/pull/3811

2. Added `withSkipExpressionEvaluation()` to `deployManifest()` in a way that could be used for artifact or text manifests. 
I can move it to an flag in `withManifestArtifact(artifact):: self + { <here> }`  if text manifests are fundamentally different.
Not sure if it's a `no-op` for text manifests if set to `true`. 

@maggieneterval would you mind advising?

Tested via:
1. `spin` CLI submitting `demo-pipeline.jsonnet` to Spinnaker and verifying in Deck
2. Our internal sponnet pipelines now all default to true and have been run.

### DIFF
--- a/sponnet/pipeline.libsonnet
+++ b/sponnet/pipeline.libsonnet
@@ -19,7 +19,7 @@
     withTemplate(templateArtifact):: self + { template: templateArtifact },
     withSchema(schema):: self + { schema: schema },
     withInherit(inheritedFields):: self + if std.type(inheritedFields) == 'array' then { inherit: inheritedFields } else { inherit: [inheritedFields] },
-    withVariableValues(variables):: self + { variables: variables }, // variables are key-value pairs of <variable name> -> <variable value>
+    withVariableValues(variables):: self + { variables: variables },  // variables are key-value pairs of <variable name> -> <variable value>
   },
 
   moniker(app, cluster):: {
@@ -52,7 +52,7 @@
     httpFile():: artifact('http/file', 'http'),
     // kubernetesObject to be tested. Where kind is Deployment/Configmap/Service/etc
     kubernetesObject(kind):: artifact('kubernetes/' + kind, 'custom'),
-    front50PipelineTemplate():: artifact('front50/pipelineTemplate', '').withArtifactAccount('front50ArtifactCredentials') // credentials are static
+    front50PipelineTemplate():: artifact('front50/pipelineTemplate', '').withArtifactAccount('front50ArtifactCredentials'),  // credentials are static
   },
 
   // expected artifacts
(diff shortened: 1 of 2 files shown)


maggieneterval:
Hey thanks for this @kskewes, LGTM but looping in @jtk54 here for final approval since he has more context around sponnet than I do.

kskewes:
Side note, perhaps I should PR a Makefile with `jsonnet fmt` and the app + pipeline submit via spin cli commands.

sponnet/pipeline.libsonnet
```diff
@@ -19,7 +19,7 @@
     withTemplate(templateArtifact):: self + { template: templateArtifact },
     withSchema(schema):: self + { schema: schema },
     withInherit(inheritedFields):: self + if std.type(inheritedFields) == 'array' then { inherit: inheritedFields } else { inherit: [inheritedFields] },
-    withVariableValues(variables):: self + { variables: variables }, // variables are key-value pairs of <variable name> -> <variable value>
+    withVariableValues(variables):: self + { variables: variables },  // variables are key-value pairs of <variable name> -> <variable value>
```
maggieneterval: might just be late in the day for my tired eyes but is this just an extra space added by accident?

kskewes: Indeed, it's just the result of `jsonnet fmt` on the file.
I think it's good to run this over the library.

maggieneterval:
@kskewes just read your updated PR description! I know we talked a little in Slack but just to clarify, the `skipExpressionEvaluation` flag will only have an effect for artifact manifests. Also not sure if the majority of sponnet users' pipelines will want to opt into this by default, going to loop in @ethanfrogers for a second look as well.

ethanfrogers:
Hmm, if the default for skipping SpEL evaluation is `false` I'd also expect a default `deployManifest` stage to default to `false`. Is there any particular reason to have the behavior here be any different?

kskewes:
@ethanfrogers,
Sponnet default is "don't do anything, don't add any JSON re: SpEL evaluation".
This results in the Spinnaker default behaviour of evaluating any SpEL in Deploy Manifest stages with artifacts.

If a user wants to skip SpEL evaluation they need to use [`withSkipExpressionEvaluation()`](https://github.com/spinnaker/spinnaker/pull/4198/files#diff-01c380160d0d012b77fc2bdbbded0996R130) within their sponnet pipeline Deploy Manifest stage. (This PR)
 
@maggieneterval,
Indeed. per reply to Ethan I think the way it is will be fine.
If we want to restrict scope of `skipExpressionEvaluation` to only artifact deploys then we would need to change this PR to be a boolean parameter on [this function](https://github.com/spinnaker/spinnaker/pull/4198/files#diff-98560c97d4b23f0cf12af197f63a07b2R227) I believe. 
In this case we would need to default the parameter to `false` to match Spinnaker default (don't skip, ie: evaluate SpEL).

maggieneterval:
Awesome, thanks for clarifying @kskewes! In that case I'm going to go ahead and merge this.

PR #4252, opened 2019-04-05
fix(appengine): Allow setting region for appengine_smoke_test.

joonlim:


dev/all_tests.yaml
```diff
@@ -314,6 +315,7 @@ tests:
       app_directory_root: appengine/standard/hello_world
       test_gcs_bucket: $storage_gcs_bucket
       branch: master
+      test_appengine_region: $test_appengine_region
```
ewiseblatt: Since you are passing this as an argument you will need to add it as a command line argument here https://github.com/spinnaker/spinnaker/blob/master/testing/citest/spinnaker_testing/appengine_scenario_support.py#L25

Where you added it below was to the validate_bom tool, not to the test fixture. They are separate stacks.

testing/citest/tests/appengine_smoke_test.py
```diff
@@ -458,4 +462,5 @@ def main():
 
 
 if __name__ == '__main__':
+  print(sys.argv)
```
ewiseblatt: remove this

dev/all_tests.yaml
ewiseblatt: Turns out this will work because unrecognized --<arg> <value> arguments are assumed to be key/value pairs.

Issue #308, opened 2015-01-11
Resolve accounts smartly or remove them entirely? 


aglover:
I was playing around w/Gate & the Spinnaker client and realized that in order for me to update the description of Hootch, I had to supply an account (see https://github.com/spinnaker/spinnaker-client/issues/22). I know why this is; nevertheless, should we care anymore? Does it make sense to update the description of Hootch in Test and not in Prod? I was envisioning a single global registry of applications where one doesn't need to care about accounts -- if I have a Hootch in test it is assumed to be the same app in Prod. 

If you look at the JSON Gate returns for an application, it has this: 

```
attributes: {
name: "hootch",
description: "A Jira issue facilitator",
email: "xxxx-xxxxx@netflix.com",
accounts: "test",
createTs: "1414384161869",
updateTs: "1420936429437"
},
```

`accounts` isn't even top level, which implies that a `description` is uniform across all of them. If I change the `description` in test and not in prod, how would we even reveal that in this structure? I think our instance on requiring an account with Gate or any service is antiquated. 

What is the plan to resolve this?

Issue #185, opened 2014-11-03
Get force cache refresh working again in oort


aglover:
@cfieber

PR #4275, opened 2019-04-10
chore(buildtool/spin): Adds logging to spin cmds.

jtk54:


dev/buildtool/spin_commands.py
```diff
@@ -190,19 +190,24 @@ def __determine_internal_version(self, repository):
     gate_major = match.group(1)
     gate_min = match.group(2)
 
-    tag_matcher = re.compile(r'version-{maj}.{min}.(\d+)'
-                             .format(maj=gate_major, min=gate_min))
+    tag_pattern = r'version-{maj}.{min}.(\d+)'.format(maj=gate_major, min=gate_min)
```
ewiseblatt: FYI you could replace #185 with SemVer.make(self.__gate_version)
https://github.com/spinnaker/spinnaker/blob/master/dev/buildtool/git_support.py#L169

dev/buildtool/spin_commands.py
```diff
@@ -190,19 +190,24 @@ def __determine_internal_version(self, repository):
...
     if tag_matches:
       patch_versions = [int(m.group(1)) for m in tag_matches]
       max_patch = max(patch_versions)
       patch = str(max_patch + 1)
     else:
       patch = '0'
-    return '{major}.{minor}.{patch}'.format(
-        major=match.group(1), minor=match.group(2), patch=patch)
 
+    internal_version = '{major}.{minor}.{patch}'.format(
```
ewiseblatt: this would be semver.to_version()

dev/buildtool/spin_commands.py
```diff
@@ -330,21 +335,28 @@ def promote_spin(self, repository):
     # and minor versions so fixes (thus patch version) are decoupled between
     # the two.
     patch = '0' # Patch is reset on a new Gate major or minor.
-    tag_matcher = re.compile(r'version-{maj}.{min}.(\d+)'
-                             .format(maj=gate_major, min=gate_min))
+    tag_pattern = r'version-{maj}.{min}.(\d+)'.format(maj=gate_major, min=gate_min)
+    tag_matcher = re.compile(tag_pattern)
```
ewiseblatt: I think you want a helper function here and above that returns the highest tag (or semver) for a given <tag_pattern>. If it doesnt have one then you create a new semver at patch 0. Otherwise you  use semver.next(semver.PATCH_INDEX)

jtk54: :+1: refactored.

dev/buildtool/spin_commands.py
jtk54: Switched over to the SemVer class, I had to formulate the raw gate versions as tags but other than that it worked swell.

dev/buildtool/spin_commands.py
```diff
@@ -312,39 +291,22 @@ def promote_spin(self, repository):
...
-    self.__stable_version = '{major}.{minor}.{patch}'.format(
-        major=match.group(1), minor=match.group(2), patch=patch)
+    next_spin_semver = bump_spin_patch(git, git_dir, self.__gate_version)
+    gate_major = next_spin_semver.major
+    gate_min = next_spin_semver.minor
+    last_patch = str(max(next_spin_semver.patch - 1, 0))
+    self.__stable_version = next_spin_semver.to_version()
+    logging.info('calculated new stable version: {}'
+                 .format(self.__stable_version))
+    last_tag = 'version-{maj}.{min}.{patch}'.format(maj=gate_major,
```
ewiseblatt: Im confused here. Why do you use next last maj/minor for the last tag?
If you just rolled one of these, then it is different from the last one.

dev/buildtool/spin_commands.py
```diff
@@ -436,6 +398,49 @@ def init_argparser(self, parser, defaults):
...
+  Spin is coupled to the Gate major and minor version.
+  Gate is a routing server, so features and breaking changes in Gate
+  must be reflected in spin since it is a client.
+
+  :param git: git support helper class.
+  :param git_dir: spin git directory.
+  :param gate_version: gate version to align spin version with in <maj>.<min>.<patch>-<buildnum> format.
+  :return: (SemanticVersion) Next semver spin version.
+  '''
+  gate_version_parts = gate_version.split('-')
```
ewiseblatt: why do you need this split? Why cant you just take the semver of gate_version and ignore the "series" part and force it to "version" if you want that?

dev/buildtool/spin_commands.py
```diff
@@ -436,6 +398,49 @@ def init_argparser(self, parser, defaults):
...
+  tag_matcher = re.compile(tag_pattern)
+  tags = git.fetch_tags(git_dir)
+
+  logging.info('searching git tags {} for patterns matching {}'.format(tags, tag_pattern))
+  matching_semvers = [SemanticVersion.make(t) for t in tags if tag_matcher.match(t)]
+  logging.info('found matching semvers: {}'.format(matching_semvers))
+  if matching_semvers:
+    max_semver = max(matching_semvers)
+    next_semver = max_semver.next(SemanticVersion.PATCH_INDEX)
+    patch = next_semver.patch
```
ewiseblatt: general question -- how do you know if you even need a new version and this is different from some previous one? Browsing the broader file this wasnt clear to me.

jtk54: `bump_spin_patch` is just the helper. We call this in two places:
1. When we record the 'internal version' the spin binary will report when asked. We always set this internal version as the Gate version + next patch if no tags are present. This is OK because we won't publish every spin binary built here, see (2).
2. When we publish the spin binary. We check if the locally checked-out HEAD commit is different from the commit at the most previous tag and roll up the patch if they aren't.

I agree it's a bit confusing and should be clarified in comments.

dev/buildtool/spin_commands.py
jtk54: This is actually cutting off the datestring build number. The input to this function is <maj>.<min>.<patch>-<buildnum> (specified in the docstring).

dev/buildtool/spin_commands.py
jtk54: Major and minor are pinned to the gate version's major and minor and will be static.

dev/buildtool/spin_commands.py
ewiseblatt: ah, gate_version is a build version, not a source version.

dev/buildtool/spin_commands.py
ewiseblatt: I think the __no_changes at #308 is what I was looking for.

dev/buildtool/spin_commands.py
```diff
@@ -312,39 +294,23 @@ def promote_spin(self, repository):
...
-      max_patch = max(patch_versions)
-      last_tag = 'version-{maj}.{min}.{max_patch}'.format(maj=gate_major,
-                                                          min=gate_min,
-                                                          max_patch=max_patch)
-      self.__no_changes = git.query_local_repository_commit_id(git_dir) == git.query_commit_at_tag(git_dir, last_tag)
-      patch = str(max_patch + 1)
-
-    self.__stable_version = '{major}.{minor}.{patch}'.format(
-        major=match.group(1), minor=match.group(2), patch=patch)
+    next_spin_semver = bump_spin_patch(git, git_dir, self.__gate_version)
```
ewiseblatt: I think the logging messages in bump_spin_patch might get confusing in this context because they suggest you are creating new versions, but really you are not. You are just trying to identify what the 'last' version was to see if you dont need to do any work because there were no changes.

jtk54: Changed up the wording here to specify that we are calculating a 'next spin patch version'.

PR #4318, opened 2019-04-19
docs(citest): reformat README.md

plumpy:
There were several places where links weren't showing up because the link's text
and its URL were on separate lines. I just ran the whole thing through the 
(internal) `mdformat` tool to fix that.

It also justified all the paragraphs to 80 characters. But I had to paste back 
in most of the tables because our internal markdown engine has more table 
features than Github.


plumpy:
@ewiseblatt

PR #4326, opened 2019-04-24
fix(sponnet): fix functions of undoRolloutManifest at pipeline.libsonnet

Mahito:
Since some functions had problems, undoRolloutManifest couldn't make a JSON.
This PR fixes those problems.

### DIFF
--- a/sponnet/pipeline.libsonnet
+++ b/sponnet/pipeline.libsonnet
@@ -272,8 +272,8 @@
       cloudProvider: 'kubernetes',
       withAccount(account):: self + { account: account },
       withNamespace(namespace):: self + { location: namespace },
-      withRevisionsBack(revisionsBack): self + { numRevisionsBack: revisionsBack },
-      withManifestName(kind, name):: self.options { manifestName: kind + ' ' + name },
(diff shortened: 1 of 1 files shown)


kskewes:
I'm not sure what the original `self.options` were supposed to be and just have missed it when adding other stages. Your fix looks correct to me. Cheers.

Mahito:
@kskewes Thank you for your review.  I got a JSON from Spinnaker GUI as below.
I decided to remove `self.options` because it seems not to be used.

```
    {
      "account": "xxxx-lab",
      "cloudProvider": "kubernetes",
      "location": "xxx-lab-app",
      "manifestName": "deployment xxxx",
      "mode": "static",
      "name": "Undo Rollout (Manifest)",
      "numRevisionsBack": 1,
      "refId": "3",
      "requisiteStageRefIds": [
        "2"
      ],
      "stageEnabled": {
        "expression": "${#stage(\"xxxxxx\")[\"status\"].toString() != \"SUCCEEDED\"}",
        "type": "expression"
      },
      "type": "undoRolloutManifest"
    },
```

PR #4382, opened 2019-05-08
fix(buildtool/halyard): Fixes versions for halyard commands.

jtk54:
Fixes regression in https://github.com/spinnaker/spinnaker/commit/50dde9eaaf8e9968304bd865f9592c956b870235. @plumpy PTAL also.

### DIFF
--- a/dev/buildtool/halyard_commands.py
+++ b/dev/buildtool/halyard_commands.py
@@ -207,21 +207,16 @@ class BuildHalyardCommand(GradleCommandProcessor):
     if not self.options.run_unit_tests:
       args.append('-x test')
 
-    if (os.path.isfile(os.path.join(repository.git_dir,
-                                    "gradle", "init-publish.gradle"))):
-      args.append('-I gradle/init-publish.gradle')
-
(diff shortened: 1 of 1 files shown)


dev/buildtool/halyard_commands.py
```diff
@@ -209,7 +209,8 @@ def _do_repository(self, repository):
 
     if (os.path.isfile(os.path.join(repository.git_dir,
                                     "gradle", "init-publish.gradle"))):
-      args.append('-I gradle/init-publish.gradle')
+      args.extend(
+          ['-I gradle/init-publish.gradle', '-Prelease.useLastTag=true'])
```
ewiseblatt: I think in the new gradle plugin there is a flag specifying which version to use. I use it in the debian or gradle command handler. That woudl be better that use last tag. If we no longer need use last tag then we probably wont need to delete all the tags anymore and can just specify the tag we want to use as we have been asking for.

jtk54:
Switched to the 'nu school' gradle invocation. PTAL.

PR #4386, opened 2019-05-08
fix(buildtool/halyard): the nebula code requires the version to be passed

plumpy:


plumpy:
cc @ewiseblatt @jtk54

PR #4413, opened 2019-05-15
fix(appengine): Add gcloud path to appengine config for localdebian

### LINKED ISSUE #3349: gcloud not in path by default on GCE Ubuntu 18.04

ezimanyi:
Fixes #3349

Gcloud is not in the path by default for systemd services in ubuntu 18.04. In order to allow appengine to still work, pass in the path to gcloud to the appengine provider.

### DIFF
--- a/dev/validate_bom__config.py
+++ b/dev/validate_bom__config.py
@@ -578,6 +578,11 @@ class AppengineConfigurator(Configurator):
       return
 
     script.append('hal -q --log=info config provider appengine enable')
+
+    if options.deploy_spinnaker_type == 'localdebian':
+      script.append(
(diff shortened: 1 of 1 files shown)


ezimanyi:
@spinnakerbot cherry-pick 1.14

PR #4816, opened 2019-08-30
fix(buildtool): log stderr of `gcloud compute instances describe`

plumpy:
By default, it was getting joined with stdout and then we tried to parse
it as JSON, which didn't work since it isn't, in fact, JSON.

My test run didn't actually have any of this stderr output, so I can't
say for certain that it actually works. But it doesn't make the
no-stderr case fail, so we're at least no worse off.

### DIFF
--- a/dev/buildtool/subprocess_support.py
+++ b/dev/buildtool/subprocess_support.py
@@ -38,7 +38,7 @@ from buildtool.base_metrics import BaseMetricsRegistry
 # this module does not offer encapsulated configuration.
 ERROR_LOGFILE_DIR = 'errors'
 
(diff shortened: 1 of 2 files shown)


plumpy:
@spinnakerbot cherrypick 1.15
@spinnakerbot cherrypick 1.14

plumpy:
@spinnakerbot cherrypick 1.15

plumpy:
@spinnakerbot cherry-pick 1.15
@spinnakerbot cherry-pick 1.14

plumpy:
@spinnakerbot cherry-pick 1.15
@spinnakerbot cherry-pick 1.14

ezimanyi:
@spinnakerbot :cherries: :pick: 1.13

PR #4832, opened 2019-09-03
fix(jsonnet): Honor "name" attribute of default artifact passed to expected artifact

gnarea:


sponnet/pipeline.libsonnet
```diff
@@ -73,15 +73,7 @@
         kind: matchArtifact.kind,
       },
     },
-    withDefaultArtifact(defaultArtifact):: self + {
-      defaultArtifact: {
-        reference: defaultArtifact.reference,
-        type: defaultArtifact.type,
-        kind: if defaultArtifact.kind == 'custom' then defaultArtifact else 'default.' + defaultArtifact.kind,
```
gnarea: I believe this value was incorrect because `kind` shouldn't have been an object when `defaultArtifact.kind == 'custom'` -- I'm guessing it should've been `"custom"` in that case.

I also suspect this field is not really used. At least it's not exported in the JSON document I get from Deck.

gnarea:
Yes, it makes sense not to break compatibility, so I've just changed the PR to simply add the `name`.

PR #4973, opened 2019-09-30
chore(github): Remove non-functional at-mention

erikmunson:
GitHub doesn't allow folks from outside the `spinnaker` org to add org-specific group mentions, so we're instructing people to do something they aren't allowed to do.


maggieneterval:
Thanks for doing this @erikmunson :hugs:

PR #4975, opened 2019-09-30
fix(build): Fix repository comparison to handle ssh/https difference

ezimanyi:
I noticed that `Push_RawChangelogGist` is re-cloning repos every time it runs (both the source repository repos and the gist which is technically also a repo).  There is logic to check if the repos have already been cloned and if so just fetch updates, but that logic is not running as we're deleting all these repos every time the script runs.  Removing the step to delete these (which is not a code change, but just a Jenkins config change) caused us to hit a bug where it doesn't realize that the repo on disk is the same repo it's considering cloning (as the one on disk has an ssh origin and the one it's considering cloning has an https origin).  This PR fixes said bug, allowing us to remove the step in the Jenkins job that deletes all these repos.

The hope is that by reducing the amount of data we're fetching from github every time this job runs we might also reduce the timeouts we've been seeing on this job.  I'm not sure if a fetch is less likely to time out than a clone, but it seems worth a try.

* refactor(build): Update determine push/pull URL to take the origin 

  We're passing a full repository to determine_pull_url and determine_push_url, but these functions only need to know the origin. To allow these to be re-used in cases where we only have an origin, change the functions to accept an origin and expect callers to pass one in.

* fix(build): Fix repository comparison to handle ssh/https difference 

  When checking to see if a directory contains the expected git repository, we compare the origin against the expected origin. This often doesn't work as we have logic to auto-translate an origin from https to SSH before cloning, so existing repositories will often have ssh origins even if the configured origin is https. Fix this by considering repositories equivalent if the origin is either the expected origin or the result of determine_pull_url on the expected origin.

### DIFF
--- a/dev/buildtool/git_support.py
+++ b/dev/buildtool/git_support.py
@@ -1078,21 +1078,21 @@ class GitRunner(object):
     self.check_run(git_dir, 'tag -d ' + ' '.join(tags_to_remove))
     logging.debug('%d of %d tags removed', len(tags_to_remove), len(all_tags))
 
-  def determine_pull_url(self, repository):
+  def determine_pull_url(self, origin):
     """Return the pull URL for a given repository from its origin."""
-    parts = self.normalize_repo_url(repository.origin)
+    parts = self.normalize_repo_url(origin)
     if len(parts) == 3:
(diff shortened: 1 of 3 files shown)


ezimanyi:

PR #4979, opened 2019-09-30
fix(build): Allow re-use of git directories

ezimanyi:
In a few places, we check if a directory containing a git repo already exists. If the directory does not exist, we clone the repo into the directory and check out the expected commit; if the directory does exist, we verify that it is at the expected commit.

This effectively makes it impossible to re-use directories, as there is no logic in the re-use path to fetch changes and check out the expected commit. Add this logic; now, if a the directory exists we'll attempt to fetch origin and check out the expected commit before moving on rather than expecting it to already be at the commit we want.


ezimanyi:

PR #4984, opened 2019-10-01
fix(citest): Add brief retry when validating a new pipeline

ezimanyi:
In a number of places in the tests, we create a new pipeline by sending a POST to gate, then immediately poll to check that the pipeline is there.  While in principle the immediate call should show the pipeline as existing (as there is logic in front50 to provide read-after-write consistency) there are a couple of edge cases.

Most of the places we try to read a new pipeline allow a few retries to give the pipeline to show up, but a few don't.  Add retries to the remaining places to reduce flakiness.


ezimanyi:

PR #4988, opened 2019-10-01
fix(citest): Allow longer service startup time

ezimanyi:
We currently time out if a service takes more than 5 minutes to start. Some of the services currently take on the order of 5 minutes to start (at least in some of the citest environments), which is leading to flakiness.

Let's add a safety factor of 2, and give the services 10 minutes to start before we time out. (Ideally they would start faster, but let's at least not have flaky tests in the meantime.) This won't decrease the best-case time as the tests start when all the services are ready, even if it's shorter than the max time.


ezimanyi:

PR #5010, opened 2019-10-09
fix(sponnet): Fix syntaxes and withClusterSize

Mahito:
- Automatic correction by jsonnet plugin for vim
- Add moniker to withClusterSize in checkPreconditions
- Fixed regions of withClusterSize in checkPreconditions


kskewes:
Formatting changes and region fix up look great. good stuff.

PR #5033, opened 2019-10-16
feat(spinbot): do not mark beginner-friendly issues as stale

maggieneterval:
After conferring with the Spinnaker SIG leads, we think it will help us curate a healthy backlog of beginner-friendly issues if they are never tagged as stale and auto-closed by Spinnakerbot.

### DIFF
--- a/spinbot/policy/stale_issue_policy.py
+++ b/spinbot/policy/stale_issue_policy.py
@@ -7,6 +7,7 @@ class StaleIssuePolicy(Policy):
         super().__init__()
         self.stale_days = self.config.get('stale_days')
         self.ignore_lifecycle_label = self.config.get('ignore_lifecycle_label', 'no-lifecycle')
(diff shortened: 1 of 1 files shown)


spinbot/policy/stale_issue_policy.py
```diff
@@ -7,6 +7,7 @@ def __init__(self):
         super().__init__()
         self.stale_days = self.config.get('stale_days')
         self.ignore_lifecycle_label = self.config.get('ignore_lifecycle_label', 'no-lifecycle')
+        self.beginner_friendly_label = self.config.get('beginner-friendly_label', 'beginner-friendly')
```
maggieneterval: Not sure if anyone has context on this, but I was curious why we're attempting to read these label values from config rather than just using a string constant? Is this so Spinbot could potentially be configured to other open-source projects with different label conventions?

emjburns:
🌵 🥖 🌱 ✈️ 🍰 I don't understand how this emoji thing works but i'm delighted by this pr

PR #5065, opened 2019-10-28
feat(sponnet): Add fields for namespaceOverride to deployManifest

Mahito:
Adds fields for `namespaceOverride` to deployManifest at pipeline.libsonnet.

Ref. feat(provider/kubernetes): namespace deployManifest
https://github.com/spinnaker/deck/pull/7016

### DIFF
--- a/sponnet/pipeline.libsonnet
+++ b/sponnet/pipeline.libsonnet
@@ -309,6 +309,7 @@
       withManifests(manifests):: self + if std.type(manifests) == 'array' then { manifests: manifests } else { manifests: [manifests] },
       withMoniker(moniker):: self + { moniker: moniker },
       withSkipExpressionEvaluation():: self + { skipExpressionEvaluation: true },
+      withNamespaceOverride(namespace):: self + { namespaceOverride: namespace },
     },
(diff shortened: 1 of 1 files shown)


ethanfrogers:
Good point, @kskewes. I think I prefer it match the property name, as well.

sponnet/pipeline.libsonnet
```diff
@@ -309,6 +309,7 @@
       withManifests(manifests):: self + if std.type(manifests) == 'array' then { manifests: manifests } else { manifests: [manifests] },
       withMoniker(moniker):: self + { moniker: moniker },
       withSkipExpressionEvaluation():: self + { skipExpressionEvaluation: true },
+      withOverrideNamespace(namespace):: self + { namespaceOverride: namespace },
```
louisjimenez: @Mahito are you open to changing the func name to match the key name as @kskewes suggested?

Mahito: @louisjimenez @kskewes Thank you for your comment and I'm sorry for the late changing. I  fixed it.

PR #5081, opened 2019-10-30
fix(spinbot): Prepends spinbot's container tagging with the branch it…

ttomsu:
I accidentally did a `make docker` when I was on the `release-1.17.x` branch and didn't realize it. In an attempt to add the branch, I discovered that the multiple `-t` flags is not supported, and that this was actually working by accident. 

This can be seen by looking in GCR: images should be tagged with their commit hashes but aren't:
m44VDH9GcqL

Spinbot appears to run as a k8s Cron Job that does an `ImagePullPolicy: Always` on the `latest` image, but I'll change it to `master-latest` to ensure someone doesn't accidentally do what I did again.

### DIFF
--- a/spinbot/Makefile
+++ b/spinbot/Makefile
@@ -1,15 +1,15 @@
 PROJECT=spinnaker-marketplace
 IMAGE=gcr.io/${PROJECT}/spinbot
 
-HASH := $(shell git rev-parse HEAD)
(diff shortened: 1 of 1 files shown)


ttomsu:
@spinnakerbot cherry-pick 1.17

ttomsu:
(just testing the `cherry-pick` labeling) :arrow_up:

PR #5100, opened 2019-11-05
feat(spinbot): Add support for 'deck-kayenta' and 'keel'

ajordens:


ajordens:
I've made sure that the bots group has access to `deck-kayenta` and `keel` (kayenta was missing).

Is this all we need to do @ttomsu?

ttomsu:
I don't think the config is pushed automatically. I can do it once merged (since I'm release manager this cycle)

ajordens:
'tis merged @ttomsu -- let me know if there's anything else I can do to help!

PR #5133, opened 2019-11-15
feat(sponnet): Support pipeline triggers

gnarea:


sponnet/pipeline.libsonnet
```diff
@@ -173,6 +173,11 @@
       payloadConstraints: {},
       addPayloadConstraints(key, value):: self + { payloadConstraints+: { [key]: value } },
     },
+    pipeline(name, application, pipeline, statuses):: trigger(name, 'pipeline') {
```
louisjimenez: Can the keys for `type: "pipeline"` and `enabled: true | false` be added?

gnarea: They are already added by the `trigger()` function: https://github.com/spinnaker/spinnaker/blob/8757a9d151f8daaaf4b91c890b6faa7cf3817420/sponnet/pipeline.libsonnet#L136-L139

Should I override them?

louisjimenez: oof, I missed that. Thanks!

PR #5150, opened 2019-11-20
fix(spin): Build the checked out version of 'spin' (as opposed to HEAD)...

ttomsu:
using the exact declared dependencies (instead of doing 'go get -u' which pulls in potentially backwards incompatible changes).

This was brought up because of a [transitive dependency](https://github.com/posener/complete) that published a backwards-incompatible `v2` as their default git branch, which is what appears to be used when `go get -u` is used. 

From `go help get`:

```
Although get defaults to using the latest version of the module containing
a named package, it does not use the latest version of that module's
dependencies. Instead it prefers to use the specific dependency versions
requested by that module. For example, if the latest A requires module
B v1.2.3, while B v1.2.4 and v1.3.1 are also available, then 'go get A'
will use the latest A but then use B v1.2.3, as requested by A.
...
The -u flag instructs get to update modules providing dependencies
of packages named on the command line to use newer minor or patch
releases when available. Continuing the previous example, 'go get -u A'
will use the latest A with B v1.3.1 (not B v1.2.3). 
``` 

In debugging this, it revealed that we were checking out `github.com/spinnaker/spin@HEAD` via `go get` and building that version, as opposed to, say the release branch. I'm still not convinced that the changes below actually build from the release branch, though.

### DIFF
--- a/dev/buildtool/spin_commands.py
+++ b/dev/buildtool/spin_commands.py
@@ -129,13 +129,12 @@ class BuildSpinCommand(RepositoryCommandProcessor):
       labels = {'repository': repository.name,
                 'dist': dist_arch.dist,
                 'arch': dist_arch.arch}
-      cmd = 'go get -v -u {}'.format(spin_package_path)
+      cmd = 'go get -v'
       self.metrics.time_call(
           'GoGet', labels, self.metrics.default_determine_outcome_labels,
           check_subprocesses_to_logfile, 'Fetching Go packages ' + context,
-          logfile, [cmd], cwd=gopath, env=env)
+          logfile, [cmd], cwd=repository.git_dir, env=env)
 
-    for dist_arch in DIST_ARCH_LIST:
       # GCS sub-directory the binaries are stored in are specified by
       # ${build_version}/${dist}.
       version_bin_path = ('spin/{}/{}/{}/{}'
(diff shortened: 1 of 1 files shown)


dev/buildtool/spin_commands.py
```diff
@@ -129,13 +129,12 @@ def build_all_distributions(self, repository):
       labels = {'repository': repository.name,
                 'dist': dist_arch.dist,
                 'arch': dist_arch.arch}
-      cmd = 'go get -v -u {}'.format(spin_package_path)
```
louisjimenez: For future reference, @ttomsu and I determined that the `-u` flag would indicate to `go get` that it should update pre-installed dependencies. In our case, there are no pre-installed dependencies on each clean build. Regardless, it is worth removing to make this command clearer.

ttomsu:
So it turns out `go help get` spits out different help text when in a directory with and without go modules. The above text applies when using go modules, but the spin hasn't been updated to use them. The transitive dependency has been fixed, so the builds can proceed, but this change still needs to go in to fix building from the right release branch.

dev/buildtool/spin_commands.py
maggieneterval: Thanks for the explanation, I was wondering about that!

PR #5194, opened 2019-12-03
fix(release): Explicitly use UTC time in release changelog

ttomsu:
Currently https://www.spinnaker.io/community/releases/versions/ displays the date with "+0000" (UTC) when in fact the time is in Eastern. This will make the frontmatter explicitly UTC.


ttomsu:
QRQe9xn5KH6 (1)

PR #5198, opened 2019-12-04
fix(bakery): Use xenial base image for GCE since trusty was deprecated

ttomsu:
Not 100% sure this will fix the failing builds, but I think it's one of the reasons bake_and_deploy is failing

### DIFF
--- a/testing/citest/tests/bake_and_deploy_test.py
+++ b/testing/citest/tests/bake_and_deploy_test.py
@@ -340,7 +340,7 @@ class BakeAndDeployTestScenario(sk.SpinnakerTestScenario):
         'type': 'bake',
         'name': 'Bake',
         'user': '[anonymous]',
-        'baseOs': 'trusty',
+        'baseOs': 'xenial' if providerType == 'gce' else 'trusty',
         'baseLabel': 'release',
         'cloudProvider': providerType,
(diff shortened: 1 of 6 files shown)


testing/citest/tests/bake_and_deploy_test.py
```diff
@@ -340,7 +340,7 @@ def make_bake_stage(self, package, providerType, requisiteStages=None,
         'type': 'bake',
         'name': 'Bake',
         'user': '[anonymous]',
-        'baseOs': 'trusty',
+        'baseOs': 'xenial' if providerType == 'gce' else 'trusty',
```
plumpy: why not just xenial everywhere?

ttomsu: AWS images (this method is called with `gce` and `aws` as the provider type) are specified per region, and thus a new one of these blocks would need to be created: https://github.com/spinnaker/rosco/blob/master/rosco-web/config/rosco.yml#L187-L257

It doesn't seem to me that anyone is maintaining that list, given that there is still an entry for Ubuntu v12.04

ttomsu: So to answer your question: I didn't want to shave that yak.

PR #5225, opened 2019-12-12
fix(boms): Do not including -validated BOMs in candidates to prune

ttomsu:
Used in `Admin_AuditBoms` and transitively in `Admin_DeleteObsoleteArtifacts` Jenkins jobs, which I am going to run to reduce the data in the `halconfig` bucket

### DIFF
--- a/dev/buildtool/inspection_commands.py
+++ b/dev/buildtool/inspection_commands.py
@@ -1012,7 +1012,8 @@ class AuditArtifactVersions(CommandProcessor):
     candidates = []
     with open(path, 'r') as stream:
       for line in stream.read().split('\n'):
-        if line.endswith('-latest-unvalidated.yml'):
+        # This matches both -unvalidated and -validated BOMs
+        if line.endswith('validated.yml'):
           continue
         bom = CollectBomVersions.url_to_bom_name(line)
         if not CollectBomVersions.RELEASED_VERSION_MATCHER.match(bom):


ttomsu:
A bit more context:

The `Admin_AuditBoms` job does a giant mapping of all top-level artifacts in Bintray, GCR, and GCS (i.e. BOMs, docker containers, and JARs). It then outputs a big list of candidates for deletion, which I noticed includes all of our `...-validated.yml` BOMs, which this PR should remove.

ezimanyi:
I'm a bit later here, but is there a reason we're no longer pruning the `-validated` BOMs?  We could be producing ~4 of these in a day (if all the integration tests pass) which will case us to keep these BOMs as well as all artifacts they reference forever.

My understanding is that we'd agreed to keep all *released* BOMs forever but that we'd retain others (and the artifacts they reference) for only ~2-3 months.

I'm curious about the motivation behind this change, as it will increase the storage space we use and (more importantly) may cause bintray indexing to be slower and increase the time between when a package is published and when it's available.

ttomsu:
I'm only aware of the branch-latest-validated BOMs, which were the ones I
was targeting to save (so an extra 18 or so right now). Are there more
you're thinking of that I don't know about?

ezimanyi:
Oh, right, I think I misunderstood---the release process creates 4-5 of these a day, but it always overwrites the same one, so we don't accumulate a backlog of them.  It also writes each of them to `{branch}-{date}` which do accumulate, but those will continue to get deleted periodically by the script even after this change.  So forget my comment, this looks good to me.

PR #5231, opened 2019-12-13
feat(auditing): Add config files to the auditing and pruning code

ttomsu:
Our current `Admin_AuditBoms` job generates a reverse index of artifacts (such as container images, debians and jars) back to their source BOMs. Coupled with `Admin_DeleteObsoleteArtifacts`, the two jobs generate a list of orphaned and deletable artifacts, and then deletes them.

This PR adds the generation of unused configs that have been collecting in the `halconfig` bucket since the beginning of the release process. These are effectively copies of all files in each service repo's `halconfig` directory ([example](https://github.com/spinnaker/gate/tree/master/halconfig)).

You can see the candidate list of files to be deleted by viewing the `prune_configs.txt` [here](https://builds.spinnaker.io/view/5%20Admin/job/Admin_AuditBoms/68/artifact/build_output/audit_artifact_versions/*config*) - you can see we're keeping around copies of yaml files from 2017-08.

While not a huge space savings (on the order of 10 megs, If I had to guess) it has been running unbounded for a few years now.

Looking at the `audit_found_configs.yml`, my best guess is that the `Admin_DeleteObsoleteArtifacts` will need to be ran twice - the first time will delete all configs that aren't referenced in a BOM. In that same pass, a whole bunch of old BOMs will also be deleted (those from 2019-08 until, say 2019-11). A second execution will clear configs that now have no matching BOM because it was deleted in the first pass.

### DIFF
--- a/dev/buildtool/inspection_commands.py
+++ b/dev/buildtool/inspection_commands.py
@@ -454,6 +454,7 @@ class CollectArtifactVersions(CommandProcessor):
      <debian_repository>__versions.yml: All the debian build versions
      <jar_repository>__versions.yml: All the jar build versions
      <docker_registry>__versions.yml: All the container build versions
+     <config_bucket>__versions.yml: All the service-specific config build versions
      missing_jars.yml: Bintray debian versions without a corresponding jar
      missing_debians.yml: Bintray jar versions witout a corresponding debian
      config.yml: The configuration values used to collect the artifacts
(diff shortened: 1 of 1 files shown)


ttomsu:
> Also, am I reading this right that this only prints out a list of delete commands to run, and doesn't actually delete the obsolete configs?

That's correct. The `Admin_DeleteObsoleteArtifacts` job first triggers the `Admin_AuditBoms` job which runs this script. In the `Admin_DeleteObsoleteArtifacts` job is the [bash script](https://builds.spinnaker.io/view/5%20Admin/job/Admin_DeleteObsoleteArtifacts/configure) that actually deletes whatever is the `prune` files - I have not added the function to delete the configs yet.

PR #5245, opened 2019-12-17
feat(build): Add support for java8 containers

plumpy:
If a `Dockerfile.java8` file exists, it (along with `Dockerfile.ubuntu-java8`) will be built, otherwise the standard `slim` and `ubuntu` containers will be tagged with `java8` suffixes.

This allows for a new `JAVA8` and `UBUNTU_JAVA8` `imageVariant` to be added to Halyard.

Once the builds are fully moved to Java 11, this PR can be reverted completely.


ttomsu:
LGTM - sorry, I had reviewed this this morning and then we ran off to the meeting

PR #5303, opened 2020-01-11
Adding pinned option for pipeline parameters

craig-rueda:
* Adds `pinned` option to Sponnet pipeline parameters

Pinned params show up in the executions of pipelines via the UI...


ttomsu:
HI @craig-rueda - thanks for the submission! I've split out the sponnet library into it's own repo. Would you mind re-opening this PR there? @louisjimenez would be the person the take a look at it then.

PR #5319, opened 2020-01-15
feat(spinrel): use spinrel to publish BOMs, profiles, and container tags

plumpy:
Did a successful test run of this and deployed the version it produced with Spinnaker for GCP...


plumpy:
I'm actually just going to close this since Travis is working on a larger reworking of the release process and no one is actually consuming these new container tags anyway...

PR #5348, opened 2020-01-23
chore(buildcop): Quick cleanup and a few inline comments.

ttomsu:


.github/build-cop-template.md
```diff
@@ -1,8 +1,10 @@
 ---
-title: Build Cop Log {{ date | date('ddd, D MMM YYYY') }}
+title: Build Cop Log - {{ date | date('ddd, D MMM YYYY') }}
 labels: build-cop-rotation
 ---
 
-1. Login to https://builds.spinnaker.io and ensure all builds are green
-2. If failing, comment on this issue any actions you had to perform to get it back to green.
-3. Close this issue when finished.
+1. Login to [https://builds.spinnaker.io](builds.spinnaker.io).
```
plumpy: This is backwards... it should be `[builds.spinnaker.io](https://builds.spinnaker.io)`.

.github/workflows/build-cop-rotation.yml
```diff
@@ -2,29 +2,31 @@ name: Build Cop
 
 on:
   schedule:
-  # - cron: "*/5 * * * *"
-  - cron: 0 11 * * MON-FRI
+  - cron: 0 11 * * MON-FRI # Rotate weekdays at 11:00 UTC
```
plumpy: wait, so you want to switch to daily builcop?

.github/build-cop-template.md
ttomsu: Good catch.

.github/workflows/build-cop-rotation.yml
ttomsu: For the time being, I'd like each new person to get into the habit of checking the build status (or at least reacting/noticing the issue filed against them)

plumpy: Fair enough, but I think the reason we went with a week is because there's no real incentive to do anything if you know it's just going to be someone else's problem tomorrow... a feeling I can certainly sympathize with.

But I suppose asking people to take a day vs. a week makes it a lot more palatable to onboarding people

PR #5656, opened 2020-04-08
feat(build-cops): Adds initial file-based schedule and syncing GH actions

ttomsu:
This PR integrates the https://github.com/spinnaker/rotation-scheduler CLI to extend a 7-day rotation schedule via GH Actions. An initial schedule is included here. All actions open a PR for review and request review from someone in `build-cops`. 

This PR also includes a GH Action to sync the schedule to  `build-cop@spinnaker.io`'s Google calendar, which I've set to be readable by users external to the `spinnaker.io` domain, but it takes 24 hrs for that change to propagate. I will follow up on this PR when build cops can subscribe to that calendar.

If users are OOO or otherwise unable to perform build cop duties (for example, I think @ethanfrogers may still be OOO for his 27-Apr thru 03-May shift), an update to that entry should look like:

```
- startDate: Mon 27 Apr 2020
  user: ethanfrogers
  userOverride: ttomsu
```
Upon any change to the schedule file, the calendar will be updated.

Occasionally we'll remove users from `build-cops` - in that case, the next time the job is invoked, all shifts from that user's next upcoming shift will be rescheduled. Existing `userOverride`s are cleared, so if you're reviewing that PR, please be on the lookout for that. See https://github.com/spinnaker/rotation-scheduler/pull/12


ttomsu:
`build-cop@spinnaker.io` can now be added to your Google Calendar - add it by following this link: https://calendar.google.com/calendar?cid=YnVpbGQtY29wQHNwaW5uYWtlci5pbw

Issue #5443, opened 2020-02-13
Migrate spinnakerbot to new K8s cluster (and/or to Cloud Run)


ttomsu:

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

ttomsu:
@ezimanyi - what is spinnakerbot's main function these days? Is this something we could achieve with other tools and without having to run our own K8s cluster?

ezimanyi:
@ttomsu : The exhaustive list is in the handlers in the `event/` and `policy/` directory in the repo, but the list of things it does (in decreasing order of usefulness in my subjective opinion):
* Tags and closes stale issues
* Adds `release/x.y` tags post-merge to PRs to indicate what release they'll be in
* Adds a message to pull requests that don't follow the commit conventions
* Adds a message to pull requests that add new files in Groovy
* Adds a message to pull requests against release branches if the commit message indicates it is a feature
* Adds a message and closes the PR if you forget to delete the "We prefer well-tested..." text from the pull request description
* Responds to `assign-issue`, `add-label`, `remove-label` commands

I'm sure a lot of this can be replaced by existing tools, or just removed (ie, I'm not sure we really need to have a command for people to assign issues...)

PR #5661, opened 2020-04-09
feat(spinbot): Replace the bulk of spinbot's issue labeling with a GHActions

ttomsu:
Setting DRY_RUN: true to see how it behaves. In theory, we may be able to turn down github.com/spinnaker/spinnakerbot after this.

tag: https://github.com/spinnaker/spinnaker/issues/5443

### DIFF
--- a/.github/workflows/stale-issues.yml
+++ b/.github/workflows/stale-issues.yml
@@ -1,0 +1,30 @@
+name: Mark and Close Stale Issues
+
+on:
+  schedule:
+  - cron: 0 */12 * * * # Every 12 hours
+
+jobs:
+  staleIssues:
+    # Only run this on repositories in the 'spinnaker' org, not on forks.
+    if: startsWith(github.repository, 'spinnaker/')
+    runs-on: ubuntu-latest
+    steps:
+    - name: stale
+      uses: gatsbyjs/stale@master
+      with:
+        DRY_RUN: true
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        DAYS_BEFORE_STALE: 45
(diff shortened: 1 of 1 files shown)


.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+  - cron: 0 */12 * * * # Every 12 hours
+
+jobs:
+  staleIssues:
+    # Only run this on repositories in the 'spinnaker' org, not on forks.
+    if: startsWith(github.repository, 'spinnaker/')
+    runs-on: ubuntu-latest
+    steps:
+    - name: stale
+      uses: gatsbyjs/stale@master
```
christopherthielen, line 14:
```suggestion
      # github release 0.1.0
      uses: gatsbyjs/stale@0dde19eb39fa2003923986adc3608c3e9178ebf3
```

.github/workflows/stale-issues.yml
```diff
@@ -0,0 +1,30 @@
...
+      with:
+        DRY_RUN: true
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
christopherthielen, line 17: This will perform actions as 'github-actions' bot.  Should we use use spinnakerbot's auth token instead?

It's also worth noting that the github-actions bot will need to be added as an oss-approver if we stick with `GITHUB_TOKEN` (spinnakerbot is an oss-approver already)

.github/workflows/stale-issues.yml
christopherthielen: I recommend tracking a specific release, and also using the exact commit sha (since tags can be deleted and re-created)

PR #6873, opened 2023-09-01
feat(gha): gha to delete untagged images older than 48 hours

link108:
In order to keeps costs under control, want to delete untagged images older than 48 hours. 

More info on GCR Cleaner can be found here: [https://github.com/GoogleCloudPlatform/gcr-cleaner/](https://github.com/GoogleCloudPlatform/gcr-cleaner/)

I have not tested this out via a GHA, but have done some dry runs locally. I've added the dry-run flag so that we can test this out on GHA and evaluate the output, before removing the dry-run flag

### DIFF
--- a/.github/workflows/clean-untagged-images.yml
+++ b/.github/workflows/clean-untagged-images.yml
@@ -1,0 +1,28 @@
+# .github/workflows/gcr-cleaner.yml
+name: 'gcr-cleaner'
+
+on:
+  schedule:
+    - cron: '0 0 */1 * *' # runs daily
+  workflow_dispatch: # allows for manual invocation
+
+jobs:
+  gcr-cleaner:
+    runs-on: 'ubuntu-latest'
+    steps:
+      # configure based on your registry
+      - name: Login to GAR
+        uses: 'docker/login-action@v2'
+        with:
+          registry: us-docker.pkg.dev
+          username: _json_key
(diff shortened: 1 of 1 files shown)


link108:
Alternatively, this could be done via [cleanup policies](https://cloud.google.com/artifact-registry/docs/repositories/cleanup-policy). For instance, we could apply this cleanup-policy:
```
[
  {
    "name": "keep-release-images",
    "action": {"type": "Keep"},
    "condition": {
      "tagPrefixes": ["spinnaker-1.1", "spinnaker-1.2", "spinnaker-1.3"],
      "tagState": "tagged"
    }
  },
  {
    "name": "delete-non-release-images-6-months-old",
    "action": {"type": "Delete"},
    "condition": {
      "olderThan": "180d"
    }
  },
  {
    "name": "delete-untagged-images-1-month-old",
    "action": {"type": "Delete"},
    "condition": {
      "tagState": "untagged",
      "olderThan": "30d"
    }
  }
]
```

This would 
- Keep all images tagged as part of a release
- Delete non-release images older than 6 months
- Delete untagged images after 30 days



Thoughts on either approach?

nemesisOsorio:
@link108 apparently using the cleanup-policy we need to update the `tagPrefixes` every time we have a new release, any idea how this policy could be updated?

dbyron-sf (approved):

PR #7035, opened 2025-04-14
fix(build): Fix Nexus library publishing

jcavanagh:
Plumbs the `publishToNexus` and related tasks up to the composite root.

### DIFF
--- a/spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/publishing/nexus/NexusPublishPlugin.groovy
+++ b/spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/publishing/nexus/NexusPublishPlugin.groovy
@@ -71,6 +71,27 @@ class NexusPublishPlugin implements Plugin<Project> {
         stagingProfileId = nexusExtension.nexusStagingProfileId()
       }
     }
+
+    project.afterEvaluate {
+      // For some reason, the nexus plugin seems to only create one of these named tasks ever
+      // As soon as the top-level task in Clouddriver is made, it won't ever define any more on other root projects
+      // However, it will happily define all the subproject versions of publishToNexus, which we can collect here
+      if(project.tasks.findByName("publishToNexus") == null) {
+        project.tasks.register("publishToNexus")
+      }
+
+      if(project.tasks.findByName("closeAndReleaseNexusStagingRepository") == null) {
+        project.tasks.register("closeAndReleaseNexusStagingRepository")
+      }
+
+      project.tasks.named("publishToNexus") {
+        it.dependsOn project.subprojects*.tasks*.findByName('publishToNexus').minus(null)
+      }
+
+      project.tasks.named("closeAndReleaseNexusStagingRepository") {
+        it.dependsOn project.subprojects*.tasks*.findByName('closeAndReleaseNexusStagingRepository').minus(null)
+      }
+    }
   }
 
   private void configureSpinnakerPublicationForNexus(Project project, NexusPublishExtension nexusExtension) {
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,18 @@ tasks.register('publish') { task ->
   }
 }
 
+tasks.register('publishToNexus') { task ->
+  mainProjects.each { build ->
+    task.dependsOn build.task(':publishToNexus')
+  }
+}
+
+tasks.register('closeAndReleaseNexusStagingRepository') { task ->
+  mainProjects.each { build ->
+    task.dependsOn build.task(':closeAndReleaseNexusStagingRepository')
+  }
+}
+
 tasks.register('test') { task ->
   mainProjects.each { build ->
     task.dependsOn build.task(':test')
--- a/spinnaker-gradle-project/spinnaker-project-plugin/build.gradle
+++ b/spinnaker-gradle-project/spinnaker-project-plugin/build.gradle
@@ -14,7 +14,7 @@ dependencies {
   implementation 'com.google.apis:google-api-services-artifactregistry:v1beta2-rev20221022-2.0.0'
   implementation 'com.google.api-client:google-api-client:2.0.0'
   implementation 'org.jetbrains.dokka:dokka-gradle-plugin:0.10.1'
-  implementation 'io.github.gradle-nexus:publish-plugin:1.3.0'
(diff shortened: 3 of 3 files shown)


spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/publishing/nexus/NexusPublishPlugin.groovy
```diff
@@ -71,6 +71,27 @@ class NexusPublishPlugin implements Plugin<Project> {
         stagingProfileId = nexusExtension.nexusStagingProfileId()
       }
     }
+
+    project.afterEvaluate {
+      // For some reason, the nexus plugin seems to only create one of these named tasks ever
+      // As soon as the top-level task in Clouddriver is made, it won't ever define any more on other root projects
+      // However, it will happily define all the subproject versions of publishToNexus, which we can collect here
+      if(project.tasks.findByName("publishToNexus") == null) {
```
dbyron-sf, line 79: The comment helps, but I'm still a little stuck on what's happening.  We're registering tasks that are somehow magically configured/defined by https://github.com/gradle-nexus/publish-plugin, but not properly registered?  Is there a bug there to fix?

And then after we register them in e.g. clouddriver, echo, fiat, etc. we declare that they depend on similar tasks in each of the individual modules (e.g. clouddriver-core, etc.) ?  I guess I can see how this stitches things together.  Blecch.

dbyron-sf: Looks like we're only using [1.3.0](https://github.com/spinnaker/spinnaker/blob/97cee4f3da768323984a4e1f404c96a9addf5306/spinnaker-gradle-project/spinnaker-project-plugin/build.gradle#L17) of https://github.com/gradle-nexus/publish-plugin, when the most recent is [2.0.0](https://github.com/gradle-nexus/publish-plugin/releases/tag/v2.0.0).  Time to upgrade?

jcavanagh: > The comment helps, but I'm still a little stuck on what's happening. We're registering tasks that are somehow magically configured/defined by https://github.com/gradle-nexus/publish-plugin, but not properly registered? Is there a bug there to fix?

The Nexus plugin defines a bunch of tasks that are named based on the destination repository (Spinnaker's is named `nexus`, so we get `publishToNexus`, etc).  

However, it behaves strangely - the Nexus plugin is defined and applied to every composite root project, but only creates a `publishToNexus` task on the first one, `clouddriver`.  `:clouddriver:publishToNexus` will exist, but `:echo:publishToNexus` and all the others won't.  

Here's what it looks like in an `afterEvaluate` hook while applying the plugin:

Clouddriver (works fine, `publishToNexus` exists):
image

Echo (doesn't work, no task with that name exists):
image

And the rest will look just like Echo - no `rootProject`-level publishing task created.  

> And then after we register them in e.g. clouddriver, echo, fiat, etc. we declare that they depend on similar tasks in each of the individual modules (e.g. clouddriver-core, etc.) ? I guess I can see how this stitches things together. Blecch.

Unfortunately pretty typical of composite builds.  Gradle has some internal magic that automatically finds named tasks on subprojects when run from the command line, and runs all of them.  From a "standard" build, when given the top-level command `./gradew spotlessApply` for example, Gradle will just find every same-named task on all subprojects and run them.  Not so for composites - the task must be explicitly created and linked to the included builds.  

e.g. for Spotless, there are actually no tasks created on the root project - tasks like `spotlessApply`, `spotlessCheck` etc, are only created on subprojects.  [I also had to shim those in the same way here](https://github.com/spinnaker/spinnaker/blob/main/spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/codestyle/SpinnakerCodeStylePlugin.groovy#L117) to account for losing that implied Gradle magic.

> Looks like we're only using [1.3.0](https://github.com/spinnaker/spinnaker/blob/97cee4f3da768323984a4e1f404c96a9addf5306/spinnaker-gradle-project/spinnaker-project-plugin/build.gradle#L17) of https://github.com/gradle-nexus/publish-plugin, when the most recent is [2.0.0](https://github.com/gradle-nexus/publish-plugin/releases/tag/v2.0.0). Time to upgrade?

Sure, seems to be about the same with some Gradle 8 compat added in.   

Also, the Nexus OSS product this is publishing to is EOL on June 30th, 2025, according [to this announcement](https://central.sonatype.org/news/20250326_ossrh_sunset/).

dbyron-sf: Well, isn't that fun.  Another migration.  Approving this and let's see what happens next.

dbyron-sf (approved):

spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/publishing/nexus/NexusPublishPlugin.groovy
jcavanagh: Sorry, I wrote the comment before pushing the actual upgrade change 😂  - could I bum a re-review?

dbyron-sf: Amazing if the new version actually works!

dbyron-sf (approved):

spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/publishing/nexus/NexusPublishPlugin.groovy
jcavanagh: I couldn't fully test since I don't actually have the credentials, but it gets as far as 401-ing all over the place when trying to upload, so I am hoping the real credentials fill in the gap 🤞

PR #7053, opened 2025-04-25
tests(local-ports): Enable dynamic port allocation for Prometheus and management server during integration tests

rahul-chekuri:
Replaces the previously hardcoded ports with dynamically assigned ports in the Kayenta integration tests.
This ensures that tests can run reliably without port conflicts, especially in parallel or shared environments.

Below are the details of the changes:

- EmbeddedPrometheusBootstrapConfiguration` now dynamically allocates and registers Prometheus container ports without using hardcoded ports.

- Removed dependency on predefined `server.port` and `management.server.port` values for integration tests.

- Management server port is dynamically obtained during runtime and exposed for Prometheus to scrape metrics.

- Prometheus container’s mapped port is dynamically registered into the environment (`embedded.prometheus.port`) at runtime.

- Fixed `ManagementTest` to use the dynamically registered Prometheus port (`embedded.prometheus.port`).

- Added logic in `BaseIntegrationTest` to update `dynamicEndpoint` to `PrometheusRemoteService`


kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/TestPrometheusConfiguration.java
```diff
@@ -0,0 +1,131 @@
+package com.netflix.kayenta.configuration;
```
dbyron-sf, line 1: Why the rename from EmbeddedPrometheusBootstrapConfiguration?  In the process, we lost the license header.

rahul-chekuri: We mainly renamed `EmbeddedPrometheusBootstrapConfiguration` to `TestPrometheusConfiguration` to clarify its intended usage. The original name included Bootstrap, which in Spring Boot contexts often implies early lifecycle loading or auto-configuration, which isn't the case now. 

This class is strictly used in test contexts to configure an embedded Prometheus instance. Renaming it with a test prefix helps prevent accidental use outside of tests and makes its role clearer to future contributors immediately.

That said, happy to revert the rename if you'd prefer we stick with the original name for consistency.

I will add the license header.

rahul-chekuri: https://github.com/spinnaker/spinnaker/pull/7053/commits/851e515d0e98a3035408fdcdc03e5dc33dd750e3

@dbyron-sf here is commit that reverted the name

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -13,66 +13,135 @@
...
+
+      prometheusContainer.start();
+
+      Map<String, Object> prometheusEnv =
+          registerEnvironment(env, prometheusContainer.getMappedPort(PROMETHEUS_INTERNAL_PORT));
+      log.info("Started Prometheus server. Connection details: {}", prometheusEnv);
+    };
   }
 
-  @Bean(name = "prometheus", destroyMethod = "stop")
```
dbyron-sf, line 53: How can we get this cleanup logic back?

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -13,66 +13,135 @@
...
+  @Bean
+  public WaitStrategy prometheusWaitStrategy() {
+    return new HttpWaitStrategy()
+        .forPath("/status")
+        .forPort(PROMETHEUS_INTERNAL_PORT)
+        .forStatusCode(200);
   }
 
-  private static final int PORT = 9090;
+  private GenericContainer<?> prometheusContainer;
```
dbyron-sf, line 57: At the moment it looks like this is only used in the prometheus method, so it could be a local variable there...Maybe that changes once there's logic to stop the container when the application context goes away though.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
```diff
@@ -16,26 +16,125 @@
...
 @AutoConfigureMetrics
 @SpringBootTest(
     classes = {MetricsReportingConfiguration.class, Main.class},
-    webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
     value = "spring.application.name=kayenta")
 @ExtendWith(SpringExtension.class)
 @ActiveProfiles({"base", "prometheus", "graphite", "cases"})
+@Import(EmbeddedPrometheusBootstrapConfiguration.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
dbyron-sf, line 50: Please add a comment explaining why this is here.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -13,66 +13,135 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package com.netflix.kayenta.configuration;
 
 import static com.playtika.test.common.utils.ContainerUtils.containerLogsConsumer;
 
-import com.netflix.kayenta.utils.EnvironmentUtils;
```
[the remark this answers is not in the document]
rahul-chekuri: done here
https://github.com/spinnaker/spinnaker/pull/7053/commits/935664dec54f435296c39f8dcd350f4110821063

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
rahul-chekuri: yes used this property to stop the container 
https://github.com/spinnaker/spinnaker/pull/7053/commits/935664dec54f435296c39f8dcd350f4110821063

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
rahul-chekuri: https://github.com/spinnaker/spinnaker/pull/7053/commits/935664dec54f435296c39f8dcd350f4110821063

Added a comment

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
```diff
@@ -16,26 +16,131 @@
...
+                        prometheusManagedAccount.getUsernamePasswordFile(),
+                        prometheusManagedAccount.getBearerToken());
+              } catch (IOException e) {
+                throw new RuntimeException(e);
+              }
+
+              prometheusManagedAccount.setPrometheusRemoteService(prometheusRemoteService);
+            });
+
+    Thread.sleep(30000); // 1/2 minute
```
dbyron-sf, line 119: Why do we need a sleep here?  And why 30 seconds?  What fails if we take this out?

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -13,66 +13,140 @@
...
-                MountableFile.forClasspathResource("/external/prometheus/prometheus.yml"),
-                "/etc/prometheus/prometheus.yml")
-            .waitingFor(prometheusWaitStrategy)
-            .withStartupTimeout(Duration.ofSeconds(30));
-    container.start();
-    Map<String, Object> env = registerEnvironment(environment, container.getMappedPort(PORT));
-    log.info("Started Prometheus server. Connection details: {}", env);
-    return container;
+  public void startPrometheusServer(ConfigurableEnvironment env) {
+    if (prometheusContainer != null && !prometheusContainer.isRunning()) {
```
dbyron-sf, line 84: Do we need to check isRunning here?  The code here would be simpler if we can depend on prometheusContainer.start() to do the right thing if the container is already started.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -13,66 +13,140 @@
...
+
+  private Map<String, Object> registerEnvironment(ConfigurableEnvironment environment, int port) {
     Map<String, Object> map = new LinkedHashMap<>();
     map.put("embedded.prometheus.port", port);
     EnvironmentUtils.registerPropertySource("embeddedPrometheusInfo", environment, map);
     return map;
   }
+
+  public void stopPrometheusContainer() {
+    if (prometheusContainer != null && prometheusContainer.isRunning()) {
```
dbyron-sf, line 146: Do we need to check isRunning here?  Why not call stop all the time?  It's simpler code for us, and also may protect against some race/edge conditions where some exception happens and we'd like to stop a partially started container.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -13,66 +13,140 @@
...
+  private GenericContainer<?> prometheusContainer;
 
-  @Bean(name = "prometheusWaitStrategy")
-  public WaitStrategy prometheusWaitStrategy() {
-    return new HttpWaitStrategy().forPath("/status").forPort(PORT).forStatusCode(200);
+  @Bean
+  public ApplicationListener<ApplicationReadyEvent> prometheus(
+      ConfigurableEnvironment env, WaitStrategy prometheusWaitStrategy) {
+    return event -> {
+      int managementPort = waitForManagementPort();
```
dbyron-sf, line 64: Do you know why it takes time for the management port to be available?  Is there some spring documentation that explains it?  I could also use some comments here about what's going on.  Like, when does spring actually execute this code that starts prometheus?  And then back to my original question, why isn't the management port available at that time?

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
rahul-chekuri: Kayenta uses schedulers to monitor Prometheus health, and these were originally configured with a fixed delay. Previously, when ports were hardcoded, Prometheus would start before the application and be ready by the time the schedulers ran.

However, with the current setup, Prometheus only starts after the application is fully up and can expose its port dynamically. As a result, the first scheduler run would fail since Prometheus wasn’t ready, causing some tests to fail as well.

To address this, I initially added a wait in the tests to allow the schedulers to run a second time—by which point Prometheus would be available, and the tests would pass. Now that the issue is clear, we can remove that wait and instead configure an initial delay (e.g., 30 seconds) for the schedulers. This ensures Prometheus is up before the first check runs.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
rahul-chekuri: ---

> **Do you know why it takes time for the management port to be available?**

```yaml
management:
  endpoints.web:
    base-path: /
    exposure.include: '*'
  endpoint.health.show-details: always
  server:
    port: 0
```

I'm not entirely sure why there's a delay, but this behavior occurs when `management.server.port` is explicitly configured. This setting was already present in the configuration when the task was handed over to me.

My assumption is that when `management.server.port` is explicitly configured, Spring Boot spins up a separate embedded server or does something extra for the management endpoints, which starts shortly after the main application context is initialized—possibly causing a delay in exposing the Prometheus metrics. I haven’t found an official Spring doc confirming this exact timing behavior, but removing the `management.server.port` configuration seems to eliminate the delay entirely.

Would you prefer that I remove this property to avoid needing the wait logic?

---

> **I could also use some comments here about what's going on.**

I will update the code with comments to explain this behavior more clearly.

---

> **When does Spring actually execute this code that starts Prometheus?**

This code starts executing immediately after the application starts. However, since Prometheus needs to bind to a port, it has to wait for the management server to initialize and expose that port—hence the delay.

---

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
```diff
@@ -16,26 +16,131 @@
...
 @AutoConfigureMetrics
 @SpringBootTest(
     classes = {MetricsReportingConfiguration.class, Main.class},
-    webEnvironment = SpringBootTest.WebEnvironment.DEFINED_PORT,
+    webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
     value = "spring.application.name=kayenta")
 @ExtendWith(SpringExtension.class)
 @ActiveProfiles({"base", "prometheus", "graphite", "cases"})
+@Import(EmbeddedPrometheusBootstrapConfiguration.class)
+@TestInstance(
```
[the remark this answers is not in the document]
rahul-chekuri: While addressing the issue mentioned in [this comment](https://github.com/spinnaker/spinnaker/pull/7053#discussion_r2066915088), we initially added logic to stop the Prometheus server in an `@AfterAll` method and used `@DirtiesContext `to spin up a fresh application context for each test class. However, we later decided against using `@DirtiesContext`, as it felt too heavy-handed and not ideal for our use case.

To improve this, I removed `@DirtiesContext` so that the same application instance and Prometheus server would be reused across all test classes. We now stop the Prometheus server in `@AfterAll`, which means it will be down after each test class finishes. To ensure tests don’t fail due to Prometheus being unavailable, we bring it back up in `@BeforeAll`.

However, since both the application and `@BeforeAll `execute around the same time, there's a race condition during the first test class execution where the Prometheus server might not be ready in time. To handle this case, we start the Prometheus server in the `EmbeddedPrometheusBootstrapConfiguration` if it's not already running.

I'll also be adding comments in the code to explain this flow for better clarity.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
```diff
@@ -16,26 +16,131 @@
...
+
+  @Autowired EmbeddedPrometheusBootstrapConfiguration prometheusConfig;
+
+  @Autowired ConfigurableEnvironment env;
+
+  @BeforeAll
+  public void setupPrometheusAccounts() throws InterruptedException {
+    prometheusConfig.startPrometheusServer(env);
+
+    int retries = 30; // wait up to 30 seconds
```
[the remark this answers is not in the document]
rahul-chekuri: As mentioned in my [previous comment](https://github.com/spinnaker/spinnaker/pull/7053#discussion_r2106534195), Prometheus is started only after the application has fully initialized, since it requires the port to be available for exposure. We're leveraging the ApplicationStartedEvent to trigger its startup.

However, since both ApplicationStartedEvent and @BeforeAll can run concurrently during the first test class execution, there's a potential race condition where the Prometheus server might not be up in time. To avoid test failures due to this timing issue, we're adding a short wait here to ensure Prometheus is running before the tests begin.

Starting from the second test class execution, this wait is no longer necessary—we can safely rely on the port returned by startPrometheusServer() since the Prometheus server is already up by then. I will try it and update here.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
rahul-chekuri: Will remove it

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
rahul-chekuri: Removed wait and added initial delay for schedulers

https://github.com/spinnaker/spinnaker/pull/7053/commits/9ab30db4a76dec4bc1afa3eb4710cca2c740f0f3

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
rahul-chekuri: removed here 

https://github.com/spinnaker/spinnaker/pull/7053/commits/9ab30db4a76dec4bc1afa3eb4710cca2c740f0f3

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
rahul-chekuri: removed here https://github.com/spinnaker/spinnaker/pull/7053/commits/9ab30db4a76dec4bc1afa3eb4710cca2c740f0f3

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
rahul-chekuri: removed custom port from properties and wait logic from the code

https://github.com/spinnaker/spinnaker/pull/7053/commits/9ab30db4a76dec4bc1afa3eb4710cca2c740f0f3

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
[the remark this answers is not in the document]
rahul-chekuri: Added comments here https://github.com/spinnaker/spinnaker/pull/7053/commits/9ab30db4a76dec4bc1afa3eb4710cca2c740f0f3

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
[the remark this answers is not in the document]
rahul-chekuri: For all test class executions except the first, we rely on the port returned by startPrometheusServer().
https://github.com/spinnaker/spinnaker/pull/7053/commits/9ab30db4a76dec4bc1afa3eb4710cca2c740f0f3

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/tests/BaseIntegrationTest.java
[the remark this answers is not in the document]
dbyron-sf: OK, I think this is starting to sink in.  Let me see if I have this right.
* spring boot exposes metrics on its (dynamic, and somehow slow to appear) management port
* prometheus scrapes metrics from spring boot's management port
* the integration tests use info from prometheus and so can't really start until all of this is up

It would be so nice if we could start prometheus in a `@BeforeAll`-annotated method and ditch the `ApplicationListener<ApplicationReadyEvent> prometheus` bean, but I guess the delay in the management port appearing makes this all super complicated.

kayenta/kayenta-integration-tests/src/test/java/com/netflix/kayenta/configuration/EmbeddedPrometheusBootstrapConfiguration.java
```diff
@@ -80,14 +87,16 @@ public ApplicationListener<ApplicationReadyEvent> prometheus(
     };
   }
 
-  public void startPrometheusServer(ConfigurableEnvironment env) {
-    if (prometheusContainer != null && !prometheusContainer.isRunning()) {
-      prometheusContainer.start();
+  public Integer startPrometheusServer(ConfigurableEnvironment env) {
+    prometheusContainer.start();
```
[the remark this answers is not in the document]
rahul-chekuri: done https://github.com/spinnaker/spinnaker/pull/7053/commits/3bcbbb2f9c67a464f7fd570a13be4974308e8378

PR #7060, opened 2025-04-30
fix(retrofit2): fix the issue of retrofit2 being unable to convert the Event object to RequestBody.

kirangodishala:
- It was observed that the igor was throwing the following error related to retrofit2 serialization. This PR addresses the issue by replacing the @Body type from an abstract `Event` type to `Map<String, Object>` and make changes accordingly at all the required places.

```
Caused by: java.lang.IllegalArgumentException: Unable to convert GenericBuildEvent(details={type=build, source=igor}, content=GenericBuildContent(project=GenericProject(name=spinnaker, lastBuild=GenericBuild(building=false, fullDisplayName=null, name=null, number=0, duration=null, timestamp=null, result=null, artifacts=null, testResults=null, url=null, id=null, genericGitRevisions=null, properties=null)), master=IgorHealthCheck, type=null)) to RequestBody (parameter #1)
    for method EchoService.postEvent
	at retrofit2.Utils.methodError(Utils.java:53) ~[retrofit-2.8.1.jar:na]
	at retrofit2.Utils.parameterError(Utils.java:58) ~[retrofit-2.8.1.jar:na]
	at retrofit2.ParameterHandler$Body.apply(ParameterHandler.java:413) ~[retrofit-2.8.1.jar:na]
	at retrofit2.RequestFactory.create(RequestFactory.java:117) ~[retrofit-2.8.1.jar:na]
	at retrofit2.OkHttpCall.createRawCall(OkHttpCall.java:194) ~[retrofit-2.8.1.jar:na]
	at retrofit2.OkHttpCall.getRawCall(OkHttpCall.java:99) ~[retrofit-2.8.1.jar:na]
	at retrofit2.OkHttpCall.execute(OkHttpCall.java:183) ~[retrofit-2.8.1.jar:na]
	at com.netflix.spinnaker.kork.retrofit.ErrorHandlingExecutorCallAdapterFactory$ExecutorCallbackCall.execute(ErrorHandlingExecutorCallAdapterFactory.java:150) ~[kork-retrofit-7.254.0.jar:7.254.0]
	at com.netflix.spinnaker.kork.retrofit.Retrofit2SyncCall.executeCall(Retrofit2SyncCall.java:47) ~[kork-retrofit-7.254.0.jar:7.254.0]
	at com.netflix.spinnaker.kork.retrofit.Retrofit2SyncCall.execute(Retrofit2SyncCall.java:34) ~[kork-retrofit-7.254.0.jar:7.254.0]
	at com.netflix.spinnaker.igor.health.EchoServiceHealthIndicator.lambda$checkHealth$1(EchoServiceHealthIndicator.java:83) ~[igor-web-4.22.0.jar:4.22.0]
```

### DIFF
--- a/igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoConverterFactory.java
+++ b/igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoConverterFactory.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2025 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.history;
+
+import com.fasterxml.jackson.databind.JavaType;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+
+public class EchoConverterFactory extends Converter.Factory {
+  private final ObjectMapper mapper;
+  private static final MediaType DEFAULT_MEDIA_TYPE =
+      MediaType.get("application/json; charset=UTF-8");
+
+  public static EchoConverterFactory create() {
+    return new EchoConverterFactory(new ObjectMapper());
+  }
+
+  private EchoConverterFactory(ObjectMapper mapper) {
+    this.mapper = mapper;
+  }
+
+  public Converter<ResponseBody, ?> responseBodyConverter(
+      Type type, Annotation[] annotations, Retrofit retrofit) {
+    if (type == Void.class) {
+      return (Converter<ResponseBody, Void>)
+          value -> {
+            value.close();
+            return null;
+          };
+    }
+    return (Converter<ResponseBody, Object>)
+        value -> {
+          JavaType javaType = mapper.getTypeFactory().constructType(type);
+          return mapper.readValue(value.charStream(), javaType);
+        };
+  }
+
+  public Converter<?, RequestBody> requestBodyConverter(
+      Type type,
+      Annotation[] parameterAnnotations,
+      Annotation[] methodAnnotations,
+      Retrofit retrofit) {
+    return (Converter<Object, RequestBody>)
+        value -> {
+          byte[] jsonValue = mapper.writeValueAsBytes(value);
+          return RequestBody.create(jsonValue, DEFAULT_MEDIA_TYPE);
+        };
+  }
+}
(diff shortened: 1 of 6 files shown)


dbyron-sf:
@kirangodishala Do you see a way forward to replace JacksonConverterFactory in [EchoConfig](https://github.com/spinnaker/spinnaker/blob/ff4c04b506f3ab320b72a0507ad3b8bb762186b0/igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/EchoConfig.groovy#L52) with something that does a simpler ObjectMapper.writeValueAsString like [retrofit1 does](https://github.com/square/retrofit/blob/parent-1.9.0/retrofit-converters/jackson/src/main/java/retrofit/converter/JacksonConverter.java#L48)?
```
.addConverterFactory(JacksonConverterFactory.create(objectMapper))
```
Seems like that could be quite a bit simpler.  Sorry for not talking about this before you've done all this work.

cristhian-castaneda (changes requested):
Noticed we’re now manually converting events to `Map<String, Object>` in quite a few places. It works, but starts to feel repetitive.

Since `@JsonTypeInfo` is already in use elsewhere, we could consider sticking with Jackson’s polymorphic support and keep `EchoService.postEvent(Event)` as is. That’d clean things up and avoid the need for explicit mapping.

If we do want to keep the `Map` approach, maybe we can centralize the logic to avoid duplicating `ObjectMapper` setup everywhere. Something like this could help:

```java
public class EventPayloadFactory {
  private static final ObjectMapper mapper = new ObjectMapper();

  public static Map<String, Object> toPayload(Object event) {
    return mapper.convertValue(event, new TypeReference<>() {});
  }
}
```

Both paths are valid, it’d just be nice to align on one so we don’t keep spreading this pattern around. Happy to help either way!

kirangodishala:
Will try to simplify the solution.

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoConverterFactory.java
```diff
@@ -0,0 +1,69 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import java.lang.annotation.Annotation;
+import java.lang.reflect.Type;
+import okhttp3.MediaType;
+import okhttp3.RequestBody;
+import okhttp3.ResponseBody;
+import retrofit2.Converter;
+import retrofit2.Retrofit;
+
+public class EchoConverterFactory extends Converter.Factory {
```
dbyron-sf, line 29: javadoc please, here and for the methods

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoService.java
```diff
@@ -24,5 +24,5 @@
 /** Posts new build executions to echo */
 public interface EchoService {
   @POST(".")
-  Call<String> postEvent(@Body Event event);
+  Call<Void> postEvent(@Body Event event);
```
dbyron-sf, line 27: Why the change in return value?

kirangodishala: It's because the echo api doesn't return anything: https://github.com/spinnaker/spinnaker/blob/main/echo/echo-web/src/main/java/com/netflix/spinnaker/echo/history/HistoryController.java#L36

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoConverterFactory.java
kirangodishala: Will update soon...

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoService.java
dbyron-sf: Fair enough, though it seems like an unrelated change...could it happen in a separate commit/PR, or does the new de/serializer code only work if this is Void?

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoConverterFactory.java
kirangodishala: Updated

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/history/EchoService.java
dbyron-sf: Ehh, let's go for it.

dbyron-sf (approved):

dbyron-sf:

PR #7066, opened 2025-05-06
fix(config): fix the issue of halyard failing to add docker registry account

kirangodishala:
- After upgrading clouddriver to retrofit2, clouddriver's DockerRegistryNamedAccountCredentials used in DockerRegistryAccountValidator needed to be updated without which the following error occurs when a docker registry account is added through hal command. 
```
Validation in default.provider.dockerRegistry.dockerhub:
! ERROR Unable to fetch tags from the docker repository:
  library/alpine, Cannot invoke method getService() on null object
? Can the provided user access this repository?

! ERROR Unable to fetch tags from the docker repository:
  xxxxx/yyyyyy, Cannot invoke method getService() on null object
? Can the provided user access this repository?
```
- This PR addresses the above  issue but since halyard is not upgraded to retrofit2 yet, the fix provided is a temporary one, constructing the ServiceClientProvider within DockerRegistryAccountValidator instead of autowiring.
-   Added a test to demonstrate the issue.

### DIFF
--- a/halyard/halyard-config/src/test/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dockerRegistry/DockerRegistryAccountValidatorTest.java
+++ b/halyard/halyard-config/src/test/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dockerRegistry/DockerRegistryAccountValidatorTest.java
@@ -1,0 +1,125 @@
+package com.netflix.spinnaker.halyard.config.validate.v1.providers.dockerRegistry;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
+import static com.github.tomakehurst.wiremock.core.WireMockConfiguration.wireMockConfig;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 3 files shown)


jasonmcintosh (dismissed):
UGH HALYARD!

halyard/halyard-config/src/test/java/com/netflix/spinnaker/halyard/config/validate/v1/providers/dockerRegistry/DockerRegistryAccountValidatorTest.java
```diff
@@ -0,0 +1,125 @@
...
+
+  @Autowired FileService fileService;
+
+  @Autowired DockerRegistryAccountValidator validator;
+
+  @RegisterExtension
+  static WireMockExtension wmDocker =
+      WireMockExtension.newInstance().options(wireMockConfig().dynamicPort()).build();
+
+  ConfigProblemSetBuilder configProblemSetBuilder = new ConfigProblemSetBuilder(context);
```
dbyron-sf, line 72: Seems safer to move this to the test method, since that's the only place it's used.  Since context is autowired, it may not be populated properly here.

kirangodishala: Done

dbyron-sf (approved):

dbyron-sf:

PR #7072, opened 2025-05-10
feature(kork-artifacts): Adds entity storage feature

xibz:
This commit adds the entity storage feature which allows large objects in the context to be stored by using the artifact store.

The entity storage uses the handler pattern which allows developers to specify how they want to handle any key or object in the execution context. There are two main handlers: the storage and expansion handlers.

The handlers are injected into our (de)serializer hooks, and are added to our "registry". The registry is just a serde modifier.

For example, when a map is about to be serialized or deserialized, it will first call the modifier to construct the appropriate (de)seriailizer, which is our new serde hooks.

These hooks house the handlers, and when the hook starts it's serde process, it will ask the handlers whether or not it can handle the object. If the handler can, the hook will then pass the object to the handler, and the handler can mutate it as needed like the common case of wanting to store large context objects. The same is true for the expansion handlers but instead of storing, it expands (does not necessarily need to expand).

To turn this feature on, add this to your spinnkaer configuration

```yaml
artifactStore:
  enabled: true
  entities:
    enabled: true
```

Additionally clouddriver requires one additional flag because it is the ONLY service so far that needs to expand the context. So this needs to also be added to the clouddriver.yaml
```yaml
artifactStore.entities.expand: true
```

In addition this commit adds:
* New exceptions to make use of Spring bubbling up and not overusing `RuntimeException`
* addresses that artifacts should always have some reference. This is now enforced with the @Nonnull annotation. Additional we will no longer serialize null values.
* Some helper functions to ArtifactTypes and adds a "CommonPrefixes" that can help identify artifact, e.g.  embedded vs remote.
* Artifact type based filtering So this change allows for filtering to happen at the artifact store level, and also by artifact type.

  ```yaml artifactStore: exclude: - type: "embedded/map/base64" value: "bad-application" ```

  This will look ONLY exclude `embedded/map/base64` and only for the application named `bad-application`. The value field is a regex so it does not need to be a simple as this example.

* Follow the commit message conventions [found here](https://spinnaker.io/community/contributing/submitting/).


clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/ArtifactReplacer.java
```diff
@@ -59,6 +62,15 @@ public class ArtifactReplacer {
 
   public ArtifactReplacer(Collection<Replacer> replacers) {
     this.replacers = ImmutableList.copyOf(replacers);
+    ObjectMapper mapper = Jackson2ObjectMapperBuilder.json().build();
```
jasonmcintosh, line 65: Is this the right mapper to use?  Don't we already pull in some object mappers fort his?  There's ... a few interesting flags particularly being JSON that should consider LIKE ignoring the unknown fields.  These may be set already... just wondering ;)

xibz: That should be the appropriate one. It's just the spring one that we register which is the primary bean

orca/orca-web/src/main/groovy/com/netflix/spinnaker/orca/web/config/WebConfiguration.groovy
```diff
@@ -67,6 +69,18 @@ class WebConfiguration {
     OrcaObjectMapper.getInstance()
   }
 
+  @Bean(name = "objectMapper") ObjectMapper orcaObjectMapper(Optional<SerializerHookRegistry> serializerModifier) {
```
jasonmcintosh, line 72: Isn't this now a duplicate bean from the core and not needed?  OR am I missing something?

kork/kork-artifacts/kork-artifacts.gradle
```diff
@@ -5,14 +5,18 @@ dependencies {
   api(platform(project(":spinnaker-dependencies")))
 
   implementation project(":kork-annotations")
+  implementation project(":kork-aws")
   implementation project(":kork-exceptions")
   implementation project(":kork-security")
   implementation "com.fasterxml.jackson.core:jackson-databind"
+  implementation platform('software.amazon.awssdk:bom')
```
jasonmcintosh, line 12: A ... concern:  Will this interact with the clouddriver AWS dependencies at all?  I'd assume not since those are set there.. .just wondering if we're breaking any other deps on the platform implementation dep here... And whether this is the "right" spot for that vs. spinnaker-dependencies which should also be applied everywhere?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStore.java
```diff
@@ -28,20 +32,44 @@ public class ArtifactStore implements ArtifactStoreGetter, ArtifactStoreStorer {
...
+  public boolean shouldExclude(String type, String application) {
+    return application == null
+        || this.exclude.containsKey(type)
+            && this.exclude.get(type).stream().anyMatch((filter) -> filter.filter(application));
   }
 
   /** Store an artifact in the artifact store */
-  public Artifact store(Artifact artifact) {
-    return artifactStoreStorer.store(artifact);
+  public Artifact store(Artifact artifact, ArtifactDecorator... decorators) {
```
jasonmcintosh, line 53: hrmm, shouldn't storage independent of app be allowed?  CERTAIN operations do operate independent of an application (mostly a few orchestration tasks).  NOT saying no :) just more wondering about this particular logic.  OR if the app should be "default" or some other thing in those cases.  NOT a "no" just me thinking about if there's certain operations here.

kork/kork-artifacts/kork-artifacts.gradle
xibz: No. This is v2 which clouddriver does not use. Once it does however, we will need to ensure they are compatible

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStore.java
xibz: Application is a requirement for artifact storage.

The reference urls are `ref://<application>/<hash>`. The reason we did this is to ensure that applications are not sharing artifacts which could get dangerous very quickly.

orca/orca-web/src/main/groovy/com/netflix/spinnaker/orca/web/config/WebConfiguration.groovy
xibz: Yep! I realized that we did not have the `@Bean{ mapper, objectMapper}` it was just `mapper`. So yea, this can go away!

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStore.java
jasonmcintosh: yeah mostly wondering if there should be a "default" similar to the way that RBAC uses an "anonymous" hidden user when no defaults are set.  This would allow "non-application" tasks to operate kinda thing. NOT that there are many of those...

xibz: There wasnt a use case for it at least for us, but it would trivial to do if anyone needs it.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/EntityPropertyAccessor.java
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    Map m = (Map) target;
+    if (!(m.containsKey(ARTIFACT_TYPE_KEY) && m.containsKey(ARTIFACT_REFERENCE_KEY))) {
+      return NO_TYPE;
+    }
+
+    Object typeObj = m.get(ARTIFACT_TYPE_KEY);
+    if (typeObj instanceof String) {
+      return (String) m.get(ARTIFACT_TYPE_KEY);
```
jasonmcintosh, line 193: NIT:  Can use the java 17 syntax
```java
Object obj = m.get(ARTIFACT_TYPE_KEY);
if (typeObj instanceof String str) { // Declare and cast in one step
    return str;
}```

File: kork/kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreStorerTest.java [hunk 6]
```diff
@@ -28,47 +28,42 @@
...
-    when(client.headObject(any(HeadObjectRequest.class)))
-        .thenThrow(S3Exception.builder().statusCode(400).build());
+    when(client.listObjectsV2(any(ListObjectsV2Request.class)))
+        .thenReturn(ListObjectsV2Response.builder().build());
+
+    verify(client, times(0)).putObject(any(PutObjectRequest.class), any(RequestBody.class));
     AuthenticatedRequest.set(Header.APPLICATION, "my-application");
     S3ArtifactStoreStorer artifactStoreStorer =
         new S3ArtifactStoreStorer(client, "my-bucket", new ArtifactStoreURISHA256Builder(), null);
-    String expectedExceptionMessage = "Failed to query artifact due to invalid request";
```
jasonmcintosh, line 53: I'm not seeing any failure tests right now on failure handling.  It's relatively minor, but we've lost a few of the exception handling test cases where or whatever reason, s3 blows up.

jasonmcintosh: Ah I see now for some reason the change in the s3storer didn't show up tracing through this.  Ignore this as you've refactored the code on this.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreStorer.java
```diff
@@ -144,33 +147,11 @@ private byte[] getReferenceAsBytes(Artifact artifact) {
...
-        // exception with a more meaningful message
-        throw new SpinnakerException(buildHeadObjectExceptionMessage(e), e);
-      }
-
-      throw new SpinnakerException("S3 head object failed", e);
-    }
+    ListObjectsV2Request request =
+        ListObjectsV2Request.builder().bucket(bucket).prefix(uri.paths()).maxKeys(1).build();
+
+    ListObjectsV2Response resp = s3Client.listObjectsV2(request);
```
jasonmcintosh, line 151: NIT:  Don't need the object declaration here.  Not a blocker, just... a note.  ALSO hasContents doesn't mean the object ACTUALLY exists.  https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/ListObjectsV2Response.html#hasContents()  probably should do a resp.getContents().isEmpty() check as well.

orca/orca-web/src/main/groovy/com/netflix/spinnaker/orca/web/config/WebConfiguration.groovy
```diff
@@ -16,6 +16,8 @@
 
 package com.netflix.spinnaker.orca.web.config
 
+import com.fasterxml.jackson.databind.module.SimpleModule
```
jasonmcintosh: This seems... to be unused set of imports?

dbyron-sf:
> To turn this feature on, add this to your spinnaker configuration
>```
> artifactStore:
>   enabled: true
>   entities:
>     enabled: true
>```

For which microservices do we need to configure this?

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/ArtifactReplacer.java
```diff
@@ -59,6 +62,15 @@ public class ArtifactReplacer {
...
+    SerializerHookRegistry serializer = SerializerHookRegistry.getINSTANCE();
+    if (serializer != null) {
+      SimpleModule module = new SimpleModule();
+      module.setSerializerModifier(serializer);
+      mapper.registerModule(module);
+    }
+
+    this.mapper = mapper;
```
dbyron-sf, line 73: Could the mapper still be static, and we do all this once, instead of once per ArtifactReplacer object?

clouddriver/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -84,10 +87,19 @@ class Main extends SpringBootServletInitializer {
...
   @Primary
-  ObjectMapper objectMapper(Jackson2ObjectMapperBuilder builder) {
-    return builder.createXmlMapper(false)
-      .mixIn(Artifact.class, ArtifactMixin.class)
-      .build()
+  ObjectMapper objectMapper(Jackson2ObjectMapperBuilder builder, Optional<DeserializerHookRegistry> deserializerHook, Optional<SerializerHookRegistry> serializerHook) {
+    builder = builder.createXmlMapper(false)
+      .mixIn(Artifact.class, ArtifactMixin.class);
+
+    SimpleModule module = new SimpleModule("registryHook")
```
dbyron-sf, line 94: Where does this magic string `registryHook` come from?  Is there some central place to store it as a string constant?

clouddriver/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
```diff
@@ -84,10 +87,19 @@ class Main extends SpringBootServletInitializer {
...
+    if (deserializerHook.isPresent()) {
+      module.setDeserializerModifier(deserializerHook.get())
+    }
+    if (serializerHook.isPresent()) {
+      module.setSerializerModifier(serializerHook.get());
+    }
+    builder.modules(l -> l.add(module))
```
dbyron-sf, line 101: Is it necessary to add the module if neither hook is present?

deck/packages/kubernetes/src/pipelines/stages/deployManifest/DeployManifestStageForm.tsx
```diff
@@ -58,12 +59,9 @@ export class DeployManifestStageForm extends React.Component<
 
   public constructor(props: IDeployManifestStageConfigFormProps & IFormikStageConfigInjectedProps) {
     super(props);
-    const stage = this.props.formik.values;
```
dbyron-sf, line 61: Seems like there's a refactor here, or possibly a fix.  I can't tell what's the refactor/fix part and what (if anything) is required to support the entity store though.  Any chance you can split the refactor/fix into a separate commit?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/ArtifactTypeDecorator.java
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts;
+
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactDecorator;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.exceptions.ArtifactStoreInvalidTypeException;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+
+public class ArtifactTypeDecorator implements ArtifactDecorator {
```
dbyron-sf, line 26: javadoc please

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/ArtifactTypeDecorator.java
```diff
@@ -0,0 +1,63 @@
...
+
+  public static ArtifactTypeDecorator toEmbedded(Artifact artifact) {
+    return new ArtifactTypeDecorator(
+        convertTo(
+            artifact,
+            ArtifactTypes.CommonAffixes.REMOTE.asString(),
+            ArtifactTypes.CommonAffixes.EMBEDDED.asString()));
+  }
+
+  private static String convertTo(Artifact artifact, String from, String to) {
```
dbyron-sf, line 90: javadoc please and perhaps some comments inside the method explaining what's going on.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/ArtifactTypes.java
```diff
@@ -18,11 +18,78 @@
...
+   * "embedded/foo/artifact" will map to a stored artifact of "remote/foo/artifact".
+   */
+  REMOTE_BASE64(CommonAffixes.REMOTE.asString() + "/" + CommonAffixes.BASE64.asString()),
+  REMOTE_MAP_BASE64(CommonAffixes.REMOTE.asString() + "/map/" + CommonAffixes.BASE64.asString()),
   ;
 
+  /**
+   * CommonAffixes describe the overall structure of the artifact.
+   *
+   * <p>The location type describes where it is located, e.g., remote or embedded. * An embedded
```
dbyron-sf, line 54: remove extra `*` ?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/ArtifactTypes.java
```diff
@@ -18,11 +18,78 @@
...
+    private final String value;
+
+    public String asString() {
+      return this.value;
+    }
+  }
+
   @Getter private final String mimeType;
+
+  public static boolean isEmbedded(ArtifactTypes t) {
```
dbyron-sf: Doesn't look like this is used.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStore.java
```diff
@@ -28,20 +32,44 @@ public class ArtifactStore implements ArtifactStoreGetter, ArtifactStoreStorer {
...
+    String application = AuthenticatedRequest.getSpinnakerApplication().orElse(null);
+    if (application == null) {
+      log.warn("failed to retrieve application from request artifact={}", artifact.getName());
+      return artifact;
```
dbyron-sf, line 72: Please update the javadoc to describe the return value, and what happens in these "failure" cases.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStore.java
```diff
@@ -28,20 +32,44 @@ public class ArtifactStore implements ArtifactStoreGetter, ArtifactStoreStorer {
...
+    }
+
+    String type = artifact.getType();
+    if (this.exclude.containsKey(type)
+        && this.exclude.get(type).stream().anyMatch((filter) -> filter.filter(application))) {
+      log.debug(
+          "filtering artifact type for application type={} application={}", type, application);
```
dbyron-sf:
```suggestion
          "excluding artifact type for application type={} application={}", type, application);
```
?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
```diff
@@ -16,14 +16,22 @@
...
+import java.util.Map;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.ComponentScan;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 
 @Configuration
+@ComponentScan
```
dbyron-sf, line 34: Please add a comment explaining whywe need this.  Seems like it's to pick up DeserializerHookRegistry and SerializerHookRegistry.  Is that right?  Do we also need `@ComponentScan` in EntityStoreConfiguration?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
```diff
@@ -56,9 +64,27 @@ public ArtifactStoreStorer artifactStoreStorer() {
...
+      ArtifactStoreStorer artifactStoreStorer,
+      Map<String, List<ApplicationStorageFilter>> exclude) {
+    ArtifactStore artifactStore =
+        new ArtifactStore(artifactStoreGetter, artifactStoreStorer, exclude);
     ArtifactStore.setInstance(artifactStore);
     return artifactStore;
   }
+
+  @Bean
+  public Map<String, List<ApplicationStorageFilter>> excludeFilters(
```
dbyron-sf, line 96: Please add javadoc that describes the keys/values in the return value.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/NoopArtifactStoreStorer.java
```diff
@@ -20,7 +20,7 @@
 /** A no-op ArtifactStoreStorer. In other words, don't actually store the artifact. */
 public class NoopArtifactStoreStorer implements ArtifactStoreStorer {
 
-  public Artifact store(Artifact artifact) {
+  public Artifact store(Artifact artifact, ArtifactDecorator... decorators) {
```
dbyron-sf, line 24: Add `@Override`?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreStorer.java
```diff
@@ -23,5 +23,5 @@
  */
 public interface ArtifactStoreStorer {
 
-  Artifact store(Artifact artifact);
+  Artifact store(Artifact artifact, ArtifactDecorator... decorators);
```
dbyron-sf, line 26: javadoc please

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ArtifactExpandHandler.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.entities;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStore;
+
+/**
+ * ArtifactExpandHandler is a handler that handles expansion of some object.
+ *
+ * <p>It is important to note that the same type MUST be returned as the original parameter v's
```
dbyron-sf: Seems like the javadoc for the handle method would be a better place for this comment.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ArtifactStorageHandler.java
```diff
@@ -0,0 +1,35 @@
...
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStore;
+
+/**
+ * ArtifactStorageHandler is used to hook into the serializers to handle a particular class, e.g. a
+ * Map, List, etc. This allows us to easily extend the functionality by providing a handler to hook
+ * into rather than writing a custom serializer for each use case.
+ *
+ * <p>It is very important that ANY handling MUST return the same type as the original parameter, v.
```
dbyron-sf: same thought here about moving the comment to handle's javadoc

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ArtifactStoragePropertyHandler.java
```diff
@@ -0,0 +1,53 @@
...
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.entities;
+
+import com.fasterxml.jackson.databind.BeanProperty;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.netflix.spinnaker.kork.artifacts.artifactstore.ArtifactStore;
+
+/**
+ * An artifact storage property handler is a special type of handler specific of handling fields in
+ * an object.
```
dbyron-sf, line 24: I could use some sample json to go along with this example.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/MapDeserializerHook.java
```diff
@@ -0,0 +1,96 @@
...
+  @Override
+  public JsonDeserializer<?> createContextual(DeserializationContext ctxt, BeanProperty property)
+      throws JsonMappingException {
+    return new MapDeserializerHook(
+        this.storage,
+        this.handlers,
+        (MapDeserializer) this.deserializer.createContextual(ctxt, property));
+  }
+
+  private Object visit(Map<Object, Object> m, ObjectMapper objectMapper) {
```
dbyron-sf, line 67: javadoc please, potentially also with comments in the method that describe what's happening.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestStorageCollectionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+    if (EntityHelper.alreadyStored(v)) {
+      return t;
+    }
+
+    Artifact artifact = EntityHelper.toArtifact(v, ArtifactTypes.EMBEDDED_MAP_BASE64.getMimeType());
+    Artifact stored = store.store(artifact, ArtifactTypeDecorator.toRemote(artifact));
+    if (ArtifactTypes.EMBEDDED_MAP_BASE64.getMimeType().equals(stored.getType())) {
+      return t;
+    }
+    // We can cast directly to T due to our check of canHandle.
```
dbyron-sf, line 90:
```suggestion
    // We can cast directly to T due to our check in canHandleProperty.
```
?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestStorageCollectionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  @Override
+  public boolean canHandleProperty(BeanProperty property, Object value) {
+    if (AuthenticatedRequest.getSpinnakerExecutionId().isEmpty() || this.exclude.shouldFilter()) {
+      return false;
+    }
+
+    String propertyName = property.getName();
+    Collection<?> v = (Collection<?>) value;
```
dbyron-sf, line 120: What if value isn't a Collection.  Safer to have an instanceof check?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestStorageCollectionHandler.java
```diff
@@ -0,0 +1,106 @@
...
+      return false;
+    }
+
+    Object obj = v.iterator().next();
+    if (!(obj instanceof Map)) {
+      return false;
+    }
+    // Good ol' clouddriver sometimes stores manifests as a list and other times as a set.
+    return this.keys.stream().anyMatch(k -> k.equals(propertyName))
+        && !EntityHelper.alreadyStored((Map) obj);
```
dbyron-sf, line 131: I'm surprised to see something dynamic like `alreadyStored` in a method like canHandleProperty.  Couldn't this code handle the object whether it's stored or not?

dbyron-sf:
@xibz please mention all the new config flags in the PR description, including:
* expression.aggressiveExpansion

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ExpandToMapHandler.java
```diff
@@ -0,0 +1,49 @@
...
+  public <T> T handle(ArtifactStore store, Object v, Class<T> clazz, ObjectMapper objectMapper) {
+    Map m = (Map) v;
+    String uri = (String) m.get("reference");
+    Artifact artifact = store.get(ArtifactReferenceURI.parse(uri));
+    String reference = artifact.getReference();
+    byte[] b = Base64.getDecoder().decode(reference);
+    try {
+      return objectMapper.readValue(b, clazz);
+    } catch (IOException e) {
+      throw new ArtifactStoreHandlerException("Failed to handle expansion", e);
```
dbyron-sf, line 46: include the artifact reference in the exception message?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOException.java
```diff
@@ -20,7 +20,7 @@
 
 /**
  * An exception used by the artifact store (de)serializers since the (de)serialize methods only
- * throw IOExceptions, and if any other exception is thrown jackson assumes it's some JSON error.
+ * throws IOExceptions, and if any other exception is thrown, jackson assumes it's some JSON error.
```
dbyron-sf, line 23: Isn't `throw` correct here?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/EntityPropertyAccessor.java
```diff
@@ -0,0 +1,197 @@
...
+
+    String type = extractType(target);
+    if (this.supportedTypes.stream().noneMatch(t -> t.equals(type))) {
+      throw new ArtifactStoreInvalidTypeException(type);
+    }
+
+    if (target instanceof Map) {
+      return new TypedValue(handleMap((Map) target, name));
+    }
+    throw new ArtifactStoreInvalidStateException(
```
dbyron-sf, line 111: If I'm following correctly, this corresponds to a 500 response code, which callers typically retry.  It doesn't make sense to retry this though, right?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestMapStorageHandler.java
```diff
@@ -0,0 +1,139 @@
...
+public class ManifestMapStorageHandler implements ArtifactStorageHandler {
+  private List<String> keys = List.of("manifests", "outputs.manifests");
+  private final ApplicationFilter exclude;
+
+  public ManifestMapStorageHandler(Map<String, List<ApplicationStorageFilter>> exclude) {
+    this.exclude =
+        new ApplicationFilter(exclude.get(ArtifactTypes.EMBEDDED_MAP_BASE64.getMimeType()));
+  }
+
+  /** Manifest storage can only be handled on a list of maps. */
```
dbyron-sf: I could use some more comments here.  From looking at the code, this returns false unless `v` is a map, which at first glance means the code handles maps, but not "a list of maps" like the comment says.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestMapStorageHandler.java
```diff
@@ -0,0 +1,139 @@
...
+    if (!(o instanceof Map)) {
+      return false;
+    }
+
+    Map<?, ?> m = (Map<?, ?>) o;
+    if (m.isEmpty()) {
+      return false;
+    }
+    Map.Entry<?, ?> e = m.entrySet().stream().findFirst().get();
+    return e.getKey() instanceof String;
```
dbyron-sf, line 149: Is this really specific enough to identify a manifest -- a map whose first key is a string?  Plus also, does `entrySet().stream()` guarantee order?  If the answer is no then this examines an arbitrary entry in the map.  I'd have expected this to look for something specific to kubernetes.  Can you add some comments explaining why this logic is the way it is?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/MapSerializerHook.java
```diff
@@ -0,0 +1,175 @@
...
+   * null. The serializer workflow will pass in the key serializer by using the withResolved method.
+   */
+  private final MapSerializer defaultSerializer;
+
+  private final ArtifactStore storage;
+  /** Handlers to handle specific keys in the JSON. */
+  private final List<ArtifactHandler> handlers;
+
+  public MapSerializerHook(
+      ArtifactStore storage, List<ArtifactHandler> handlers, MapSerializer defaultSerializer) {
```
dbyron-sf, line 54: Check that handlers isn't null?  Otherwise visit could end up throwing a NullPointerException.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/ArtifactTypeDecorator.java
```diff
@@ -0,0 +1,63 @@
...
+    String t = artifact.getType();
+    if (t == null || !t.startsWith(from + "/")) {
+      throw new ArtifactStoreInvalidTypeException(
```
dbyron-sf, line 93: This corresponds to a 500 response code, but isn't worth retrying, is it?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/filters/ApplicationFilter.java
```diff
@@ -0,0 +1,43 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.filters;
+
+import com.netflix.spinnaker.security.AuthenticatedRequest;
+import java.util.List;
+
+public class ApplicationFilter {
```
dbyron-sf, line 25: javadoc please, for the class and method.  Especially please clarify whether filter here means include or exclude.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/filters/ApplicationStorageFilter.java
```diff
@@ -0,0 +1,20 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.filters;
+
+public interface ApplicationStorageFilter {
+  boolean filter(String application);
```
dbyron-sf, line 36: javadoc please

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/filters/RegexApplicationStorageFilter.java
```diff
@@ -0,0 +1,31 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.filters;
+
+import java.util.regex.Pattern;
+
+public class RegexApplicationStorageFilter implements ApplicationStorageFilter {
```
dbyron-sf, line 27: javadoc please

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreConfiguration.java
```diff
@@ -72,20 +77,25 @@ public ArtifactStoreGetter artifactStoreGetter(
   }
 
   @Bean
-  public S3Client artifactS3Client(ArtifactStoreConfigurationProperties properties) {
-    S3ClientBuilder builder = S3Client.builder();
+  public S3Client artifactS3Client(
```
dbyron-sf, line 80: This feels like a separate change from the entity store.  Is it?  What does it do?   Worth separating into its own commit, potentially with a test?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreStorer.java
```diff
@@ -144,33 +147,11 @@ private byte[] getReferenceAsBytes(Artifact artifact) {
...
-        // due to this being a HEAD request, there is no message giving a clear
-        // indication of what failed. Rather than seeing a useful message back
-        // to gate, we instead see just null. To alleviate this, we wrap the
-        // exception with a more meaningful message
-        throw new SpinnakerException(buildHeadObjectExceptionMessage(e), e);
-      }
-
-      throw new SpinnakerException("S3 head object failed", e);
-    }
+    ListObjectsV2Request request =
```
dbyron-sf, line 148: Why the change from HeadObjectRequest?

Plus also, at the moment buildHeadObjectExceptionMessage isn't used.  We might be better off with something similar for listing objects.

kork/kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreTest.java
```diff
@@ -24,13 +24,22 @@
...
 
   private ArtifactStoreGetter artifactStoreGetter = mock(ArtifactStoreGetter.class);
   private ArtifactStoreStorer artifactStoreStorer = mock(ArtifactStoreStorer.class);
-  private ArtifactStore artifactStore = new ArtifactStore(artifactStoreGetter, artifactStoreStorer);
+  private ArtifactStore artifactStore =
+      new ArtifactStore(artifactStoreGetter, artifactStoreStorer, new HashMap<>());
+
+  @BeforeAll
+  static void init() {
+    AuthenticatedRequest.setApplication("applicaiton");
```
dbyron-sf:
```suggestion
    AuthenticatedRequest.setApplication("application");
```

kork/kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/DeserializerHookRegistryTest.java
```diff
@@ -0,0 +1,186 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.kork.artifacts.artifactstore.entities;
+
+import static org.junit.jupiter.api.Assertions.*;
```
dbyron-sf: please avoid star imports.

kork/kork-artifacts/src/test/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/DeserializerHookRegistryTest.java
```diff
@@ -0,0 +1,186 @@
...
+                              "type": "remote/map/base64",
+                              "reference": "ref://application/hash"
+                            }
+                         ]
+                     }
+                    """,
+            Map.of("manifests", List.of(Map.of("hello", "world!")))));
+  }
+
+  @SneakyThrows
```
dbyron-sf: Is this better than adding `throws Exception` to the method signature.  Seems like that's pretty typical for test methods.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreStorer.java
xibz: Oh, turns out `HeadObject` is painfully slow than list objects v2.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestStorageCollectionHandler.java
xibz: Ah this part is a little tricky, but I believe it can ONLY be a collection due to where this is registered.

However, I could add the check in the event someone wants to use this elsewhere

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreConfiguration.java
xibz: This commit was old, and I didn't want to dissect the difference. I believe open source handled custom credentials to AWS differently than how we implemented it internally, but I know we've tested our change a lot internally, so I went with ours, but I can revert if needed.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestMapStorageHandler.java
xibz: We are just type checking. Less about determining the contents of the manifests. If the keys are "manifests" or "outputs.manifests" we just make the assumption it's either remote or proper manifests

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestStorageCollectionHandler.java
xibz: It can, but we want to avoid calling the serializers when we can. And we could do this check within the handle property, but it's better to just rely on the original deserializer.

But let me play around with it

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/exceptions/ArtifactStoreIOException.java
xibz: Yep!

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreConfiguration.java
dbyron-sf: Looks like perhaps you guys have an AwsCredentialsProvider bean that, with this new code, takes precedence over the one `getCredentialsProvider` builds.  That seems generally fine.  It would be great to have in a separate commit, or even a separate PR, so we could see that all the existing tests still pass...and as a bonus, have some additional test coverage for this new code path (i.e. non-null AwsCredentialsProvider argument).

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestMapStorageHandler.java
dbyron-sf: Aaah, ok.  Up the call stack, canHandle uses `keys` which contains `manifests` and `output.manifests`.  So if either of those keys exist and is a list that contains maps...we then look at the first map in the list, and an arbitrary property in that map.  If it's a string, we're good....

Am I right that:
```
    Map.Entry<?, ?> e = m.entrySet().stream().findFirst().get();
    return e.getKey() instanceof String;
```
is the part that distinguishes a remote manifest from a proper one?  Can you add a comment about that...and what non-String thing remote manifests are?

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/ArtifactTypeDecorator.java
xibz: Hmm, why? The type is invalid?

dbyron-sf: Yeah, if it failed once, isn't it gonna fail on retry too?

jasonmcintosh: I've seen this fail due to an internal bug that's later retried & fixed.  USUALLY when an async operation hasn't completed, and a null or similar check on that operation being there.  I'd lean towards keeping a retry on these myself...

clouddriver/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
xibz: Oh, it actually is completely optional, but helps with debugging if you look in your object mapper you can see what modules are registered and if they are named, it makes it easier to vet.

clouddriver/clouddriver-web/src/main/groovy/com/netflix/spinnaker/clouddriver/Main.groovy
xibz: No, but I did feel like this was easier to read. Given this is a bean and only is done once at startup, I did not think I need to wrap it in an if. However, if you feel strongly about it, I can change it

dbyron-sf: I'm not sure of the consequences of having more modules in the ObjectMapper...My reflex is that performance will be the best with only the modules we really need, but I don't have data.  I feel like it's worth an if statement though.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/ArtifactStoreConfiguration.java
xibz: Yes, it's to pick up the hook registries. We only need `ComponentScan` on `EntityStoreConfiguration` if we ever want to just utilize this somehow without the artifact store. I can add it, but it is currently not needed, but would allow us to not remember needing to add it if it ever needed to be a stand alone package

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/entities/ManifestMapStorageHandler.java
xibz: > Am I right that:

    Map.Entry<?, ?> e = m.entrySet().stream().findFirst().get();
    return e.getKey() instanceof String;


> is the part that distinguishes a remote manifest from a proper one? Can you add a comment about that...and what non-String thing remote manifests are?

I dont think so. All this is doing is saying is the key a string. The `canHandle` method do not check anything about whether or not anything is remote or not.

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/artifact/ArtifactReplacer.java
xibz: But since we are building it from the jackson object, this means we will either need an atomic set or allow it to have multiple stomp over the static mapper. I think if we really want it to be static, we need to use an atomic.

kork/kork-artifacts/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/s3/S3ArtifactStoreStorer.java
jasonmcintosh: Following up on the:
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/s3/model/ListObjectsV2Response.html#hasContents() probably should do a resp.getContents().isEmpty() check as well.

? @xibz

xibz: Not necessarily. So, this has always been confusing in the list objects v2 API, but `hasContents` says did something get returned from S3? That's all we care about. Since we always provide a full, non-prefix key (never a folder or partial path), the presence of any returned object implies the object exists. If S3 returns no contents, then the object does not exist.

Issue #7065, opened 2025-05-06
Clouddriver fails to fetch new docker images from Azure Container Registry


abreunig359:
### Issue Summary:
New docker image tags that are pushed to the container registry no longer trigger pipelines and can no longer be selected  for a manual execution after deploying clouddriver version 5.91.0 (first version with Retrofit 2) and later.
Older image tags can still be selected for manual execution.

### Cloud Provider(s):
Azure

### Environment:
Private Azure container registry configured with username, password and address `https://<registry_name>.azurecr.io`

### Feature Area:
Pipelines with trigger by container registry

### Description:
- Expected behavior: Pipelines are triggered when a new docker image tag is pushed to the private Azure container registry and newly pushed tags can be selected for manual execution
- Actual behavior: When a new image is pushed to the private Azure container registry the pipelines are no longer triggered and the new tag cannot be selected for manual execution

### Steps to Reproduce:
- Launch clouddriver with a version starting from 5.91.0 with the existing configuration for the private Azure container registry

### Additional Details:
Exception during startup:
```
2025-05-06 08:16:15.701  WARN 1 --- [ecutionAction-4] c.n.s.c.cache.LoggingInstrumentation     : dockerRegistry:risingregistry/DockerRegistryImageCachingAgent[1/1] completed with one or more failures in 0s

java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at org.codehaus.groovy.runtime.dgmimpl.arrays.ObjectArrayGetAtMetaMethod.invoke(ObjectArrayGetAtMetaMethod.java:44) ~[groovy-4.0.15.jar:4.0.15]
	at groovy.lang.MetaMethod.doMethodInvoke(MetaMethod.java:328) ~[groovy-4.0.15.jar:4.0.15]
	at org.codehaus.groovy.vmplugin.v8.IndyInterface.fromCache(IndyInterface.java:321) ~[groovy-4.0.15.jar:4.0.15]
	at com.netflix.spinnaker.clouddriver.docker.registry.api.v2.client.DockerRegistryClient$_parseForQueryParams_closure20.doCall(DockerRegistryClient.groovy:502) ~[clouddriver-docker-2025.0.0.jar:2025.0.0]
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:na]
[...]

PR #7073, opened 2025-05-12
fix(docker): fix the issue of fetching docker images

kirangodishala:
- A docker registry when responding with next link to query for catalog repositories or tags, it is expected to send two query parameters : 1. `n` - number of items per page and 2. `last` - last item rendered in the current response. But Azure Container Registry was also sending one more query parameter (`orderby`) with no value which is causing the failure while parsing the `next` link.  https://github.com/spinnaker/spinnaker/issues/7065
```
java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
...
	at com.netflix.spinnaker.clouddriver.docker.registry.api.v2.client.DockerRegistryClient$_parseForQueryParams_closure20.doCall(DockerRegistryClient.groovy:502) ~[clouddriver-docker-2025.0.0.jar:2025.0.0]

```
- This PR addresses the issue by accepting blank value for a query param.

### DIFF
--- a/clouddriver/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
+++ b/clouddriver/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
@@ -499,7 +499,8 @@ class DockerRegistryClient {
     }
     if (queryParamsString) {
       queryParams = queryParamsString.split("&").collectEntries { param ->
-        def (key, value) = param.split("=")
+        // ensures two strings are returned even when a query parameter has no value(returns empty string).
+        def (key, value) = param.split("=", 2)
         [key, value]
       }
     }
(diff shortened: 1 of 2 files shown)


christosarvanitis (approved):
This will work. For reference a blank orderby in ACR means order by alphabetical order. Other accepted values are time_asc,time_desc

clouddriver/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/api/v2/client/DockerRegistryClient.groovy
```diff
@@ -499,7 +499,7 @@ class DockerRegistryClient {
     }
     if (queryParamsString) {
       queryParams = queryParamsString.split("&").collectEntries { param ->
-        def (key, value) = param.split("=")
+        def (key, value) = param.split("=", 2)
```
dbyron-sf, line 502: I could use a comment here about why 2 is correct/important.

kirangodishala: Will add a comment. It means, it returns two items even when there is no param value.

dbyron-sf:

PR #7078, opened 2025-05-16
feat(kayenta): adding outlierStrategy field in metric config modal

zrashwani:
Add outlier strategy field to the canary config modal, this config exists in Kayenta config schema [here](https://github.com/spinnaker/spinnaker/blob/93ea3ff5b9c804d727797e871101b679bd8b1595/kayenta/docs/canary-config.md#properties-7) but was not exposed in the modal when metric configuration is created/updated.

Screenshot 2025-05-16 at 9 53 55 PM

### DIFF
--- a/deck-kayenta/package.json
+++ b/deck-kayenta/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@spinnaker/kayenta",
-  "version": "2.5.0",
+  "version": "2.6.0",
   "license": "Apache-2.0",
   "repository": {
     "type": "git",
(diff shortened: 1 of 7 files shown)


jasonmcintosh:
I'm pretty good with this - though wondering if any tests? :P

zrashwani:
> I'm pretty good with this - though wondering if any tests? :P

Yes, it is already [there](https://github.com/spinnaker/spinnaker/pull/7078/files#diff-44635fa3004a9491a4bf6d4995d2c8d4b169024b5a961e308175e9ce30152e0dR109-R115), let me know if additional tests needed

jasonmcintosh:
Ahh thanks!  I'm not as good at UI stuff, but I think that'll cover it :) I just missed that in my review... thanks again!

jasonmcintosh (approved):

PR #7090, opened 2025-05-23
fix(build): actually publish debians

dbyron-sf:
See https://spinnakerteam.slack.com/archives/C0DPVDMQE/p1748013379193469 for background / evidence that the monorepo wasn't publishing debians before this.


jasonmcintosh (approved):
This looks like the spot ;)

dbyron-sf:
replaced by https://github.com/spinnaker/spinnaker/pull/7091 so we only publish debs on release, not on every dev/pr build.

PR #7109, opened 2025-06-16
feat(header): new module for authenticating based on the X-SPINNAKER-USER header

dbyron-sf:
enabled when `header.enabled` is true (defaults to false).

Note: this is very permissive.  Only enable it when there's authentication external to gate.


dbyron-sf:
@mergifyio update

PR #7115, opened 2025-06-20
perf(roles): make fiat full role sync synchronized across fiat instances

kirangodishala:
This PR adds one more improvement to full role sync operation.
Here are the details:
- Only the schedule role sync was done with distributive lock so far. Now,  /roles/sync api calls with empty roles list will also make role sync synchronized across all fiat instances.
- Other competing threads trying to perform full role sync would wait until the thread performing role sync is completed and  get the results (last sync count) from that instead of performing the operation again.
- With the global latch and local latch mechanism, the code makes sure only one thread within a fiat instance is allowed to go ahead with acquiring distributed lock which makes sure only one thread among all the fiat instances perform the full role sync.
- Redis client is added into Synchronizer.java to write the sync count and sync completion time every time there is a full role sync operation and those values for the last sync operation are fetched to verify if enough time is elapsed to attempt/skip a fresh sync operation.


fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/Synchronizer.java
```diff
@@ -55,24 +65,45 @@ public class Synchronizer {
...
   }
 
   public long syncAndReturn(List<String> roles) {
+    if (roles == null || roles.isEmpty()) {
+      return syncAllUserRoles();
+    }
+    return syncUserRoles(roles);
+  }
+
+  private long syncUserRoles(List<String> roles) {
```
dbyron0, line 107: javadoc here and elsewhere please.

dbyron-sf: Thank you!

christosarvanitis:
Thanks for tackling Fiat perf issues 🚀  
QQ: Do we have added benefit on top of the previous work done here https://github.com/spinnaker/fiat/pull/962

kirangodishala:
> Thanks for tackling Fiat perf issues 🚀 QQ: Do we have added benefit on top of the previous work done here [spinnaker/fiat#962](https://github.com/spinnaker/fiat/pull/962)

Previous PR added `CachedSynchronizationStrategy` which user can enable using `fiat.role.sync-cache.enabled=true`. ~~But this PR is an improvement to the `DefaultSynchronizationStrategy` when the all the user roles need to be synced.~~
This PR is an improvement to both  `DefaultSynchronizationStrategy` and `CachedSynchronizationStrategy` when the all the user roles need to be synced as distributed locking ensures only single instance of Fiat can do the full role sync at any time.

PR #7120, opened 2025-06-24
fix(web): add ProvidedIdRequestFilter

dbyron-sf:
enabled by setting provided-id-request-filter.enabled to true (default: false).

ProvidedIdRequestFilter is configured early enough in the filter chain so gate uses
X-SPINNAKER-REQUEST-ID and X-SPINNAKER-EXECUTION-ID headers from incoming http requests
when it makes downstream http requests (e.g. to fiat) during authentication.


dbyron-sf:
@mergifyio update

Jaimeoby (approved):

PR #7122, opened 2025-06-25
fix(build): Enable publishing for Gradle plugins

jcavanagh:
- Publishing tasks are now exposed via the `spinnaker-gradle-project` composite, so they can be properly addressed from the root build
- Gradle plugins now have a required `description` field, so that is now set in the metadata
- Automatically publishes on `main` pushes to its subtree, which will produce a version like `main-1`

### DIFF
--- a/spinnaker-gradle-project/build.gradle
+++ b/spinnaker-gradle-project/build.gradle
@@ -51,6 +51,6 @@ tasks.register('clean') {
 }
 
 tasks.register('publish') {
-  dependsOn ":spinnaker-project-plugin:publish"
-  dependsOn ":spinnaker-extensions:publish"
+  dependsOn ":spinnaker-project-plugin:publishPlugins"
+  dependsOn ":spinnaker-extensions:publishPlugins"
 }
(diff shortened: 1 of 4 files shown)


spinnaker-gradle-project/build.gradle
```diff
@@ -51,6 +51,11 @@ tasks.register('clean') {
 }
 
 tasks.register('publish') {
-  dependsOn ":spinnaker-project-plugin:publish"
-  dependsOn ":spinnaker-extensions:publish"
+  dependsOn ":spinnaker-project-plugin:publishPlugins"
+  dependsOn ":spinnaker-extensions:publishPlugins"
+}
+
+// Alias for publish
```
dbyron-sf: Why do we need this if the github action invokes the `publish` task?

jcavanagh: We don't - I had that in before deciding changing the workflow was cleaner.  Removed it.

dbyron-sf:

PR #7124, opened 2025-06-25
fix(front50/korkversion): remove older duplicate entry of korkVersion

kirangodishala:
Found duplicate entry for korkVersion in front50/gradle.properties. This PR fixes that.


dbyron-sf (approved):
I don't believe this is used in the monorepo, but still good to clean up.  Have you seen bad consequences from this?

PR #7088, opened 2025-05-22
feat(retrofit2): Add CallAdapter for legacy Retrofit signatures

jcavanagh:
This adds a CallAdapter to support non-Call<..> signatures on Retrofit interfaces, similar to retrofit1.

This eliminates the need for `Retrofit2SyncCall` at every Retrofit method call point and eliminates the need to wrap all interface return types in Call<...>

This does lose some potential async and streaming capability, but could offer a much more light-touch migration path for anything that might remain.

PR #7131, opened 2025-06-30
fix(retrofit2): fix a broken scenario related to LegacySignatureCallAdapter

kirangodishala:
- With the introduction of `LegacySignatureCallAdapter` in https://github.com/spinnaker/spinnaker/pull/7088, the logic to verify whether the request is for retrofit2 client is no longer valid because the retrofit2 enabled API interface need not have method return types with Call wrapper.

- This PR addresses the issue by introducing the RetrofitVersion enum field in all the ServiceClientProvider.getService() calls. 

### DIFF
--- a/kork/kork-retrofit2/src/test/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryTest.java
+++ b/kork/kork-retrofit2/src/test/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryTest.java
@@ -23,8 +23,6 @@ import static com.github.tomakehurst.wiremock.client.WireMock.getRequestedFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
 import static com.github.tomakehurst.wiremock.client.WireMock.verify;
-import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
-import static org.assertj.core.api.AssertionsForClassTypes.catchThrowable;
 import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
@@ -40,7 +38,6 @@ import com.netflix.spinnaker.config.okhttp3.DefaultOkHttpClientBuilderProvider;
 import com.netflix.spinnaker.config.okhttp3.OkHttpClientProvider;
 import com.netflix.spinnaker.kork.client.ServiceClientFactory;
 import com.netflix.spinnaker.kork.client.ServiceClientProvider;
-import com.netflix.spinnaker.kork.exceptions.SystemException;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerConversionException;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException;
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
@@ -220,14 +217,8 @@ public class Retrofit2ServiceFactoryTest {
     assertDoesNotThrow(
         () -> serviceClientProvider.getService(Retrofit2TestService.class, serviceEndpoint));
 
-    // FIXME: This should not throw exception
-    Throwable thrown =
-        catchThrowable(
-            () ->
-                serviceClientProvider.getService(Retrofit2CallTestService.class, serviceEndpoint));
-    assertThat(thrown).isInstanceOf(SystemException.class);
-    assertThat(thrown.getMessage())
-        .contains("No service client provider found for url (http://localhost:1234)");
+    assertDoesNotThrow(
+        () -> serviceClientProvider.getService(Retrofit2CallTestService.class, serviceEndpoint));
   }
 
   @Configuration
--- a/kork/kork-retrofit2/src/test/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryTest.java
+++ b/kork/kork-retrofit2/src/test/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryTest.java
@@ -23,6 +23,9 @@ import static com.github.tomakehurst.wiremock.client.WireMock.getRequestedFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.stubFor;
 import static com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo;
 import static com.github.tomakehurst.wiremock.client.WireMock.verify;
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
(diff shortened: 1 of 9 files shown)


kork/kork-retrofit2/src/test/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryTest.java
```diff
@@ -218,16 +231,18 @@ void testRetrofit2Client_withLegacySignatureCallAdapter() {
...
-    assertThat(thrown).isInstanceOf(SystemException.class);
-    assertThat(thrown.getMessage())
-        .contains("No service client provider found for url (http://localhost:1234)");
+        () ->
+            serviceClientProvider.getService(
+                Retrofit2TestService.class,
+                serviceEndpoint,
+                ServiceClientProvider.RetrofitVersion.RETROFIT2));
+
+    assertDoesNotThrow(
```
dbyron-sf, line 220: duplicate assertion?

kork/kork-web/src/main/java/com/netflix/spinnaker/kork/client/ServiceClientProvider.java
```diff
@@ -25,7 +25,11 @@
 
 @NonnullByDefault
 public interface ServiceClientProvider {
-
+  /** Which version of retrofit to use. Retrofit 1.x and 2.x are supported. */
```
dbyron-sf: Could you add some detail about how to decide which version to use?  It needs to match where the annotations come from in the interface class, right?

kork/kork-web/src/main/java/com/netflix/spinnaker/kork/client/ServiceClientProvider.java
```diff
@@ -25,7 +25,11 @@
...
+  enum RetrofitVersion {
+    RETROFIT1,
```
dbyron-sf, line 43: I don't see anything using RETROFIT1.  Can you add a test, or something so we can see how this works?

kork/kork-retrofit2/src/test/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryTest.java
kirangodishala: There are two interfaces one with Call and other without - Retrofit2TestService and Retrofit2CallTestService. So the test confirms both types.

kork/kork-web/src/main/java/com/netflix/spinnaker/kork/client/ServiceClientProvider.java
kirangodishala: Added

kork/kork-web/src/main/java/com/netflix/spinnaker/kork/client/ServiceClientProvider.java
kirangodishala: Here it is - [RetrofitServiceProviderTest.kt](https://github.com/spinnaker/spinnaker/pull/7131/files#diff-2e4624eda13fb9fa674ea7025abb5c7ceb4d3fee76dd12ca7c4546bc9ff8fd7e)

kork/kork-web/src/main/java/com/netflix/spinnaker/kork/client/ServiceClientProvider.java
```diff
@@ -34,7 +51,7 @@ public interface ServiceClientProvider {
    * @param <T> type of client , usually an interface with all the remote method definitions.
    * @return the retrofit interface implementation
    */
-  public <T> T getService(Class<T> type, ServiceEndpoint serviceEndpoint);
+  public <T> T getService(Class<T> type, ServiceEndpoint serviceEndpoint, RetrofitVersion version);
```
jcavanagh, line 56: We could significantly reduce the footprint of this change by having this interface have two signatures.  The implementation in `DefaultServiceClientProvider` can just have `RETROFIT2` be the default for the original signature.  Clients that have Retrofit1 annotations can use the second signature.

I think that a small footprint is desirable, since this change is just temporary, right?  We'd have to remove every `RETROFIT2` argument once Retrofit1 is fully removed.

jcavanagh (changes requested):

kork/kork-web/src/main/java/com/netflix/spinnaker/kork/client/ServiceClientProvider.java
kirangodishala: @jcavanagh - "defaulting to RETROFIT2" approach was considered but this PR is the beginning of a larger change of separating retrofit1 and retrofit2 beans. And instead of leaving doors open for mistakes it was decided to force the change for correctness even at the expense of larger footprint. And this approach will be fool-proof when handling retrofit3 and later versions in future.

dbyron-sf: Thoughts @jcavanagh ?  I wasn't planning to remove the enum argument once retrofit1 disappears, though I suppose we could.  I'd rather leave it in place (likely with a use of retrofit1 in test-only code) to pave the way for a future retrofit3.

dbyron-sf: With a default, it really does feel easy to screw up.

jcavanagh: Well, we don't have to guess, because Retrofit 3 already exists: https://github.com/square/retrofit/discussions/4379

It was a Kotlin-focused release, and allows things like coroutines to work naturally instead of Call wrappers there too.  

As far as can tell, it's drop-in for Java, and a RETROFIT3 enum value does not imply any new logic in the client factory.  So we'd still have the toil of updating all these call points with RETROFIT3, but there's actually nothing that would change in the internals.

I think that illustrates how it's an unnecessary encapsulation break - the point of this factory is so the caller does not have to be aware of the client internals.  The enum's existence is anticipating that Retrofit upgrades will break in specific ways, which we can already see is not the case - a classic case of YAGNI (You Ain't Gonna Need It).

As for the need to detect Retrofit1, we can also do things like reflect on the provided interface and see what the annotations are, or just keep the encapsulation break for the odd clients.

It's also the type of change that makes fork-maintainers like me a little sad, because changing commonly used APIs like this creates an hour or two of reintegration work.

dbyron-sf: Yeah, given that we're planning to backport this to the release-2025.0.x branch, I like the idea of reducing breakage for folks.  Custom code that uses a retrofit1 interface with the no-enum version of this method that defaults to retrofit2 is already broken, but at least uses of retrofit2 interfaces with it will stay working.

dbyron-sf (approved):

dbyron-sf:
@Mergifyio backport release-2025.0.x

PR #7137, opened 2025-07-05
fix(retrofit2): fix retrofit2 interceptors mixing up with retrofit1 client

kirangodishala:
Due to the complex configuration around retrofit clients, the retrofit2 interceptors are getting mixed up with retrofit1 client and this PR addresses the issue by making the following changes:

- Move interceptors and log level beans to respective retrofit modules from OkHttpClientComponents and RetrofitConfiguration
- Make RawOkHttpClientConfiguration public as it provides the basic OkHttpClient which can be customized based retrofit version.
- Remove redundant Retrofit2ConfigurationProperties from kork-web
- Group interceptors into three groups with qualifiers - retrofit-common, retrofit1 and retrofit2.
- Cleanup OkHttp3ClientConfiguration to make use of grouped interceptors
- Refactor halyard's DockerRegistryAccountValidator to replace constructed ServiceClientProvider with  autowired ServiceClientProvider
- Adjust various tests based on the above code changes


kork/kork-web/src/main/java/com/netflix/spinnaker/config/OkHttpClientComponents.java
```diff
@@ -72,6 +73,7 @@ public class OkHttpClientComponents {
   private final OkHttpMetricsInterceptorProperties metricsProperties;
 
   @Bean
+  @Qualifier("retrofit-common")
```
dbyron-sf, line 76: I thought we were going with "marker interfaces" so we get help at compile time?  These qualifiers seem too easy to miss / mess up, and then we might not learn that anything's wrong until startup time, or worse, when a particular API call fails.

dbyron-sf:
superceded by https://github.com/spinnaker/spinnaker/pull/7173

PR #7151, opened 2025-07-15
refactor(front50/retrofit2): upgrade the lone EchoService interface to retrofit2 in front50

kirangodishala:
- Front50 has a single retrofit1 interface which is broken due to default retrofit2 client behaviour.  This PR fixes the issue by upgrading to retrofit2.
- Since the default jackson convertor factory can't handle `Event` object implementations like `PluginEvent` when calling postEvent(), a `CustomConverterFactory` is created in `kork-retrofit`.
- Two similar Convertor Factories - EchoConvertorFactory and HelmConverterFactory in igor - are now replaced with CustomConverterFactory
- Added EchoConfiguration.EchoProperties class for better testability.


front50/front50-core/src/main/java/com/netflix/spinnaker/front50/echo/EchoService.java
```diff
@@ -17,17 +17,17 @@
...
-import retrofit.client.Response;
-import retrofit.http.Body;
-import retrofit.http.Headers;
-import retrofit.http.POST;
+import okhttp3.ResponseBody;
+import retrofit2.http.Body;
+import retrofit2.http.Headers;
+import retrofit2.http.POST;
 
 public interface EchoService {
```
dbyron-sf, line 25: I suppose because [EchoConfiguration](https://github.com/spinnaker/spinnaker/blob/df936003726c65eb9871878ea44bbee3b82170fc/front50/front50-core/src/main/java/com/netflix/spinnaker/front50/config/EchoConfiguration.java#L37) uses [ServiceClientProvider.getService](https://github.com/spinnaker/spinnaker/blob/df936003726c65eb9871878ea44bbee3b82170fc/kork/kork-web/src/main/java/com/netflix/spinnaker/config/DefaultServiceClientProvider.java#L59), this is enough to get things working.  I'd like to have a failing test first to make sure though.

Looks like this is only used with plugins, so potentially not so often.

dbyron-sf: I'd especially like a test since [RetrofitVersion enough isn't backported to release-2025.0.x yet](https://github.com/spinnaker/spinnaker/pull/7135), and it's not clear if the default-to-retrofit2 will remain.

dbyron-sf: If it wasn't clear, I'd like to backport this PR to release-2025.0.x.

kirangodishala: I added a test and few more unexpected changes. The Event object to be passed to EchoService.postEvent() is abstract and retrofit2's jackson converter factory is not able to handle RequestBody conversion. So I added a reusable custom converter to kork-retrofit module.

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/echo/EchoServiceTest.java
```diff
@@ -0,0 +1,148 @@
...
+      DefaultOkHttpClientBuilderProvider.class,
+      OkHttp3MetricsInterceptor.class,
+      RawOkHttpClientFactory.class,
+      RetrofitConfigurationProperties.class,
+      OkHttpClientConfigurationProperties.class,
+      OkHttpMetricsInterceptorProperties.class,
+      ObjectMapper.class
+    })
+@TestPropertySource(
+    properties = {"services.echo.enabled=true", "services.echo.base-url=https://localhost:56789"})
```
dbyron-sf: hard-coded ports give me the willies.  Can you switch to a dynamic port, likely with `@DynamicPropertySource` ?

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/config/EchoServiceTest.java
```diff
@@ -14,33 +14,26 @@
  * limitations under the License.
  */
 
-package com.netflix.spinnaker.front50.echo;
+package com.netflix.spinnaker.front50.config;
```
dbyron-sf: Why rename this file?

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/EchoConfig.groovy
```diff
@@ -18,7 +18,7 @@ package com.netflix.spinnaker.igor.config
 
 import com.netflix.spinnaker.config.OkHttp3ClientConfiguration
 import com.netflix.spinnaker.igor.IgorConfigurationProperties
-import com.netflix.spinnaker.igor.history.EchoConverterFactory
+import com.netflix.spinnaker.kork.retrofit.util.CustomConverterFactory
```
dbyron-sf, line 21: Could we have this refactor in a separate commit please?  That way it'll be easier to see that all the existing tests continue to pass, and to see how different the code is in kork from what it's replacing.

kork/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/util/CustomConverterFactory.java
```diff
@@ -27,25 +28,27 @@
...
+ * in the Echo API. The Event parameter is annotated with @Body and is an abstract class, which
+ * caused the default Retrofit2 Jackson converter Factory to fail.
  *
- * <p>This Factory is specifically designed to handle the conversion of Event objects for the Echo
- * API's postEvent method. The Event parameter is annotated with @Body and is an abstract class,
- * which caused the default Retrofit2 Jackson converter Factory to fail.
- *
- * <p>This Factory uses a custom ObjectMapper to serialize and deserialize Event objects, ensuring
- * compatibility with the Echo API.
+ * <p>This Factory uses a custom ObjectMapper to serialize and deserialize Event like objects.
```
dbyron-sf: Except when the type is a String, which I suppose isn't Event-like, but the javadoc needs to mention String somewhere.

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/echo/EchoServiceTest.java
kirangodishala: Actually this was the test before fix where the test fails before even the call is made. And I made necessary arrangements to use dynamic port post fix.

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/config/EchoServiceTest.java
kirangodishala: Instead of making echoService() method public, I moved the test file to the same package as EchoConfiguration

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/EchoConfig.groovy
kirangodishala: Agreed. Will do.

kork/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/util/CustomConverterFactory.java
kirangodishala: Will update.

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/echo/EchoServiceTest.java
dbyron-sf: Would you mind moving the dynamic port to the initial commit?  Besides removing a fixed port it'll also reduce the diffs for the fix, which makes it easier to see what's changing.

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/config/EchoServiceTest.java
dbyron-sf: Hmmm...but why call echoService() directly?  Why not let Spring construct the bean for you and autowire it?

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/echo/EchoServiceTest.java
kirangodishala: I made changes to EchoConfiguration for better testability. Dynamic port is not possible like the way it does post fix.

kirangodishala: I need to change the EchoConfiguration first and add test to demonstrate the issue. Shall I do that?

dbyron-sf: Maybe it's easier to chat?  I'd like to understand why a dynamic port isn't possible in the "before" test.

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/config/EchoServiceTest.java
kirangodishala: We should give the base url for the bean to get created and wiremock starts later.

dbyron-sf: We're in a bit of a time warp with the force-pushing, so the question of renaming the file isn't really relevant anymore.  Now the question is....can EchoServiceTest live in the com.netflix.spinnaker.front50.echo package since that's where EchoService lives?

front50/front50-core/front50-core.gradle
```diff
@@ -24,6 +24,7 @@ dependencies {
   implementation "org.springframework.security:spring-security-web"
   implementation "io.spinnaker.kork:kork-web"
   implementation "io.spinnaker.kork:kork-retrofit"
+  implementation "io.spinnaker.kork:kork-retrofit2"
```
dbyron-sf, line 27: Is this change really necessary for the new test?

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/config/EchoServiceTest.java
```diff
@@ -0,0 +1,155 @@
...
+    @Bean
+    OkHttpClient okhttp3Client(
+        OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
+        List<Interceptor> interceptors,
+        HttpTracing httpTracing) {
+      return new RawOkHttpClientFactory()
+          .create(okHttpClientConfigurationProperties, interceptors, httpTracing);
+    }
+
+    @Bean
```
dbyron-sf: It's unfortunate that we need this retrofit1 interceptor...I have a feeling we won't after https://github.com/spinnaker/spinnaker/pull/7137.  Can you add a comment here about removing it at some point?

front50/front50-core/front50-core.gradle
kirangodishala: Retrofit2ServiceFactoryAutoConfiguration is from kork-retrofit2 which is needed for retrofit2 client.

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/config/EchoServiceTest.java
kirangodishala: It holds all common and retrofit2 specific interceptors.  And after https://github.com/spinnaker/spinnaker/pull/7137, it holds only the common interceptors excluding retrofit2 specific interceptors. There is only one retrofit1 specific interceptor(SpinnakerRequestInterceptor) which doesn't extend okhttp3.Interceptor, so the above doesn't hold this.

front50/front50-core/front50-core.gradle
dbyron-sf: but since this commit only adds a test, wouldn't this be `testImplementation` ?

front50/front50-core/src/test/java/com/netflix/spinnaker/front50/config/EchoServiceTest.java
dbyron-sf: Sorry, are you saying that after https://github.com/spinnaker/spinnaker/pull/7137, the SpinnakerRequestInterceptor bean can disappear?

kirangodishala: It's my bad. I thought you were referring interceptors variable in okhttp3Client(). 
Yes, after https://github.com/spinnaker/spinnaker/pull/7137, we don't need SpinnakerRequestInterceptor. And I will add a comment now.

front50/front50-core/front50-core.gradle
kirangodishala: Even EchoConfiguration needs it. front50-core's runtime classpath doesn't have kork-retrofit2.

dbyron-sf: OK, as a nit I suppose it could start in testImplementation and then move to implementation in a subsequent commit.  If you have the stamina for that, great.  Otherwise let's go with it like this.

kirangodishala: Done. I hope nothing wrong for me to wish for the world to be perfect now :-)

dbyron-sf (approved):

dbyron-sf:

PR #7140, opened 2025-07-07
chore(gha): run clouddriver k8s integration tests when the workflow is modified

dbyron-sf:

PR #7156, opened 2025-07-18
chore(gha): run clouddriver integration tests when kork or fiat change

dbyron-sf:
The "regular" clouddriver tests run on kork or fiat change, so this does the same for the integration tests.

In addition to improving test coverage, this helps mergify since it wants an it-test test to have run before it'll merge.

While we're at it, run clouddriver integration tests when the github workflow changes so we get better feedback when editing workflows.  https://github.com/spinnaker/spinnaker/pull/7140 did the same for the k8s tests.


dbyron-sf:

PR #7159, opened 2025-07-20
fix(retrofit2): remove leading slashes from all the retrofit2 api interfaces

kirangodishala:
- This PR addressed the issue that would occur if the base url of any retrofit2 service has multiple slashes in the url which cause the final URL to omit all the text in the base url after the first slash.
- Added a test to demonstrate the issue
- Added ErrorHandlingExecutorCallAdapterFactory to retrofit2 clients wherever missing.


rosco/rosco-core/src/test/java/com/netflix/spinnaker/rosco/services/ClouddriverServiceTest.java
```diff
@@ -0,0 +1,130 @@
...
+    })
+public class ClouddriverServiceTest {
+  @RegisterExtension
+  static WireMockExtension wmClouddriver =
+      WireMockExtension.newInstance().options(wireMockConfig().dynamicPort()).build();
+
+  @Autowired ClouddriverService clouddriverService;
+
+  static String WM_CORRECT_ENDPOINT =
+      "/api/artifacts/fetch/"; // since the baseUrl has /api/ at the end.
```
dbyron-sf, line 77: I don't understand this variable name/comment.  Isn't the correct endpoint `/artifacts/fetch` from clouddriver's [ArtifactController](https://github.com/spinnaker/spinnaker/blob/a5c65ed59d9eab54f9abdb01ba6f2bbdc46d0df8/clouddriver/clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java#L76)?  What does /api have to do with it?

gate/gate-web/src/test/groovy/com/netflix/spinnaker/gate/FunctionalSpec.groovy
```diff
@@ -119,7 +119,7 @@ class FunctionalSpec extends Specification {
 
     def localPort = ctx.environment.getProperty("local.server.port")
     api = new Retrofit.Builder()
-        .baseUrl("http://localhost:${localPort}")
+        .baseUrl("http://localhost:${localPort}/")
```
dbyron-sf: Omit the trailing slash and use RetrofitUtils.getBaseUrl here?

rosco/rosco-core/src/test/java/com/netflix/spinnaker/rosco/services/ClouddriverServiceTest.java
kirangodishala: For wiremock, the end point starts from the first slash in the final URL so with the base url: `http://localhost:<port>/api/`  and endpoint: `/artifacts/fetch` the wiremock's endpoint becomes `/api/artifacts/fetch`.  But since retrofit2 removes /api from the final URL(`http://localhost:<port>/artifacts/fetch`) when leading slash exists on the endpoint, the wiremock's endpoint becomes `/artifacts/fetch` and this happened to be same as couddriver's endpoint on the interface method but in the given context(base url having multiple slashes) it's incorrect.

gate/gate-web/src/test/groovy/com/netflix/spinnaker/gate/FunctionalSpec.groovy
kirangodishala: Done

rosco/rosco-core/src/test/java/com/netflix/spinnaker/rosco/services/ClouddriverServiceTest.java
dbyron-sf: OK, we talked and I think I understand this better.  `/api` is relevant because this test explicitly adds it.  And in the real world it would take some spring boot config property on the clouddriver side to match, or perhaps running some reverse proxy, etc. in front of clouddriver.  And the point of the test is to exercise this case, where the base URL has a path component after the hostname.

dbyron-sf:

PR #7163, opened 2025-07-22
fix(deck): Fix RunJob external logs when component is not mounted and interpolation is needed

christosarvanitis:
Deck is failing to render the external logs link when interpolation is needed but the async operation to Clouddriver to mount the manifest. This PR fixes this issue but keeps the ExternalLog link when interpolation is not needed


radu-dragon (approved):

deck/packages/core/src/manifest/stage/JobStageExecutionLogs.tsx
```diff
@@ -52,7 +52,7 @@ export class JobStageExecutionLogs extends React.Component<IJobStageExecutionLog
     const { manifest } = this.state;
     const { externalLink, podNamesProviders, location, account } = this.props;
     // prefer links to external logging platforms
-    if (externalLink) {
+    if (externalLink && (!externalLink.includes('{{') || !isEmpty(manifest))) {
```
jasonmcintosh, line 55: Think we need a rewrite or rethink on this... too complicated of logic.  I could see passing all of this down a bit and being a bit smarter on the checks internally.

jasonmcintosh:
<ignore>ALSO given we're doing this in single repo deck for older releases, we shouldn't need this as we'll sync from there.</ignore> we're not syncing from the single deck repo.  SO really just... see if we can't make this logic a touch better.  It'd be nice if we could test this too...

jasonmcintosh (approved):
Approved - have tests now!  I'd love to simplify this but that's separate issue.

christosarvanitis:
@jasonmcintosh possible backport on 2025.1.x?

PR #7164, opened 2025-07-22
feat(core): add rootId to PipelineExecution

dbyron-sf:
so it's cheap to find the root id of a pipeline execution. Without it, it's possible to look through triggers and find parents until the root, but this is easier.

It's implemented so that if a pipeline's root id is the same as its id, it's a top-level pipeline. Otherwise (root id != id), it's a child.


jasonmcintosh (approved):
ONE call out... I don't know that this will be backwards compatible if it adds this on new instances, since the field won't exist on prior versions for serialization purposes.  PROBABLY worth a call out in the release notes that any pipeline created with this can't be read by prior versions after without removing this field.

dbyron-sf:
> ONE call out... I don't know that this will be backwards compatible if it adds this on new instances, since the field won't exist on prior versions for serialization purposes. PROBABLY worth a call out in the release notes that any pipeline created with this can't be read by prior versions after without removing this field.

I don't think there's an issue with Spinnaker itself...the only code that reads the root id does it for freshly constructed executions.  But yes, code that's processing completed executions can't assume it's present all the time.

jasonmcintosh:
https://github.com/spinnaker/spinnaker/blob/313fa6d4ea5a3dee52495e9f43f9999cc0caf5a1/orca/orca-pipelinetemplate/src/main/java/com/netflix/spinnaker/orca/config/PipelineTemplateConfiguration.java#L58 we're good - we disable failing on unknown properties so ;)

dbyron-sf:
release note: https://github.com/spinnaker/spinnaker.io/pull/524

Issue #7022, opened 2025-04-01
Echo: Gchat notifications not working after spinnaker 1.37 update


rrondeau:
Hi,

We upgrade our spinnaker cluster and gchat notifications stopped working.

echo logs from one notification :

```
2025-03-27 17:20:58.810  WARN 1 --- [oScheduler-1315] c.n.s.okhttp.OkHttp3MetricsInterceptor   : Request POST:https://chat.googleapis.com/v1/spaces/xxxxx/messages%3Fkey=xxxx&token=xxxxx is missing [X-SPINNAKER-USER, X-SPINNAKER-ACCOUNTS] authentication headers and will be treated as anonymous.

com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException: Status: 404, Method: POST, URL: https://chat.googleapis.com/v1/spaces/xxxxx/messages%3Fkey=xxxxx&token=xxxxx, Message:
```

It looks like the `?` is now encoded by spinnaker which result into a 404 error from chat.googleapis.com.
I search a little bit and it seems to be the retrofit upgrade that cause the issue : https://github.com/spinnaker/echo/commit/6a0e7847fa6b651909aa83b15add0d254da743c7#diff-a08887a49a23122e876957918cc6866f40713c21eb2ec82b70bbee376c8eed39

Thanks !

jasonmcintosh:
SOoo there's a number of retrofit fixes incoming.  You might try asking in slack - and there were a few recent ones to echo in particular.  At a guess a lot of the user base isn't using GCHAT so testing this may require work.  Would welcome tests & PRs on this!

mbergo:
Is this still an issue @jasonmcintosh?

mbergo:
@rrondeau https://github.com/spinnaker/echo/pull/1505 cc: @jasonmcintosh

dbyron-sf:
Can you try [1.37.10](https://spinnaker.io/changelogs/1.37.10-changelog/) please?

rrondeau:
Hi,

I will try this week, maybe tomorow, its a public holiday today ;)

Thanks !

dbyron-sf:
@kirangodishala can you take a look when you're back please?

dbyron-sf:
@rrondeau Can you try [1.37.11](https://spinnaker.io/changelogs/1.37.11-changelog/) please?

rrondeau:
Issue still there !
```
[...]

PR #7165, opened 2025-07-22
fix(googlechat): fix googlechat encoding issue

kirangodishala:
- This PR addressed the issue https://github.com/spinnaker/spinnaker/issues/7022

- The issue occurred because the partial webhook url which contains query parameters(after two terms separated by slashes) was being sent as path variable due to which incorrect encoding was happening. This used to work with retrofit1 but retrofit2 is strict in this regard.

- From the [google chat official documentation](https://developers.google.com/workspace/chat/quickstart/webhooks?utm_source=chatgpt.com#python_1), found the webhook url is of the form - https://chat.googleapis.com/v1/spaces/SPACE_ID/messages?key=KEY&token=TOKEN so based on this, fix is provided by separated the path and query parameters.


dbyron-sf (approved):

dbyron-sf:

PR #7168, opened 2025-07-24
feat(orca/webhook): add account property to webhook stages and introduce WebhookAccountProcessor interface

dbyron-sf:
There are a series of changes here:
* feat(clouddriver/web): add an endpoint to fetch account details
* feat(orca/webhook): add new account property to webhook stages
* feat(orca/webhook): add webhook.validateAccount
* feat(orca/webhook): add WebhookAccountProcessor interface

with these orca config flags to control behavior:
* `webhook.requireAccount` (default: false). If true and the account property is not present (i.e. a non-empty string),
the stage fails.
* `webhook.validateAcount` (default: false). If true, and an account property is present, call clouddriver's GET /credentials/{account}/authorized endpoint to validate that the current user has permission to use the account.

The bigger picture here is that this is a new mechanism to help lock down webhook stages.  The addition of the WebhookAccountProcessor interface could go in a separate PR, but it's so close to the rest of what's going on here that I included it.

Note that this code doesn't actually a WebhookAccountProcessor bean, so that part of the change is a no-op without custom code / a plugin that adds one.

### DIFF
--- a/clouddriver/clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.java
+++ b/clouddriver/clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.java
@@ -34,6 +34,7 @@ import java.util.Objects;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import javax.annotation.CheckForNull;
+import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PathVariable;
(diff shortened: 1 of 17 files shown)


clouddriver/clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/CredentialsController.java
```diff
@@ -82,6 +83,18 @@ public Map<String, Object> getAccountCredentialsDetails(@PathVariable String acc
...
+  /**
+   * Return account details only when the caller has write permissions for the account
+   *
+   * @param accountName the account name for which to retrieve details
+   * @return a map of account details
+   */
+  @GetMapping("/{accountName}/authorized")
+  @PreAuthorize("hasPermission(#accountName, 'ACCOUNT', 'WRITE')")
+  public Map<String, Object> getAccountCredentialsWithAuth(@PathVariable String accountName) {
+    return getAccountCredentialsDetails(accountName);
```
jasonmcintosh, line 95: Note this isn't adding a new endpoint, this is just adding AUTH around an existing endpoint.  There's a question in my mind - why not JUST add auth to the above endpoint?  This would break systems querying that endpoint internally - just not seen who uses that, or to what impact, but users SHOULD be authorized.  Just would need to check internal usage.
`@GetMapping("/{accountName}")`
Is the intent here to call this to FORCE a permissions check & allow other processes to use the other endpoint instead?  Just a ... maybe broader question around "internal only" endpoints and requiring auth (thinking Igor issue right now where an endpoint got auth that didn't and it broke things)

dbyron-sf: Yeah, I didn't want to risk breaking uses of the no-auth endpoint.  The intent is for the webhook code to do a permissions check, and yeah, to leave other code as it was.

Looks like gate has some calls to the unauthorized endpoint.  Would love to leave that auditing/improvement to a separate effort.

PR #7172, opened 2025-07-29
feat(webhook): add new webhook allow list property safeToRetry

dbyron-sf:
which defaults to false.

Before this, CreateWebhookTask would retry SocketTimeoutException only for GET requests. Now, it also retries other request methods (e.g. PUT, POST, etc.) if safeToRetry is true in the relevant allow list entry.


orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookProperties.java
```diff
@@ -159,6 +159,12 @@ public static class AllowedRequest {
 
     /** The url must match this pattern to be considered valid */
     private String urlPattern;
+
+    /**
+     * Whether it's safe to retry requests to urls that match. GET requests are always considered
+     * safe, but this enables retries of other methods as well.
+     */
+    private boolean safeToRetry;
```
mattgogerly, line 167: Is this config per webhook, or for all webhooks? Seems like it would be valuable to be able to configure it for specific patterns?

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/service/RestTemplateData.java
```diff
@@ -0,0 +1,58 @@
...
+    this.restTemplate = restTemplate;
+    this.validatedUri = validatedUri;
+    this.httpMethod = httpMethod;
+    this.payloadEntity = payloadEntity;
+    this.stageData = stageData;
+    this.allowedRequest = allowedRequest;
+  }
+
+  public ResponseEntity<Object> exchange() {
+    return this.restTemplate.exchange(validatedUri, httpMethod, payloadEntity, Object.class);
```
mattgogerly, line 56: Not sure if we even have access to it yet (probably not..), but preference would be `RestClient` or `WebClient` over `RestTemplate`.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/service/WebhookService.java
```diff
@@ -231,17 +241,22 @@ private RestTemplateData getRestTemplateData(
...
-          String message =
-              String.format(
-                  "http method '%s', uri: '%s' not allowed",
-                  httpMethod.toString(), validatedUri.toString());
-          log.info(message);
-          throw new IllegalArgumentException(message);
+        Optional<WebhookProperties.AllowedRequest> allowedRequest =
+            getAllowedRequest(httpMethod, validatedUri);
+        if (webhookProperties.isAllowedRequestsEnabled()) {
+          if (allowedRequest.isEmpty()) {
```
mattgogerly, line 247: nit: any reason for pulling this logic out of the `isAllowedRequest` method? it reads a bit clunky

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/tasks/WebhookResponseProcessor.java
```diff
@@ -137,19 +142,19 @@ private TaskResult processClientException(Exception e) {
               "Remote host resolution failure in webhook for pipeline %s to %s, will retry.",
               executionId, stageData.url);
       executionStatus = ExecutionStatus.RUNNING;
-    } else if (stageData.method == HttpMethod.GET
-        && (e instanceof SocketTimeoutException
-            || e.getCause() instanceof SocketTimeoutException)) {
+    } else if ((e instanceof SocketTimeoutException
+            || e.getCause() instanceof SocketTimeoutException)
+        && ((stageData.method == HttpMethod.GET) || safeToRetry(restTemplateData))) {
```
mattgogerly, line 147: nit: could move `stageData.method == HttpMethod.GET` into `safeToRetry` as a short circuit to simplify this if statement

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookProperties.java
dbyron-sf: It's per allow list entry.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/service/RestTemplateData.java
dbyron-sf: This is the same code as before this PR, only moved from WebhookService.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/service/WebhookService.java
dbyron-sf: Yeah, this is clunky.  We need to distinguish between "not allowed" (i.e. not corresponding allow list entry) and "the allowed request feature is disabled".  I could remove the isAllowedRequestsEnabled check from getAllowedRequest but that wouldn't really simplify this code.  Combining the two ifs into one with `&&` would help.  Is that worth it?  Open to suggestions here.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/tasks/WebhookResponseProcessor.java
dbyron-sf: Yeah, I'd tempted to leave it like this.  I dunno, six of one, half a dozen of the other.

mattgogerly: Fine by me

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/service/WebhookService.java
mattgogerly: Really splitting hairs I guess, but could we move line 244 to line 247, above the second if? I don't particularly mind; if you're happy with it as is that's fine.

dbyron-sf: Because allowedRequest is used on line 259, I don't think we can do that.  I say let's ship it as it is.

mattgogerly (approved):

PR #7173, opened 2025-07-29
fix(retrofit1): remove the retrofit1 code

kirangodishala:
- This PR fully deletes Retrofit1 remnants, as all Spinnaker services are upgraded to retrofit2
-  retrofit1 client creation code including all the supporting beans and classes from kork is removed
-  Code referring to RetrofitError in SpinnakerServerException and it's subclasses is removed
- All retrofit1-specific logic has been stripped from exception handlers in Orca
- finally removed all the retrofit1 related dependencies


orca/orca-web/config/orca.yml
```diff
@@ -59,7 +59,7 @@ resilience4j.retry:
     # katoRetrofitServiceWriter is any write operation to Clouddriver using Retrofit.
     katoRetrofitServiceWriter:
       retryExceptions:
-        - retrofit.RetrofitError
+        - com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException
```
dbyron-sf: I think we'd be better off with this in a separate PR since it's not really related to removing retrofit1.

orca/orca-queue/src/test/java/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerExceptionHandlerTest.java
```diff
@@ -308,16 +250,21 @@ void taskThrowsSpinnakerHttpExceptionNoRetry(SpinnakerHttpException spinnakerHtt
   }
 
   private static Stream<SpinnakerHttpException> nonRetryableSpinnakerHttpExceptions() {
-    return nonRetryableRetrofitErrors()
-        .map(retrofitError -> makeSpinnakerHttpException(retrofitError));
+    // irrespective of the exception details, http code 500 is non-retryable
```
dbyron-sf, line 253: It seems a little strange that orca doesn't retry GETs/idempotent requests on 500, but it does look like that's the way it's implemented.

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -144,10 +143,7 @@ dependencies {
     api("com.squareup.retrofit2:converter-jackson:${versions.retrofit2}")
     api("com.squareup.retrofit2:retrofit-mock:${versions.retrofit2}")
     api("com.squareup.retrofit2:retrofit:${versions.retrofit2}")
-    api("com.squareup.retrofit:converter-jackson:${versions.retrofit}")
```
dbyron-sf, line 147: the definition of versions.retrofit above also needs to disappear.

echo/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GoogleChatNotificationAgent.groovy
```diff
@@ -20,10 +20,10 @@ import com.netflix.spinnaker.echo.googlechat.GoogleChatMessage
 import com.netflix.spinnaker.echo.googlechat.GoogleChatService
 import com.netflix.spinnaker.echo.api.events.Event
 import groovy.util.logging.Slf4j
+import okhttp3.ResponseBody
```
dbyron-sf: This commit could go in a separate PR, yes?  Like, it's an actual bug?  And one we could backport to release-2025.0.x?

rosco/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
```diff
@@ -40,7 +39,7 @@ class BakeRequest {
   @Schema(description= "A generated UUID which will be used to identify the effective packer bake", accessMode = Schema.AccessMode.READ_ONLY)
   String request_id = UUID.randomUUID().toString()
   String user
-  @Schema(description="The package(s) to install, as a space-delimited string") @JsonProperty("package") @SerializedName("package")
+  @Schema(description="The package(s) to install, as a space-delimited string") @JsonProperty("package")
```
dbyron-sf: How is this related to retrofit1?

kirangodishala: We got Gson from retrofit1 libraries. So when retrofit1 dependencies are removed, compile errors showed up.

echo/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GoogleChatNotificationAgent.groovy
kirangodishala: Agree. I will raise a separate PR for it and remove it from here.

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
kirangodishala: will do

orca/orca-web/config/orca.yml
kirangodishala: Ok, will raise a separate PR.

rosco/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/api/BakeRequest.groovy
dbyron-sf: But...we could add an explicit dependency on gson for this, right?  And, isn't removing `@SerializedName` a behavior change?  Or, if we're not actually de/serializing this with gson, then perhaps this annotation is effectively ignored.  Would like to understand more.

kirangodishala: We are not using gson anywhere so removed it. But as you mentioned, I will add dependency in case any plugin code is using.

dbyron-sf: OK, so this `@SerializedName` annotation was being ignored.  Seems good to remove it.  Thanks.

dbyron-sf: I don't think I'd add the gson dependency...If plugins depend on gson I think it's fine to expect them to have an explicit dependency.

kirangodishala: I added gson dependency so that any plugins still using gson for conversions would not break.

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
kirangodishala: Done

orca/orca-queue/src/test/java/com/netflix/spinnaker/orca/q/handler/RunTaskHandlerExceptionHandlerTest.java
kirangodishala: Yeah, refactored the test in compliance with the implementation.

echo/echo-notifications/src/main/groovy/com/netflix/spinnaker/echo/notification/GoogleChatNotificationAgent.groovy
dbyron-sf: https://github.com/spinnaker/spinnaker/pull/7175

kork/kork-web/src/main/groovy/com/netflix/spinnaker/retrofit/RetrofitConfigurationProperties.java
```diff
@@ -1,33 +0,0 @@
...
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-package com.netflix.spinnaker.retrofit;
-
-import org.springframework.boot.context.properties.ConfigurationProperties;
-import retrofit.RestAdapter;
-
-@ConfigurationProperties("retrofit")
```
dbyron-sf, line 22: There are some config files that still reference `retrofit:` properties:
* [halyard.yaml](https://github.com/spinnaker/spinnaker/blob/3574bbeb8daae54340c182b4da8bcce0c87cd32a/halyard/halyard-web/config/halyard.yml#L36-L37)
* [kayenta.yml](https://github.com/spinnaker/spinnaker/blob/3574bbeb8daae54340c182b4da8bcce0c87cd32a/kayenta/kayenta-graphite/src/integration-test/resources/config/kayenta.yml#L4-L5) (commented out, but it's a good time to remove it)
* [halyard design](https://github.com/spinnaker/spinnaker/blob/3574bbeb8daae54340c182b4da8bcce0c87cd32a/halyard/docs/design.md?plain=1)
* [gate-test-app.yml](https://github.com/spinnaker/spinnaker/blob/3574bbeb8daae54340c182b4da8bcce0c87cd32a/gate/gate-api-tck/src/main/resources/gate-test-app.yml#L38-L39)
* [application-echo.properties](https://github.com/spinnaker/spinnaker/blob/3574bbeb8daae54340c182b4da8bcce0c87cd32a/gate/gate-web/src/test/resources/application-echo.properties#L39)
* [SqlConfigurationTest.java](https://github.com/spinnaker/spinnaker/blob/3574bbeb8daae54340c182b4da8bcce0c87cd32a/orca/orca-sql/src/test/java/com/netflix/spinnaker/orca/testconfig/SqlConfigurationTest.java#L65)

kork/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
```diff
@@ -83,36 +83,14 @@ class OkHttp3ClientConfiguration {
 
   public OkHttp3ClientConfiguration(OkHttpClientConfigurationProperties okHttpClientConfigurationProperties,
                                     OkHttp3MetricsInterceptor okHttp3MetricsInterceptor) {
-    this(okHttpClientConfigurationProperties, okHttp3MetricsInterceptor, null, null, null,
+    this(okHttpClientConfigurationProperties, okHttp3MetricsInterceptor, HttpLoggingInterceptor.Level.BASIC, null, null,
```
dbyron-sf: Not sure what's happening in the code before this PR, but wouldn't we be better off adding an HttpLoggingInterceptor.Level constructor argument instead of hard-coding this?

orca/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
```diff
@@ -37,10 +34,6 @@ abstract class BaseRetrofitExceptionHandler implements ExceptionHandler {
       return true
     }
 
-    if(httpMethod == null) {
```
dbyron-sf, line 40: I think `boolean shouldRetry(Exception e, String kind, Integer responseCode)` needs to disappear.  I don't think it's used, but it no longer makes sense to have httpMethod null here.

dbyron-sf:
Trying to decide if there's anything to do with this from [Retrofit2ServiceFactoryAutoConfiguration](https://github.com/spinnaker/spinnaker/blob/3574bbeb8daae54340c182b4da8bcce0c87cd32a/kork/kork-retrofit2/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryAutoConfiguration.java#L32):
```
@ConditionalOnProperty(value = "retrofit2.enabled", havingValue = "true", matchIfMissing = true)
```
If someone does set retrofit2.enabled to false, will their builds break, or code fail to start?
Is it more pleasing to remove this altogether, or leave it for when we migrate to retrofit3/4/5/etc. ?

I'm pretty tempted to remove it and worry about that migration when it happens.

kork/kork-web/src/main/groovy/com/netflix/spinnaker/config/OkHttp3ClientConfiguration.groovy
kirangodishala: Adding constructor argument means changing the signature of the two constructors and the corresponding callers. That will be a bigger change. Do you want me to proceed with that? Even if that is the case, I feel it can be done in a different PR in future.

dbyron-sf: What log level did this code path use before this PR?

kirangodishala: It must be retrofit1's usage which didn't need retrofit2's log level hence was null.

dbyron-sf: If only retrofit1 used the constructor, can we remove it altogether?

kirangodishala: The code previously used the constructor for retrofit1 client is now getting retrofit2 client and hence having null for log level fails while creating the new client.

kirangodishala: I will try to add the log level constructor parameter.

kork/kork-web/src/main/groovy/com/netflix/spinnaker/retrofit/RetrofitConfigurationProperties.java
kirangodishala: Made all the changes

dbyron-sf: https://github.com/spinnaker/spinnaker/pull/7173/commits/55bf9d22f070dc0a5904f76697fde23d067895f1

orca/orca-retrofit/src/main/groovy/com/netflix/spinnaker/orca/retrofit/exceptions/BaseRetrofitExceptionHandler.groovy
dbyron-sf: removed in https://github.com/spinnaker/spinnaker/pull/7173/commits/b7ad47544c3603d2a2d5d02fa07ec6a282d03aa4

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
```diff
@@ -60,14 +65,36 @@ public class Front50PluginsConfiguration {
...
+  public static PluginOkHttpClientProvider pluginsOkHttpClient(
+      Environment environment,
+      Provider<Registry> registry,
+      Retrofit2EncodeCorrectionInterceptor correctionInterceptor) {
     OkHttpClientConfigurationProperties okHttpClientProperties =
         Binder.get(environment)
             .bind("ok-http-client", Bindable.of(OkHttpClientConfigurationProperties.class))
             .orElse(new OkHttpClientConfigurationProperties());
 
+    OkHttpMetricsInterceptorProperties okHttpMetricsInterceptorProperties =
```
dbyron-sf, line 77: I could use an education here.  Wouldn't we get the same behavior much more simply by adding an OkHttpMetricsInterceptorProperties argument to this method?

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
```diff
@@ -60,14 +65,36 @@ public class Front50PluginsConfiguration {
...
+        Binder.get(environment)
+            .bind(
+                "ok-http-client.interceptor", Bindable.of(OkHttpMetricsInterceptorProperties.class))
+            .orElse(new OkHttpMetricsInterceptorProperties());
+
+    OkHttp3MetricsInterceptor okHttp3MetricsInterceptor =
+        new OkHttp3MetricsInterceptor(registry, okHttpMetricsInterceptorProperties);
+
+    okHttpClientProperties.getRefreshableKeys().setEnabled(false);
```
dbyron-sf, line 86: Why set this to false here?

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/HttpClientSdkConfiguration.java
```diff
@@ -62,13 +63,23 @@ public static SdkFactory httpClientSdkFactory(
                 "ok-http-client.interceptor", Bindable.of(OkHttpMetricsInterceptorProperties.class))
             .orElse(new OkHttpMetricsInterceptorProperties());
 
+    Retrofit2ConfigurationProperties retrofit2ConfigurationProperties =
```
dbyron-sf, line 66: Same question here.

kork/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/internal/DefaultOkHttp3ClientFactory.kt
```diff
@@ -69,6 +72,8 @@ class DefaultOkHttp3ClientFactory(
       config.connection.readTimeout?.let {
         readTimeoutMs = it.toMillis()
       }
+      refreshableKeys.enabled = false
```
dbyron-sf, line 75: And here...why is this important?

kork/kork-retrofit2/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryAutoConfiguration.java
```diff
@@ -19,32 +19,16 @@
...
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.core.Ordered;
-import org.springframework.core.annotation.Order;
 
 @Configuration
-@ConditionalOnProperty(value = "retrofit2.enabled", havingValue = "true", matchIfMissing = true)
 @EnableConfigurationProperties(Retrofit2ConfigurationProperties.class)
```
dbyron-sf: Am I right that this annotation can disappear too, as well as the entire Retrofit2ConfigurationProperties class from kork-retrofit2?

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
kirangodishala: OkHttpClient.Builder bean in OkHttpClientComponents is equipped with handling refreshable identity which is not available to the above configuration. But there is a chance that a new OkHttpClient.Builder from the OkHttp3ClientConfiguration may get returned if "ok-http-client" configuration has refreshable keys enabled. So while creating OkHttp3ClientConfiguration object, explicitly disabling ensures that the interceptors are added to the  OkHttpClient

kork/kork-retrofit2/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryAutoConfiguration.java
kirangodishala: yes, it's a miss. I saved it for separate commit but apparently missed.

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
dbyron-sf: Why isn't handling refreshable identity available here?

kork/kork-retrofit2/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryAutoConfiguration.java
dbyron-sf: No worries...deleting code is so exciting :)

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
kirangodishala: In static @Bean methods, Spring cannot perform autowiring of configuration properties classes

dbyron-sf: Aaah, it's because the `@Bean` method here is static.  Not sure why that is, but so be it.

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
kirangodishala: It was never there. But didn't matter with retrofit1. But now we need Retrofit2EncodeCorrectionInterceptor with retrofit2, we need to make sure the first `if` block in `OkHttp3ClientConfiguration.createForRetrofit2()` shouldn't be executed.

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/HttpClientSdkConfiguration.java
kirangodishala: With retrofit2 client, Retrofit2ConfigurationProperties is needed to avoid any encoding issues

kork/kork-plugins/src/main/kotlin/com/netflix/spinnaker/kork/plugins/sdk/httpclient/internal/DefaultOkHttp3ClientFactory.kt
kirangodishala: As mentioned above, to not return a OkHttpClient.Builder that doesn't have retrofit2 interceptors

kork/kork-retrofit2/src/main/java/com/netflix/spinnaker/kork/retrofit/Retrofit2ServiceFactoryAutoConfiguration.java
dbyron-sf: https://github.com/spinnaker/spinnaker/pull/7173/commits/483aa691b89d44fd3f91630893520ee22f8a4dfe

kork/kork-plugins/src/main/java/com/netflix/spinnaker/config/Front50PluginsConfiguration.java
dbyron-sf: Aaah, ok.  Seems a little wacky that OkHttp3ClientConfiguration.createForRetrofit2() works way, but this code does look correct.

PR #7179, opened 2025-07-31
feat(orca): plumb additional headers from incoming requests into webhook stages

dbyron-sf:
These changes:

* enable ProvidedIdRequestFilter in orca, getting incoming additional headers (e.g. X-FOO and X-BAR into the MDC, configured with (for example):
```
    provided-id-request-filter:
      enabled: true
      additionalHeaders:
        - X-FOO
        - X-BAR
```
* add an additionalHeaders property to the pipeline execution context, populated with additional headers from ProvidedIdRequestFilter (if provided-id-request-filter.enabled is true)
* add additional headers from the pipeline execution context into the MDC of running tasks
* include additional headers from the MDC in outgoing http requests that webhook stages make if new config property `webhook.includeAdditionalHeaders` is true (it defaults to false).

Together with mTLS (external to gate) and [header auth](https://github.com/spinnaker/spinnaker/pull/7109) + ProvidedIdRequestFilter in gate, this is a mechanism to get trusted headers from outside Spinnaker into the http requests that webhook stages make, which provides extra context to the receivers of those requests.

release notes: https://github.com/spinnaker/spinnaker.io/pull/533

### DIFF
--- a/kork/kork-web/src/main/java/com/netflix/spinnaker/okhttp/SpinnakerRequestHeaderInterceptor.java
+++ b/kork/kork-web/src/main/java/com/netflix/spinnaker/okhttp/SpinnakerRequestHeaderInterceptor.java
@@ -18,31 +18,52 @@ package com.netflix.spinnaker.okhttp;
 
 import com.netflix.spinnaker.kork.common.Header;
 import com.netflix.spinnaker.security.AuthenticatedRequest;
+import io.micrometer.core.instrument.util.StringUtils;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
 import okhttp3.Interceptor;
 import okhttp3.Request;
 import okhttp3.Response;
+import org.slf4j.MDC;
 
 /**
  * As {@link retrofit.RequestInterceptor} no longer present in retrofit2, we have to use {@link
  * okhttp3.Interceptor} to add authenticated headers to requests.
  */
+@Slf4j
 public class SpinnakerRequestHeaderInterceptor implements Interceptor {
 
+  /**
+   * Whether to propagate headers to outgoing HTTP requests. If false, no headers are propagated.
+   * Not X-SPINNAKER-*, not additionalHeaders.
+   */
   private final boolean propagateSpinnakerHeaders;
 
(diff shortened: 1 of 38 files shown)


kork/kork-web/src/main/java/com/netflix/spinnaker/okhttp/SpinnakerRequestHeaderInterceptor.java
```diff
@@ -18,31 +18,52 @@
...
 import okhttp3.Interceptor;
 import okhttp3.Request;
 import okhttp3.Response;
+import org.slf4j.MDC;
 
 /**
  * As {@link retrofit.RequestInterceptor} no longer present in retrofit2, we have to use {@link
  * okhttp3.Interceptor} to add authenticated headers to requests.
  */
+@Slf4j
```
jasonmcintosh, line 36: https://github.com/spinnaker/spinnaker/blob/8b5c743e650a8d706bb7079147cfda75305cdd93/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/filters/CorsFilter.java#L32 should probably use log4j2

orca/orca-queue-tck/src/main/kotlin/com/netflix/spinnaker/orca/q/LogMessageTask.kt
```diff
@@ -0,0 +1,32 @@
...
+import com.netflix.spinnaker.orca.api.pipeline.TaskResult
+import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution
+import org.slf4j.LoggerFactory
+import org.slf4j.MDC
+
+class LogMessageTask : Task {
+  private val log = LoggerFactory.getLogger(javaClass)
+
+  override fun execute(stage: StageExecution): TaskResult {
+    log.info("MDC: ${MDC.getCopyOfContextMap()}")
```
jasonmcintosh, line 29: I'maI may have missed this... but... should this REALLY be a log.info?  MDC can have a LOT of data potentially in some systems.

dbyron-sf: This is test code.

jasonmcintosh (approved):

PR #7188, opened 2025-08-05
fix(clouddriver/aws): fix corner cases in AmazonNamedImageLookupController

dbyron-sf:
Share in the journey where we wade into some groovy code and tests that we really wish were written in java.  After a couple more PRs, perhaps things will have finally stabilized enough so that makes sense.

In the meantime, here are some fixes for some yucky "find ami" corner (or perhaps not so corner) cases, and potentially some others too.
* find by - name and tags - two amis - different accounts - same region - same name - same id - only one has tags
* find by - name and tags - two amis - same account - same region - same name - different ids - only one has tags

Here are some more details:

**Previous Behavior**
Query by name and tags was performed but account was not specified could result in not finding an image when one existed:
* Was nondeterministic and could return an image sometimes, but not other times.
* Depended on the order in which images were returned from the cache.
* Would overwrite already set tags for an image id if multiple images for the id existed with different tag sets, including clearing already set tags.

**New Behavior**
* If no tag filter is specified, tags on images are ignored
* If a tag filter is specified, only images for which all tags match are considered
* If no region is specified, images in all regions are considered
* if a region is specified, only images in the specified region are considered


clouddriver/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/controllers/AmazonNamedImageLookupController.groovy
```diff
@@ -64,77 +68,223 @@ class AmazonNamedImageLookupController {
...
 
     Collection<CacheData> matchesByName = cacheView.getAll(NAMED_IMAGES.ns, namedImageIdentifiers, RelationshipCacheFilter.include(IMAGES.ns))
-
     Collection<CacheData> matchesByImageId = cacheView.getAll(IMAGES.ns, imageIdentifiers)
 
-    List<NamedImage> allFilteredImages = filter(
-      render(matchesByName, matchesByImageId, lookupOptions.q, lookupOptions.region),
-      extractTagFilters(request)
-    )
+    Map<String, String> tagFilters = extractTagFilters(request)
```
jasonmcintosh, line 94: NIT: Not sure need a variable tagFilters here as it's not re-used.  Just a preference, not a "need".

PR #7189, opened 2025-08-05
fix(orca/retrofit2): fix wildcard in parameter type in echo's retrofit2 client's api interface method

kirangodishala:
- With retrofit2, parameter type must not include a type variable or wildcard so EchoService.recordEvent method parameter is updated. 
- Changed from 
`recordEvent(@Body Map<String, ?> notification)` 
to 
`recordEvent(@Body Map<String, Object> notification)`
- Added a test to demonstrate the issue
- Also changed the method return type to match the actual return of API

### DIFF
--- a/orca/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
+++ b/orca/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
@@ -62,14 +62,14 @@ class EchoNotifyingExecutionListener implements ExecutionListener {
         }
         AuthenticatedRequest.allowAnonymous({
           Retrofit2SyncCall.execute(
-            echoService.recordEvent(
+            echoService.recordEvent([
               details: [
                 source     : "orca",
                 type       : "orca:${execution.type}:starting".toString(),
                 application: execution.application,
               ],
               content: buildContent(execution)
-            )
+            ] as Map)
           )
         })
       }
@@ -92,14 +92,14 @@ class EchoNotifyingExecutionListener implements ExecutionListener {
         }
         AuthenticatedRequest.allowAnonymous({
           Retrofit2SyncCall.execute(
-            echoService.recordEvent(
+            echoService.recordEvent([
               details: [
                 source     : "orca",
                 type       : "orca:${execution.type}:${wasSuccessful ? "complete" : "failed"}".toString(),
                 application: execution.application,
               ],
               content: buildContent(execution)
-            )
+            ] as Map)
           )
         })
       }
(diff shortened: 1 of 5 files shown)


orca/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
```diff
@@ -62,14 +62,14 @@ class EchoNotifyingExecutionListener implements ExecutionListener {
...
-            echoService.recordEvent(
+            echoService.recordEvent([
               details: [
                 source     : "orca",
                 type       : "orca:${execution.type}:starting".toString(),
                 application: execution.application,
               ],
               content: buildContent(execution)
-            )
+            ] as Map)
```
dbyron-sf, line 72: Can you explain this a bit more?  What was happening before this change...some kind of magic groovy literal Map?  Why is the cast required?

orca/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListener.groovy
```diff
@@ -144,7 +144,7 @@ class EchoNotifyingStageListener implements StageListener {
         MDC.put(Header.EXECUTION_ID.header, stage.execution.id)
         MDC.put(Header.USER.header, stage.execution?.authentication?.user ?: "anonymous")
         AuthenticatedRequest.allowAnonymous({
-          Retrofit2SyncCall.execute(echoService.recordEvent(event))
+          Retrofit2SyncCall.execute(echoService.recordEvent(event as Map))
```
dbyron-sf, line 147: Curious about the casts here too, but I can probably close my eyes.  One day we'll convert this to java...

orca/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingExecutionListener.groovy
kirangodishala: `Map<String, ?>` accepts any type for the values but when I changed it to `Map<String, Object>`, it expects all values to be exactly Object, but Groovy sometimes infers more specific types (like Map<String, String> for details). So cast is required.

orca/orca-echo/src/main/groovy/com/netflix/spinnaker/orca/echo/spring/EchoNotifyingStageListener.groovy
kirangodishala: Again, groovy's map literal type inference is more flexible with wildcards. So when I replaced wildcard with specific type, groovy complained.

PR #7190, opened 2025-08-07
fix(keel): add executor pool for asynchronous requests

mattgogerly:
Keel was failing to start as it tries to make an asynchronous call to `Front50Service.allApplications` to prime it's cache on startup. No `Executor` was provided to the `ErrorHandlingExecutorCallAdapterFactory`, so the call never completed and blocked startup.


dbyron-sf:
> Keel was failing to start as it tries to make an asynchronous call to `Front50Service.allApplications` to prime it's cache on startup.

Is [this](https://github.com/spinnaker/spinnaker/blob/48a43bd0baa442ee81b525892b8cc8857a729850/keel/keel-front50/src/main/kotlin/com/netflix/spinnaker/keel/front50/Front50Cache.kt#L88) the call?  That call looks synchronous to me.

dbyron-sf:

Issue #7170, opened 2025-07-25
GCP updateBackendServices Request Does Not Wait for the GCP Operation to Finish


chuanzhenwu:
We are using the Spinnaker Deploy stage to create new GCP Compute Instance Group and Attach groups to a Compute Backend Service.

We noticed that the Deploy stage does not wait and check for the actual operation status for updating GCP Compute Backend Service, but instead just continue if the update requests received no exception. 

Recently this has caused an outage to us, as the update backend service request actually failed after 10 mins, but our deploy stage succeeded and proceeded to delete the old instance groups.

Can we modify the logic to actually introduce waiting and checking for the operation task's status before wrapping the BasicGoogleDeployHandler.updateBackendServices() stage?

Thanks!

jasonmcintosh:
Closing as resolved with merged fixes & releases

PR #7191, opened 2025-08-11
fix(google): Add retry and status polling logic

mrusan:
Addresses https://github.com/spinnaker/spinnaker/issues/7170 similarly to https://github.com/spinnaker/clouddriver/pull/6379 but for the master branch code in the monorepo, where the code was refactored to Java.

The `enableAsyncOperationWait` flag was omitted for this PR, because part of the operations were already using `googleOperationPoller` correctly, and we're just making the behaviour consistent.

Also covers a couple of regression bugs discovered during implementation and testing.


dbyron-sf:
The commit message:
```
chore(google): Fix regression bugs and improve tests
```
needs more detail.  First, was it a chore, a fix, or perhaps some additional test coverage?  What was actually broken?

If the initial commit is `test` and adds whatever coverage we need to demonstrate the bug, and then a subsequent test fixes the bug, and adjusts the test so we can see the behavior change...then we can actually see what's going on.  There's always a tension about whether the initial test fails or passes.  It's typically harder for the person writing the code to make it pass, but easier for everyone else to see the behavior change that way, so please try to make it pass.

If you did add some additional test coverage outside any fix, please include that in a separate `test` commit as well.

mrusan:
> The commit message:
> 
> ```
> chore(google): Fix regression bugs and improve tests
> ```
> 
> needs more detail. First, was it a chore, a fix, or perhaps some additional test coverage? What was actually broken?
> 
> If the initial commit is `test` and adds whatever coverage we need to demonstrate the bug, and then a subsequent test fixes the bug, and adjusts the test so we can see the behavior change...then we can actually see what's going on. There's always a tension about whether the initial test fails or passes. It's typically harder for the person writing the code to make it pass, but easier for everyone else to see the behavior change that way, so please try to make it pass.
> 
> If you did add some additional test coverage outside any fix, please include that in a separate `test` commit as well.

Hello @dbyron-sf 👋 
The commit you're seeing in the PR was a squash of several - are you concerned, that the previous commit messages are visible? 

There's probably some confusion, since the initial development commit message was 

```
chore(google): Add retry and status polling logic for autoscalers and backend services
```

There were additional tests added, to make sure we don't hit this again - are you asking to split them out into a separate Pull Request?

dbyron-sf:
I'm concerned because I can't tell what was actually fixed.  I'm asking to split the code changes into multiple commits, but it's fine if they're in the same PR.

clouddriver/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.java
```diff
@@ -1079,34 +1143,47 @@ private void updateRegionalBackendServices(
...
+                            SCOPE_REGIONAL,
+                            TAG_REGION,
+                            region),
+                        registry);
             task.updateStatus(
                 BASE_PHASE,
                 String.format(
                     "Done associating server group %s with backend service %s.",
                     serverGroupName, backendService.getName()));
+
```
jasonmcintosh: This doesn't seem to quite match the other PR where the status was set to "Waiting for backend service..."?

mrusan:
@dbyron-sf I split up the single commit into several, as discussed
@jasonmcintosh I updated the status messages to mirror the previous Groovy implementation, including ordering and exact phrasing, also verified existing messages were already aligned

dbyron-sf:
I'm still struggling to see what's going on.  With any luck @jasonmcintosh has enough context to figure it out.
The first commit (ignore the hash, I rebased on main in my local clone)
```
* b5271c8b86 - fix(google): Align polling logic to wait on all async operations (2 minutes ago) <Maksim Rusan>
```
doesn't have any test changes, so it's difficult/impossible to see what behavior changes users will see.  Seems like we have a testing gap since all the tests pass.  Later on, there are commits that add tests.  What would help is to add the tests first to demonstrate the current behavior.  Then, when the fix comes later, I'd expect some change to the tests that make it clear what behavior is changing.

For
```
* 0c0fa304d8 - fix(google): Regressions for lbPolicy and updateBackendServices() (5 minutes ago) <Maksim Rusan>
```
there's still not enough detail in the commit message.  What was broken?  How does this particular change fix it?  Again, there are no test changes here, so it's basically impossible to tell what's changed.

dbyron-sf:

PR #7196, opened 2025-08-12
feat(plugins): Log output of the version requirements when a mismatch

jasonmcintosh:
Was debugging a weird semver parsing...and the log messages were NOT helpful.  Just adding some basic logging on parsing of plugins and when version parsing fails, wrap it and output the information.  Could see making the .info a debug potentially....


edgarulg (approved):
LGTM

PR #7206, opened 2025-08-20
fix(halyard/retrofit2): fix daemon service getTask API definition

kirangodishala:
- Retrofit2 does not support type variables in method return type. So `DaemonService.getTask()` is failing with the following error:
```
Method return type must not include a type variable or
  wildcard: com.netflix.spinnaker.halyard.core.tasks.v1.DaemonTask<C, T>
for method DaemonService.getTask 
```

- This PR fixes the issue by manually deserializing the ResponseBody so that retrofit2 doesn't need to know the actual generic types at compile time. 

### DIFF
--- a/halyard/halyard-cli/src/test/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonServiceTest.java
+++ b/halyard/halyard-cli/src/test/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonServiceTest.java
@@ -1,0 +1,67 @@
+package com.netflix.spinnaker.halyard.cli.services.v1;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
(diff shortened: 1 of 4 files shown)


halyard/halyard-cli/src/test/java/com/netflix/spinnaker/halyard/cli/services/v1/DaemonServiceTest.java
```diff
@@ -0,0 +1,67 @@
+package com.netflix.spinnaker.halyard.cli.services.v1;
```
dbyron-sf, line 18: missing license header?

kirangodishala: added now

dbyron-sf:

PR #7212, opened 2025-08-21
fix(echo/rest): make it possible to send events to URLs with no trailing slash

dbyron-sf:
The move to retrofit2 made it impossible to send events to URLs with no trailing slash.  https://github.com/spinnaker/echo/pull/1476 changed POST("/") to POST(".") and ensured the base URL ends with a trailing slash with RetrofitUtils.getBaseUrl.  retrofit2 rejects baseUrls that don't end with a trailing slash though, so we need a different solution.

echo moved to retrofit2 in http://github.com/spinnaker/echo/pull/1466, so this fix needs to go to release-2025.0.x, release-2025.1.x (and release-1.37.x in http://github.com/spinnaker/echo if we change plans and decide to release 1.37.next).

### DIFF
--- a/echo/echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventService.java
+++ b/echo/echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventService.java
@@ -45,7 +45,8 @@ public class RestEventService {
    */
   public void sendEventWithCircuitBreaker(
       Map<String, Object> eventMap, RestUrls.Service service, CircuitBreaker circuitBreaker) {
-    circuitBreaker.executeRunnable(() -> sendEvent(eventMap, service));
+    circuitBreaker.executeRunnable(
+        () -> sendEvent(service.getConfig().getUrl(), eventMap, service));
   }
 
   /**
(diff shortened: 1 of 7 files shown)


echo/echo-rest/src/main/java/com/netflix/spinnaker/echo/events/RestEventService.java
```diff
@@ -54,9 +55,11 @@ public void sendEventWithCircuitBreaker(
    * @param event The event data to be sent.
    * @param service The REST service to which the event should be sent.
    */
-  public void sendEvent(Map<String, Object> event, RestUrls.Service service) {
+  public void sendEvent(String url, Map<String, Object> event, RestUrls.Service service) {
```
jasonmcintosh: Since the service is passed, actually looking at this today, wondering if we need url in the "sendEvent" call...

dbyron-sf: Good point!  https://github.com/spinnaker/spinnaker/pull/7212/commits/d765305b6dd36b66fc33cfe12e02a336959f7035

echo/echo-rest/src/test/java/com/netflix/spinnaker/echo/rest/RestServiceTest.java
```diff
@@ -86,20 +87,23 @@ void testRetService_withPostMappingSingleSlash() {
   }
 
   @Test
-  void testRetService_withPostMappingDot() {
+  void testRestService_withPostMappingDot() {
```
jasonmcintosh, line 90: Honestly this is a bit... questionable as a test in general.  This is testing RetroFit itself in a sense vs. anything that's actually in spinnaker?   Not a huge deal just an observation :) useful in that it DOES check some of the behaviors of the event but... wonder if could have tested the standard RestService...

dbyron-sf: Yeah, it was something we learned along the way of adopting retrofit2...

dbyron-sf:
@mergifyio backport release-2025.0.x release-2025.1.x

PR #7216, opened 2025-08-21
chore(gradle): Upgraded and changed a number of settings to work with latest gradle and plugins

jasonmcintosh:
this is currently a work in progress.  Upgrading a few plugins, removing dead config (java 11 support), and a few other key aspects.  This ALSO bumps a number of versions and spotless language usage to the latest versions, leading to some slight formatting changes as part of this PR.


jasonmcintosh:
Closed due to other upgrades bringing some of this in!

PR #7218, opened 2025-08-22
fix(rosco): BakeryService bakeManifest request body fix

christosarvanitis:
Post retrofit2 update Orca doesnt send the correct body to Rosco causing the Helm/Kustomize bakes to fail with:

`Status: 400, Method: POST, URL: http://localhost:8087/api/v2/manifest/bake/KUSTOMIZE, Message: Exactly one input artifact must be provided to bake.`


dbyron-sf:
replaced by https://github.com/spinnaker/spinnaker/pull/7224

PR #7230, opened 2025-08-29
fix(halyard/retrofit2): fix DaemonService's convertor factory

kirangodishala:
- The default retrofit2 convertor factory doesn't serialize an object passed to @ Body param whose type is an abstract class. This PR fixes the issue by passing a custom convertor factory which works in similar lines as retrofit1. 


dbyron-sf:

PR #7238, opened 2025-09-04
feat(clouddriver/aws): Implement the account API for aws accounts

dbyron-sf:
This change allows clouddriver AWS accounts to use the account management API. Specifically, it allows clouddriver to load AWS accounts from the AccountDefinitionRepository used by the account API.

The new config flag `account.storage.aws.enabled`, in addition to the existing config flag `account.storage.enabled`, control when this feature is enabled.  Both are false by default.

### DIFF
--- a/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/config/AccountsConfiguration.java
+++ b/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/config/AccountsConfiguration.java
@@ -21,6 +21,7 @@ import static lombok.EqualsAndHashCode.Include;
 import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.netflix.spinnaker.clouddriver.security.AccessControlledAccountDefinition;
 import com.netflix.spinnaker.fiat.model.resources.Permissions;
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Comparator;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/config/AccountsConfiguration.java
```diff
@@ -57,8 +58,8 @@ public static class Account implements AccessControlledAccountDefinition {
     @Include private Boolean enabled;
     @Include private List<CredentialsConfig.Region> regions;
     @Include private List<String> defaultSecurityGroups;
-    private List<String> requiredGroupMembership;
-    @Include private Permissions.Builder permissions;
+    private List<String> requiredGroupMembership = new ArrayList<>();
+    @Include private Permissions.Builder permissions = new Permissions.Builder();
```
dbyron-sf, line 62: We don't have a test for this.  The motivator is the `@Nonnull` annotation in [AccessControlledAccountDefinition](https://github.com/spinnaker/spinnaker/blob/9d86a95abc517ea909735f0ee0884bd5f502131d/clouddriver/clouddriver-api/src/main/java/com/netflix/spinnaker/clouddriver/security/AccessControlledAccountDefinition.java#L29).

requiredGroupMembership is the deprecated version of permissions. We didn't see any references to it but decided to give it a default value anyway.

jasonmcintosh: Any chance of marking this deprecated too?  I'd love to get rid of that field... :)

jasonmcintosh:
Be nice to get these in for GCP/others too... I thought we'd had it more than just for k8s/docker at ONE point but maybe not :(

PR #7239, opened 2025-09-04
feat(clouddriver/aws): log aws sdk endpoints

dbyron-sf:
If if the new configuration flag `aws.client.logEndpoints` is true (default: false), log the aws endpoint used by all aws clients from AmazonClientBuilder once per aws service.

Example log messages:
```    
LogEndpointRequestHandler::beforeRequest: service name: 'AmazonEC2', endpoint: 'https://ec2.us-west-2.amazonaws.com'
```

As well, log each aws sdk endpoint used retrieve s3 artifacts once if the new config property `artifacts.s3.logEndpoints` is true.  It defaults to false.

Example log message:
```
S3ArtifactRequestHandler::beforeRequest: name: <s3-artifact-account-name>, endpoint: 'https://<bucket>.us-east-2.amazonaws.com'
```


jasonmcintosh (approved):
I'd probably have gone DEBUG vs. info but it's disabled by default so :) that's fine

PR #7240, opened 2025-09-04
feat(clouddriver/aws): log the endpoint that aws sdk sts clients use

dbyron-sf:
when the existing config flag `aws.client.logEndpoints` is true.  It defaults to false.  https://github.com/spinnaker/spinnaker/pull/7239 added that config flag.

### DIFF
--- a/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/NetflixSTSAssumeRoleSessionCredentialsProvider.java
+++ b/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/NetflixSTSAssumeRoleSessionCredentialsProvider.java
@@ -22,6 +22,7 @@ import com.amazonaws.auth.AWSSessionCredentialsProvider;
 import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
 import com.amazonaws.client.builder.AwsClientBuilder.EndpointConfiguration;
 import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient;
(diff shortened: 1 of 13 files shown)


clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/NetflixSTSAssumeRoleSessionCredentialsProvider.java
```diff
@@ -73,6 +75,10 @@ public NetflixSTSAssumeRoleSessionCredentialsProvider(
       stsClientBuilder.withRegion(this.region);
     }
 
+    if (awsConfigurationProperties.getClient().getLogEndpoints()) {
```
dbyron-sf, line 78: This whole PR is actually about this tiny change here.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/config/AmazonCredentialsParser.java
```diff
@@ -602,7 +611,8 @@ public CopyConstructorTranslator(ObjectMapper objectMapper, Class<T> credentialT
       this.credentialType = credentialType;
       try {
         copyConstructor =
-            credentialType.getConstructor(credentialType, AWSCredentialsProvider.class);
+            credentialType.getConstructor(
+                credentialType, AWSCredentialsProvider.class, AwsConfigurationProperties.class);
```
jasonmcintosh, line 615: NOT complaining... but this MIGHT break some of the plugins that is out there (which I'm hoping goes away anyways).  Might be worth a doc note though just in case.

jasonmcintosh (approved):

PR #7241, opened 2025-09-04
fix(echo): Fix NullPointerException when REST is enabled but no endpoints configured

shlomodaari:
This fix addresses an issue where the Echo service crashes with a NullPointerException when rest.enabled=true but no REST endpoints are defined. The fix includes:

1. Initialize the endpoints list as an empty ArrayList in RestProperties to ensure it's never null

2. Add a defensive null check in RestConfig with a warning log message for better error handling

This provides a robust solution with defence in depth to prevent the service from crashing due to missing REST endpoint configuration.

This was introduced as part of this PR: https://github.com/spinnaker/spinnaker/commit/0e8c45b45a7374f4ddacf579aa12dd1d67915b79

### DIFF
--- a/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentials.java
+++ b/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentials.java
@@ -22,6 +22,7 @@ import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.netflix.spinnaker.clouddriver.aws.AwsConfigurationProperties;
 import com.netflix.spinnaker.clouddriver.security.AbstractAccountCredentials;
 import com.netflix.spinnaker.fiat.model.resources.Permissions;
 import java.util.ArrayList;
@@ -31,7 +32,7 @@ import java.util.List;
 import java.util.Objects;
 
 /**
- * Basic set of Amazon credentials that will a provided {@link
+ * Basic set of Amazon credentials that use a provided {@link
  * com.amazonaws.auth.AWSCredentialsProvider} to resolve account credentials. If none provided, the
  * {@link com.amazonaws.auth.DefaultAWSCredentialsProviderChain} will be used. The account's active
  * regions and availability zones can be specified as well.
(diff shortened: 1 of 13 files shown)


shlomodaari:
The error I encountered when rest.enabled: true with no endpoints configured:
```Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
2025-09-04 19:03:50.969 ERROR 1 --- [           main] o.s.boot.SpringApplication               : Application run failed

org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'restEventListener' defined in URL [jar:file:/opt/echo/lib/echo-rest-2025.8.29.22.13.36.main.jar!/com/netflix/spinnaker/echo/events/RestEventListener.class]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'restServices' defined in class path resource [com/netflix/spinnaker/echo/config/RestConfig.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [com.netflix.spinnaker.echo.config.RestUrls]: Factory method 'restServices' threw exception; nested exception is java.lang.NullPointerException: Cannot invoke "java.util.List.iterator()" because the return value of "com.netflix.spinnaker.echo.config.RestProperties.getEndpoints()" is null
        at org.springframework.beans.factory.support.ConstructorResolver.createArgumentArray(ConstructorResolver.java:801) ~[spring-beans-5.3.31.jar:5.3.31]
        at org.springframework.beans.factory.support.ConstructorResolver.autowireConstructor(ConstructorResolver.java:224) ~[spring-beans-5.3.31.jar:5.3.31]
        at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.autowireConstructor(AbstractAutowireCapableBeanFactory.java:1372) ~[spri

File: echo/echo-rest/src/main/java/com/netflix/spinnaker/echo/config/RestConfig.java [hunk 1]
```diff
@@ -83,6 +83,13 @@ RestUrls restServices(
 
     RestUrls restUrls = new RestUrls();
 
+    // Add null check to prevent NullPointerException when no endpoints are configured
+    if (restProperties.getEndpoints() == null) {
```
dbyron-sf, line 87: I'm a little torn here.  Things are arguably better without this check, right?  Echo fails to start because echo is misconfigured.  As in, someone set `rest.enabled` to true, but didn't configure any endpoints.  Isn't failing fast the more helpful thing here?  That way folks have a (much) better chance of fixing their config, as opposed to running for a potentially long time before realizing they're not getting events.

shlomodaari: This is a second layer of security, but the   @Valid List<RestEndpointConfiguration> endpoints = new ArrayList<>(); should fix it! 
We can remove the if, if needed :)

dbyron-sf: We disagree.  To me, this PR makes it easier to miss configuration errors.

jasonmcintosh: I kinda agree with DB - this SHOULD have been failing.  I THINK replacing the exception with a GOOD exception (InvalidConfig or similar) and a GOOD ERROR message ... and then just release notes "IF You've previously had rest.enable true without endpoints, this will not fail on this.".  

I can see SOME people making a mistake or an insert mistake... OR "we leave it default but only set the endpoint in prod" is the opposite argument.  That "we left it on everywhere but only set it in prod" is the only case I can think of... and if you're doing that, probably should just set rest.enabled in prod to true IMHO.  

Either way it DOES break things (apparently our internal systems left this on... and never set endpoints, which is where this came from).  AKA mistake on armory internals.

shlomodaari: I understand both of you, and I will close this PR! 
Maybe we can just add a comment for this in the next release, so people are aware of it! :) 

Thank you for taking the time to review and comment!

PR #7242, opened 2025-09-05
fix(sql): Update account type column during upserts

dbyron-sf:
If a user creates an account with an initial type, say "aws", and then creates another account with the same name but a different type, say "kubernetes", the operation will succeed but the type given by the "type" column will not match the actual type of the account. This causes clouddriver to fail to start as it cannot marshal the account definition jsons into the correct Java objects. Fix this by also updating the "type" column during an update operation.


jasonmcintosh (approved):
Oh man - we'd not dug too much but this looks like an exact issue we were kinda troubleshooting.  THANK YOU! :)

PR #7243, opened 2025-09-09
feat(echo/aws): upgrade echo from aws sdk v1 to v2

kirangodishala:
- AWS SDK for Java v1.x will reach end-of-support on December 31, 2025. So spinnaker needs to be upgraded to AWS SDK v2 at the earliest. 
- This PR upgrades echo service to AWS SDK v2 and removes all the references to AWS SDK v1 libraries. 
- A couple of classes from Kork are also affected with this change.


echo/echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/pubsub/aws/AmazonMessageAcknowledger.java
```diff
@@ -53,17 +54,15 @@ public AmazonMessageAcknowledger(
...
       registry.counter(getProcessedMetricId(subscriptionName)).increment();
-    } catch (ReceiptHandleIsInvalidException e) {
-      log.warn(
-          "Error deleting message: {}, queue: {}, reason: {} (receiptHandle: {})",
-          message.getMessageId(),
-          queueUrl,
-          e.getMessage(),
-          message.getReceiptHandle());
+    } catch (SqsException e) {
+      log.warn("Failed to delete message from queue {}: {}", queueUrl, e.getMessage());
```
dbyron-sf: Possible to reduce diffs in the message?  I see SqsException has a requestId property.  Is that the same as `message.getMessageId()` ?  If the receipt handle was helpful before, I figure it's still helpful.  I don't see that in SqsException.

kirangodishala: Updated. We have `message.messageId` in v2 as well.

echo/echo-pubsub-aws/src/main/java/com/netflix/spinnaker/echo/pubsub/aws/AmazonMessageAcknowledger.java
```diff
@@ -53,17 +54,20 @@ public AmazonMessageAcknowledger(
...
-    // Delete from queue
     try {
-      amazonSQS.deleteMessage(queueUrl, message.getReceiptHandle());
+      sqsClient.deleteMessage(
+          DeleteMessageRequest.builder()
+              .queueUrl(queueUrl)
+              .receiptHandle(message.receiptHandle())
+              .build());
       registry.counter(getProcessedMetricId(subscriptionName)).increment();
-    } catch (ReceiptHandleIsInvalidException e) {
```
dbyron-sf, line 60: Looks like ReceiptHandleIsInvalidException exists in v2 as well.  It does seem more correct to catch the more general exception.

kork/kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/AwsComponents.java
```diff
@@ -35,6 +37,12 @@ AWSCredentialsProvider awsCredentialsProvider() {
     return new DefaultAWSCredentialsProviderChain();
   }
 
+  @Bean
+  @ConditionalOnMissingBean(AwsCredentialsProvider.class)
```
dbyron-sf, line 41: `@ConditionalOnMissingBean` probably isn't correct here.  I suspect it needs to be `@ConditionalOnProperty` instead, so we're not depending on the order that Spring processes bean-providing code.

Plus also, can we have a test that demonstrates the need to provide this bean?

kirangodishala: Modules like clouddriver can have their own AwsCredentialsProvider but others rely on this. Hence I feel  `@ConditionalOnMissingBean` is appropriate. It has been the case with [AWS V1](https://github.com/spinnaker/spinnaker/blob/main/kork/kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/AwsComponents.java#L33) as well. 

I will work on the test.

dbyron-sf: Yeah, it's wrong there too.  From the docs for `@ConditionalOnMissingBean`:
```
The condition can only match the bean definitions that have been processed by the application context so far and, as such, it is strongly recommended to use this condition on auto-configuration classes **only.**
```
Neither this nor the v1 place are auto-configuration classes.

kork/kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/V2ARN.java
```diff
@@ -0,0 +1,62 @@
...
+
+package com.netflix.spinnaker.kork.aws;
+
+import software.amazon.awssdk.arns.Arn;
+
+/**
+ * Util for easy parsing of an Amazon Resource Names (ARNS)
+ * https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
+ */
+public class V2ARN {
```
jasonmcintosh: This is nearly a duplicate of https://github.com/spinnaker/kork/blob/master/kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/ARN.java MINUS the SDK change.  I know it's somewhat nitpicking but be nice if they both shared the same properties & just overloaded the constructors kinda thing or similar.  Then deprecate the ARN one.  Given what this does i"d just replace the V1 ARN with the V2 package parsing.  Per https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-side-by-side.html

kork/kork-pubsub-aws/src/main/java/com/netflix/spinnaker/kork/pubsub/aws/PubSubUtils.java
```diff
@@ -82,6 +135,23 @@ public static String subscribeToTopic(AmazonSNS amazonSNS, ARN topicARN, ARN que
         EXPONENTIAL);
   }
 
+  /** Returns the subscription arn resulting from subscribing the queueARN to the topicARN */
+  public static String subscribeToTopic(SnsClient snsClient, V2ARN topicARN, V2ARN queueARN) {
```
jasonmcintosh: I know there were limited tests before on this but be nice to have tests...

jasonmcintosh:
I'm MOSTLY ok with the code ;) minus.. one nit pick on the ARN & that tests are rather non-existent.  I'm not really worried about keeping sdk1 & 2 in the same base at least for now though it's nice to see it gone in echo.

kork/kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/AwsComponents.java
kirangodishala: [AwsComponents](https://github.com/spinnaker/spinnaker/blob/main/kork/kork-aws/src/main/resources/META-INF/spring/org.springframework.boot.autoconfigure.AutoConfiguration.imports) is an auto-configuration class

kork/kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/V2ARN.java
kirangodishala: I think, I can use the existing ARN class with SDK change. That works even with existing V1 code. Because there is absolutely no change in the functionality with the use of  v1's ARN vs v2's ARN. @dbyron-sf - what do you say?

dbyron-sf: Sure, let's use the existing ARN class.

kork/kork-aws/src/main/java/com/netflix/spinnaker/kork/aws/AwsComponents.java
dbyron-sf: OK, fair enough...and we're in the exact same file, providing one more bean...and have the joy of AWSCredentialsProvider (v1) vs AwsCredentialsProvider (v2).

PR #7244, opened 2025-09-10
chore(graphql): Remove unused and legacy graphql libraries and sample endpoints

jasonmcintosh:


edgarulg (approved):
LGTM, I am not seeing any test updated as part of the graphql removed it might not have tests at all.

jasonmcintosh:
Never had tests and was removed from a lot of downstream services 5+ years ago - this was just never cleaned up.  Keel is the only system that still utilizes graphql for stuff and it works completely differently.

PR #7249, opened 2025-09-17
fix(oracle): A few oracle operations call themselves google which causes some class path bean issues in rare cases

jasonmcintosh:


jasonmcintosh:
NOTE:  The test on this is ... hard to repeat - as you'd need to enable both providers and do some class path scanning and it's... also a touch odd how it works.  I caught this via an extension scan on when both are in the class path at the same time in different orders.

dbyron-sf (approved):

PR #7250, opened 2025-09-19
feat(front50/aws): upgrade front50 to aws sdk v2

kirangodishala:
- AWS SDK for Java v1.x will reach end-of-support on December 31, 2025. So spinnaker needs to be upgraded to AWS SDK v2 at the earliest.
- This PR upgrades front50 service to AWS SDK v2 and removes all the references to AWS SDK v1 libraries. 
-  Also part of halyard that makes uses front50-s3 module is updated
- `SignerOverride` is removed when replacing AmazonS3 with S3Client in `S3ClientFactory`. In AWS SDK v2, the signer is chosen automatically by the SDK based on the service and region. The default and recommended signer is AWS Signature Version 4 (SigV4). AWS SDK v1 supported legacy signers such as AWS Signature Version 2, S3SignerType, and QueryStringSigner, but these are not available in SDK v2. Even most S3-compatible storage systems now support SigV4, so there is typically no need for a custom implementation of legacy signers.
- So the property `spinnaker.s3.signerOverride` property is ignored with this aws v2 upgrade.


front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
```diff
@@ -34,51 +35,51 @@
...
 
-      Optional.ofNullable(s3Properties.getProxyHost()).ifPresent(clientConfiguration::setProxyHost);
-      Optional.ofNullable(s3Properties.getProxyPort())
-          .map(Integer::parseInt)
-          .ifPresent(clientConfiguration::setProxyPort);
-    }
+      ProxyConfiguration proxyConfig =
+          ProxyConfiguration.builder().endpoint(URI.create(proxyUri)).build();
 
-    if (!StringUtils.isEmpty(s3Properties.getSignerOverride())) {
```
dbyron-sf, line 54: Looks like we're not doing signer override anymore.  If that's right, please update the PR description with details about the specific config property we're ignoring (and why), and what the consequences are for people who have that property set.

halyard/halyard-config/halyard-config.gradle
```diff
@@ -23,8 +23,8 @@ dependencies {
   implementation "io.spinnaker.kork:kork-docker"
   implementation "io.spinnaker.kork:kork-web"
   implementation "io.spinnaker.kork:kork-cloud-config-server"
-  implementation 'com.amazonaws:aws-java-sdk-core:1.11.534'
-  implementation 'com.amazonaws:aws-java-sdk-s3:1.11.534'
+  implementation "software.amazon.awssdk:auth"
```
dbyron-sf, line 26: Why change halyard at the same time?  I'm concerned there are other uses of e.g. S3Validator or AwsAccountValidator where the underlying Spinnaker code is still on v1 of the sdk.  If that doesn't matter, then it seems like we could make this change in a separate PR.

kirangodishala: `S3Validator` extends `Validator<S3PersistentStore>` which is connected only to front50. `AwsAccountValidator.getAwsCredentialsProvider` method is used only by S3Validator. So there is no other impact on existing v1 code.

S3Validator from halyard calls methods of front50 classes (`S3Config` & `S3MetadataStorageProperties`). So with the change in front50, the code in halyard is bound to change to escape complier errors. The only other option left would be to keep the existing front50 classes and create v2 versions of them but I don't think that is necessary.

kirangodishala: There is no other v1 code in `halyard-config`. In only remaining aws code is in halyard-deploy and they too use only one single util (`com.amazonaws.util.IOUtils` ) class which has nothing to do with above changes.

dbyron-sf: OK, makes sense.

PR #7253, opened 2025-09-22
feat(healthcheck): Add GRPC and HTTP/2 support for GCP load balancers

shlomodaari:
feat(healthcheck): Add GRPC and HTTP/2 support for GCP load balancers
This change adds complete support for GRPC and HTTP/2 health checks in Google Cloud load balancers. The implementation includes both backend and frontend changes to fully support these protocols.

Added grpcServiceName field to GoogleHealthCheck class
Fixed GRPC and HTTP/2 target format in getTarget()
Updated health check caching and Google API integration
Added HTTP/2 and GRPC options to UI components
Added proper GRPC service name handling in UI and backend
Fixed class references and imports
Added support in DeleteGoogleInternalLoadBalancerAtomicOperation
image
image
image
image
image


clouddriver/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
```diff
@@ -106,6 +117,7 @@ class GoogleHealthCheck {
     int healthyThreshold
     int port
     String requestPath
+    String grpcServiceName
```
christosarvanitis, line 120: Do we need a new string for this or we can use the requestPath since anyone using the gRPC would now what the endpoint should look like? Im good either way

clouddriver/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheckTest.groovy
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright 2024 Armory, Inc.
```
christosarvanitis: YYYY Company ;)

clouddriver/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheck.groovy
shlomodaari: ## Why the `grpcServiceName` field was necessary:

The `grpcServiceName` field was added to properly support GRPC health checks, which have fundamentally different target format requirements compared to HTTP-based health checks.

### Technical Details:

1. **Different Target Formats:**
   - HTTP/HTTPS: `"HTTP:${port}${requestPath ?: '/'}"`  
   - GRPC: `"GRPC:${port}${grpcServiceName ?: ''}"`

2. **Key Differences:**
   - **Path vs Service Name**: HTTP uses URL paths (e.g., `/health`, `/api/status`) while GRPC uses service names (e.g., `com.example.HealthService`)
   - **Slash Handling**: HTTP paths require a leading slash when empty (`/`), but GRPC service names should not have leading slashes
   - **Protocol Semantics**: GRPC service names follow protobuf service naming conventions, not HTTP URL conventions

3. **Why Reusing `requestPath` Wouldn't Work:**
   - GRPC service names like `"com.example.HealthService"` would be incorrectly formatted as `"GRPC:9090/com.example.HealthService"` 
   - Empty GRPC service names would get a `/` suffix: `"GRPC:9090/"` instead of the correct `"GRPC:9090"`
   - This would cause health check failures as the GRPC protocol expects proper service name formatting

4. **Solution:**
   - Added dedicated `grpcServiceName` field for GRPC-specific service names
   - Maintained `requestPath` for HTTP/HTTPS compatibility
   - Updated target generation logic to use the appropriate field based on health check type

This separation ensures protocol compliance and prevents health check failures when using GRPC load balancers.

clouddriver/clouddriver-google/src/test/groovy/com/netflix/spinnaker/clouddriver/google/model/GoogleHealthCheckTest.groovy
shlomodaari: LOL good catch! 

I Updated that

cristhian-castaneda (approved):
LGTM 🌮

jasonmcintosh (approved):

PR #7254, opened 2025-09-23
fix(oauth2): move tokenResponseClient bean to static inner config to avoid circular dependency

rahul-chekuri:
Previously, the OAuth2AccessTokenResponseClient bean was defined directly inside OAuth2SsoConfig while also being injected into the same class. This caused a circular dependency at runtime in Spring Boot 3, which prohibits circular references by default.

**Error observed:**
```
APPLICATION FAILED TO START
Description:

The dependencies of some of the beans in the application context form a cycle:

┌──->──┐
|  OAuth2SsoConfig (field private OAuth2AccessTokenResponseClient tokenResponseClient)
└──<-──┘

Action:
Relying upon circular references is discouraged and they are prohibited by
default. Update your application to remove the dependency cycle.
```

**Resolution:**

- Moved tokenResponseClient bean definition into a static inner @Configuration class (OAuth2ClientBeans).
- This separates bean creation from injection and breaks the cycle.
- Keeps DefaultAuthorizationCodeTokenResponseClient for prod.
- Still allows overriding with TestAuthorizationCodeTokenResponseClient in tests.

**Test note:**

The OAuth2Test did not fail earlier because OAuth2TestConfiguration was importing a
mock OAuth2AccessTokenResponseClient unnecessarily, which suppressed this
issue. Without that mock, the test suite would have caught the circular
dependency during startup.

### DIFF
--- a/gate/gate-web/src/test/java/com/netflix/spinnaker/gate/security/oauth/TestAuthorizationCodeTokenResponseClient.java
+++ b/gate/gate-web/src/test/java/com/netflix/spinnaker/gate/security/oauth/TestAuthorizationCodeTokenResponseClient.java
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2025 OpsMx, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


gate/gate-web/src/test/java/com/netflix/spinnaker/gate/security/oauth/OAuth2Test.java
```diff
@@ -42,7 +40,6 @@
       "spring.security.oauth2.client.registration.github.client-secret=client-secret"
     })
 @TestPropertySource(properties = {"spring.config.location=classpath:gate-test.yml"})
-@Import(OAuth2TestConfiguration.class)
```
rahul-chekuri, line 45: @dbyron-sf The OAuth2Test did not fail earlier because `OAuth2TestConfiguration` was importing a mock `OAuth2AccessTokenResponseClient`, which suppressed this issue. Without that mock, the test suite would have caught the circular dependency during startup, so I removed the import as it is not needed here.

dbyron-sf:
@mergifyio backport release-2025.2.x

PR #7245, opened 2025-09-12
Various kork updates (secrets, utilities, prereqs for credentials updates)

jvz:
I've broken up these changes into a few commits to cover the various things this includes. I still have further changes to make that depend on these which are related to credentials/accounts updates (moving the clouddriver code into the updated kork API along with updates to Igor and Kayenta for credentials self service), but I figured this was already large enough to start with.

- Kotlin update to 1.8
- Combined the editor config files into a top level file (which helps with formatting files not in subdirectories)
- Updates to `AuthenticatedRequest` to consider using the spring security principal classes for obtaining a username
- Extraction of a `ClassScanner` utility class to replace some repetitive use of class scanning APIs from Spring
- Addition of a `NamedTypeProvider` SPI for simplifying how to scan for named JSON types (such as used by credentials, but also used in various places in ad hoc fashions)
- Extraction of `BootstrapComponents` configuration class that `PlatformComponents` and elsewhere depend on (duplicate beans)
- *New feature:* `boot-secret://` resolution of external secrets at startup (configures before the `Environment` is resolved)

PR #7256, opened 2025-09-29
feature(kork-secrets-aws): Support secrets-manager boot secrets

jvz:
Follow up to https://github.com/spinnaker/spinnaker/pull/7245 where I add `boot-secret://` support to AWS Secrets Manager as a reference implementation. Similar updates can be made for the other `SecretEngine` implementations, but they require some refactoring and added tests due to how the existing implementations are structured (boot secrets resolve a string first before optionally being parsed by a YAML/JSON parser if a `k` query parameter is present; this differs slightly from how `SecretEngine` implementations support the same parameter in engine-specific code).


jasonmcintosh:
Overall I like the path this is going - but figuring out spring bean order/loading is a PIA :( Waiting on some of that to approve...

PR #7085, opened 2025-05-21
refactor(retrofit2): upgrade all retrofit clients to retrofit2 in Orca

kirangodishala:
- This PR upgrades all retrofit1 clients in `Orca` to retrofit2 and makes required changes in the code and tests accordingly. 

- All the retrofit1 dependencies and references are removed.

- Retrofit2 boilerplate code was already implemented in kork's kork-retrofit module and this PR makes use of it.

PR #7262, opened 2025-10-01
fix(clouddriver/cloudformation): fix an API failure caused by encoded colon char post retrofit2 upgrade

kirangodishala:
- This PR fixes the issue of orca not able to receive aws cloud formation stack for the supplied stackId. 
- The root cause of the issue is that when retrofit is upgraded, all the special characters including colon(`:`) are encoded and cloud formation api expects the stackId to be decoded in the request URI itself which is not the case. Hence explicit decoding is now added to fix the issue.
- This API is a special case as it doesn't get the stackId as a path param but used AntPathMatcher to extract the stackId from the request URI. Now the arrangement is changed to receive stackId as a query param.
- Updated a previously added test to demonstrate the issue. Test initially fails when wiremock expects decoded URL and it passes when changed to encoded url.
- Added an integration test that verifies the functionality of the `CloudFormationController's' API when a stackId arn is passed as a query param.

### DIFF
--- a/orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/OortService.java
+++ b/orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/OortService.java
@@ -178,7 +178,7 @@ public interface OortService {
   Call<List<Map>> getEntityTags(@QueryMap Map parameters);
 
   @GET("aws/cloudFormation/stacks/stack")
-  Call<Map> getCloudFormationStack(@Query("stackId") String stackId);
+  Call<Map<String, Object>> getCloudFormationStack(@Query("stackId") String stackId);
 
   @GET("servicebroker/{account}/serviceInstance")
   Call<Map<String, Object>> getServiceInstance(
--- a/orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
+++ b/orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/DelegatingOortService.java
@@ -177,7 +177,7 @@ public class DelegatingOortService extends DelegatingClouddriverService<OortServ
   }
 
   @Override
-  public Call<Map> getCloudFormationStack(String stackId) {
+  public Call<Map<String, Object>> getCloudFormationStack(String stackId) {
     return getService().getCloudFormationStack(stackId);
   }
 
--- a/orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
+++ b/orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/providers/aws/cloudformation/WaitForCloudFormationCompletionTask.java
@@ -50,9 +50,8 @@ public class WaitForCloudFormationCompletionTask implements OverridableTimeoutRe
       Map task = ((List<Map>) stage.getContext().get("kato.tasks")).iterator().next();
       Map result = ((List<Map>) task.get("resultObjects")).iterator().next();
       String stackId = (String) result.get("stackId");
-      Map<String, ?> stack =
-          (Map<String, Object>)
-              Retrofit2SyncCall.execute(oortService.getCloudFormationStack(stackId));
+      Map<String, Object> stack =
+          Retrofit2SyncCall.execute(oortService.getCloudFormationStack(stackId));
       log.info(
           "Received cloud formation stackId "
               + stackId


orca/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/OortServiceTest.java
```diff
@@ -72,8 +77,10 @@ void getServerGroupFromClusterPathMustNotComeAfterQuery() {
   void verifyCloudFormationStackApi() {
     String stackId =
         "arn:aws:cloudformation:us-west-2:123456789012:stack/my-stack/50d6f6c0-e4a3-11e4-8f3c-500c217fbb7a";
+    String encodedStackId =
+        "arn%3Aaws%3Acloudformation%3Aus-west-2%3A123456789012%3Astack/my-stack/50d6f6c0-e4a3-11e4-8f3c-500c217fbb7a";
```
jasonmcintosh: The colons of course in this are encoded, but missing the slashes seems... problematic?  Is there a reason the WHOLE thing isn't encoded including the colons?  What about changing Orca with this instead to make this NOT so goofy?

kirangodishala: In a path variable, when slashes are present and "encoded=true" for that path param, retrofit treats the strings appearing between slashes  as path segments so the encoding is limited to the strings between slashes.

orca/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/OortServiceTest.java
```diff
@@ -80,7 +85,15 @@ void verifyCloudFormationStackApi() {
...
+    //
+    //                                                Request was not matched
+    //                                               =======================
+    // Closest stub:
+    // /aws/cloudFormation/stacks/arn:aws:cloudformation:us-west-2:123456789012:stack/my-stack/50d6f6c0-e4a3-11e4-8f3c-500c217fbb7a
+    // Request:
+    // /aws/cloudFormation/stacks/arn%3Aaws%3Acloudformation%3Aus-west-2%3A123456789012%3Astack/my-stack/50d6f6c0-e4a3-11e4-8f3c-500c217fbb7a
 
-    assertThat(map).isNotNull();
+    assertThat(map).isNull();
```
dbyron-sf: Can you add a verify call to demonstrate which wiremock endpoint actually gets hit?  The comment helps, but I'd rather have it in code.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/controllers/CloudFormationController.java
```diff
@@ -55,12 +57,13 @@ AmazonCloudFormationStack get(HttpServletRequest request) {
     String pattern = (String) request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
     String stackId =
         new AntPathMatcher().extractPathWithinPattern(pattern, request.getRequestURI());
-    log.debug("Cloud formation get stack with id {}", stackId);
+    String decodedStackId = URLDecoder.decode(stackId, StandardCharsets.UTF_8);
+    log.debug("Cloud formation get stack with id {}", decodedStackId);
     return cloudFormationProvider
-        .get(stackId)
+        .get(decodedStackId)
```
dbyron-sf: If the fix is in clouddriver, can we have a test in clouddriver too?

kirangodishala: will try. Not possible with wiremock though.

orca/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/OortServiceTest.java
kirangodishala: Will do

orca/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/OortServiceTest.java
jasonmcintosh: Yeah my question is... should we redo this similar to what's been done for igor/jenkins where this isn't a path variable but instead done as a query parameter or similar to allow better handling?  OR is there another alternative.  As said, this will work, I just wonder if there's a "better" solution or a "cleaner" way to reduce the headaches on this.  It may not, just... "irksome" on the way this is handled.

kirangodishala: I change it from path param to query param. This way the changes are minimal and cleaner.

orca/orca-clouddriver/src/test/java/com/netflix/spinnaker/orca/clouddriver/OortServiceTest.java
kirangodishala: Done

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/controllers/CloudFormationController.java
kirangodishala: Added a clouddriver test

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/controllers/CloudFormationController.java
```diff
@@ -50,11 +47,8 @@ List<AmazonCloudFormationStack> list(
     return cloudFormationProvider.list(accountName, region);
   }
 
-  @RequestMapping(method = RequestMethod.GET, value = "/**")
-  AmazonCloudFormationStack get(HttpServletRequest request) {
-    String pattern = (String) request.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
-    String stackId =
-        new AntPathMatcher().extractPathWithinPattern(pattern, request.getRequestURI());
+  @RequestMapping(method = RequestMethod.GET, value = "/stack")
```
dbyron-sf, line 50: I like this change since it seems simpler.  I'm concerned about what happens during deployment of the new code though.  A mismatch of old orca and new clouddriver (or vice-versa) will end up with failed requests.  I think we'd be better off adding the new endpoint and then, later, removing the old one.

dbyron-sf:
If I'm following things correctly, the bug in clouddriver has been present since https://github.com/spinnaker/clouddriver/pull/6340 (first released in [1.37.0](https://spinnaker.io/changelogs/1.37.0-changelog/#retrofit2-upgrade), so we'll backport the fix to release-2025.0.x and release-2025.1.x.  orca still used retrofit1 on the release-2025.0.x branch, so there will be conflicts, but probably not difficult to resolve.

orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/OortService.java
```diff
@@ -177,8 +177,8 @@ Call<List<Map<String, Object>>> getEntityTags(
   @GET("tags")
   Call<List<Map>> getEntityTags(@QueryMap Map parameters);
 
-  @GET("aws/cloudFormation/stacks/{stackId}")
-  Call<Map> getCloudFormationStack(@Path(value = "stackId", encoded = true) String stackId);
+  @GET("aws/cloudFormation/stacks/stack")
+  Call<Map> getCloudFormationStack(@Query("stackId") String stackId);
```
dbyron-sf: If we're making a new endpoint, how feasible is it to use specify the type of the key/value in the Map (e.g. `Map<String, Object>` ?

kirangodishala: Yes we can change it to return Map with generics. I will do that.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/controllers/CloudFormationController.java
kirangodishala: Even now the API call from orca to clouddriver is failing due to encode issue. So I don't think keeping the old code would avoid failure.

dbyron-sf: OK.  I figured it was only a subset of calls that fails, but if it's all of them, then it makes sense to replace the endpoint.

dbyron-sf:
> If I'm following things correctly, the bug in clouddriver has been present since [spinnaker/clouddriver#6340](https://github.com/spinnaker/clouddriver/pull/6340) (first released in [1.37.0](https://spinnaker.io/changelogs/1.37.0-changelog/#retrofit2-upgrade), so we'll backport the fix to release-2025.0.x and release-2025.1.x. orca still used retrofit1 on the release-2025.0.x branch, so there will be conflicts, but probably not difficult to resolve.

Now that I think about it, the bug has more likely been present since orca upgraded to retrofit2 in https://github.com/spinnaker/spinnaker/pull/7085, first released in 2025.1.0.  If that's right, I expect this to only make sense to backport to release-2025.1.x and release-2025.2.x, and I expect this to backport cleanly to both.

dbyron-sf:
@mergifyio backport release-2025.1.x release-2025.2.x

dbyron-sf:
> Now that I think about it, the bug has more likely been present since orca upgraded to retrofit2 in #7085, first released in 2025.1.0. If that's right, I expect this to only make sense to backport to release-2025.1.x and release-2025.2.x, and I expect this to backport cleanly to both.

@kirangodishala please confirm this is how you see it as well.

kirangodishala:
> > Now that I think about it, the bug has more likely been present since orca upgraded to retrofit2 in #7085, first released in 2025.1.0. If that's right, I expect this to only make sense to backport to release-2025.1.x and release-2025.2.x, and I expect this to backport cleanly to both.
> 
> @kirangodishala please confirm this is how you see it as well.

Yes, I confirm @David. The bug was introduced with the orca retrofit2 upgrade in 2025.1.0.

PR #7268, opened 2025-10-07
fix(front50/s3): always configure region in s3 client, when specified

kirangodishala:
- While upgrading front50 to aws sdk v2 in https://github.com/spinnaker/spinnaker/pull/7250, it was mistakenly assumed that the regionOverride was meant for the s3 compatible storage but in reality it was meant only for signerOverride in sdk v1. So with the existing code, for s3 compatible storages, region won't be set and will cause the following error:
```
software.amazon.awssdk.regions.providers.SystemSettingsRegionProvider@629bee3b: 
Unable to load region from system settings. Region must be specified either via environment variable (AWS_REGION) or  system property (aws.region)
```
- This PR fixes the issue by assigning region to S3Builder even for S3 compatible storages.
- A test is added to demonstrate the issue.

### DIFF
--- a/front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
+++ b/front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
@@ -72,13 +72,9 @@ public class S3ClientFactory {
 
     if (!StringUtils.isEmpty(s3Properties.getEndpoint())) {
       s3Builder.endpointOverride(URI.create(s3Properties.getEndpoint()));
-
-      if (!StringUtils.isEmpty(s3Properties.getRegionOverride())) {
-        s3Builder.region(Region.of(s3Properties.getRegionOverride()));
-      }
-    } else {
-      Optional.ofNullable(s3Properties.getRegion()).map(Region::of).ifPresent(s3Builder::region);
     }
+
+    Optional.ofNullable(s3Properties.getRegion()).map(Region::of).ifPresent(s3Builder::region);
 
     return s3Builder.build();
   }
--- a/front50/front50-s3/src/test/java/com/netflix/spinnaker/front50/config/S3ClientFactoryTest.java
+++ b/front50/front50-s3/src/test/java/com/netflix/spinnaker/front50/config/S3ClientFactoryTest.java
@@ -1,0 +1,24 @@
+package com.netflix.spinnaker.front50.config;
+
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.Mockito.mock;
+
+import org.junit.jupiter.api.Test;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.core.exception.SdkClientException;
+
+public class S3ClientFactoryTest {
+
+  @Test
+  void create_shouldNotThrowSdkClientException_whenRegionIsProvidedWithEndpoint() {
+    AwsCredentialsProvider mockProvider = mock(AwsCredentialsProvider.class);
+    S3Properties props = new S3MetadataStorageProperties();
+    props.setRegion("us-west-2");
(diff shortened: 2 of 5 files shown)


front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
```diff
@@ -72,14 +72,10 @@ public static S3Client create(
 
     if (!StringUtils.isEmpty(s3Properties.getEndpoint())) {
       s3Builder.endpointOverride(URI.create(s3Properties.getEndpoint()));
-
-      if (!StringUtils.isEmpty(s3Properties.getRegionOverride())) {
```
dbyron-sf, line 76: Is it time to remove regionOverride and signerOverride from [S3BucketProperties](https://github.com/spinnaker/spinnaker/blob/d0f0620ac1d415e4c23f6fd8e31a4f661331a2a4/front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3BucketProperties.java#L24-L27) ?  It doesn't look like they're used.

front50/front50-s3/src/test/java/com/netflix/spinnaker/front50/config/S3ClientFactoryTest.java
```diff
@@ -15,10 +14,6 @@ void create_shouldNotThrowSdkClientException_whenRegionIsProvidedWithEndpoint()
     S3Properties props = new S3MetadataStorageProperties();
     props.setRegion("us-west-2");
     props.setEndpoint("https://minio-host:9000");
-    // FIXME: this should not throw an exception
-    assertThatThrownBy(() -> S3ClientFactory.create(mockProvider, props))
-        .isInstanceOf(SdkClientException.class)
-        .hasMessageContaining(
-            "Unable to load region from system settings. Region must be specified either via environment variable (AWS_REGION) or  system property (aws.region)");
+    assertThatCode(() -> S3ClientFactory.create(mockProvider, props)).doesNotThrowAnyException();
```
dbyron-sf: It would be stronger to assert that the region in the resulting client matches the configured value.  How feasible is that?

front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
kirangodishala: There is [S3PluginStorageProperties](https://github.com/spinnaker/spinnaker/blob/d0f0620ac1d415e4c23f6fd8e31a4f661331a2a4/front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3PluginStorageProperties.java) that extends `S3BucketProperties` which may get impacted. But as the `AmazonS3` itself is upgraded to `S3Client`, any plugin would have been broken already. So shall I go ahead and remove those fields or wait for some more time?

front50/front50-s3/src/test/java/com/netflix/spinnaker/front50/config/S3ClientFactoryTest.java
kirangodishala: Done

front50/front50-s3/src/main/java/com/netflix/spinnaker/front50/config/S3ClientFactory.java
dbyron-sf: I'd say remove them, and say we did in a release note.

kirangodishala: ok

dbyron-sf: Done in https://github.com/spinnaker/spinnaker/pull/7268/commits/1b4720367dfe4e461bf3593e02c1804e6f89e2b0.

dbyron-sf (approved):

PR #7273, opened 2025-10-10
perf(front50): fetch pipeline by id in StartPipelineTask

qunayu:
Hi,

I'm working with @dbyron-sf to try and port some of their changes over from the salesforce fork.

This PR improves the performance of the `StartPipelineTask` to fetch pipelines by ID rather than the whole lot. Updated original testing to include happy path tests for triggering pipelines as well.

Thanks!

### DIFF
--- a/orca/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy
+++ b/orca/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.orca.front50.tasks
 
 import com.netflix.spinnaker.kork.exceptions.ConfigurationException
 import com.netflix.spinnaker.kork.retrofit.Retrofit2SyncCall
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionStatus
 import com.netflix.spinnaker.orca.api.pipeline.Task
 import com.netflix.spinnaker.orca.api.pipeline.models.PipelineExecution
@@ -34,6 +35,8 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Component
 
 import javax.annotation.Nonnull
+
+import static java.net.HttpURLConnection.HTTP_NOT_FOUND
 
 @Component
 @Slf4j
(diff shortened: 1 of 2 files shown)


orca/orca-front50/src/main/groovy/com/netflix/spinnaker/orca/front50/tasks/StartPipelineTask.groovy
```diff
@@ -66,10 +69,9 @@ class StartPipelineTask implements Task {
...
-    List<Map<String, Object>> pipelines = isStrategy ?
-        Retrofit2SyncCall.execute(front50Service.getStrategies(application)) :
-        Retrofit2SyncCall.execute(front50Service.getPipelines(application, false))
-    Map<String, Object> pipelineConfig = pipelines.find { it.id == pipelineId }
+    Map<String, Object> pipelineConfig = isStrategy
+      ? Retrofit2SyncCall.execute(front50Service.getStrategies(application)).find { it.id == pipelineId }
+      : getPipelineById(pipelineId)
 
     if (!pipelineConfig) {
       throw new ConfigurationException("The referenced ${isStrategy ? 'custom strategy' : 'pipeline'} cannot be located (${pipelineId})")
```
qunayu, line 77: Hey @dbyron-sf , I ended up not converting `ConfigurationException` to `IllegalArgumentException` - I didn't quite understand the intent there and it feels like a bigger change to the interface. Let me know what you think!

dbyron-sf: ConfigurationException inherits from SpinnakerException, so by default RetrySupport retries it (because SpinnakerException.retryable defaults to null and [RetrySupport.retry](https://github.com/spinnaker/spinnaker/blob/c79666126ee96f04eb7a1ad79d9898d12d96a9fe/kork/kork-exceptions/src/main/java/com/netflix/spinnaker/kork/core/RetrySupport.java#L36-L41) retries unless the retryable property is explicitly false.  This doesn't really matter here since the retry mechanism isn't RetrySupport, it's the one in [RunTaskHandler](https://github.com/spinnaker/spinnaker/blob/c79666126ee96f04eb7a1ad79d9898d12d96a9fe/orca/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/RunTaskHandler.kt#L205-L207) that, for both ConfigurationException and IllegalArgumentException, ends up calling [DefaultExceptionHandler.handle](https://github.com/spinnaker/spinnaker/blob/c79666126ee96f04eb7a1ad79d9898d12d96a9fe/orca/orca-core/src/main/java/com/netflix/spinnaker/orca/exceptions/DefaultExceptionHandler.java#L45-L46) which passes `false` for `shouldRetry`.

I'm torn about what's the better exception to use given that neither one retries, and that I don't think we want orca to retry / wait around in the hopes that a pipeline happens to appear in front50.  ConfigurationException feels more correct, or at least equally correct, so in the spirit of minimizing diffs in this PR, let's stick with it.

PR #7298, opened 2025-10-20
fix(orca/clouddriver): clarify log line in AbstractCheckIfApplicationExistsTask

dbyron-sf:
to be more specific.

Log message introduced in https://github.com/spinnaker/orca/pull/4788 / Spinnaker version 1.36.0.

### DIFF
--- a/orca/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/AbstractCheckIfApplicationExistsTask.java
+++ b/orca/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/AbstractCheckIfApplicationExistsTask.java
@@ -99,16 +99,14 @@ public abstract class AbstractCheckIfApplicationExistsTask implements Task {
     }
     if (fetchedApplication == null) {
       if (this.config.isAuditModeEnabled()) {
-        String pipelineName = "unknown";
-        if (stage.getParent() != null) {
-          pipelineName = stage.getParent().getName();
-        }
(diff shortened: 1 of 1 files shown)


orca/orca-clouddriver/src/main/groovy/com/netflix/spinnaker/orca/clouddriver/tasks/AbstractCheckIfApplicationExistsTask.java
```diff
@@ -99,16 +99,14 @@ public TaskResult execute(@Nonnull StageExecution stage) {
...
-            "Warning: stage: {}, pipeline: {}, message: {}. "
-                + "This will be a terminal failure in the near future.",
+            "Warning: {} in stage: {}, type: {}, pipeline: {}. "
+                + "This would be a terminal failure if"
+                + " 'tasks.clouddriver.check-if-application-exists.audit-mode-enabled: false'",
             errorMessage,
             stage.getName(),
-            pipelineName);
+            stage.getType(),
+            stage.getExecution().getName());
```
jasonmcintosh, line 109: Note, I THINK it's "ok" but... there's a change here on a few levels on the output that's... interesting (e.g. using the pipeline name vs. the parent name otherwise defaulting to unknown).

dbyron-sf: Yeah, I generally think the new log line is better/more helpful.

PR #7300, opened 2025-10-21
feat(gate/web): add GET /executions/failedStages endpoint

dbyron-sf:
and corresponding GET /pipelines/failedStages endpoint to orca.

The idea is for the UI to call this endpoint to give a link to a leaf failing stage.  This saves users from clicking through a potentially large number of child pipelines to find the one that failed.

This PR introduces a new configuration property: `tasks.controller.failedStages.onlyIncludeStagesThatFailPipelines` that defaults to false.  FWIW, we set this to true at Salesforce.

### DIFF
--- a/gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/ExecutionsController.java
+++ b/gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/ExecutionsController.java
@@ -168,4 +168,31 @@ public class ExecutionsController {
                 reverse,
                 expand));
   }
+
+  @Operation(
+      summary =
+          "Retrieves the subset of pipeline stages in a failed state. "
+              + "For stages which are nested pipelines the execution graph is traversed and failed child stages are also returned. "
+              + "An empty list is returned if no failed stages map to the given pipeline execution id.")
+  @RequestMapping(value = "/executions/failedStages", method = RequestMethod.GET)
+  List<Object> getFailedStagesForPipelineExecution(
+      @Parameter(description = "Pipeline execution id for which to retrieve failed stages.")
+          @RequestParam(value = "executionId")
+          String executionId,
+      @Parameter(
+              description =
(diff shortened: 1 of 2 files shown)


gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/ExecutionsController.java
```diff
@@ -164,4 +168,31 @@ List searchForPipelineExecutionsByTrigger(
...
+          String deckOrigin,
+      @Parameter(
+              description =
+                  "The maximum number of nested pipeline executions to return for failing stages. The default value is 1.")
+          @RequestParam(value = "limit", defaultValue = "1")
+          Integer limit) {
+    return Retrofit2SyncCall.execute(
+        orcaServiceSelector
+            .select()
+            .getFailedStagesForPipelineExecution(executionId, deckOrigin, limit));
```
jasonmcintosh, line 196: ONE concern:  We're passing the deck origin aka URL path... down to orca.  This exposes some information I'd NOT normally consider good to pass down?  Reading through still so... maybe this is really needed, just... seems off.

orca/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/FailedStageExecution.java
```diff
@@ -0,0 +1,362 @@
...
+        stageExecution.getType().equalsIgnoreCase("pipeline") ? "pipelineConfig" : "runJobConfig";
+
+    final StringBuilder stringBuilder = new StringBuilder();
+
+    if (!Strings.isNullOrEmpty(deckOrigin)) {
+      stringBuilder.append(deckOrigin);
+    }
+
+    return stringBuilder
+        .append("/#/applications/")
```
jasonmcintosh, line 211: Ditto on above... not sure these UI configurations should be in Orca.  Gate yes, not sure in orca...

jasonmcintosh (approved):

PR #7302, opened 2025-10-23
perf(orca): query for individual pipelines from front50

dbyron-sf:
This cleans up some places in orca that query front50 for all pipelines in an application, only to use one of them.

### DIFF
--- a/clouddriver/clouddriver-artifacts/clouddriver-artifacts.gradle
+++ b/clouddriver/clouddriver-artifacts/clouddriver-artifacts.gradle
@@ -22,7 +22,6 @@ dependencies {
   implementation project(":clouddriver-api")
   implementation project(":clouddriver-core")
 
-  implementation "com.amazonaws:aws-java-sdk-s3"
   implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-xml"
   implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-yaml"
   implementation "com.google.apis:google-api-services-storage"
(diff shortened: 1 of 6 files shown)


orca/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -486,6 +498,10 @@ class TaskController {
 
     List<PipelineExecution> matchingExecutions = new ArrayList<>()
 
+    if (pipelineConfigIds.isEmpty()) {
```
jasonmcintosh, line 501: Nice little short circuit :)

orca/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/TaskController.groovy
```diff
@@ -461,19 +462,30 @@ class TaskController {
...
-      pipelines = pipelines.stream().filter({ pipeline ->
-        if (pipelineName != null && pipelineName != "") {
-          return pipeline.get("name") == pipelineName
-        } else {
-          return true
+      if (pipelineName != null && pipelineName != "") {
+        try {
+          Map<String, Object> pipeline = Retrofit2SyncCall.execute(front50Service.getPipeline(application, pipelineName, false))
+          // double check that the name actually matches since that's what the previous code did
+          if (pipeline.get("name") == pipelineName) {
```
jasonmcintosh: MINOR concern... I know TECHNICALLY this works, but dear lord I dislike the "==" vs. ".equals" on these cases.

dbyron-sf: Done in https://github.com/spinnaker/spinnaker/pull/7302/commits/28e3a0711864b9d70e11d53a44931215fe73630c.

jasonmcintosh (approved):

PR #7308, opened 2025-10-27
feat(clouddriver-lambda): Add configuration for tcpKeepAlive on the AWS Lambda client

manuelaguir:
For long running lambdas there is a network condition that mades clouddriver to lost the TCP connection after 350 seconds when invoking the function as explained on this blog post
https://aws.amazon.com/blogs/networking-and-content-delivery/implementing-long-running-tcp-connections-within-vpc-networking/
There are 2 parts for allowing this long running functions to be successfully invoked and retrieve the result:
Add sysctl configuration on the pod's security context 
securityContext:
    sysctls:
        - name: "net.ipv4.tcp_keepalive_time"
          value: "45"
        - name: "net.ipv4.tcp_keepalive_intvl"
          value: "45"
And enable TCP Keep alive on the lambda client configuration. This PR add that configuration.

aws:
     lambda:
        tcpKeepAlive: true

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -39,7 +39,7 @@ jobs:
       uses: actions/checkout@v5
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v3
+      uses: github/codeql-action/init@v4
       with:
         languages: ${{ matrix.language }}
         config-file: ${{ matrix.project }}/codeql-config.yml
@@ -58,4 +58,4 @@ jobs:
         ./gradlew :${{ matrix.project }}:assemble
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v3
+      uses: github/codeql-action/analyze@v4


clouddriver/clouddriver-lambda/src/test/java/com/netflix/spinnaker/clouddriver/lambda/deploy/ops/AwsLambdaClientConfigurationTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+  @Test
+  void tcpKeepAliveEnabled_setsClientFlag() {
+    LambdaServiceConfig props = new LambdaServiceConfig();
+    props.setTcpKeepAlive(true); // Enable
+
+    ClientConfiguration config = new ClientConfiguration();
+    config.setUseTcpKeepAlive(props.isTcpKeepAlive());
+
+    assertTrue(config.useTcpKeepAlive(), "TCP KeepAlive should be enabled");
```
jasonmcintosh: SO this isn't actually testing anything other than the AWS object.  E.g. the props isn't used or even tested as part of this.  Nominally what would be tested is that the generation of the provider after creation would have it's client configuration set.  OR at the least... this test should actually test the configuration is parsed & applied from the properties.  There's another recent PR that validates some of this configuration.  I think the raw CODE is fine with the property & provider generation.

manuelaguir: Removed this class and add tests into AbstractLambdaAtomicOpeation

PR #7309, opened 2025-10-30
perf(sql): Simplify sql query in retrievePipelineExecutionDetailsForApplication

dbyron-sf:
Since the query already specifies ids, the additional where condition for application is redundant and can be removed.

Before the change, the query uses pipeline_application_idx.  After it uses the primary key (id).

While we're at it, switch to use the read pool since replication lag is acceptable here. It's only used by gate / the UI to retrieve pipelines to display.


jasonmcintosh (approved):
Absolutely approved.  I'd not looked if that index used elsewhere - wondering if it could be dropped?  We had recent issues which LOOKED index update related, but our refId changes fixed that so didn't dive deep.

dbyron-sf:
> Absolutely approved. I'd not looked if that index used elsewhere - wondering if it could be dropped? We had recent issues which LOOKED index update related, but our refId changes fixed that so didn't dive deep.

Yeah, we still have a backlog item to review the indices, but haven't gotten to it.

PR #7312, opened 2025-11-03
feat(docker): Start publishing to GHCR as well as GAR for future removal of GAR for docker

jasonmcintosh:
To note:
* In past, I had to "pre-create" the container.  Docs state you don't HAVE to.  We WILL have to change visibility post publish to enable access going forward.  
* Added a few more open container labels on license just to be more "clear"
* HAVE NOT touched halyard or other systems.  This at this point is JUST publishing the images to GHCR.

### DIFF
--- a/orca/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
+++ b/orca/orca-sql/src/main/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepository.kt
@@ -562,7 +562,15 @@ class SqlExecutionRepository(
       val baseQuery = jooq.select(field("config_id"), field("id"))
         .from(table)
         .where(baseQueryPredicate)
-         .orderBy(field("config_id"))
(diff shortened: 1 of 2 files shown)


.github/workflows/clean-untagged-images.yml
```diff
@@ -7,6 +7,33 @@ on:
...
+        with:
+          registry: us-docker.pkg.dev
+        username: ${{ github.actor }}
+        password: ${{ secrets.GITHUB_TOKEN }}
+
+      # delete untaggd images older than 48 hours
+      - uses: 'docker://us-docker.pkg.dev/gcr-cleaner/gcr-cleaner/gcr-cleaner-cli'
+        with:
+          args: >-
+            -repo=ghcr.io/spinnaker/clouddriver
```
dbyron-sf, line 25: alphabetical order?  kayenta?

jasonmcintosh: Good call :) I figured I'd missed something... this is "dry run" on the GAR and not seen it show much, so not sure this is needed.  Will add this & sort

PR #7314, opened 2025-11-03
perf(gate/web): increase performance of the GET /{application}/pipelineConfigs/{pipelineName:.+} endpoint

dbyron-sf:
by changing it to query front50 for a specific pipeline instead of all pipelines in an application, and then filtering.

### DIFF
--- a/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/Front50Service.java
+++ b/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/Front50Service.java
@@ -47,7 +47,7 @@ public interface Front50Service {
   @GET("pipelines")
   Call<List<Map>> getAllPipelineConfigs();
 
-  @GET("/pipelines/{app}")
+  @GET("pipelines/{app}")
   Call<List<Map<String, Object>>> getPipelineConfigsForApplication(
       @Path("app") String app,
       @Query("pipelineNameFilter") String pipelineNameFilter,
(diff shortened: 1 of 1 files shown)


gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/Front50Service.java
```diff
@@ -47,8 +47,8 @@ Call<List<Map>> getApplicationHistory(
   @GET("pipelines")
   Call<List<Map>> getAllPipelineConfigs();
 
-  @GET("pipelines/{app}")
-  Call<List<Map>> getPipelineConfigsForApplication(
+  @GET("/pipelines/{app}")
```
jasonmcintosh: This shouldn't impact but this just bit me other day. "/" vs "" as had missed the prefix in a pR and it lead to a conflict.  :) JUST a "callout" - what I tested, it had no impact.    Absolutely like the more specific reference - also fixed a weird object mapper bug I was hitting!

jasonmcintosh (dismissed):

gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/Front50Service.java
dbyron-sf: Good catch...https://github.com/spinnaker/spinnaker/pull/7159 removed all the leading slashes, so lemme leave it that way.

dbyron-sf: done in https://github.com/spinnaker/spinnaker/pull/7314/commits/18c32d5f69c687c798f607ee84bcbefde202b1a9

jasonmcintosh (approved):

PR #7319, opened 2025-11-04
perf(executions): add pipelineLimit query parameter to gate's GET /applications/{application}/pipelines and GET /applications/{application}/pipelineConfigs endpoints

dbyron-sf:
Pass through the pipelineLimit query param to front50 when fetching pipelineConfigs. This allows the UI to control how many configs are returned in an attempt to optimize the executions page for larger applications.

Pass through the pipelineLimit query param to orca when fetching pipeline executions.  This allows the UI to control how many executions are returned when displaying a list of executions per application.

### DIFF
--- a/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/OrcaService.java
+++ b/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/OrcaService.java
@@ -35,7 +35,8 @@ public interface OrcaService {
       @Query("limit") Integer limit,
       @Query("statuses") String statuses,
       @Query("expand") Boolean expand,
-      @Query("pipelineNameFilter") String pipelineNameFilter);
+      @Query("pipelineNameFilter") String pipelineNameFilter,
+      @Query("pipelineLimit") Integer pipelineLimit);
 
   @Headers("Accept: application/json")
   @GET("projects/{projectId}/pipelines")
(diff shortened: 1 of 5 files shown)


front50/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -249,13 +252,20 @@ private List<Pipeline> sortPipelines(List<Pipeline> pipelines) {
...
 
+    if ((pipelineLimit != null) && (pipelineLimit < pipelines.size())) {
+      pipelines = pipelines.subList(0, pipelineLimit);
+    }
+
+    List<Pipeline> copiedPipelines = new ArrayList<>();
     int i = 0;
     for (Pipeline p : pipelines) {
-      p.setIndex(i);
+      Pipeline copied = p.clone();
```
jasonmcintosh, line 262: Ok was originally a bit "what's going on here" but... looking and it wasn't exactly obvious, but this seems like you're doing this to prevent the original pipelineDAO (which contains cached results) from getting modified, by cloing it and returning that new list of cloned objects which have their roder set.  A question that comes to mind is that... the previous system did a sort operation on the cache that this now removes.  I'm wondering how much this is REALLY needed (or if this setIndex should continue to be called always & just return the sub list?).  I'm not against this per se, just... trying to think about the potential impacts both good/bad.

dbyron-sf: I suppose there are other ways to fix this.  If I remember right there is a bunch of pretty unsavory stuff going on with setIndex, but it's all pretty vague at the moment...but it does seem like multiple calls to retrieve pipelines could (by specifying different criteria), end up returning different pipelines.  And then does it really make sense to change the cache each time (either by setting the index, or sorting)?  My reflex is no.

jasonmcintosh: yeah... I'm ... kinda fuzzy myself.  I'm inclined to merge right now & move on.   This is one of those areas that makes me uncomfortable, but not necessarily because of this PR but really legacy stuff.

PR #7330, opened 2025-11-05
fix(echo): Fix rest event config to work with splunk style URLs or those without a trailing slash on the URL

jasonmcintosh:


echo/echo-rest/src/test/groovy/com/netflix/spinnaker/echo/config/RestConfigSpec.groovy
```diff
@@ -118,4 +118,21 @@ class RestConfigSpec extends Specification {
...
+    wireMockServer.stubFor(WireMock.post(WireMock.urlEqualTo("/hec/services/collector/event"))
+      .willReturn(WireMock.aResponse()
+        .withStatus(200).withBody("{\"confirmed\"}")))
+    RestProperties.RestEndpointConfiguration endpoint = new RestProperties.RestEndpointConfiguration(
+      url: actualUrl)
+    RestUrls restUrls = configureRestServices(endpoint, EmptyHeadersFile)
+
+
+    then:
+      restUrls.getServices().get(0).getClient().recordEvent(actualUrl, [:]).execute().body().string().contains("confirmed")
```
dbyron-sf, line 138: Can you add a verify call to show what wireMockServer endpoints actually get hit?

jasonmcintosh: It's already there as that's validating the return of the mock server in the response body but... an extra check incoming to also check that the server was called not just the response of the server.

jasonmcintosh:
@Mergifyio backport release-2025.2.x release-2025.1.x

PR #7335, opened 2025-11-07
fix(spin): Fix spin cli docker publishing to GHCR

jasonmcintosh:
Spin was ONLY published on release and used a custom set of passed logins vs. rosco which passed to the build then to docker and thus got auth secrets.  Spin didn't pass the GHCR auth secret (GITHUB_TOKEN).  This should fix this.


jasonmcintosh:
@Mergifyio backport release-2025.3.x

PR #7337, opened 2025-11-09
Add GitHub App Authentication Support for Team Membership

nandak522:
This PR adds GitHub App authentication support to Fiat service, modernizing authentication from individual GitHub user account's Personal Access Tokens (PATs) to GitHub App tokens. This enhancement provides better rate limits ([15,000/hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-authenticated-users) vs [5,000/hour](https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28#primary-rate-limit-for-authenticated-users)), enhanced security with fine-grained permissions, and maintains full backward compatibility with existing PAT configurations.

#### Key Features
- [x] Dual Authentication Support - Automatically uses GitHub App when configured, falls back to PAT
- [x] Automatic Token Management - Installation tokens cached and refreshed automatically
- [x] Configuration Validation - Comprehensive validation at startup with clear error messages
- [x] Enhanced Logging - Informative logs showing authentication method, benefits, and recommendations
- [x] Zero Breaking Changes - Complete backward compatibility with existing deployments
- [x] Production Ready - Robust error handling and comprehensive test coverage

#### Configuration
GitHub App Configuration (Recommended) in `halconfig` file
```
auth:
  group-membership:
    service: github
    github:
      baseUrl: https://api.github.com
      organization: your-org
      authMethod: AUTO  # Default - prefers GitHub App if configured, GITHUB_APP/PAT are the other methods
   
      # GitHub App (preferred)
      appId: "12345"
      privateKeyPath: "/path/to/private-key.pem"
      installationId: "789012"

      # PAT (fallback during migration)
      accessToken: "ghp_xxxxxxxxxxxx"
```

#### Migration Strategy:

1. Start with both PAT and GitHub App configured
2. `authMethod: AUTO` automatically uses GitHub App (higher priority)
3. PAT serves as fallback during testing
4. Remove PAT after successful validation

### Bonus 1: kork-github Module Addition
A new reusable module `kork-github` has been added to centralize GitHub integration logic across Spinnaker services. Housed in `kork/kork-github/`, this provides a unified GitHub authentication and client factory for use across Fiat, Clouddriver, Igor, Echo, Orca, and other services.

#### Architecture
```
kork-github (Reusable Core)
├── GitHubAppAuthenticator    - JWT generation & token management
├── GitHubClientFactory        - Factory for creating authenticated clients
└── Dependencies
    ├── org.kohsuke:github-api:1.321 (hub4j)
    ├── io.jsonwebtoken (JWT generation)
    └── org.bouncycastle (PEM parsing)
```

#### Before (Retrofit-based):
```
fiat-github/
├── GitHubConfig.java                (manual Retrofit setup)
├── GithubTeamsUserRolesProvider.java (manual pagination)
├── client/
│   ├── GitHubClient.java             (Retrofit interface)
│   ├── GitHubAppAuthService.java     (JWT + tokens)
│   └── GitHubAppRequestInterceptor.java (auth injection)
└── model/
    ├── Member.java                   (custom model)
    └── Team.java                     (custom model)

Dependencies: 8 external
```
#### After (hub4j-based):
```
kork-github/ (NEW - Reusable!)
├── GitHubAppAuthenticator.java      (JWT + tokens)
└── GitHubClientFactory.java         (factory methods)

fiat-github/ (SIMPLIFIED!)
├── GitHubConfig.java                (simple factory calls)
└── GithubTeamsUserRolesProvider.java (automatic pagination)

Dependencies: 1 external (kork-github, which includes hub4j)
```

<!--We prefer small, well tested pull requests.

* Follow the commit message conventions [found here](https://spinnaker.io/community/contributing/submitting/).

Note that we are unlikely to accept pull requests that add features without prior discussion. The best way to propose a feature is to open an issue first and discuss your ideas there before implementing them./>


fiat/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/client/GitHubAppRequestInterceptor.java
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2025 Netflix, Inc.
```
jasonmcintosh: DON'T think y'all are netflix so probably should change this to the org doing the contributions ;)

fiat/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/client/GitHubAppAuthService.java
```diff
@@ -0,0 +1,179 @@
...
+      return true;
+    }
+  }
+
+  private PrivateKey loadPrivateKey(String privateKeyPath) {
+    try {
+      String content = new String(Files.readAllBytes(Paths.get(privateKeyPath)));
+
+      // Remove PEM headers and footers, and whitespace
+      String privateKeyPEM =
```
jasonmcintosh: Technically this should work, but... there's a few other alternative libraries that make loading private keys in PEM format much simpler/easier.  Look up bouncy castle - a quick google search gives some decent code via AI generation on using it's PemParser class.  There's some alternatives as well in Spring (aka https://docs.spring.io/spring-boot/3.3/api/java/org/springframework/boot/ssl/pem/PemContent.html which PRETTY sure exists in older versions).

fiat/fiat-github/src/test/java/com/netflix/spinnaker/fiat/config/GitHubConfigTest.java
```diff
@@ -0,0 +1,287 @@
...
+  @BeforeEach
+  void setUp(@TempDir Path tempDir) throws Exception {
+    gitHubConfig = new GitHubConfig();
+    gitHubProperties = new GitHubProperties();
+    gitHubProperties.setBaseUrl("https://api.github.com");
+    gitHubProperties.setOrganization("test-org");
+    gitHubConfig.setGitHubProperties(gitHubProperties);
+
+    // Create test private key
+    KeyPairGenerator keyGen = KeyPairGenerator.getInstance("RSA");
```
jasonmcintosh: So this would work but I kinda wonder about just generating a VERY long lived key instead of creating the key every test.  This would reduce a fair bit of complexity.

nandak522: Even for testing purposes, committing a fake private key to the codebase is not optimal I guess. (Infact in our company, such commits are straight up rejected by our Security team.) Would you absolutely recommend it @jasonmcintosh ?

christosarvanitis:
@nandak522 Thank you so much for this and im excited you decided to tackle this! Some thoughts: 
1. Have you considered using the library https://hub4j.github.io/github-api/? From previous experience while developing our proprietary Github Integration plugin I found this library very well maintained/tested and containing all the necessary methods for making everything easier. Have a look i think it will fit in this PR very nicely. 
2. Extracting the Github App Auth flow into Kork will give us the basis for leveraging Github Apps across Spinnaker - maybe not just for the Fiat Oauth2 but also for Clouddriver github provider or new features like integrating the CI with Github Actions etc

fiat/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/client/GitHubAppRequestInterceptor.java
nandak522: Handled in https://github.com/spinnaker/spinnaker/pull/7337/commits/b09421a6926ca560d684641a487ca0f28b4506cf.

fiat/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/client/GitHubAppAuthService.java
nandak522: Handled in https://github.com/spinnaker/spinnaker/pull/7337/commits/2018566cba47b4326864cae462c405bb6e751bdf

nandak522:
@christosarvanitis 
1. To your point on adopting hub4j/github-api, I can see that `org.kohsuke.github.GitHub` could become a replacement for existing `com.netflix.spinnaker.fiat.roles.github.client.GitHubClient`, similarly `org.kohsuke.github.GHUser` for `com.netflix.spinnaker.fiat.roles.github.model.Member` and `org.kohsuke.github.GHTeam` for `com.netflix.spinnaker.fiat.roles.github.model.Team`. Do you recommend updating these references to the hub4j equivalents in this PR ? It would be a fairly wide‑ranging change in terms of impact.
2. I loved your suggestion on moving GitHub app code into Kork so other modules can reuse it. Do you think it's better to couple that change with this PR or handle it as a follow‑up refactor ? I’m trying to avoid a big‑bang PR, but I'm happy to adjust based on your inputs.

christosarvanitis:
> 1. To your point on adopting hub4j/github-api, I can see that `org.kohsuke.github.GitHub` could become a replacement for existing `com.netflix.spinnaker.fiat.roles.github.client.GitHubClient`, similarly `org.kohsuke.github.GHUser` for `com.netflix.spinnaker.fiat.roles.github.model.Member` and `org.kohsuke.github.GHTeam` for `com.netflix.spinnaker.fiat.roles.github.model.Team`. Do you recommend updating these references to the hub4j equivalents in this PR ? It would be a fairly wide‑ranging change in terms of impact

That is a good point and i agree with you that we should stick to the bare minimum model of `com.netflix.spinnaker.fiat.roles.github.model`. It will still work for the scope of this change. 

>  I can see that org.kohsuke.github.GitHub could become a replacement for existing com.netflix.spinnaker.fiat.roles.github.client.GitHubClient

It gets interesting here though since the Spinnaker interface is oriented towards a Github Org (see org.kohsuke.github. GHOrganization). For the initial work i would recommend avoiding it to be honest since as you mentioned small incremental changes are preferred.  

> I loved your suggestion on moving GitHub app code into Kork so other modules can reuse it. Do you think it's better to couple that change with this PR or handle it as a follow‑up refactor ? I’m trying to avoid a big‑bang PR, but I'm happy to adjust based on your inputs

For this PR if other's agree as well I see it as a good fit for kork. For the GithubAppAuthenticator and clientFactory piece definitely. I would love to hear the thoughts of others in the community though :)

nandak522:
@christosarvanitis, I gave a shot at refactoring based on your comments. Please take a look and let me know.

nandak522:
Hi @jasonmcintosh , @christosarvanitis, Just following up on my replies to your comments.

fiat/fiat-github/src/test/java/com/netflix/spinnaker/fiat/config/GitHubConfigTest.java
jasonmcintosh: *ponder* I've generated this in the past, it's not END Of the world ;) Just simplifies that you're testing key decoding vs. key creation.  I'd be more ok if we were testing standard key handling in GENERIC.  This isn't a "no" per se, but... more of "hrmm" moment.

jasonmcintosh:
Will try to take a look more this week ;) I know christos was out this week from other conversations so give him some time.  VERY much appreciate the PR and work - this has come up a LOT in clouddriver-artifacts and think this'll be a great path to enhancing that flow as well!

fiat/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -173,193 +187,374 @@ private void initializeTeamMembershipCache() {
...
+          core.getLimit(),
+          core.getResetDate());
+
+      // Warn if rate limit is getting low
+      if (core.getRemaining() < core.getLimit() * 0.1) {
+        log.warn(
+            "GitHub API rate limit is running low: {}/{} remaining ({}%). Resets at {}",
+            core.getRemaining(),
+            core.getLimit(),
+            (core.getRemaining() * 100 / core.getLimit()),
```
jasonmcintosh, line 434: I'm going to hopefully assume "limit" is never 0 - but... i do wonder about GH Enterprise on premise rate limits possibly being 0?  AKA divide by zero bug potential here.  Not looked if this API always returns a higher number than 0? It's caught down below.. just an interesting error possibility.

fiat/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
```diff
@@ -173,193 +187,374 @@ private void initializeTeamMembershipCache() {
...
+      // For authentication/authorization errors (401/403), throw exception instead of returning
+      // empty set
+      // This allows Guava cache to keep using stale cached values until issue is resolved
+      if (message.contains("401") || message.contains("403") || message.contains("rate limit")) {
+        throw new RuntimeException(
+            "Critical GitHub API error - using cached values if available", e);
+      }
+
+      // For other transient errors, return empty set
+      return Collections.emptySet();
```
jasonmcintosh, line 392: Given this is ONLY used in the initialize cache method call that I can see... returning an empty set should be fine :)

jasonmcintosh: Hrmm, though WILL note... looking... i think previously we'd throw an exception on this vs. just an empty list.  I'm not totally sure that it's really a concern here, just... definitely a difference.

fiat/fiat-github/src/test/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProviderTest.java
```diff
@@ -16,66 +16,684 @@
...
+  @Mock private PagedIterable<GHUser> team1Members;
+
+  @Mock private PagedIterable<GHUser> team2Members;
+
+  private GithubTeamsUserRolesProvider provider;
+  private GitHubProperties gitHubProperties;
 
   @BeforeEach
-  void setUp() {
+  void setUp() throws IOException {
```
jasonmcintosh, line 59: So the ONE disadvantage RIGHT NOW... is this test no longer does REAL API calls against a mock of the github APIs, but does mock of the github client.  Given the library change to use a real GitHub client library, I'm ok with testing that the client is setup correctly.  SO i don't consider this a "negative" but a callout.

kork/kork-github/kork-github.gradle
```diff
@@ -0,0 +1,39 @@
...
+  // hub4j/github-api for GitHub interactions
+  api "org.kohsuke:github-api:1.321"
+
+  // JWT for GitHub App authentication
+  implementation "io.jsonwebtoken:jjwt-api:0.11.5"
+  implementation "io.jsonwebtoken:jjwt-impl:0.11.5"
+  implementation "io.jsonwebtoken:jjwt-jackson:0.11.5"
+
+  // Bouncy Castle for PEM parsing
+  implementation "org.bouncycastle:bcpkix-jdk15on:1.70"
```
jasonmcintosh, line 18: NIT: https://github.com/spinnaker/spinnaker/blob/main/kork/spinnaker-dependencies/spinnaker-dependencies.gradle#L12 version is defined here, shouldn't specify 1.70 - it SHOULD replace via the platform platform definition up above.  Still better to NOT explicitly set it IMHO...

kork/kork-github/src/main/java/com/netflix/spinnaker/kork/github/GitHubAppAuthenticator.java
```diff
@@ -0,0 +1,198 @@
...
+   * Returns an authenticated GitHub client using installation token. Handles token caching and
+   * automatic refresh.
+   *
+   * @return Authenticated GitHub client
+   * @throws IOException if authentication fails
+   */
+  public GitHub getAuthenticatedClient() throws IOException {
+    String token = getInstallationToken();
+
+    return new GitHubBuilder().withEndpoint(baseUrl).withAppInstallationToken(token).build();
```
jasonmcintosh, line 93: So this gets a cached token's token if it's not expired.  I'm wondering and don't know this SDK AS well... will this "token" passed here ever cause the GitHub app to expire?  I'd ASSUME the GitHub client this returns will be "smart" enough to refresh tokens as needed, but just want to confirm (without digging into the GH Client code being used)

kork/kork-github/src/test/java/com/netflix/spinnaker/kork/github/GitHubAppAuthenticatorTest.java
```diff
@@ -0,0 +1,224 @@
...
+
+    // When & Then - should load successfully
+    assertNotNull(authenticator);
+  }
+
+  @Test
+  void shouldSupportPKCS1Format(@TempDir Path tempDir) throws IOException {
+    // Given - PKCS#1 format (BEGIN RSA PRIVATE KEY)
+    String pkcs1Key =
+        "-----BEGIN RSA PRIVATE KEY-----\n"
```
jasonmcintosh, line 127: NIT:  And definitely not a big deal but should be able to use multiline strings with java 17 vs. the newline and quote sorta thing here.

fiat/fiat-github/src/test/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProviderTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 OpsMx, Inc.
+ * Copyright 2025 Netflix, Inc.
```
jasonmcintosh, line 2: To note, since this is basically a rewrite of these tests... Netflix is probably the WRONG attribution vs. RazorPay.

fiat/fiat-github/src/main/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProvider.java
nandak522: This is handled in https://github.com/spinnaker/spinnaker/pull/7353/commits/dd1330c6b37bded9527a1330e8abea5326d09a7e

fiat/fiat-github/src/test/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProviderTest.java
nandak522: Sure, added an integration test covering this. It's here: https://github.com/spinnaker/spinnaker/pull/7353/commits/9c909e0299869a5edf7add7338030cc1dcfdee52

kork/kork-github/kork-github.gradle
nandak522: Sure, addressed it in https://github.com/spinnaker/spinnaker/pull/7353/commits/3fbfed72c36ebed06821a7b4e342c9fd21d9bb22

kork/kork-github/src/main/java/com/netflix/spinnaker/kork/github/GitHubAppAuthenticator.java
nandak522: Brilliant catch. Addressed it here: https://github.com/spinnaker/spinnaker/pull/7353/commits/240f2a37228be3c5faf26de01c7105f9a3295b19

kork/kork-github/src/test/java/com/netflix/spinnaker/kork/github/GitHubAppAuthenticatorTest.java
nandak522: Can fix it. But currently waiting for our security team to approve this, as they have to make an exception to allow commiting RSA private keys, one more time.

fiat/fiat-github/src/test/java/com/netflix/spinnaker/fiat/roles/github/GithubTeamsUserRolesProviderTest.java
nandak522: This was a typo, fixed it in https://github.com/spinnaker/spinnaker/pull/7353/commits/b2cc4e93f08857b987563e44243d3425d1823e65

kork/kork-github/src/test/java/com/netflix/spinnaker/kork/github/GitHubAppAuthenticatorTest.java
nandak522: Addressed in https://github.com/spinnaker/spinnaker/pull/7353/commits/0202e0378432621582d1999312252d07859d2fe6.

PR #7340, opened 2025-11-14
fix(web): provide json error responses in more cases

dbyron-sf:
Previously, SpinnakerTomcatErrorValve was only used with header authentication. Now it's used all the time. And, it generates a json response when there's no underlying throwable, which it didn't before.


jasonmcintosh (approved):

jasonmcintosh:
Noice! :)

jasonmcintosh:
NOT sure it's a "fix" but ;) I'll go with it!

dbyron-sf:
> NOT sure it's a "fix" but ;) I'll go with it!

Yeah, fair enough.  It's probably a feature...

PR #7359, opened 2025-12-05
chore(clouddriver): log unhandled exceptions with a stack trace during HTTP URL Restrictions check

igcherkaev:
When something is missing or unfilled in configuration for an artifact the error is then obscured with a generic "URI not valid: " in http body between orca/rosco and clouddriver, this PR fixes it by providing a full exception message and stack trace.


igcherkaev:
Hmm, not sure what happens :)

PR #7360, opened 2025-12-05
chore(clouddriver): log unhandled exceptions with a stack trace during HTTP URL Restrictions check

igcherkaev:
When something is missing or unfilled in configuration for an artifact the error is then obscured with a generic "URI not valid: " in http body between orca/rosco and clouddriver, this PR fixes it by providing a full exception message and stack trace.

### DIFF
--- a/clouddriver/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/HttpUrlRestrictions.java
+++ b/clouddriver/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/HttpUrlRestrictions.java
@@ -31,6 +31,7 @@ import lombok.AllArgsConstructor;
 import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.StringUtils;
 import org.springframework.security.web.util.matcher.IpAddressMatcher;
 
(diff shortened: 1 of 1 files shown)


jasonmcintosh (approved):

clouddriver/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/HttpUrlRestrictions.java
```diff
@@ -223,6 +225,7 @@ public URI validateURI(URI url) throws IllegalArgumentException {
     } catch (IllegalArgumentException iae) {
       throw iae;
     } catch (Exception ex) {
+      log.error("HttpUrlRestrictions Exception: {}", ex.getMessage(), ex);
```
dbyron-sf, line 228: I guess I'm too late, but I feel like we're now in a weird in-between stage.  We could show more detail to the user by including `ex.getMessage()` in the IllegalArgumentException message below.

Or, we could decide that those details are potentially sensitive and keep them hidden, but then don't we want/need the url in the log message?

igcherkaev: Maybe. The url is visible in Deck, but with a cryptic `"URI not valid: " + url` message. What it really means is "Unhandled exception with the given URL". And whether sending the whole stack trace back is a good idea or not I'd leave it to you guys to decide. There are multiple assumptions in that restrictions check function, like, several configuration parameters are provided and not null, or that all network address resolutions will succeed, etc... Either way, the end goal is to make it easier for spinnaker operators figure out what's missing or broken or misconfigured. Having a stack trace is very helpful.

PR #7365, opened 2025-12-11
fix(fiat): properly backoff groups on retry

perek:
Currently the retry logic on the google groups provider is incorrectly per try which means that it doesn't actually retry properly, this pr fixes this and adds a test which failed originally and now passes

### DIFF
--- a/halyard/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringConfig.java
+++ b/halyard/halyard-deploy/src/main/java/com/netflix/spinnaker/halyard/deploy/spinnaker/v1/profile/SpringConfig.java
@@ -85,8 +85,7 @@ class SpringConfig {
 
     if (oauth2.getClient().getPreEstablishedRedirectUri() != null
         && !oauth2.getClient().getPreEstablishedRedirectUri().isEmpty()) {
-      registration.put(
-          "redirect-uri", "\"" + oauth2.getClient().getPreEstablishedRedirectUri() + "\"");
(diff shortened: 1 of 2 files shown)


fiat/fiat-google-groups/src/test/java/com/netflix/spinnaker/fiat/roles/google/BatchRequestBackoffTest.java
```diff
@@ -0,0 +1,216 @@
...
+import java.util.concurrent.atomic.AtomicInteger;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+/**
+ * Test to verify that batch request retry logic works correctly with exponential backoff. This
+ * tests the behavior of GoogleDirectoryUserRolesProvider.multiLoadRoles() when encountering rate
+ * limiting (403) errors from Google's API.
+ */
+public class BatchRequestBackoffTest {
```
jasonmcintosh, line 46: Overall i think this tests the logic effectively.  The ONE challenge ... and trying to think of the best way to do this is that it does NOT actually test that the "GoogleDirectoryUserRolesProvider" class DOES the batch operations.  I'm SORELY tempted to include this anyway since tests ARE passing (minus a non related issue) AND this fixes the crux of the issue.  Let me see if I can't tweak this SLIGHTLY to make sure that the actual provider class DOES "invoke" this correctly?

jasonmcintosh:
Fixed in https://github.com/spinnaker/spinnaker/pull/7385 so closing this PR at this time.

PR #7371, opened 2025-12-18
fix(deck): fix artifact not bound error on deck

qunayu:
This is something we ran into since the June 2023 release - we've ran it on our fork for a year and a half now with no issues.

This comes up when:
1. We have processes that changed the expectedArtifact IDs often
2. In deck, pipelines configs are cached and not refreshed upon a simple refresh
3. When people run their pipelines manually, the old expectedArtifactIds are used to trigger the pipeline, which fails to bind to the new artifact ids because of [this line here](https://github.com/spinnaker/spinnaker/blob/main/orca/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/ArtifactUtils.java#L207), which checks that the artifact ids match.

The solution here is to simply re-fetch the pipeline from backend instead of relying on caches in the frontend.


deck/packages/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -153,7 +154,10 @@ export class ExecutionGroup extends React.PureComponent<IExecutionGroupProps, IE
             .then((plan) => resolve(plan))
             .catch(() => resolve(config));
         } else {
-          resolve(config);
+          PipelineConfigService.getPipelinesForApplication(this.props.application.name).then((pipelines) => {
```
dbyron-sf: This retrieves all pipeline configs in an application, yes?  Possible to change this to only retrieve the pipeline config with the specific id the code cares about here?  That'll reduce the load a ton on front50 for big applications.

qunayu: Good call out - I've pushed a new commit that does exactly this with some tests. Tested manually and seemed to work fine. LMK what you think!

deck/packages/core/src/pipeline/executions/executionGroup/ExecutionGroup.tsx
```diff
@@ -153,7 +154,9 @@ export class ExecutionGroup extends React.PureComponent<IExecutionGroupProps, IE
             .then((plan) => resolve(plan))
             .catch(() => resolve(config));
         } else {
-          resolve(config);
+          PipelineConfigService.getPipelineForApplication(this.props.application.name, config.name)
+            .then((pipeline) => resolve(pipeline))
+            .catch(() => resolve(config));
```
dbyron-sf, line 159: Can you explain this fallback case a bit more?  As in, what's the difference between `resolve(pipeline)` and `resolve(config)`?  Is it really OK to do a "best effort" thing here?

dbyron-sf: I see that's what the code above does, but it still seems a little fishy.

qunayu: Yeah, `config` here would be the potentially old pipeline. I figure if this fails, we just revert to the current behaviour (which most of the time works, unless the expected artifact changed between last load and now). If we make this throw an error it would be a breaking change, which feels a bit too strict. What do you think?

dbyron-sf: Yeah, fair enough.

PR #7372, opened 2025-12-18
fix(orca-core): fix evaluateVariables stage when evaluation has already failed prior

qunayu:
Another small one that's been running in our fork for more than a year:

- Sometimes the evaluate variables stage doesn't evaluate the first variable correctly.
- I think this can only happen if a notification is added thatdoesn’t evaluate correctly 
- This fails because of a bad assumption in the stage, which assumes that number of evaluations failed before the evaluation starts for the stage is always 0
- Fixed by using actual summary’s number of failed evaluations first. 
- Added a test which fails without the code change.

### DIFF
--- a/orca/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesStageSpec.groovy
+++ b/orca/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesStageSpec.groovy
@@ -135,6 +135,13 @@ class EvaluateVariablesStageSpec extends Specification {
 
     then:
     shouldContinue == false
+
+    // If the processDefaultEntries failure isn't accounted for properly, the
+    // first variable can be treated as failed to evaluate, even when it's
+    // correct.  Explicitly assert that the first variable is correct so it's
+    // easier to see when something isn't working than with assert on all
+    // variables.
+    stage.context.variables[0].value == correctVars[0].value
     stage.context.variables == correctVars
     summary.totalEvaluated == 5
     summary.failureCount == 1


orca/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/EvaluateVariablesStage.java
```diff
@@ -77,7 +77,7 @@ public boolean processExpressions(
     EvaluateVariablesStageContext context = stage.mapTo(EvaluateVariablesStageContext.class);
     StageContext augmentedContext = contextParameterProcessor.buildExecutionContext(stage);
     Map<String, Object> varSourceToEval = new HashMap<>();
-    int lastFailedCount = 0;
+    int lastFailedCount = summary.getFailureCount();
```
dbyron-sf, line 80: I'm confused here.  If I'm following the code correctly, we get here via [StageExecution.withMergedContext](https://github.com/spinnaker/spinnaker/blob/cccc1fcd6db86a76d8481b09f80f4df14ddbf73b/orca/orca-queue/src/main/kotlin/com/netflix/spinnaker/orca/q/handler/ExpressionAware.kt#L50), which passes a newly constructed ExpressionEvaluationSummary to this method.  If that's true, then this code change doesn't actually make a real difference.  I mean, it does in the test because the test builds a summary object with a failure in it, but is this what actually happens when orca is running?

Plus also, does this change really help the code determine whether the first variable was successfully evaluated?  When it fails to evaluate, I figure summary.getFailureCount() is 1 after calling contextParameterProcessor.process.  And so initializing to 0 feels like the right way to detect that.  Changing to summary.getFailureCount() here would break that, wouldn't it?

I think the test below needs to have more detailed assertions about which expression failed...and also a new case added for when the summary starts with no failures.

qunayu: Hey David, thanks for reviewing this my PRs! 

Yes, a fresh `ExpressionEvaluationSummary` is passed in, however [line 74 of this file](https://github.com/spinnaker/spinnaker/pull/7372/files#diff-056d6fe5330510dd076152f98f93dae60d84d6ae81998431e1a3ef23ee67d212R74) can potentially modify that variable, e.g. in cases where processing of the notification fields fail evaluation, leading to the failure count being incremented there. Once the first variable fails to evaluated, the failure is added to the `summary` and so the count is incremented. So if evaluation processing failed when processing default entries, then regardless of whether the first variable has been evaluated correctly, this stage will think that evaluation failed.

Hopefully that makes sense. I've included a new test which actually fails evaluation at the notifications fields to demonstrate this (which is really what I should've included from the start). Let me know if you think I should include some more tests.

dbyron-sf: Aaah, ok.  I get it.  What do you think about a diff like this?  It helped me understand what was really wrong with the previous behavior:
```
diff --git a/orca/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesStageSpec.groovy b/orca/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesStageSpec.groovy
index c39e6def24..2d88560728 100644
--- a/orca/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesStageSpec.groovy
+++ b/orca/orca-core/src/test/groovy/com/netflix/spinnaker/orca/pipeline/tasks/EvaluateVariablesStageSpec.groovy
@@ -135,6 +135,13 @@ class EvaluateVariablesStageSpec extends Specification {
 
     then:
     shouldContinue == false
+
+    // If the processDefaultEntries failure isn't accounted for properly, the
+    // first variable can be treated as failed to evaluate, even when it's
+    // correct.  Explicitly assert that the first variable is correct so it's
+    // easier to see when something isn't working than with assert on all
+    // variables.
+    stage.context.variables[0].value == correctVars[0].value
     stage.context.variables == correctVars
     summary.totalEvaluated == 5
     summary.failureCount == 1
```

qunayu: Yeah, I like that! Thanks, applied!

PR #7375, opened 2025-12-19
chore(build): remove unused properties

mattgogerly:
Remove unused properties now that dependent builds are in use.


clouddriver/gradle.properties
```diff
@@ -1,7 +1,4 @@
-korkVersion=7.254.0
```
dbyron-sf, line 1: halyard/gradle.properties still mentions korkVersion, fiatVersion, spinnakerGradleVersion.

mattgogerly: Serves me right for not including halyard in our fork.

PR #7384, opened 2026-01-02
fix(build): produce production build of Deck

mattgogerly:
`NODE_ENV` is set to `production` for the `webpack` build but not for the `rollup` builds triggered by running `yarn modules`. This results in a partial development React build.

Set `NODE_ENV` when running `yarn modules` too so we get a full production build.

**Before**
image

**After**
image

### DIFF
--- a/deck/build.gradle
+++ b/deck/build.gradle
@@ -28,6 +28,9 @@ task modules(type: YarnTask) {
   dependsOn "yarn"
 
   yarnCommand = ["modules"]
+  environment = [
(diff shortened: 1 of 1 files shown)


dbyron-sf (approved):
Does it make sense to change e.g. [.github/actions/publish-npm/action.yml](https://github.com/spinnaker/spinnaker/blob/bbbfd9fae85340cc0d5963458e70c0b1ce94610d/.github/actions/publish-npm/action.yml#L51) too?

PR #7385, opened 2026-01-06
fix(gcp): Fiat batch queries on rate limit would randomly fail

jasonmcintosh:
GCP batch operations are... somewhat buggy on how they operate.  The batch would fail and not retry appropriately NOR rate limit appropriately.  This  Fixes how groups are pulled from GCP to really use rate limits AND retries appropriately at the expense of moving from batch logic, but ALSO then supports recursive lookups of groups on these AND handles pagination correctly.


fiat/fiat-google-groups/src/main/java/com/netflix/spinnaker/fiat/roles/google/GoogleDirectoryUserRolesProvider.java
```diff
@@ -332,6 +355,8 @@ public static class Config {
      */
     private RoleSource[] roleSources = new RoleSource[] {Config.RoleSource.NAME};
 
+    private int groupLookupConcurrency = 10;
```
perek, line 358: maybe should make this a property, not sure this is always a safe number?

jasonmcintosh: So this is technically a property that could be set via
```
auth:
  group-membership:
    google:
      groupLookupConcurrency: 5
```

PR #7388, opened 2026-01-09
fix(echo): actually increment trigger.errors.accessdenied counter

qunayu:
Very small PR to actually increment the `trigger.errors.accessdenied` counter in echo.


mattgogerly:
We actually just fixed this on our fork too, but you beat us to upstreaming it 😄 Thanks!

mattgogerly (approved):

PR #7052, opened 2025-04-25
Integrate gate oauth2

rahul-chekuri:
Oauth configuration properties are changed for new java way of dsl.

Old Config Properties:
```
security:
  authn:
    oauth2:
      enabled: true
      client:
        clientId: <client-id>
        clientSecret: <client-secret>
        accessTokenUri: https://www.googleapis.com/oauth2/v4/token
        userAuthorizationUri: https://accounts.google.com/o/oauth2/v2/auth
        scope: profile email
      userInfoRequirements:
        hd: <domain>
      resource:
        userInfoUri: https://www.googleapis.com/oauth2/v3/userinfo
      userInfoMapping:
        email: email
        firstName: given_name
        lastName: family_name
      provider: GOOGLE
```

New Config Properties:

Google:

```
spring:
  security:
    oauth2:
      client:
        registration:
          google:
            client-id: <client-id>
            client-secret: <client-secret>
            authorization-grant-type: authorization_code
            redirect-uri: "https://<your-domain>/login/oauth2/code/google"
            scope: profile,email,openid
            client-name: google
        provider:
          google:
            authorization-uri: https://accounts.google.com/o/oauth2/auth
            token-uri: https://oauth2.googleapis.com/token
            user-info-uri: https://www.googleapis.com/oauth2/v3/userinfo
            user-name-attribute: sub

```
Github:

```
spring:
  security:
    oauth2:
      client:
        registration:
          userInfoMapping:
            email: email
            firstName: ''
            lastName: name
            username: login
          github:
            client-id: <client-id>
            client-secret: <client-secret>
            authorization-grant-type: authorization_code
            redirect-uri: "https://<your-domain>/login/oauth2/code/github"
            scope: user,email
            client-name: github
        provider:
          github:
            authorization-uri: https://github.com/login/oauth/authorize
            token-uri: https://github.com/login/oauth/access_token
            user-info-uri: https://api.github.com/user
            user-name-attribute: login
```

PR #7392, opened 2026-01-13
fix(oauth2): re-implement ExternalAuthTokenFilter to fix spin-cli login

mattgogerly:
https://github.com/spinnaker/spinnaker/pull/7052 removed `ExternalAuthTokenFilter`, which allowed an arbitrary Bearer token to be provided and exchanged for a valid OAuth session. This functionality is [used](https://github.com/spinnaker/spinnaker/blob/4790896f658a8fec16a39773bdc19f4de66f6e70/spin/cmd/gateclient/client.go#L545) by `spin-cli` to login, and so `spin-cli` does not work when using OAuth post `2025.2`.

This PR re-implements the `ExternalAuthTokenFilter`, re-using the new `UserInfoServiceHelper.getSpinnakerOAuth2User()` to convert a bearer token to a `SpinnakerOAuth2User`.

Before:
```
$ spin account whoami
Caching oauth2 token.
{}
```

After
```
$ spin account whoami
Caching oauth2 token.
{
 "accountNonExpired": true,
 "accountNonLocked": true,
 "allowedAccounts": [
  ...
 ],
 "credentialsNonExpired": true,
 "enabled": true,
 "lastName": "Matt Gogerly",
 "roles": [
  ...
 ],
 "username": "mattgogerly"
}
```
Note that this adds two new configuration properties, with these defaults:
```
security:
  oauth2:
    external-auth-token-filter:
      connectTimeoutMs: 5000
      readTimeoutMs: 5000
```

### DIFF
--- a/gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/ExternalAuthTokenFilterConfigurationProperties.java
+++ b/gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/ExternalAuthTokenFilterConfigurationProperties.java
@@ -24,7 +24,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  *
  * <p>These properties control the HTTP client timeouts when calling the OAuth2 user-info endpoint.
  */
-@ConfigurationProperties("security.oauth2.external-auth-token-filter")
+@ConfigurationProperties(prefix = "security.oauth2.external-auth-token-filter")
 @Data
 public class ExternalAuthTokenFilterConfigurationProperties {
 
--- a/gate/gate-oauth2/src/test/java/com/netflix/spinnaker/gate/security/oauth2/OAuth2ConfigurationTest.java
+++ b/gate/gate-oauth2/src/test/java/com/netflix/spinnaker/gate/security/oauth2/OAuth2ConfigurationTest.java
@@ -113,11 +113,5 @@ class OAuth2ConfigurationTest {
     OAuthUserInfoServiceHelper oAuthUserInfoServiceHelper() {
       return mock(OAuthUserInfoServiceHelper.class);
     }
-
-    @Bean
-    ExternalAuthTokenFilterConfigurationProperties
-        externalAuthTokenFilterConfigurationProperties() {
-      return new ExternalAuthTokenFilterConfigurationProperties();
-    }
   }
 }


gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/ExternalAuthTokenFilter.java
```diff
@@ -0,0 +1,161 @@
...
+
+  private String extractBearerToken(HttpServletRequest request) {
+    String authHeader = request.getHeader(AUTHORIZATION_HEADER);
+    if (authHeader != null && authHeader.startsWith(BEARER_PREFIX)) {
+      return authHeader.substring(BEARER_PREFIX.length());
+    }
+    return null;
+  }
+
+  private OAuth2AuthenticationToken authenticateWithToken(String accessToken) {
```
dbyron-sf, line 148: javadoc please, here and elsewhere.

gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/OAuth2SsoConfig.java
```diff
@@ -43,10 +53,18 @@ public class OAuth2SsoConfig {
   @Autowired private SpinnakerOAuth2UserInfoService customOAuth2UserService;
   @Autowired private SpinnakerOIDCUserInfoService oidcUserInfoService;
   @Autowired private DefaultCookieSerializer defaultCookieSerializer;
+  @Autowired private ClientRegistrationRepository clientRegistrationRepository;
+  @Autowired private OAuthUserInfoServiceHelper userInfoServiceHelper;
 
   @Autowired
   private OAuth2AccessTokenResponseClient<OAuth2AuthorizationCodeGrantRequest> tokenResponseClient;
 
+  @Value("${spring.security.oauth2.external-auth-token-filter.connect-timeout-ms:5000}")
```
dbyron-sf: This isn't actually a spring property, is it?  If not, let's name it something specific to Spinnaker.  And if you have the stamina, please use a properties class instead of `@Value`.  Were there similar params before https://github.com/spinnaker/spinnaker/pull/7052?

mattgogerly: Updated. The properties didn't exist before.

gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/ExternalAuthTokenFilter.java
```diff
@@ -0,0 +1,205 @@
...
+   *
+   * <p>This method retrieves the client registration, fetches user information from the configured
+   * user-info endpoint, and creates an authenticated OAuth2AuthenticationToken.
+   *
+   * @param accessToken the OAuth2 access token to authenticate with
+   * @return an OAuth2AuthenticationToken if authentication succeeds, or {@code null} if
+   *     authentication fails due to missing configuration or invalid token
+   */
+  private OAuth2AuthenticationToken authenticateWithToken(String accessToken) {
+    ClientRegistration clientRegistration =
```
dbyron-sf, line 149: @mattgogerly I realize we've already merged this, but I've got a question...Could this code move to the constructor?  Do we really need to do it on every request?  I'm talking about all of:
```
    ClientRegistration clientRegistration =
        clientRegistrationRepository.findByRegistrationId(registrationId);

    if (clientRegistration == null) {
      log.warn("No client registration found for registrationId: {}", registrationId);
      return null;
    }

    String userInfoUri = clientRegistration.getProviderDetails().getUserInfoEndpoint().getUri();
    if (userInfoUri == null || userInfoUri.isEmpty()) {
      log.warn("No user-info-uri configured for registrationId: {}", registrationId);
      return null;
    }
```

PR #7397, opened 2026-01-20
chore(cloud): Disable non-maintained cloud providers

jasonmcintosh:
Proposing this.  We stop automatically including:
* Titus
* huaweicloud
* tencentcloud
* yandex
I'm open to other ideas - this is mostly to open the discussion on "should these be in the project".  TO NOTE:  Consul is used in the GCP cloud so I've NOT disabled that - it's the only place that's used today, but I haven't looked at how much it's used.


mattgogerly:
Could you remove these from Deck too? Should be a case of removing them from the build modules script and removing them from the app package.json.

jasonmcintosh:
Converting to draft for now.  Given some conversations may delay this a bit longer as seen some recent conversations that maybe one of these is actually in use.

jasonmcintosh:
Closing this for now, as it turns out some ARE used.  Not as commonly as anticipated, but still in use.

PR #7399, opened 2026-01-21
feat(cats): Concept for a pub/sub scheduler.

jasonmcintosh:
New system with deterministic ordered, controlled throughput for cache operations.
* Scheduling NEW agents is handled via a state machine & scheduler.
* Agent operations pull from the stream to handle requests.
* RESULTS are NOT sent over a queue but direct write by the clouddriver that got the agent execution request.

Plan is to move some of this to kork in a follow on PR that lets us abstract the different processing and loop handling and such in a generic manner with support for:
* Kafka
* RabbitMQ
* Redis Stream
* SQS/SNS
* GCP pub/sub
And then to move some other operations to this framework.


kork/kork-sql/kork-sql.gradle
```diff
@@ -13,7 +13,7 @@ dependencies {
   api "org.liquibase:liquibase-core"
   api "com.zaxxer:HikariCP"
 
-  implementation 'com.github.blagerweij:liquibase-sessionlock:1.5.0'
+  implementation 'com.github.blagerweij:liquibase-sessionlock:1.6.9'
```
jasonmcintosh, line 16: Though not STRICTLY related to the pub/sub - this fixed a log message on output on locks.

jasonmcintosh:
TODO: Fix for latest spring versions

clouddriver/cats/cats-pubsub/src/main/java/com/netflix/spinnaker/cats/pubsub/controllers/PubSubAdminController.java
```diff
@@ -0,0 +1,64 @@
...
+  @Autowired
+  public void setObjectMapper(ObjectMapper mapper) {
+    this.mapper = mapper.copy();
+    SimpleModule simpleModule = new SimpleModule("convertLongsToDates");
+    simpleModule.addSerializer(long.class, this);
+    this.mapper.registerModule(simpleModule);
+  }
+
+  @GetMapping("/agents")
+  public String getAgents() throws Exception {
```
jasonmcintosh, line 41: Arguably... this should be admin restricted as well AND probably exposed in gate vs. an internal endpoint.  IF there's interest I'll add a gate controller for this...

christosarvanitis (approved):
A good addition for a scheduling system

PR #7400, opened 2026-01-21
feat(gate): add API endpoint to force cancel execution

mattgogerly:
Adds an admin only endpoint to Gate to force cancel an execution, removing the need to port-forward or exec to Orca to hit Orca's endpoint.


jasonmcintosh:
FYI I had https://github.com/spinnaker/spinnaker/pull/7313 ;) that looks similar.

mattgogerly:
Superseded by https://github.com/spinnaker/spinnaker/pull/7313

PR #7422, opened 2026-01-26
fix(publishing): Fix image publishing for deck due to NPM trusted publishing

jasonmcintosh:


.github/workflows/deck-kayenta.yml
```diff
@@ -40,6 +40,7 @@ on:
 permissions:
   id-token: write
   contents: write
+  packages: write
```
dbyron-sf, line 43: What fails without this?  What's the error message?

jasonmcintosh: https://github.com/spinnaker/spinnaker/actions/runs/21362381976/job/61484827236
```
Error: buildx failed with: ERROR: failed to build: failed to solve: failed to push ghcr.io/spinnaker/deck:2025.3-2: denied: installation not allowed to Write organization package
```

jasonmcintosh: NPM set the permissions on identity which was needed to do oauth... but.. apparently that then removed the default write permissions on packages/etc.  Seems a bit annoying to me on GHA handling on that point.

jasonmcintosh:

PR #7431, opened 2026-01-28
fix(cats-sql): make SqlUnknownAgentCleanupAgent sharding-aware

mrusan:
### Problem

`SqlUnknownAgentCleanupAgent` was not sharding-aware, causing it to delete cache records belonging to agents handled by other pods in a sharded clouddriver deployment.

### Solution

Integrate the cleanup agent with the existing sharding infrastructure to ensure each pod only cleans up records for agents it is responsible for.

---

#### 1. SqlCachingPodsObserver — Expose sharding state

Add `getPodIndex()` and `getPodCount()` accessors to allow consumers to query current sharding topology.

#### 2. SqlUnknownAgentCleanupProperties — New configuration class

```yaml
sql:
  unknown-agent-cleanup-agent:
    enabled: false              # opt-in, disabled by default
    pollIntervalSeconds: 120
    timeoutSeconds: 60
    minRecordAgeSeconds: 300    # age guard before deletion
    deleteBatchSize: 100
    dryRun: false
    excludedDataTypes: []
```

#### 3. SqlUnknownAgentCleanupAgent — Core changes

- Consult `ShardingFilter` before deleting records to respect pod boundaries
- Add safety guards:
  - Skip cleanup if agent registry is empty (prevents mass deletion on startup race)
  - Skip cleanup if sharding enabled but `NoopShardingFilter` is injected (misconfiguration detection)
  - Skip cleanup if `SqlCachingPodsObserver` sharding state is not yet established
- Add `last_updated` age check to avoid deleting recently-written records
- Add `dryRun` mode for validation without deletion
- Add `excludedDataTypes` for operational flexibility
- Make poll/timeout intervals configurable

#### 4. SqlCacheConfiguration — Wire new dependencies

- Inject `ShardingFilter` and `shardingEnabled` flag into cleanup agent bean
- Conditional bean creation: `!sql.read-only && sql.unknown-agent-cleanup-agent.enabled`

---

### Backward Compatibility

- Agent is **disabled by default** — no behavior change for existing deployments
- Existing configuration properties unaffected
- No schema changes required

---

### Test Coverage

Added tests for:

- Sharding filter integration (preserves other pods' data)
- Safety guards (empty registry, uninitialized sharding state, noop filter with sharding enabled)
- Configuration properties (age guard, dry-run, exclusions, batching, poll/timeout overrides)

### DIFF
--- a/.github/workflows/spinnaker-release.yml
+++ b/.github/workflows/spinnaker-release.yml
@@ -13,6 +13,10 @@ on:
         required: true
       spinnaker-previous-version:
         description: Override the previous version used for changelog generation (picks the last patch version if unset)
+permissions:
+  id-token: write
+  contents: write
+  packages: write
 
 jobs:
   check-preconditions:
--- a/deck-kayenta/package.json
+++ b/deck-kayenta/package.json
@@ -6,10 +6,6 @@
   },
   "version": "2.6.0",
   "license": "Apache-2.0",
-  "repository": {
-    "type": "git",
-    "url": "https://github.com/spinnaker/deck-kayenta.git"
-  },
   "module": "build/dist/index.js",
   "typings": "build/dist/index.d.ts",
   "files": [


jasonmcintosh (approved):
I think I'm good with this PR.  Looks like this'll fix an outstanding issue.... we've had...

PR #7438, opened 2026-01-30
fix(ldap): publish firstName and lastName for users authenticated via LDAP

KrasimirAntonov:
When using LDAP for authentication in Spinnaker, the user attributes firstName and lastName are not available or published.

These attributes are available with other identity providers such as SAML or OAuth2. This PR clarifies that LDAP users do not have these fields populated, which can affect downstream systems or UI components that rely on full names.

No other providers are impacted by this behavior.

### DIFF
--- a/.github/workflows/spinnaker-release.yml
+++ b/.github/workflows/spinnaker-release.yml
@@ -94,17 +94,41 @@ jobs:
       version-override: ${{ needs.version.outputs.version }}
   deck:
     needs: version
-    uses: ./.github/workflows/deck.yml
-    secrets: inherit
-    with:
-      publish-debs: true
-      version-override: ${{ needs.version.outputs.version }}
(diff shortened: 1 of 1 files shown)


jasonmcintosh:
So we really need tests on these.  I'm not against adding these as fields pulled from the ldap response... but would like some example tests showing this ;) The givenName & sn look like part of the inetOrgPerson so I'm mostly ok with assuming they're there, but ... be good to confirm with some tests that we DO add these.

KrasimirAntonov:
@jasonmcintosh Added the requested coverage - we now have Spock specs that explicitly assert we pull `mail, givenName, and sn` from the LDAP context and map them onto the returned User (email, firstName, lastName).

jasonmcintosh (approved):
I am good with this ;) Thanks!

PR #7439, opened 2026-01-30
fix(deck): Preserve execution selection when editing form fields

mrusan:
# fix(deck): Preserve execution selection when editing form fields

## Problem

When manually executing a pipeline with a pipeline trigger, selecting a specific execution from the dropdown and then editing any other form field (e.g., Change Number) caused the selection to revert to the default (latest) execution.

## Root Cause

`PipelineTriggerTemplate.componentWillReceiveProps` compared `command` object references:

```typescript
if (nextProps.command !== this.props.command) {
  this.initialize();
}
```

Formik creates new `command` object references on every form field change. The reference comparison was always true, triggering `initialize()` on every keystroke, which refetched executions and reset the dropdown to the default selection.

## Solution

Compare the semantic value that matters: `trigger.pipeline` (the source pipeline ID).

```typescript
if (nextTrigger?.pipeline !== currentTrigger?.pipeline) {
  this.initialize(nextProps.command);
}
```

Re-initialization now occurs only when the user selects a different source pipeline, not when editing unrelated form fields.

Secondary fix: `initialize()` now accepts an optional `command` parameter to avoid using stale `this.props` when called from `componentWillReceiveProps`.

## Files Changed

- `PipelineTriggerTemplate.tsx` - Core fix
- `PipelineTriggerTemplate.spec.tsx` - New test suite

## Test Coverage

| Category | Tests |
|----------|-------|
| Component lifecycle | 4 |
| Selection preservation | 2 |
| Re-initialization behavior | 3 |
| Re-run scenarios | 1 |
| User interaction | 4 |
| Edge cases | 4 |
| Async behavior | 3 |
| State consistency | 3 |
| Formik integration | 2 |
| Rendering | 3 |

## Known Limitation (Out of Scope)

Race condition: if user switches pipelines while a request is in flight, a late-arriving response can overwrite the correct data. Documented in tests with TODO for future fix.


jasonmcintosh:
Though I'm... not great at UI, I think the AI here has done a decent job on this and identifying the issue.  I'm good with this though love to get @mattgogerly or someone who's more on the "UI" side to confirm.

mattgogerly (approved):

jasonmcintosh:
@mergifyio backport release-2025.3.x release-2025.4.x

PR #7453, opened 2026-02-03
fix(saml): Restore allowedAccounts

KrasimirAntonov:
**Description**

This fix restores support for allowedAccounts when using the SAML authentication provider.

The allowedAccounts behavior was unintentionally lost during the upgrade to version 1.35.0. In version 1.34.2, this feature still works as expected.
Other authentication providers were not affected and continue to handle allowedAccounts correctly.

With this change, the logic for handling allowedAccounts is unified across all authentication providers, ensuring consistent behavior regardless of the provider in use.

Tests have also been added to validate the correct behavior of allowedAccounts when using SAML.

### DIFF
--- a/gate/gate-ldap/src/main/groovy/com/netflix/spinnaker/gate/security/ldap/LdapSsoConfig.groovy
+++ b/gate/gate-ldap/src/main/groovy/com/netflix/spinnaker/gate/security/ldap/LdapSsoConfig.groovy
@@ -116,6 +116,8 @@ class LdapSsoConfig {
 
       return new User(username: username,
         email: ctx.getStringAttribute("mail"),
+        firstName: ctx.getStringAttribute("givenName"),
+        lastName: ctx.getStringAttribute("sn"),
         roles: roles,
         allowedAccounts: allowedAccountsSupport.filterAllowedAccounts(username, roles))
     }
--- a/gate/gate-ldap/src/test/groovy/com/netflix/spinnaker/gate/security/ldap/LdapUserContextMapperSpec.groovy
+++ b/gate/gate-ldap/src/test/groovy/com/netflix/spinnaker/gate/security/ldap/LdapUserContextMapperSpec.groovy
@@ -1,0 +1,83 @@
+package com.netflix.spinnaker.gate.security.ldap
+
+import com.netflix.spinnaker.gate.security.AllowedAccountsSupport
+import com.netflix.spinnaker.gate.services.PermissionService
+import com.netflix.spinnaker.security.User
+import org.springframework.ldap.core.DirContextAdapter
+import org.springframework.ldap.core.DirContextOperations
+import org.springframework.security.core.authority.SimpleGrantedAuthority
+import spock.lang.Specification
+
+class LdapUserContextMapperSpec extends Specification {
+
+  def permissionService = Mock(PermissionService)
+  def allowedAccountsSupport = Mock(AllowedAccountsSupport)
+
+  def subject = new LdapSsoConfig.LdapUserContextMapper(
+    permissionService: permissionService,
+    allowedAccountsSupport: allowedAccountsSupport
+  )
+
+  def "mapUserFromContext should map user details, sanitize roles, and call PermissionService and AllowedAccountsSupport"() {
+    given:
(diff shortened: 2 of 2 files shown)


jasonmcintosh:
To note... I wonder about this value in GENERAL.  AKA... this is a "Point-in-time" set of accounts, and it ignores new/created accounts post creation.  MOST pipelines don't use this field for example but get from a cache which is regularly updated on new requests from fiat for a user data... and fiat's user data is synced on new app/account creation.  

E.g. we can add this back, but I wonder... is this REALLY useful/valuable?  Possibly INITIALLY...

KrasimirAntonov:
This field is actually very useful in our CD flow.

We have an artifact promotion workflow from non-production environments to production, where we explicitly validate that the users who executed the pipeline have deployment permissions on the target cluster. At first glance, it might seem that restricting permissions at the application level would be sufficient, but in our use case that is not enough.

We operate in a large organization (115k+ users) and also work with external users who share groups. Because of this, we need to perform authorization checks at the account level per user, not just at the application level.

We have reviewed and experimented with alternative pipeline logic, but none of the available approaches fully cover this requirement. Restoring the accounts in the allowedAccounts field would directly address this gap and allow us to properly enforce these validations in our promotion workflows.

jasonmcintosh:
So might be worth a slack discussion.  NORMALLY you'd set access controls on the account itself as part of the spinnaker config.  E.g. this blocks the deploy to the group who has access to that cluster explicitly.  App permissions are for pipelines - but account permissions are used to handle this.  SO if someone "approved" and didn't have that permission, despite approving the deploy would fail.  

Note, for manual invocation the pipeline normally adds this as part of the execution and pulls it from fiat data - aka the auth flow shouldn't NEED to set this as fiat has access to this information as part of it's sync/login flows.  There's a flag in recent versions that removes this that could be added to turn this back on.  

I'm not AGAINST this PR, but would be worth making sure this is "the right thing"

KrasimirAntonov:
Regarding the first paragraph, we are aligned that permissions are configured and used exactly in that way.

For the second paragraph, I don’t understand why `allowedAccounts` is being populated for providers like [OAuth](https://github.com/spinnaker/spinnaker/blob/main/gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/OAuthUserInfoServiceHelper.java#L125) and [LDAP](https://github.com/spinnaker/spinnaker/blob/main/gate/gate-ldap/src/main/groovy/com/netflix/spinnaker/gate/security/ldap/LdapSsoConfig.groovy#L122), but not for SAML. We need the `allowedAccounts` value because we perform a validation like the following:

image<br><br>



`hasSpinnakerAccount`: ${#root.execution.authentication.allowedAccounts.contains(spinnakerAccount)}

The `allowedAccounts` list from version 1.34.2:
image

jasonmcintosh:
So at ONE POINT the goal is to remove "allowedAccounts" from the user identity ENTIRELY.  When you have more and more users, more and more accounts, this just doesn't scale.  The permissions checks still need to validate a user is in the account LIST, but... there's a higher risk of including this data on user LOOKUPS.  Further, this data has to be refreshed REGULARLY as part of sync operations.  I'll NOTE ... "root.execution.authentication.user" is probably not where I'd look to validate this.  There's a... "execution.allowedAccounts" which is optionally passed in.  

I'd be more interested honestly in a ... validation step that validates that the user is authorized & alerts/rejects with an appropriate error message.  To this, we can merge this... then maybe come up with a better approach than the pipeline just failing to run or failing on some of the stages.  Possibly a "isUserAuthorized" expression maybe.  I'd want to think about that one.

Frankly the whole auth system... really needs a "re-think".  I've had a FEW ideas on changing the auth flow, just not on the priority.  BUT as said I'm ok with merging this for now until we figure out a better path, would welcome some involvement on this LONG TERM.

jasonmcintosh (approved):
Approving - long term... like to redo this whole "allowedAccounts" approach in GENERAL & provide better feedback.... or expressions to validate the requests kinda things.

jasonmcintosh:
@mergifyio backport release-2025.4.x

PR #7454, opened 2026-02-03
fix(kayenta): Send Authorization header proactively for Prometheus-compatible APIs

rahulraj0504:
**## Issue Summary**

This PR fixes an issue where Kayenta fails to authenticate with Prometheus-compatible metrics providers (like Coralogix) that do not return HTTP 401 for missing authentication.

When configuring a Prometheus account with `bearerToken` authentication:
```
yaml
kayenta:
  prometheus:
    accounts:
      - name: coralogix_prom
        endpoint:
          baseUrl: https://your-prometheus-api.example.com
        bearerToken: <token>
        supportedTypes:
          - METRICS_STORE
```
Kayenta fails with:
```
Caused by: com.fasterxml.jackson.core.JsonParseException: Unrecognized token 'OK': 
was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
```
### **Root Cause**

The current implementation in RetrofitClientFactory.java uses OkHttp's .authenticator() method, which only triggers after receiving an HTTP 401 response:

```
return okHttp3ClientBuilder
    .authenticator(
        (route, response) ->
            response.request().newBuilder().header("Authorization", credential).build())
    .build();
```

This challenge-response pattern assumes all APIs return 401 for missing/invalid authentication. However, some providers like Coralogix return HTTP 200 with an empty or non-JSON response body when authentication is missing, causing:

1. The authenticator never triggers (no 401 received)
2. The Authorization header is never added
3. Kayenta receives an unexpected response body
4. JSON parsing fails when trying to parse non-JSON content

## Changes

- Added AuthorizationInterceptor class
- Modified createAuthenticatedClient() to use .addInterceptor()
- Added unit tests using MockServer

## Test Results — RetrofitClientFactoryTest

All **9 tests passed** across 3 test groups, proving the bug and validating the fix.

```
BUILD SUCCESSFUL
Tests: 9, Failures: 0, Errors: 0, Skipped: 0
```

### OLD behavior (.authenticator()) — demonstrates the bug

| Test | Result | What it proves |
|------|--------|----------------|
| authenticatorFailsWhenServerReturns200WithoutAuth | PASSED | **Bug reproduced**: authenticator never fires when Coralogix returns 200 instead of 401 → JsonParseException |
| authenticatorWorksOnlyWhenServerReturns401 | PASSED | Old approach only works if server returns 401 first |
| authenticatorCausesDoubleRequestOn401Servers | PASSED | Even on 401 servers, the old approach wastes a round-trip (2 requests instead of 1) |

### NEW behavior (.addInterceptor()) — proves the fix

| Test | Result | What it proves |
|------|--------|----------------|
| interceptorWorksWithServerThatNeverReturns401 | PASSED | **Fix verified**: interceptor sends auth proactively — works with Coralogix-like servers |
| interceptorAlsoWorksWithStandard401Servers | PASSED | Backwards-compatible with standard 401 servers |
| basicAuthIsSentProactively | PASSED | Basic auth also works proactively |
| authorizationHeaderMatchesConfiguredCredential | PASSED | Exact Authorization header value is sent correctly |

### Side-by-side comparison — same server, old vs new

| Test | Result | What it proves |
|------|--------|----------------|
| oldFailsNewSucceedsOnSameServer | PASSED | **Definitive before/after**: same Coralogix-like server → OLD fails with JsonParseException, NEW succeeds |
| reproducesExactProductionError | PASSED | Reproduces the exact JsonParseException: Unrecognized token 'OK' error from production |

### Key takeaway

- **OLD (.authenticator())**: Request sent **without** Authorization header → server returns 200 OK (plain text) → authenticator **never fires** → JsonParseException
- **NEW (.addInterceptor())**: Request sent **with** Authorization header on first attempt → server returns valid JSON → success

### DIFF
--- a/.github/actions/publish-npm/action.yml
+++ b/.github/actions/publish-npm/action.yml
@@ -58,4 +58,7 @@ runs:
         NODE_ENV: production
       shell: bash
       working-directory: ${{ inputs.project }}
-      run: npm publish
+      ## We have to do this... since otherwise "latest" tag is applied, and backports
+      ## then fail because they'll be on older versions... and you can't tag an older
+      ## semver version as "latest".  NPM rejects it.
+      run: npm publish --tag release-${{ inputs.version }}
--- a/fiat/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
+++ b/fiat/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatPermissionEvaluator.java
@@ -501,7 +501,7 @@ public class FiatPermissionEvaluator implements UserPermissionEvaluator {
    */
   @SuppressWarnings("unused")
   public boolean isAdmin() {
-    return true; // TODO(ttomsu): Chosen by fair dice roll. Guaranteed to be random.
+    return isAdmin(SecurityContextHolder.getContext().getAuthentication());
   }
 
   public boolean isAdmin(Authentication authentication) {
--- a/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/OrcaService.java
+++ b/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/OrcaService.java
@@ -105,6 +105,18 @@ public interface OrcaService {
       @Body String ignored);
 
   @Headers("Accept: application/json")
+  @PUT("/admin/forceCancelExecution")
+  Call<Map> forceCancelPipeline(
+      @Query("executionId") String executionId,
+      @Query("executionType") String executionType,
+      @Query("canceledBy") String canceledBy);
+
+  @Headers("Accept: application/json")
+  @POST("/admin/queue/hydrate")
+  Call<Map> rehydrateExecution(
+      @Query("executionId") String executionId, @Query("dryRun") boolean dryRun);
+
+  @Headers("Accept: application/json")
   @PUT("pipelines/{id}/pause")
   Call<Void> pausePipeline(@Path("id") String id, @Body String ignored);
 
--- a/gate/gate-saml/src/main/java/com/netflix/spinnaker/gate/security/saml/ResponseAuthenticationConverter.java
+++ b/gate/gate-saml/src/main/java/com/netflix/spinnaker/gate/security/saml/ResponseAuthenticationConverter.java
@@ -17,6 +17,7 @@
 
 package com.netflix.spinnaker.gate.security.saml;
 
+import com.netflix.spinnaker.gate.security.AllowedAccountsSupport;
 import com.netflix.spinnaker.gate.services.AuthenticationService;
 import com.netflix.spinnaker.security.User;
 import java.util.Collection;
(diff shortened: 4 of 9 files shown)


kayenta/kayenta-core/src/main/java/com/netflix/kayenta/retrofit/config/RetrofitClientFactory.java
```diff
@@ -109,9 +133,12 @@ private static OkHttpClient createAuthenticatedClient(
     }
 
     return okHttp3ClientBuilder
+        // Add interceptor to proactively send Authorization header on all requests
+        .addInterceptor(new AuthorizationInterceptor(credential))
+        // Keep authenticator as fallback for 401-based authentication flows
         .authenticator(
```
jasonmcintosh: Note, would want to remove the authenticator block entirely.  Not needed, just the interceptor.

jasonmcintosh:
Basically this ;) just need tests to show the issue.  There's examples using this in a few places with real server calls.

kayenta/kayenta-core/src/main/java/com/netflix/kayenta/retrofit/config/RetrofitClientFactory.java
rahulraj0504: @jasonmcintosh Sure, Removed authenticator fallback

rahulraj0504:
## Test Results — `RetrofitClientFactoryTest`

All **9 tests passed** across 3 test groups, proving the bug and validating the fix.

```
BUILD SUCCESSFUL
Tests: 9, Failures: 0, Errors: 0, Skipped: 0
```

### OLD behavior (`.authenticator()`) — demonstrates the bug

| Test | Result | What it proves |
|------|--------|----------------|
| `authenticatorFailsWhenServerReturns200WithoutAuth` | ✅ PASSED | **Bug reproduced**: authenticator never fires when Coralogix returns 200 instead of 401 → `JsonParseException` |
| `authenticatorWorksOnlyWhenServerReturns401` | ✅ PASSED | Old approach only works if server returns 401 first |
| `authenticatorCausesDoubleRequestOn401Servers` | ✅ PASSED | Even on 401 servers, the old approach wastes a round-trip (2 requests instead of 1) |

### NEW behavior (`.addInterceptor()`) — proves the fix

| Test | Result | What it proves |
|------|--------|----------------|
| `interceptorWorksWithServerThatNeverReturns401` | ✅ PASSED | **Fix verified**: interceptor sends auth proactively — works with Coralogix-like servers |
| `interceptorAlsoWorksWithStandard401Servers` | ✅ PASSED | Backwards-compatible with standard 401 servers |
| `basicAuthIsSentProactively` | ✅ PASSED | Basic auth also works proactively |
| `authorizationHeaderMatchesConfiguredCredential` | ✅ PASSED | Exact Authorization header value is sent correctly |

### Side-by-side comparison — same server, old vs new

| Test | Result | What it proves |
|------|--------|----------------|
| `oldFailsNewSucceedsOnSameServer` | ✅ PASSED | **Definitive before/after**: same Coralogix-like server → OLD fails with `JsonParseException`, NEW succeeds |
| `reproducesExactProductionError` | ✅ PASSED | Reproduces the exact `JsonParseException: Unrecognized token 'OK'` error from production |

### Key takeaway

The MockServer logs confirm the root cause:
- **OLD (`.authenticator()`)**: Request sent **without** `Authorization` header → server returns `200 OK` (plain text) → authenticator **never fires** → `JsonParseException`
- **NEW (`.addInterceptor()`)**: Request sent **with** `Authorization` header on first attempt → server returns valid JSON → success

PR #7456, opened 2026-02-07
fix(kayentadeck): Fix NPM publishing to NOT tag by latest. 

jasonmcintosh:
This prevents us from hitting an issue where backports fail to publish due to being older versions.  Specifically the NPM publishing by default if tags are NOT set uses a "latest' tag on every publish.  The result is that when you publish an older release, which shows up as a lower revision, NPM rejects that tag since it's older breaking publishing.  This is a VERY simple "just use the version" vs. "latest" approach.  IF we want a more advanced option, can consider it - I just went simple on this one.

### DIFF
--- a/.github/actions/publish-npm/action.yml
+++ b/.github/actions/publish-npm/action.yml
@@ -58,4 +58,7 @@ runs:
         NODE_ENV: production
       shell: bash
       working-directory: ${{ inputs.project }}
-      run: npm publish
+      ## We have to do this... since otherwise "latest" tag is applied, and backports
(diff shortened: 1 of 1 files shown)


jasonmcintosh:
```
npm error Cannot implicitly apply the "latest" tag because previously published version 2025.4.1 is higher than the new version 2025.2.4. You must specify a tag using --tag.
```
E.g. this error.

dbyron-sf (approved):

jasonmcintosh:
@mergifyio backport release-2025.4.x release-2025.3.x release-2025.2.x

PR #7460, opened 2026-02-07
fix(publishing): Tags cant be semver so prefix with release

jasonmcintosh:
NPM publishing on tags doesn't like a "semver" tag.  OK so we don't use semver, but the semver with a label in front... silly NPM


jasonmcintosh:
@mergifyio backport release-2025.4.x release-2025.3.x

PR #7502, opened 2026-02-21
fix(spring): remove LocalVariableTableParameterNameDiscoverer warnings from kork, fiat, and clouddriver

dbyron-sf:
This includes changes mainly in kork, fiat, and clouddriver to add `-parameters` as compiler argument to remove warnings on startup from org.springframework.core.LocalVariableTableParameterNameDiscoverer like:
```
"Using deprecated '-debug' fallback for parameter name resolution. Compile the affected code with '-parameters' instead or avoid its introspection: com.netflix.spinnaker.config.DefaultServiceClientProvider
```
on startup.  These warnings are new with spring boot 3.0.  See https://stackoverflow.com/questions/74600681/warning-printed-after-migrating-to-spring-boot-3-0-spring-integration-6-0 and https://ricma.co/posts/tech/dev/migrating-to-spring-61-javac-parameters/ for background.

Note that there is some non-trivial conversion of groovy to java here, and a change in clouddriver to compile java code separately from groovy (i.e. with JavaCompile instead of GroovyCompile).  All of this is supposed to be a no-op (other than removing the warnings), but it does touch some pretty fundamental parts of clouddriver.

Note also there are remaining warnings from LocalVariableTableParameterNameDiscoverer in clouddriver-aws.  Splitting groovy from java, converting to java, etc. remains to do there.

### DIFF
--- a/fiat/build.gradle
+++ b/fiat/build.gradle
@@ -49,6 +49,18 @@ subprojects {
       }
     }
 
+    // Avoid warnings like
+    //
+    // Using deprecated '-debug' fallback for parameter name resolution. Compile
+    // the affected code with '-parameters' instead or avoid its introspection:
+    // com.netflix.spinnaker.fiat.controllers.RolesController
+    //
+    // FWIW, the spring boot gradle plugin enables this since 2.0 via
+    // https://github.com/spring-projects/spring-boot/pull/9839.
+    tasks.withType(JavaCompile) {
+      options.compilerArgs << "-parameters"
+    }
+
     dependencies {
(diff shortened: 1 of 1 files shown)


jasonmcintosh (approved):
Looks good :) PROBABLY could do most of the srcDirs handling via the gradle project, but... I'm good with this ;)

jasonmcintosh:
Please note:  I've NOT looked AS much... as I'm TRULY testing more of the integration tests vs. introspecting each bit of change.  AKA this is a "preliminary overall view" that the changes look good.

PR #7503, opened 2026-02-21
fix(iap): Fix iap auth post spring 3.0 upgrade

jasonmcintosh:
The PRIMARY impact is that line 130 was skipped if the user was previously authenticated, not setting as a result auth on the context.  This fixes this AND adds full integration tests to verify IAP in the future.

### DIFF
--- a/.github/actions/oidc-debugger-action/action.yml
+++ b/.github/actions/oidc-debugger-action/action.yml
@@ -1,0 +1,18 @@
+name: 'oidc-debugger-action'
+description: 'Print decoded OIDC token claims generated through Github Actions.'
+inputs:
+  audience:
+    description: 'The audience to use when requesting the JWT. Your Github server URL and repository owner (e.g. https://github.com/github).'
+    required: true
+    default: '${{ github.server_url }}/${{ github.repository }}'
+runs:
+  using: "composite"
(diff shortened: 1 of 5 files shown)


jasonmcintosh:
Backport candidate: 2025.4.x and 2026.0.x

rahul-chekuri (commented):
@jasonmcintosh Do you think it would be useful to add a test to confirm that IAP works correctly when configured alongside other IDPs like SAML, OAuth, or basic authentication?

jasonmcintosh:
SO Auth has NEVER supported MULTIPLE auth providers in a single configuration.  MINUS TLS.  It would be worth adding a test around TLS & other configs at the same time.  Specifically:
* You cannot have IAP auth AND oauth2 in the SAME configuration.  This has been tried several times, and it's tied to the way spring security filter chains operate.
* Spinnaker DOES support cert auth & oauth2 in the same flows (or certs & saml or certs and XYZ).  As it normally starts a secondary port for auth and separate LBs for this purpose.  

E.g. I don't think this is needed as that use case never worked.

jasonmcintosh:
@Mergifyio backport release-2026.0.x release-2025.4.x

PR #7504, opened 2026-02-22
feat(ci/gcb): Add region in gcb account def for regional endpoints

himanhsugusain:
My last PR for privatePool https://github.com/spinnaker/igor/commit/9dbe94dae8a4e63bca33fc9ed9e9aede805406b8 (reverted in this PR ) only supports starting a gcb build  and it only works if its the default region ( us-central1 ) for gcp. GCP uses regional endpoints for non default pools, so we need regional endpoints to monitor and cancel build via spinnaker UI.
Explored several option how to support this and I conclude with PR being simplest change

Note: since this uses a different account in the gcb config, it also needs a separate cloud build subscription, to avoid race condition b/w the accounts, since one will discard the update if its not associated with its build artifacts . We can probably optimize the cloud build subscriber to send to correct account.
This does not break existing functionalities.

```overlays/config/files/spinnaker-local.yml
# Add configuration that should be mounted into all JVM services
gcb:
  enabled: true
  accounts:
    - name: dev
      permission: {}
      project: dev
      subscriptionName: spinnaker-cloud-build
    - name: dev-us-east1
      permission: {}
      cloudBuildRegion: us-east1
      project: dev
      subscriptionName: spinnaker-test-us-east1
```

**Test Setup**
```
# Account: dev with no region in config
# build.yaml
steps:
  - args:
      - bash
      - '-c'
      - 'echo "FOO=${_FOO}"'
    name: ubuntu
substitutions:
  _FOO: bar
```

```
# Account: dev-us-east1 with region us-east1
# build.yaml
options:
  pool:
    name: >-
      projects/dev/locations/us-east1/workerPools/test-pool-private
steps:
  - args:
      - bash
      - '-c'
      - 'sleep 60; echo "FOO=${_FOO}"'
    name: ubuntu
substitutions:
  _FOO: bar
```
Test performed manually
- Trigger
- Cancellation ( from spinnaker and from GCB )

**This is not AI slop, its my personal slop**


igor/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/CloudBuildFactory.java
```diff
@@ -66,6 +66,16 @@ CloudBuild getCloudBuild(GoogleCredentials credentials, String applicationName)
     return builder.build();
   }
 
+  CloudBuild getCloudBuild(
```
jasonmcintosh, line 61: I note there's... a fair bit of duplication between the other getCloudBuild method.  I'd prefer to collapse these into a single method that determines which type of cloudbuild to use.

igor/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildClient.java
```diff
@@ -52,15 +52,17 @@ GoogleCloudBuildClient create(GoogleCredentials credentials, String projectId) {
       Storage cloudStorage = cloudBuildFactory.getCloudStorage(credentials, applicationName);
       return new GoogleCloudBuildClient(projectId, cloudBuild, cloudStorage, executor);
     }
+
+    GoogleCloudBuildClient create(
```
jasonmcintosh, line 55: Same issue here - duplicate methods that really shouldn't be.  Give the factory could handle this and give the correct response...

igor/igor-web/src/test/groovy/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildTest.java
```diff
@@ -137,41 +136,6 @@ public void presentAccountTest() throws Exception {
     assertThat(stubCloudBuildService.findUnmatchedRequests().getRequests()).isEmpty();
   }
 
-  @Test
-  public void presentAccountTestWithPoolOption() throws Exception {
```
jasonmcintosh, line 141: I'm... looking and wondering - isn't this set of tests still quasi needed to verify that build calls are made?  Not sure if the move/change to the tests in the above covers these tests?

jasonmcintosh:
Few minor things ;) AND... we don't mind AI slop - AI has actually gotten FAR less sloppy lately.  OR you use it then clean up the slop to get good results!  Thanks for the PR!

igor/igor-web/src/test/groovy/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildTest.java
himanhsugusain: This test is not relevant anymore. We don't care what the build spec is anymore.
If we have private pool then we can optionally configure region to use correct gcb regional endpoint.

igor/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildClient.java
himanhsugusain: updated

igor/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/CloudBuildFactory.java
himanhsugusain: updated

igor/igor-web/src/test/groovy/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildTest.java
jasonmcintosh: K I guess i was wondering (not knowing the GCP SDK/apis as well) if we were essentially removing functionality by removing the pool stuff.  https://github.com/spinnaker/spinnaker/pull/7504/changes#diff-3eec797cd0d376fd39becf47d176dded75fc91b0d84c081c94d3fb381f42f629L58 - e.g. this block.  I'm GUESSING ... b/c of the way the regional endpoint works this covers this... :) I'm ok with that if so.

igor/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildAccountFactory.java
```diff
@@ -35,9 +35,14 @@ final class GoogleCloudBuildAccountFactory {
 
   GoogleCloudBuildAccount build(GoogleCloudBuildProperties.Account account) {
     GoogleCredentials credentials = getCredentials(account);
-
-    GoogleCloudBuildClient client =
-        googleCloudBuildClientFactory.create(credentials, account.getProject());
+    GoogleCloudBuildClient client;
+    if (account.getCloudBuildRegion().isPresent()) {
```
jasonmcintosh, line 39: So we can look at merging this.  But a code path I prefer on this is just pass the "region" down kinda thing vs. do the client or optional config here.  I can see about refactoring this on a later PR... similar to the way you're doing the null check.  Just one of those minor things I prefer.

igor/igor-web/src/test/groovy/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildTest.java
himanhsugusain: yes, If the regional endpoint is set. It will just work, we don't need to parse region out of the build spec.

igor/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildAccountFactory.java
himanhsugusain: what I understand is that the client is created with account, so Builds w/ one account cannot have different client endpoints, unless we inject endpoint down the stack when the cloud build request is made, but that makes requires more changes.
I think the new version of the library have something better to handle this, I will look into that too.

PR #7510, opened 2026-02-25
feat(provider/google, deck/google): GCE Compute v1 migration and instance flexibility policy

### LINKED ISSUE #7304: Clouddriver Supports Deploying GCE Instance Group with New Instance Flexibility Config
GCE has announced new feature Instance Flexibility, which can greatly add stability to the MIG.

From reading their documentation (https://cloud.google.com/compute/docs/instance-groups/configure-instance-flexibility#terraform_1), a new _instance_flexibility_policy_ field is introduced to the instance template, which Spinnaker doesn't know how to parse and configure. It will be best that we add in support for this feature.

The involved components will be Deck and Clouddriver.

mrusan:
# feat(provider/google, deck/google): GCE Compute v1 migration and instance flexibility policy

## Summary

Migrate `clouddriver-google` from Compute API beta to stable v1, add instance flexibility policy support for regional MIGs, and align backend and Deck payloads with the stable-v1 API surface.

This includes the Compute dependency and endpoint cutover, `shieldedInstanceConfig` adoption, patch-based MIG auto-healing updates, `partnerMetadata` deprecation, explicit-zone intent round trips, and validation for flexibility constraints that would otherwise fail only after creating an instance template.

## Changes

| Area | Details |
| --- | --- |
| Compute v1 migration | Bump `google-api-services-compute` from `beta-rev20240430-2.0.0` to `v1-rev20250415-2.0.0`; route default clients through `compute/v1` and `batch/compute/v1`; migrate `ShieldedVmConfig` to `ShieldedInstanceConfig`; remove the beta-only instance-template `view=FULL` option. |
| Auto-healing | Replace the dedicated beta `setAutoHealingPolicies` RPC with exact zonal and regional MIG `PATCH` requests. `InstanceGroupManager#setAutoHealingPolicies` is retained only as the model setter used to construct the PATCH body. |
| Instance flexibility | Add `GoogleInstanceFlexibilityPolicy` with named instance selections and optional ranks; map it to stable-v1 regional MIG requests; cache and display it; inherit or explicitly clear it during clone. Target shape remains owned by `distributionPolicy.targetShape`. |
| Flexibility validation | Require a regional MIG and `BALANCED`, `ANY`, or `ANY_SINGLE_ZONE`; allow omitted ranks; reject negative ranks, null selections, empty/blank machine types, and duplicate machine types after normalizing whitespace, case, and URL-qualified names. |
| Regional placement | Emit `updatePolicy.instanceRedistributionType=NONE` for Spinnaker's regional non-`EVEN`/flexibility contract. Preserve observed GCP zones separately from Spinnaker's explicit-zone intent marker; explicit `selectZones=false` removes inherited intent. |
| Shielded config | Write `shieldedInstanceConfig`. Read `shieldedInstanceConfig` first and fall back to legacy `shieldedVmConfig` in new backend and Deck paths so pre-cutover cached data remains usable during rollout. Preserve unspecified flags on partial template updates. |
| `partnerMetadata` deprecation | Retain a permissive description field so old payloads deserialize, but strip the field in deploy/clone converters and Deck command builders before stable-v1 instance-template writes. Remove the UI editor. |
| Deck alignment | Add an accessible React flexibility editor and server-group details view on the post-Angular wizard and details surfaces; bind policy, regional mode, and target shape separately; deep-clone cached policy state before editing; remove unsupported auto-healing fields from command hydration. |
| Wire-contract tests | Capture real generated Compute requests and assert HTTP method, stable-v1 path, and JSON body for MIG insert/PATCH, shielded config, auto-healing, placement, and metadata behavior. |

### Files Added

* `GoogleInstanceFlexibilityPolicy.java` — Spinnaker model for stable-v1 MIG instance selections.
* `InstanceFlexibilityPolicyValidationSupport.groovy` — shared create/clone/operation validation.
* `GoogleRmigRedistributionContract.java` — regional placement and redistribution contract.
* `CapturingComputeTransport.java` — raw generated-Compute request capture for boundary tests.
* `GceInstanceFlexibilityConfigurer.tsx` — Deck wizard editor and submit guard.
* `GceInstanceFlexibilityPolicyDetails.tsx` — Deck server-group details rendering.
* Focused backend and Deck specs for credentials routing, validation, serialization, cache/clone behavior, auto-healing, and accessibility.

### Files Modified

* `spinnaker-dependencies.gradle` — stable Compute v1 dependency revision.
* `ComputeVersion.groovy` — stable default service and batch paths while retaining alpha routing; credential-routing specs cover both channels.
* `BasicGoogleDeployHandler.java` — stable-v1 shielded writes, flexibility mapping, target-shape canonicalization, explicit-zone metadata synchronization, and regional redistribution assembly.
* `AbstractGoogleServerGroupCachingAgent.java` — flexibility, autoscaler visibility, zones, and explicit-intent projection.
* `BasicGoogleDeployDescription.groovy` / `BaseGoogleInstanceDescription.groovy` / `GoogleServerGroup.groovy` — request and view models.
* `GCEUtil.groovy`, clone, snapshot, and template-update operations — stable-v1 shielded handling with legacy read fallback.
* Auto-healing operations and validators — MIG PATCH bodies and rejection of unsupported `autoHealingPolicy.maxUnavailable`.
* `InstanceTemplates.java` and server-group manager wrappers — stable-v1 API alignment.
* Deck command builders, wizard, details, and auto-healing components — payload cleanup, compatibility reads, and flexibility UX.

## Backward Compatibility

| Change | Upgrade impact | Rollback considerations |
| --- | --- | --- |
| Compute beta → stable v1 | Existing GCP resources remain the same resources; only the API/client surface changes. | Roll Deck and Clouddriver back together. Resources created through v1 continue to exist and can be reached through the beta API, but representative deploy/clone/autoscaling smoke tests are still required. |
| `partnerMetadata` removal | Existing payloads deserialize and execute, but the field is deliberately stripped and no longer reaches instance-template writes. This is a behavioral break for pipelines that relied on it. | Saving a stage in new Deck can remove the field from persisted pipeline configuration. A rollback does not recreate it; restore a previous Front50 pipeline version or re-add it manually if the old behavior is required. |
| `autoHealingPolicy.maxUnavailable` | Existing pipelines that send this field now fail validation instead of silently losing the setting. Stable-v1 MIG auto-healing supports `healthCheck` and `initialDelaySec`; update-policy `maxUnavailable` is a different contract. | Old beta code can accept the old field only if it is still present. New Deck strips it when rebuilding/saving the stage, so use Front50 history if rollback must restore the prior setting. |
| `shieldedVmConfig` → `shieldedInstanceConfig` | New code reads both key shapes and writes only the stable-v1 shape, supporting forward rollout over old cached data. | Reverse compatibility is not symmetric for shared cached objects: old code does not explicitly understand the new key. Refresh Google server-group/instance-template caches before clone, edit, or snapshot operations after rollback. |
| Instance flexibility policy | Additive for existing pipelines when absent. New pipelines gain regional multi-machine-type behavior and stricter validation. | Old Clouddriver ignores unknown description properties, so a flexibility-enabled pipeline can appear to succeed while deploying without flexibility. Existing flexible MIGs keep running, but do not clone or mutate them with the rolled-back version. |
| Explicit zones | `selectZones=true` now requires a regional group and at least one explicit zone. `selectZones=false` clears inherited Spinnaker intent while still retaining observed GCP zones in the view. | Pipelines that previously sent `selectZones=true` without zones must be corrected before upgrade; rollback restores the old permissive behavior but also restores its stale-intent risk. |
| MIG auto-healing RPC → PATCH | The resulting GCP auto-healing resource state is shared across API versions; PATCH bodies are covered at the raw HTTP boundary. | The old beta RPC can operate after rollback. Verify one zonal and one regional policy update during rollback smoke testing. |

There is no database or GCP resource migration, so an early binary rollback is practical, but some pipeline inputs intentionally change behavior and new feature state is not understood by old code.

## Test Coverage

* Stable and alpha Compute client service/batch routing.
* Flexibility validation for regional mode, target shape, optional/negative ranks, null entries, blank machine types, normalized duplicates, and distinct selections.
* Deploy-handler and raw-request coverage for flexibility mapping, machine-type trimming, regional redistribution, explicit zones, stale intent removal, and shielded v1 serialization.
* Cache/view coverage for flexibility, autoscaler status, observed zones, and explicit-zone intent.
* Clone coverage for policy inheritance, explicit empty-policy clearing, zone override precedence, and shielded v1/legacy fallback.
* Exact zonal and regional auto-healing PATCH method/path/body assertions for upsert and delete.
* Deck coverage for command hydration, deep-clone isolation, React wizard rerenders, accessibility, help integration, duplicate warnings, clear/inherit behavior, and unsupported-field stripping.

## References

1. [About instance flexibility in MIGs](https://cloud.google.com/compute/docs/instance-groups/about-instance-flexibility)
2. [Configure instance flexibility](https://cloud.google.com/compute/docs/instance-groups/configure-instance-flexibility)
3. [Regional MIG target distribution shape](https://cloud.google.com/compute/docs/instance-groups/regional-mig-set-target-distribution-shape)
4. [Stable-v1 regional MIG resource](https://cloud.google.com/compute/docs/reference/rest/v1/regionInstanceGroupManagers)
5. [Stable-v1 regional MIG PATCH](https://cloud.google.com/compute/docs/reference/rest/v1/regionInstanceGroupManagers/patch)
6. [Stable-v1 instance-template list](https://cloud.google.com/compute/docs/reference/rest/v1/instanceTemplates/list)
7. [Stable-v1 shielded instance configuration](https://cloud.google.com/compute/docs/reference/rest/v1/instances/insert)
8. [Stable Compute v1 API](https://cloud.google.com/compute/docs/reference/rest/v1)

## Other

Addresses https://github.com/spinnaker/spinnaker/issues/7304


clouddriver/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/converters/CopyLastGoogleServerGroupAtomicOperationConverter.groovy
```diff
@@ -23,17 +23,26 @@ import com.netflix.spinnaker.clouddriver.google.security.GoogleNamedAccountCrede
...
 class CopyLastGoogleServerGroupAtomicOperationConverter extends AbstractAtomicOperationsCredentialsConverter<GoogleNamedAccountCredentials> {
 
   AtomicOperation convertOperation(Map input) {
     new CopyLastGoogleServerGroupAtomicOperation(convertDescription(input))
   }
 
   BasicGoogleDeployDescription convertDescription(Map input) {
+    // Saved pipelines may still contain partnerMetadata from older payload shapes. Strip it here so
+    // clones continue without propagating unsupported partner metadata into instanceTemplates.insert.
+    if (input.containsKey('partnerMetadata')) {
```
christosarvanitis, line 41: This is a breaking change and we need to consider users having partner integration. Add release notes:
"partnerMetadata (Compute beta-only) is no longer propagated on stable v1. Pipelines containing it keep working; the field is ignored with a warning. Migrate to resourceManagerTags." Also provide an example in the release notes how the resourceManagerTags can be used because it is not a drop-in replacement

deck/packages/google/src/serverGroup/configure/serverGroupCommandBuilder.service.js
christosarvanitis, line 1: Probably something missing here is the deep clone of the instanceFlexibilityPolicy to avoid mutating on edit through the wizard

clouddriver/clouddriver-google/src/main/java/com/netflix/spinnaker/clouddriver/google/deploy/handlers/BasicGoogleDeployHandler.java
```diff
@@ -1096,14 +1115,92 @@ protected void setDistributionPolicyToInstanceGroup(
       }
 
       if (StringUtils.isNotBlank(description.getDistributionPolicy().getTargetShape())) {
-        distributionPolicy.setTargetShape(description.getDistributionPolicy().getTargetShape());
+        // Canonicalize before the GCP call so Deck/API casing and whitespace cannot produce
+        // rejected or inconsistently stored distribution shapes.
+        distributionPolicy.setTargetShape(
+            description.getDistributionPolicy().getTargetShape().trim().toUpperCase(Locale.ROOT));
```
christosarvanitis: NIT extract to a canonicalizeTargetShape(string) in GCEUtil to replace this and in the clouddriver/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/InstanceFlexibilityPolicyValidationSupport.groovy

clouddriver/clouddriver-google/src/main/groovy/com/netflix/spinnaker/clouddriver/google/deploy/validators/StandardGceAttributeValidator.groovy
```diff
@@ -512,30 +513,51 @@ class StandardGceAttributeValidator {
...
-            validateNonNegativeLong(fixed as int, "autoHealingPolicy.maxUnavailable.fixed")
-          } else if (percent != null) {
-            validateInRangeInclusive(percent as int,
-              0, 100, "autoHealingPolicy.maxUnavailable.percent")
-          } else if (rejectEmptyMaxUnavailable) {
-            this.errors.rejectValue("autoHealingPolicy.maxUnavailable",
-              "${this.context}.autoHealingPolicy.maxUnavailable.neitherFixedNorPercent")
-          }
-        }
+      // Stable Compute v1 autoHealingPolicies only support healthCheck and initialDelaySec.
```
christosarvanitis, line 526: This is a breaking change as well for existing pipelines. Needs release notes. The use case would be a saved pipeline JSON with autoHealingPolicy.maxUnavailable in a deploy stage, never opened in deck since upgrade → clouddriver's validator hard-rejects at runtime. Open and re-save the pipeline in Deck to remove it automatically.

christosarvanitis (changes requested):
Overall a well tested PR, thanks! Some comments to address the breaking changes and help users navigate the beta -> v1

mrusan:
> Overall a well tested PR, thanks! Some comments to address the breaking changes and help users navigate the beta -> v1

Thank you for the review, Christos.

1. Canonicalization: good idea, implemented in follow-up commit.
2. Deep clone: I checked the code again - it is already implemented at [serverGroupCommandBuilder.service.js:476-479](https://github.com/spinnaker/spinnaker/blob/45389437fa8af2258a8721fa4c4dfb8e8fdf29bf/deck/packages/google/src/serverGroup/configure/serverGroupCommandBuilder.service.js#L476-L479) - pipeline hydration deep-clones the entire original cluster. Coverage spec [L149-153](https://github.com/spinnaker/spinnaker/blob/45389437fa8af2258a8721fa4c4dfb8e8fdf29bf/deck/packages/google/src/serverGroup/configure/serverGroupCommandBuilder.service.spec.js#L151-L153) and L222-225 mutates the nested command and verifies the source remains unchanged. Anything you feel needs to be added? 
3. Docs: https://github.com/spinnaker/spinnaker.io/pull/643

christosarvanitis (approved):

PR #7511, opened 2026-02-25
fix(gate/web): remove BakeController.handleBakeOptionsException

dbyron-sf:
The catch-all @ExceptionHandler in BakeController masks all exceptions as HTTP 404 with a generic error message. This hides real errors (including 500s from downstream services) and makes debugging difficult.

Remove it so exceptions are handled by kork's GenericExceptionHandlers, which maps exceptions to appropriate HTTP status codes and preserves error details in the response.

and some additional test coverage and conversion from groovy to java while we're at it.

### DIFF
--- a/gate/gate-web/src/main/java/com/netflix/spinnaker/gate/controllers/BakeController.java
+++ b/gate/gate-web/src/main/java/com/netflix/spinnaker/gate/controllers/BakeController.java
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.gate.controllers;
 
 import com.netflix.spinnaker.gate.services.BakeService;
 import io.swagger.v3.oas.annotations.Operation;
+import java.util.List;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -32,7 +33,7 @@ public class BakeController {
 
   @Operation(summary = "Retrieve a list of available bakery base images, grouped by cloud provider")
   @RequestMapping(value = "/options", method = RequestMethod.GET)
-  public Object bakeOptions() {
+  public List bakeOptions() {
     return bakeService.bakeOptions();
   }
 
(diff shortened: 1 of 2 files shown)


gate/gate-web/src/main/java/com/netflix/spinnaker/gate/controllers/BakeController.java
```diff
@@ -0,0 +1,52 @@
...
+
+@RequestMapping("/bakery")
+@RestController
+public class BakeController {
+
+  @Autowired private BakeService bakeService;
+
+  @Operation(summary = "Retrieve a list of available bakery base images, grouped by cloud provider")
+  @RequestMapping(value = "/options", method = RequestMethod.GET)
+  public List bakeOptions() {
```
rahul-chekuri, line 36: It might be helpful to change the return type to `List<BakeOptions>` instead of `List` to improve type safety.

gate/gate-web/src/main/java/com/netflix/spinnaker/gate/controllers/BakeController.java
```diff
@@ -0,0 +1,52 @@
...
+    return bakeService.bakeOptions();
+  }
+
+  @Operation(summary = "Retrieve a list of available bakery base images for a given cloud provider")
+  @RequestMapping(value = "/options/{cloudProvider}", method = RequestMethod.GET)
+  public Object bakeOptions(@PathVariable("cloudProvider") String cloudProvider) {
```
rahul-chekuri, line 42: It might be helpful to change the return type to `BakeOptions` instead of `Object` to improve type safety.

gate/gate-web/src/main/java/com/netflix/spinnaker/gate/controllers/BakeController.java
dbyron-sf: BakeService.bakeOptions sometimes returns `List<BakeOptions>`, but it also sometimes returns `List` because RoscoService.bakeOptions returns `List`.  Unless we make some common type between rosco and gate for this, I think we're stuck where we are.

gate/gate-web/src/main/java/com/netflix/spinnaker/gate/controllers/BakeController.java
dbyron-sf: Similar struggle here.  `BakeService.bakeOptions(String cloudProvider)` sometimes returns `BakeOptions` and sometimes returns a `Map`, since that's what RoscoService.bakeOptions(String cloudProvider) returns.

PR #7516, opened 2026-02-27
build(deck): upgrade webpack to 5.104.1, webpack-dev-server to 5.2.1 & eslint to 9.39.2

mattgogerly:
Upgrade `webpack-dev-server` to 5.2.1, `webpack` to 5.2.1 and `eslint` to 9.39.2. This required a whole lot more work than I originally expected, including upgrading all the webpack plugins, and then `eslient` and all of its plugins.

Dependabot tried its best but these upgrades were all required for this to work for us. We've had this going for about a month now and haven't seen any problems. It resolved about 60 CVEs flagged by Dependabot for us.

### DIFF
--- a/.github/workflows/rebuild-all.yml
+++ b/.github/workflows/rebuild-all.yml
@@ -25,7 +25,7 @@ jobs:
           app-id: ${{ secrets.RELEASE_APP_ID }}
           private-key: ${{ secrets.RELEASE_PRIVATE_KEY }}
       - name: Trigger Deck Release
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           github-token: ${{ steps.app-token.outputs.token }}
           script: |
@@ -48,7 +48,7 @@ jobs:
           app-id: ${{ secrets.RELEASE_APP_ID }}
           private-key: ${{ secrets.RELEASE_PRIVATE_KEY }}
       - name: Trigger Deck Kayenta Release
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           github-token: ${{ steps.app-token.outputs.token }}
           script: |
--- a/.github/workflows/spinnaker-release.yml
+++ b/.github/workflows/spinnaker-release.yml
@@ -104,7 +104,7 @@ jobs:
 
 
       - name: Trigger Deck Release
-        uses: actions/github-script@v7
+        uses: actions/github-script@v8
         with:
           github-token: ${{ steps.app-token.outputs.token }}
           script: |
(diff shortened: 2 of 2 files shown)


deck/packages/core/src/application/search/react-ultimate-pagination.d.ts
```diff
@@ -22,7 +22,7 @@ declare module 'react-ultimate-pagination' {
 
   export function createUltimatePagination(options: {
     itemTypeToComponent: ItemTypeToComponent;
-    WrapperComponent: React.ComponentClass = 'div';
+    WrapperComponent?: React.ComponentClass | string;
```
jasonmcintosh, line 25: Hrmm, I don't know react - but... is this intentinoal here to say "class is a string" vs. div?  I'd ASSUME yes, just caught my eye.

mattgogerly: iirc this was something caught by the newer version of eslint, default values aren't valid in type declarations. The library [itself](https://github.com/ultimate-pagination/react-ultimate-pagination/blob/944c8376a8959ad5a37fcaa697994eccb4f96252/src/react-ultimate-pagination.js#L21) sets a default `div` value, so we just need to accept `string` in the typing.

deck/packages/eslint-plugin/utils/utils.ts
```diff
@@ -1,17 +1,17 @@
...
-  Expression,
-  Identifier,
-  Literal,
-  MemberExpression,
-  NewExpression,
-  Node,
-  Program,
-  SpreadElement,
-} from 'estree';
+import type * as ESTree from 'estree';
```
jasonmcintosh, line 2: GENERALLY we've avoided wildcards... at least in java code

deck/packages/scripts/package.json
```diff
@@ -41,7 +41,7 @@
     "rollup-plugin-external-globals": "0.6.1",
     "rollup-plugin-less": "1.1.3",
     "rollup-plugin-postcss": "^4.0.2",
-    "rollup-plugin-terser": "7.0.2",
+    "@rollup/plugin-terser": "^0.4.0",
```
jasonmcintosh, line 44: Did... we go back a version?  I'm guessing not and this is just a ... impact of the change from the previous path... just caught my eye ;)

jasonmcintosh (approved):
SO I think I'm good - don't see any big concerns - a few wildcard imports should fix, but.. overall this looks like a painful upgrade done well!

deck/packages/scripts/package.json
mattgogerly: Nope, rollup-plugin-terser is deprecated and @rollup/plugin-terser is the replacement, they just started back at version 0.1.

PR #7534, opened 2026-03-10
fix(gateclient): send OAuth2 Bearer on every request so Gate authenticates

RevinKey:
Based on discussion https://spinnakerteam.slack.com/archives/C091CCWRJ/p1773086087421309 

This change will fix the use-case when using github oauth to authenticate the spin cli, when the user needs to be derived from the bearer token only:

Attach OAuth2 token source to context so every API request sends Authorization: Bearer.
Gate's ExternalAuthTokenFilter only runs when Bearer is present; without it, FiatSessionFilter sees no user and logs "Authenticated user was not present in authenticated request."

### DIFF
--- a/spin/cmd/gateclient/client_test.go
+++ b/spin/cmd/gateclient/client_test.go
@@ -1,0 +1,152 @@
+// Copyright (c) 2018, Google, Inc.
+// Copyright (c) 2019, Noel Cower.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


jasonmcintosh:
ONLY thing is can we get some tests ;) I'd love to refactor this code/cleanup the auth flow stuff as this has hit a few other oauth2 flows as well...

jasonmcintosh:
Similar to PR: https://github.com/spinnaker/spinnaker/pull/7532/changes

PR #7535, opened 2026-03-11
feat(igor): allow overriding artifact extensions in Artifactory monitors

mattgogerly:
Igor's Artifactory monitor searches with hard-coded extensions based on the type (`.pom` or `.tgz`). Allow overriding this on specific monitors if needed.


jasonmcintosh:
Looks like an import missing :) BUT overall looks good!

mattgogerly:
> Looks like an import missing :) BUT overall looks good!

Whoops, fixed.

PR #7536, opened 2026-03-11
chore(deps): update jinjava to 2.7.6

ian-fox:
This fixes [CVE-2026-25526](https://nvd.nist.gov/vuln/detail/CVE-2026-25526)

I chose `2.7.6` rather than the other patched version of `2.8.3` to minimize the chance of a breaking change, since spinnaker is already on `2.7.x`.

### DIFF
--- a/halyard/halyard-core/halyard-core.gradle
+++ b/halyard/halyard-core/halyard-core.gradle
@@ -16,7 +16,7 @@ dependencies {
   implementation 'org.apache.commons:commons-compress:1.20'
   implementation 'commons-io:commons-io'
   implementation 'io.reactivex.rxjava3:rxjava'
-  implementation 'com.hubspot.jinjava:jinjava:2.7.6'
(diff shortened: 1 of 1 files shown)


halyard/halyard-core/halyard-core.gradle
```diff
@@ -16,7 +16,7 @@ dependencies {
   implementation 'org.apache.commons:commons-compress:1.20'
   implementation 'commons-io:commons-io'
   implementation 'io.reactivex.rxjava3:rxjava'
-  implementation 'com.hubspot.jinjava:jinjava:2.2.3'
+  implementation 'com.hubspot.jinjava:jinjava:2.7.6'
```
dbyron-sf: The better thing is:
```suggestion
  implementation 'com.hubspot.jinjava:jinjava'
```
so halyard picks up the version from spinnaker-dependencies.

ian-fox: That makes sense, thanks!

PR #7538, opened 2026-03-15
fix(security): restore AnonymousConfig after Spring Boot 3 / Security6 upgrade

rahul-chekuri:
User Impact:
After upgrading to Spring Boot 3.0 / Spring Security 6 (Spinnaker 2025.4.0), Kubernetes accounts were not visible in Deck UI. All accounts were marked `authorized: false`.

Root Cause:
Spring Security 6 removes WebSecurityConfigurerAdapter and requires SecurityFilterChain beans. During migration, multiple configuration classes defined a @Bean method named `configure(HttpSecurity http)`, which resulted in bean name collision.

Because `spring.main.allow-bean-definition-overriding=true` was enabled, MetricsEndpointConfiguration.configure silently overrode AnonymousConfig.configure.

As a result:
- AnonymousConfig never executed
- @SpinnakerUser resolved to null
- CredentialsController marked all accounts unauthorized

Fix:
Rename SecurityFilterChain bean methods to unique names to prevent bean overriding and ensure AnonymousConfig is properly registered.

Verified:
- Kubernetes accounts are visible in Deck
- authorized flag is correctly set
- No regression in metrics endpoint

### DIFF
--- a/clouddriver/clouddriver-aws/clouddriver-aws.gradle
+++ b/clouddriver/clouddriver-aws/clouddriver-aws.gradle
@@ -34,6 +34,7 @@ dependencies {
   implementation "com.github.ben-manes.caffeine:guava"
   implementation "com.netflix.awsobjectmapper:awsobjectmapper"
   implementation "com.netflix.frigga:frigga"
+  implementation "commons-io:commons-io"
   implementation "io.spinnaker.fiat:fiat-api"
   implementation "io.spinnaker.fiat:fiat-core"
   implementation "io.spinnaker.kork:kork-cloud-config-server"
@@ -53,9 +54,6 @@ dependencies {
   implementation "org.apache.groovy:groovy"
   implementation "org.springframework.boot:spring-boot-actuator"
   implementation "org.springframework.boot:spring-boot-starter-web"
-  implementation 'com.aestasit.infrastructure.sshoogr:sshoogr:0.9.25'
-  implementation 'com.jcraft:jsch.agentproxy.connector-factory:0.0.9'
-  implementation 'com.jcraft:jsch.agentproxy.jsch:0.0.9'
   implementation "com.github.wnameless.json:json-flattener:0.11.1"
 
   testImplementation "io.spinnaker.kork:kork-exceptions"


jasonmcintosh:
Be great for a few more tests including better integration tests and I think a few OTHER Places have the same "configuration".  Ideally a few more integration tests on these.  That said, I think this addresses the issues on these core areas.

gate/gate-core/src/main/java/com/netflix/spinnaker/gate/security/anonymous/AnonymousConfig.java
```diff
@@ -48,7 +47,7 @@
 @Configuration
 @Log4j2
 @EnableWebSecurity
-@Order(Ordered.LOWEST_PRECEDENCE)
+@Order()
```
dbyron-sf: This seems like an unrelated change.  Also, isn't it important to keep AnonymousConfig at lowest precedence?

rahul-chekuri: By default it is at lowest precedence so no need for **Ordered.LOWEST_PRECEDENCE**

dbyron-sf: Perhaps, but why not minimize the diffs and also be explicit?  I have no idea what the default is, but I'd rather not depend on it.

rahul-chekuri: Ok will put back the code

dbyron-sf: ping @rahul-chekuri

rahul-chekuri: done and amended the commit

dbyron-sf: Thanks!

dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-2025.4.x release-2025.5.x release-2025.6.x

dbyron-sf:
@mergifyio backport release-2026.0.x

PR #7525, opened 2026-03-05
fix(saml): Fix wiring issue on saml.  Add FULL end to end integration test using keycloak & htmlunitdriver

jasonmcintosh:

PR #7539, opened 2026-03-17
fix(gate): rename x509 SecurityFilterChain bean to prevent SAML override when both are enabled

### LINKED ISSUE #7518: SAML and x509 SecurityFilterChain bean conflict causes 403 when both are enabled
## Summary

When both SAML and x509 authentication are enabled in Gate, the x509 `SecurityFilterChain` bean overrides the SAML one because both are registered under the same bean name `securi
tyFilterChain`. This results in **403 Forbidden** for all browser-based SAML authentication attempts.

This is a consequence of the Spring Boot 3 / Spring Security 6 migration, where `WebSecurityConfigurerAdapter` (which supported multiple ordered instances) was replaced by `Securi
tyFilterChain` `@Bean` methods.

## Affected versions

- **2025.4.1** — confirmed
- **2026.0.0** — confirmed (with additional issues, see below)
- Likely all versions since the Spring Boot 3 migration

## Steps to reproduce

1. Configure Gate with both SAML and x509 enabled:
   ```bash
   hal config security authn saml enable
   hal config security authn x509 enable
   hal deploy apply
   ```

2. Access the Spinnaker UI in a browser — the SAML redirect to the IdP results in:
   ```json
   {"error":"Forbidden","message":"Access Denied","status":403}
   ```

3. Disabling x509 (`hal config security authn x509 disable`) makes SAML work immediately.

## Root cause

Both `SAMLConfiguration$WebSecurityConfig` (in `gate-saml`) and `X509Config` (in `gate-x509`) define a `@Bean` method named `securityFilterChain` that returns a `SecurityFilterCha
in`:

**gate-saml** — `SAMLConfiguration$WebSecurityConfig`:
```java
@Bean
public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
    // Configures SAML2 login...
    return http.build();
}
```

**gate-x509** — `X509Config`:
```java
@Bean
public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
    // Configures x509 authentication...
    return http.build();
}
```

Since both beans have the same name, Spring's `DefaultListableBeanFactory` silently overrides the first one registered (SAML) with the second (x509). At TRACE level, this message
confirms it:

```
Overriding bean definition for bean 'securityFilterChain' with a different definition:
replacing [Root bean: class [null]; ... factoryBeanName=SAMLConfiguration$WebSecurityConfig; factoryMethodName=securityFilterChain; ...]
with      [Root bean: class [null]; ... factoryBeanName=X509Config; factoryMethodName=securityFilterChain; ...]
```

As a result, the active security filter chain only contains `X509AuthenticationFilter`. When a browser request arrives without a client certificate, the `Http403ForbiddenEntryPoin
t` rejects it with 403.

## Additional issue: filter chain ordering

Even after separating the two beans (by renaming one), there is a second problem: the x509 chain uses an `anyRequest()` matcher and is evaluated **before** the SAML chain. This me
ans x509 intercepts all requests (including port 8084 browser traffic) before SAML gets a chance to handle them.

The SAML chain uses a port-based `RequestMatcher` via `MultiAuthSupport` (matching port 8084), while x509 should only handle port 8085 (the API port). But since x509's `anyRequest
()` comes first, SAML is never reached.

## Suggested fix

The `@Bean` methods should have **unique names** to prevent the override:

```java
// In X509Config:
@Bean("x509SecurityFilterChain")  // or any unique name
public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception { ... }
```

Additionally, the x509 chain should either:
- Have `@Order(Ordered.LOWEST_PRECEDENCE)` to ensure it's evaluated after SAML
- Or use a port-based `RequestMatcher` matching only the API port (8085) instead of `anyRequest()`

## Workaround

We implemented a `BeanDefinitionRegistryPostProcessor` packaged as a JAR that:
1. Detects the x509 override and renames the bean to `x509SecurityFilterChain`
2. Restores the SAML bean definition as `securityFilterChain`
3. Reorders the filter chains in `FilterChainProxy` so SAML is evaluated before x509

The JAR is added to Gate's classpath via a Kubernetes deployment patch. Full source and scripts available.

## Related note: 2026.0.0 additional bug

Version 2026.0.0 has a separate issue in `SAMLConfiguration$WebSecurityConfig`: the `@EnableWebSecurity` annotation (which includes `@Configuration(proxyBeanMethods = false)`) pre
vents `@Autowired` field injection on beans created via inter-method calls. This causes a `NullPointerException` on `AllowedAccountsSupport.filterAllowedAccounts()` in `ResponseAu
thenticationConverter` after successful SAML authentication:

```
Cannot invoke "AllowedAccountsSupport.filterAllowedAccounts(String, Collection)" because "this.allowedAccountsSupport" is null
```

## Environment

- Halyard-deployed Spinnaker on Kubernetes
- SAML IdP: Auth0
- x509: mutual TLS for machine-to-machine API access
- Dual-port setup: 8084 (SAML/browser), 8085 (x509/API) via `default.apiPort`

vjda:
## Summary
- Fixes a bean name collision between SAML and x509 authentication in Gate when both are enabled.
- Renames x509's `SecurityFilterChain` bean so Spring no longer overrides the SAML chain.
- Moves x509 to `@Order(3)` so SAML's port-based chain runs before x509's catch-all matcher.

## Context
Issue #7518 reports two separate bugs in the X509/SAML configuration path. The NPE portion was already fixed in #7525. This PR addresses the remaining bean collision bug.

When both SAML and x509 are enabled, both configurations expose a bean named `securityFilterChain`. Spring silently overrides the SAML bean with the x509 one, which causes browser-based SAML authentication to fail with `403 Forbidden`.

## Changes
- rename `securityFilterChain` to `x509SecurityFilterChain` in `X509Config`
- change x509 filter chain order from `@Order(2)` to `@Order(3)`
- clarify the ordering comment to document that SAML must run before x509 when both are enabled

## Testing
- patch verified in a live environment using a custom container image (forked from 95829010)
- verified the resulting intended order is:
  1. Managed Delivery schema chain
  2. SAML chain
  3. x509 chain

Closes #7518


gate/gate-x509/src/main/java/com/netflix/spinnaker/gate/security/x509/X509Config.java
```diff
@@ -55,9 +55,11 @@ public class X509Config {
...
 
   @Bean
-  // ManagedDeliverySchemaEndpointConfiguration#schemaSecurityFilterChain should go first
-  @Order(2)
-  public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
+  // ManagedDeliverySchemaEndpointConfiguration#schemaSecurityFilterChain and
+  // SAMLConfiguration#securityFilterChain
+  // have higher priority when they are enabled.
+  @Order(3)
+  public SecurityFilterChain x509SecurityFilterChain(HttpSecurity http) throws Exception {
```
jasonmcintosh, line 62: Any chance of getting some tests for these?  I've got a SAML integration test, showing this with that would be awesome!

vjda:
I've implemented some tests and refactorized the SAMLConfigurationIntegrationTest class to extract SAML setup and some assertion to a common abstract class.

### What changed

The original `SAMLConfigurationIntegrationTest` was a monolithic class carrying ~380 lines of infrastructure alongside its actual test cases. With the new `SAMLAndX509ConfigurationIntegrationTest` being introduced to cover the SAML + X.509 coexistence scenario, duplicating all that scaffolding was simply not an option.

The refactor introduces **`AbstractSAMLConfigurationIntegrationTest`** — an abstract base class that centralizes all shared test infrastructure using the [Template Method pattern](https://refactoring.guru/design-patterns/template-method):

- **Keycloak Testcontainer lifecycle** — `@BeforeAll` / `@AfterAll` setup/teardown, realm provisioning via the Keycloak Admin REST API
- **`@DynamicPropertySource`** — wires the SAML metadata URL dynamically once the container is up
- **Mock beans** — `ClouddriverService`, `FiatService`, `Front50Service` stubs, shared across all subclasses
- **Selenium/HtmlUnitDriver flow** — the full SAML redirect + login assertion extracted into a reusable `assertSamlRedirectFlowWorks()` helper
- **`assertSamlPropertiesLoaded()`** — common property assertion for subclass use

Both concrete test classes now **extend the base** and focus exclusively on what makes them unique. Adding future SAML variant tests (e.g., SAML + OAuth2, SAML + LDAP) will be a matter of extending the base and writing the relevant assertions — no infrastructure copy-paste required.

This makes the test suite as maintainable as the production code it covers. 🙌

@jasonmcintosh, let me know if there's anything else you think is missing or if anything catches your eye in the approach I've taken. Thanks in advance!!

jasonmcintosh:
AWESOME let me look!!  Thank you!!

jasonmcintosh:
I'll probably end up doing some more... direct tests here later, but I'm good with this ;) Thank you!!

jasonmcintosh:
What I'd LIKE to get added and can do in other PRs later, or work on... is actually generating certs & setting up the secondary endpoint that normallly use with MTLS & testing it actually DOES auth correctly, but this is good for now!

jasonmcintosh:
@mergifyio backport release-2026.0.x release-2025.4.x

PR #7549, opened 2026-03-20
fix(builds): Fix spinnaker.io publishing

jasonmcintosh:
PRIMARY change is the git.ts where we strip the x-access-token for API operations to oktakat.  Bearer is just a change per Github docs recommendations for JWT usage.  The dist/ files was an NPM update of the files from dependabot upgrades.


jcavanagh:
Just wanted to note that the Action does need to be rebuilt after every code change (as well as Dependabot upgrades)  via `./gradlew buildGhaJs`.  GHA does not natively run TypeScript code, so the compiled JS is checked in.

jasonmcintosh:
@Mergifyio backport release-2025.4.x release-2026.0.x

PR #7554, opened 2026-03-23
feat(echo): add OTLP/gRPC transport for CDEvents notifications

tridentsx:
## Summary

Add support for sending CDEvents as OpenTelemetry spans via OTLP/gRPC, as an alternative to the existing HTTP CloudEvents broker transport.

## Motivation

Organizations using OTel-based observability backends (e.g. SigNoz, Jaeger, Grafana Tempo) can now receive Spinnaker pipeline execution events as OTel spans, enabling distributed tracing of pipeline executions without deploying a separate CloudEvents broker.

## Changes

- **`echo-notifications.gradle`** — Add OpenTelemetry SDK and OTLP exporter dependencies
- **`CDEventsConfigProperties`** — New config class for `cdevents.transport` (`http` or `otlp`)
- **`CDEventsOTelSender`** — Converts CloudEvents to OTel spans, exports via OTLP/gRPC. Caches SDK instances per endpoint.
- **`CDEventsNotificationAgent`** — Routes to HTTP or OTel sender based on transport config
- **`CDEventsNotificationAgentSpec`** — Tests for OTel transport routing and HTTP transport routing

## Configuration

```yaml
cdevents:
  enabled: true
  transport: otlp  # default: "http" (existing behavior, fully backward compatible)
```

The per-pipeline notification `address` field is reused as the OTLP endpoint when transport is `otlp`, so each pipeline can target a different OTel collector. No Deck/UI changes required.

## Backward Compatibility

- Default transport is `http` — existing behavior is unchanged
- No new UI fields or notification types
- Additive change behind a config flag

### DIFF
--- a/.github/actions/spinnaker-release/package-lock.json
+++ b/.github/actions/spinnaker-release/package-lock.json
@@ -693,9 +693,9 @@
       }
     },
     "node_modules/fast-xml-parser": {
-      "version": "5.5.6",
-      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.5.6.tgz",
-      "integrity": "sha512-3+fdZyBRVg29n4rXP0joHthhcHdPUHaIC16cuyyd1iLsuaO6Vea36MPrxgAzbZna8lhvZeRL8Bc9GP56/J9xEw==",
+      "version": "5.5.7",
+      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.5.7.tgz",
+      "integrity": "sha512-LteOsISQ2GEiDHZch6L9hB0+MLoYVLToR7xotrzU0opCICBkxOPgHAy1HxAvtxfJNXDJpgAsQN30mkrfpO2Prg==",
       "funding": [
         {
           "type": "github",
@@ -705,7 +705,7 @@
       "dependencies": {
         "fast-xml-builder": "^1.1.4",
         "path-expression-matcher": "^1.1.3",
-        "strnum": "^2.1.2"
+        "strnum": "^2.2.0"
       },
       "bin": {
         "fxparser": "src/cli/cli.js"
@@ -1187,9 +1187,9 @@
       }
     },
     "node_modules/strnum": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/strnum/-/strnum-2.1.2.tgz",
-      "integrity": "sha512-l63NF9y/cLROq/yqKXSLtcMeeyOfnSQlfMSlzFt/K73oIaD8DGaQWd7Z34X9GPiKqP5rbSh84Hl4bOlLcjiSrQ==",
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/strnum/-/strnum-2.2.1.tgz",
+      "integrity": "sha512-BwRvNd5/QoAtyW1na1y1LsJGQNvRlkde6Q/ipqqEaivoMdV+B1OMOTVdwR+N/cwVUcIt9PYyHmV8HyexCZSupg==",
       "funding": [
         {
           "type": "github",
@@ -1965,13 +1965,13 @@
       }
     },
     "fast-xml-parser": {
-      "version": "5.5.6",
-      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.5.6.tgz",
-      "integrity": "sha512-3+fdZyBRVg29n4rXP0joHthhcHdPUHaIC16cuyyd1iLsuaO6Vea36MPrxgAzbZna8lhvZeRL8Bc9GP56/J9xEw==",
+      "version": "5.5.7",
+      "resolved": "https://registry.npmjs.org/fast-xml-parser/-/fast-xml-parser-5.5.7.tgz",
+      "integrity": "sha512-LteOsISQ2GEiDHZch6L9hB0+MLoYVLToR7xotrzU0opCICBkxOPgHAy1HxAvtxfJNXDJpgAsQN30mkrfpO2Prg==",
       "requires": {
         "fast-xml-builder": "^1.1.4",
         "path-expression-matcher": "^1.1.3",
-        "strnum": "^2.1.2"
+        "strnum": "^2.2.0"
       }
     },
     "form-data": {
@@ -2298,9 +2298,9 @@
       }
     },
     "strnum": {
-      "version": "2.1.2",
-      "resolved": "https://registry.npmjs.org/strnum/-/strnum-2.1.2.tgz",
-      "integrity": "sha512-l63NF9y/cLROq/yqKXSLtcMeeyOfnSQlfMSlzFt/K73oIaD8DGaQWd7Z34X9GPiKqP5rbSh84Hl4bOlLcjiSrQ=="
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/strnum/-/strnum-2.2.1.tgz",
+      "integrity": "sha512-BwRvNd5/QoAtyW1na1y1LsJGQNvRlkde6Q/ipqqEaivoMdV+B1OMOTVdwR+N/cwVUcIt9PYyHmV8HyexCZSupg=="
     },
     "stubs": {
       "version": "3.0.0",
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,1 @@
+../AGENTS.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,133 @@
+# AI Coding Agent Instructions
+
+## Purpose
+- Spinnaker is an open-source continuous delivery platform for releasing software with high velocity and confidence.
+- Provides multi-cloud deployment support (AWS, GCP, Kubernetes, Azure, CloudFoundry, etc.) with automated canary analysis.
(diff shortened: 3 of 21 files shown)


echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  private OpenTelemetrySdk getOrCreateSdk(String endpoint) {
+    return sdkCache.computeIfAbsent(
+        endpoint,
+        ep -> {
+          OtlpGrpcSpanExporter exporter =
+              OtlpGrpcSpanExporter.builder()
+                  .setEndpoint(ep)
+                  .setTimeout(10, TimeUnit.SECONDS)
```
jasonmcintosh: Does this really need to be set or per https://javadoc.io/static/io.opentelemetry/opentelemetry-sdk/0.7.1/index.html?io/opentelemetry/sdk/trace/export/BatchSpanProcessor.html should this be done via system properties kinda thing?

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -0,0 +1,101 @@
...
+import java.nio.charset.StandardCharsets;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.stereotype.Component;
+
+/** Sends CDEvents as OTel spans to an OTLP/gRPC endpoint (e.g. SigNoz on port 4317). */
+@Slf4j
+@Component
```
jasonmcintosh, line 41: A concern... though it's minor - this may always be generated vs. conditional on the configuration.  A lot of spinnaker uses beans in the configuration class with a ConditionalOn enabled being true kinda thing.  This reduces the amount of objects loaded UNLESS they're FULLY enabled.  I'm not AGAINST this as it'd be needed for both cases, just calling this out.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -0,0 +1,101 @@
...
+public class CDEventsOTelSender {
+
+  private static final String INSTRUMENTATION_NAME = "spinnaker-cdevents";
+  private final Map<String, OpenTelemetrySdk> sdkCache = new ConcurrentHashMap<>();
```
jasonmcintosh, line 48: An additional note - this is an unbounded cache with no cleanup, so if URLs are very dynamic, there'll be several entries.  Given how this is MEANT to be used, I don't think that many and so not worried about it really.

echo/echo-notifications/echo-notifications.gradle
```diff
@@ -34,6 +34,9 @@ dependencies {
   implementation "io.cloudevents:cloudevents-http-basic:3.0.0"
   implementation "io.cloudevents:cloudevents-json-jackson:3.0.0"
   implementation ("dev.cdevents:cdevents-sdk-java:0.3.1")
+  implementation "io.opentelemetry:opentelemetry-api:1.40.0"
```
jasonmcintosh: NOMINALLY these get added via the spinnaker-dependenceis in kork.  ALSO it'd be nice if there were a BOM to define 1.40.0 as an overall constraint vs. needing to set the versions PER library.  NOT checked if the OTEL supports that (most libraries lately are adding BOMs for platform constraints)

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
jasonmcintosh: An alternative is to added it to the configuration properties just to allow this to be set.  I've seen LONG timeouts cause issues on backlog on failures on remote URLs

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -0,0 +1,101 @@
...
+
+    Attributes attrs =
+        Attributes.of(
+            AttributeKey.stringKey("cdevents.type"), cdEvent.getType(),
+            AttributeKey.stringKey("cdevents.source"), cdEvent.getSource().toString(),
+            AttributeKey.stringKey("cdevents.id"), cdEvent.getId());
+
+    if (cdEvent.getData() != null) {
+      String payload = new String(cdEvent.getData().toBytes(), StandardCharsets.UTF_8);
+      attrs = attrs.toBuilder().put(AttributeKey.stringKey("cdevents.data"), payload).build();
```
jasonmcintosh, line 66: ONE thing I've not checked, and it's been an old concern.  The amount of potential data here as an "attribute" as  I'm not sure the size of this data, what's IN this data, etc.  IF this is say the full PIPELINE output, this could be MASSIVE and... potentially a problem I'd THINK?  Need to look at this a bit more - I think i raised this in the past on the normal CDEvents but not looked recently...

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Good point. I've made the timeout configurable via cdevents.otlpTimeoutSeconds in the config properties (defaults to 10s),  so its possible to tune it without needing system properties.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Agreed,
added @ConditionalOnProperty(name = "cdevents.transport", havingValue = "otlp") so the bean is only loaded when OTLP transport is actually configured. 
The autowire in the agent is now required = false to match.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Agreed it's fine in practice,
endpoints come from static per-pipeline notification config. 
Added a comment in the code noting this for future readers.

echo/echo-notifications/echo-notifications.gradle
tridentsx: Switched to platform("io.opentelemetry:opentelemetry-bom:1.40.0") so versions are managed centrally.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: The CDEvent data payload is constructed from a fixed set of metadata fields (execution ID, name, URL, status/outcome) plus an explicit customData map that operators configure per-notification. 
It doesn't include the full pipeline execution output. The data size is bounded by what the user intentionally puts in customData — same data that's already sent over the HTTP transport today.

echo/echo-notifications/echo-notifications.gradle
```diff
@@ -34,6 +34,10 @@ dependencies {
   implementation "io.cloudevents:cloudevents-http-basic:3.0.0"
   implementation "io.cloudevents:cloudevents-json-jackson:3.0.0"
   implementation ("dev.cdevents:cdevents-sdk-java:0.3.1")
+  implementation platform("io.opentelemetry:opentelemetry-bom:1.40.0")
```
jasonmcintosh: I think last thing - can this move to kork/spinnaker-dependencies/spinnaker-dependencies.gradle :) This'll then impact the whole project to set the OTEL library version restrictions.

tridentsx: Since kork is a separate repo, should I submit a separate PR to kork/spinnaker-
dependencies first and then rebase this one on top? Or is it fine to modify the kork subtree directly in this PR?

jasonmcintosh: so it's in this monorepo - whole thing is a monorepo :) AKA no need to do separate PR, just this file:
https://github.com/spinnaker/spinnaker/blob/main/kork/spinnaker-dependencies/spinnaker-dependencies.gradle

tridentsx: Done, moved the OTel BOM to kork/spinnaker-dependencies/spinnaker-dependencies.gradle and removed it from echo-notifications. Thanks for clarifying the monorepo setup

jasonmcintosh:
OK code looks all good - but needs a formatting fix
```./gradlew spotlessApply``` should do it!

tridentsx:
> OK code looks all good - but needs a formatting fix `./gradlew spotlessApply` should do it!

Ran spotlessApply, formatting fixed. Thanks!

jasonmcintosh (approved):

jasonmcintosh:
Thanks for the PR!!

PR #7556, opened 2026-03-24
feat(auth): Api token system

jasonmcintosh:
VERY preliminary concept allowing users to generate an API token for access instead of needing to setup MTLS.  This GREATLY simplifies API operations for the average user.  Tokens could be issued UP TO a predefined limit.  Admins can list all tokens/delete tokens, while users can list their tokens & delete their tokens only.  

This would obviously need tests & additional input, but seeing if there's enough interest to add this to spinnaker...


jasonmcintosh:
Closing in favor of https://github.com/spinnaker/spinnaker/pull/7683 :)

PR #7557, opened 2026-03-24
perf(gate/header): skip login of service account users when using header authentication

dbyron-sf:
Before this PR, header auth in gate (implemented in https://github.com/spinnaker/spinnaker/pull/7109) calls [permissionService.login](https://github.com/spinnaker/spinnaker/blob/e83241968f76d83adebd0e516a73d0e383fce5a4/gate/gate-header/src/main/java/com/netflix/spinnaker/gate/security/header/HeaderAuthenticationUserDetailsService.java#L81), which becomes a [POST to fiat's /roles/{userId} endpoint](https://github.com/spinnaker/spinnaker/blob/e83241968f76d83adebd0e516a73d0e383fce5a4/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/PermissionService.java#L82) (via [FiatService.loginUser](https://github.com/spinnaker/spinnaker/blob/e83241968f76d83adebd0e516a73d0e383fce5a4/fiat/fiat-api/src/main/java/com/netflix/spinnaker/fiat/shared/FiatService.java#L101)). This effectively means that every call to gate that uses header auth causes a call to fiat. This is a lot more traffic than happens using oauth. [The logic there](https://github.com/spinnaker/spinnaker/blob/e83241968f76d83adebd0e516a73d0e383fce5a4/gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/OAuthUserInfoServiceHelper.java#L192-L193) is similar, but there's no call to fiat for service accounts.

This PR adds similar logic for header auth, skipping login with fiat for service accounts. It determines service accounts by periodically calling front50 and caching the results in memory.  There's a new configuration property `header.serviceAccountCacheTimeoutMinutes` that defaults to 60 to control this behavior.

### DIFF
--- a/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/model/front50/ServiceAccountPojo.java
+++ b/gate/gate-core/src/main/java/com/netflix/spinnaker/gate/model/front50/ServiceAccountPojo.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2026 Salesforce, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.gate.model.front50;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * A simple POJO representing a service account with just the name field. This is a subset of
+ * front50's ServiceAccount class (which implements Timestamped) and is used to deserialize
+ * responses from front50's /serviceAccounts endpoint when only the name is needed.
+ */
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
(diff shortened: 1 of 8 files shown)


jasonmcintosh:
SO there's.... another implementation that was done in MTLS that probably should look at as an alternative approach
https://github.com/spinnaker/spinnaker/blob/e83241968f76d83adebd0e516a73d0e383fce5a4/gate/gate-x509/src/main/groovy/com/netflix/spinnaker/gate/security/x509/X509AuthenticationUserDetailsService.groovy#L145-L156
Login debounce handling here.

jasonmcintosh:
I'll note to the above, spinnaker service accounts were NEVER supposed to be IDP associated accounts, aka they were nominally fixed internal accounts independent of IDP configured accounts.  Otherwise you'd get into privilege potential issues where the service account would have it's permissions scopes suddenly expanded/changed due to a login, impacting triggers referencing said accounts...

dbyron-sf:
> SO there's.... another implementation that was done in MTLS that probably should look at as an alternative approach

Looks like this defaults to off...and at the moment I can't quite see exactly how it's different / better.  Can we start with this and if change it down the road if necessary?

jasonmcintosh:
NOT against this for now given that it should only trigger on that header auth which ... I don't see many using?

jasonmcintosh:
ALSO IMHO debounce SHOULD default to on - when using MTLS makes a WORLD of perf difference.

dbyron-sf:
> NOT against this for now given that it should only trigger on that header auth which ... I don't see many using?

Yeah, it's pretty new.  It could just be us / Salesforce.

jasonmcintosh:
Yeah IAP auth i see FAIRLY regularly - google's got a nice setup there :) OK approved

PR #7560, opened 2026-03-24
fix(clouddriver/k8s): throw an exception if kubectl returns an empty output

mattgogerly:
If kubectl returns nothing (e.g. when there was an issue with the k8s CDN causing the kubectl binary to be an empty file) Clouddriver currently returns a 404 on getManifest. Teach it to throw an exception instead.


jasonmcintosh:
INTERESTINGLY enough - I think this USED to throw an exception back before... 1.36...

jasonmcintosh:
E.g. there's a job output area which currently ignores empty instead of throwing an exception.  That USUALLY has indicated an issue with the API server and IS a real failure, but sometimes was also when there was no job output....

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
```diff
@@ -444,7 +444,18 @@ public KubernetesManifest get(
     }
 
     try {
-      return gson.fromJson(status.getOutput(), KubernetesManifest.class);
+      String output = status.getOutput();
```
dbyron-sf, line 447: Wouldn't it make more sense to put this inside executeKubectlCommand?  Isn't it an issue for more than just `kubectl get` ?

PR #7561, opened 2026-03-25
feat(echo): add mTLS support for OTLP/gRPC CDEvents transport

tridentsx:
## Summary

Add optional mTLS (mutual TLS) support for the OTLP/gRPC CDEvents transport introduced in #7554.

## Motivation

Organizations running OTel collectors behind mTLS (e.g. in zero-trust environments) need Spinnaker to present a client certificate when exporting spans. This enables secure, authenticated telemetry export without relying on network-level trust.

## Changes

- **`CDEventsConfigProperties`** — Add `otlpCaCertPath`, `otlpClientCertPath`, `otlpClientKeyPath` for PEM file paths
- **`CDEventsOTelSender`** — Add `configureTls()` method that calls the OTel SDK's native `setTrustedCertificates`/`setClientTls` APIs when cert paths are configured
- **`CDEventsNotificationAgentSpec`** — Add test for mTLS config routing

## Configuration

```yaml
cdevents:
  enabled: true
  transport: otlp
  otlp-ca-cert-path: /etc/otel/certs/ca.crt
  otlp-client-cert-path: /etc/otel/certs/client.crt
  otlp-client-key-path: /etc/otel/certs/client.key
```

## Backward Compatibility

- When no cert paths are set, behavior is unchanged (plain gRPC or system-default TLS)
- No new dependencies — uses the OTel SDK's built-in TLS support
- Additive change behind optional config


echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsConfigProperties.java
```diff
@@ -35,4 +35,13 @@ public class CDEventsConfigProperties {
 
   /** OTLP exporter timeout in seconds. */
   private long otlpTimeoutSeconds = 10;
+
+  /** Path to PEM file containing trusted CA certificate(s) for verifying the OTLP collector. */
+  private String otlpCaCertPath;
+
+  /** Path to PEM file containing the client certificate for mTLS. */
```
jasonmcintosh, line 42: I'll note from the docs:

Sets ths client key and the certificate chain to use for verifying client when TLS is enabled. The key must be PKCS8, and both must be in PEM format.

May be ... well understood :) AFTER this a docs update would be helpful!!

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -91,20 +95,38 @@ private OpenTelemetrySdk getOrCreateSdk(String endpoint) {
...
+                  .addSpanProcessor(BatchSpanProcessor.builder(builder.build()).build())
                   .setResource(
                       Resource.builder().put("service.name", "spinnaker-echo-cdevents").build())
                   .build();
 
           return OpenTelemetrySdk.builder().setTracerProvider(tracerProvider).build();
         });
   }
+
+  private void configureTls(OtlpGrpcSpanExporterBuilder builder) {
```
jasonmcintosh, line 135: ONE concern at a glance - this is reloading the certs on EVERY send.  Wondering... if that's really needed.  Loading the bytes on init would likely reduce some cost on these operations.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsConfigProperties.java
tridentsx: Updated the Javadoc to note the PKCS8 requirement. Happy to do a docs update as a follow-up once this
lands.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: The file reads were actually inside sdkCache.computeIfAbsent() so it was once per endpoint, not every send — but your point still stands. Moved the file reads to the constructor so bytes are loaded once at startup. configureTls() now just passes the pre-loaded byte arrays to the builder. 

This offcourse requires a restart, if thats not desirable I could hot-reload by reading the files in configureTls() but caching with a TTL — but I keept it simple for now.

jasonmcintosh: NORMALLY spring apps have a /refresh endpoint - I've actually tested that myself that config properties support for these kinds of things but for now, I tend to vote "restarts".  This tends to allow a careful "rotation" process for new certs that is ran through health checks & deployments & separate config volumes vs. updating an existing config. This is actually a core design on spinnaker deployments even.

PR #7567, opened 2026-03-26
fix(gate/oauth2): restore /login endpoint behavior

dbyron-sf:
to what was present before https://github.com/spinnaker/spinnaker/pull/7052.

Specifically:
* GET /login with a bearer token authenticates the user via the user-info endpoint and redirects to the base URL.
* GET /login without a bearer token redirects to the OAuth2 provider's authorization endpoint

See https://github.com/spinnaker/spinnaker/pull/7567/commits/d6c43666b4dc1936c3d03824169af0706e3ff4b0 for the behavior after https://github.com/spinnaker/spinnaker/pull/7052, but before this PR.

### DIFF
--- a/gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/OAuth2SsoConfig.java
+++ b/gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/OAuth2SsoConfig.java
@@ -34,7 +34,6 @@ import org.springframework.security.config.annotation.web.builders.HttpSecurity;
 import org.springframework.security.config.annotation.web.configuration.EnableWebSecurity;
 import org.springframework.security.oauth2.client.endpoint.OAuth2AccessTokenResponseClient;
 import org.springframework.security.oauth2.client.endpoint.OAuth2AuthorizationCodeGrantRequest;
-import org.springframework.security.oauth2.client.registration.ClientRegistration;
 import org.springframework.security.oauth2.client.registration.ClientRegistrationRepository;
 import org.springframework.security.oauth2.client.registration.InMemoryClientRegistrationRepository;
 import org.springframework.security.oauth2.client.web.OAuth2AuthorizationRequestRedirectFilter;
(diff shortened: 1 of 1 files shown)


gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/ExternalAuthTokenFilter.java
```diff
@@ -74,10 +75,28 @@ public ExternalAuthTokenFilter(
...
+    // configured at startup and does not change.
+    ClientRegistration clientRegistration =
+        clientRegistrationRepository.findByRegistrationId(registrationId);
+    if (clientRegistration == null) {
+      throw new IllegalArgumentException(
+          "No client registration found for registrationId: " + registrationId);
+    }
+    this.clientRegistration = clientRegistration;
+
+    String userInfoUri = clientRegistration.getProviderDetails().getUserInfoEndpoint().getUri();
```
jasonmcintosh, line 94: I WILL note this is NORMALLY found via the well-known endpoints, aka this should be automatic via discovery mechanisms...

gate/gate-oauth2/src/main/java/com/netflix/spinnaker/gate/security/oauth2/ExternalAuthTokenFilter.java
```diff
@@ -111,6 +130,13 @@ protected void doFilterInternal(
           if (authentication != null) {
             SecurityContextHolder.getContext().setAuthentication(authentication);
             log.debug("Successfully authenticated user via external bearer token");
+
+            // /login has no controller on the oauth2Login() stack, so redirect
+            // authenticated users to the base URL instead of continuing into a 404.
+            if ("/login".equals(request.getRequestURI())) {
```
jasonmcintosh, line 136: FYI I've got some of this flow for SAML documented.  We could probably replicate what I did with saml using keycloak to get a real oauth2 integration test as well to make sure this doesn't break again in the future...

PR #7591, opened 2026-03-31
chore(clouddriver): centralize version of json-flattener

dbyron-sf:
since everything gets the newest version anyway, 0.14.2 in this case, e.g.
```
|    +--- com.github.wnameless.json:json-flattener:0.11.1 -> 0.14.2 (*)
```

No change to the actual version used by clouddriver.

### DIFF
--- a/.github/workflows/clean-untagged-images.yml
+++ b/.github/workflows/clean-untagged-images.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       # configure based on your registry
       - name: Login to GHCR
-        uses: docker/login-action@v3 # Use the latest stable version
+        uses: docker/login-action@v4 # Use the latest stable version
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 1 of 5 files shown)


clouddriver/clouddriver-configserver/clouddriver-configserver.gradle
```diff
@@ -17,5 +17,5 @@ dependencies {
   implementation "org.apache.commons:commons-lang3"
   implementation "org.springframework.cloud:spring-cloud-context"
   implementation "org.springframework.cloud:spring-cloud-config-server"
-  implementation "com.github.wnameless.json:json-flattener:0.14.2"
+  implementation "com.github.wnameless.json:json-flattener:${jsonFlattenerVersion}"
```
jasonmcintosh, line 20: SHOULD this be moved to spinnaker-dependencies?

jasonmcintosh: OR the bom?

dbyron-sf: It's clouddriver-only, so...it could, but I think it's OK in clouddriver's gradle.properties.

PR #7595, opened 2026-04-01
feat(spin/iap): add token caching for IAP authentication

himanhsugusain:
Cache IAP ID tokens to disk to avoid repeated authentication flows on
   each CLI invocation. Follows the existing OAuth2/GSA caching pattern.

   - Add CachedIDToken struct with expiry validation
   - Parse JWT exp claim to determine token validity
   - Store cached token in ~/.spin/config
   - Capture refresh token from interactive OAuth flow
   - Request offline access to obtain refresh tokens

### DIFF
--- a/.github/workflows/clean-untagged-images.yml
+++ b/.github/workflows/clean-untagged-images.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       # configure based on your registry
       - name: Login to GHCR
-        uses: docker/login-action@v3 # Use the latest stable version
+        uses: docker/login-action@v4 # Use the latest stable version
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
@@ -43,7 +43,7 @@ jobs:
     steps:
       # configure based on your registry
       - name: Login to GAR
-        uses: 'docker/login-action@v3'
+        uses: 'docker/login-action@v4'
         with:
           registry: us-docker.pkg.dev
           username: _json_key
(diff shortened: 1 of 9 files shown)


spin/config/auth/iap/token_utils.go
```diff
@@ -0,0 +1,68 @@
...
+
+// jwtClaims represents the relevant claims from an ID token.
+type jwtClaims struct {
+	Exp int64 `json:"exp"`
+}
+
+// ParseIDTokenExpiry extracts the expiration time from a JWT ID token.
+// JWTs have three parts separated by dots: header.payload.signature.
+// The payload contains the exp claim as a Unix timestamp.
+func ParseIDTokenExpiry(idToken string) (time.Time, error) {
```
jasonmcintosh, line 25: WONDERING if there isn't a library for this already someplace?

jasonmcintosh:
Barring if there's a better library to use vs. custom parsing I"m good with this ;)

spin/config/auth/iap/token_utils_test.go
```diff
@@ -0,0 +1,161 @@
+// Copyright (c) 2018, Snap Inc.
```
jasonmcintosh: ONE nit:  THINKING this should be glean since y'all wrote it ;)

spin/config/auth/iap/token_utils.go
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) 2018, Snap Inc.
```
jasonmcintosh: Ditto here ;)

spin/config/auth/iap/token_utils.go
himanhsugusain: there was option to use google's library, but did not want it to be gcp-iap specific so used `github.com/golang-jwt/jwt/v5"`

spin/config/auth/iap/token_utils_test.go
himanhsugusain: Don't know much how the licences work, but I have the commits signed w/ my personal github so will just attribute this to me. :)

PR #7616, opened 2026-04-10
Fix Bearer Token auth for Spring Boot 3

### LINKED ISSUE #7390: Spring Boot 3 Breaks Bearer Token API Authentication
After upgrading to Spinnaker 2025.4.0, Bearer token authentication for Spinnaker API access no longer works:
* Looks like it is related to the Spring Boot 3 upgrade.  See: [7377](https://github.com/spinnaker/spinnaker/issues/7377).
* The OAuth2 client filter intercepts all requests before bearer tokens can be validated.  AFAICT, there is a no workaround in `gate-local.yml`
* I suspect we need to implement custom `OpaqueTokenIntrospector` for Google's non-standard tokeninfo endpoint, or configure separate security filter chains for API vs browser access.

utafrali:
Fixes #7390

Spring Boot 3 removed `OAuth2LoginAuthenticationFilter`, which broke Bearer Token API authentication. Switched to `OAuth2AuthorizationRequestRedirectFilter` when setting up the external auth token filter.

Tested with Spring Boot 3, Bearer token auth works correctly.


jasonmcintosh:
So... wondering if we couldn't get an integration test to show this.  I'm inclined to approve it regardless, given details, but really wondering how we can NOT have this hit in the future.  I've created one for SAML - I think we could use almost the same exact test but for an OAuth2 configuration (using keycloak) to show this issue better.

dbyron-sf:
@utafrali Can you explain exactly what scenario / sequence of http requests / cookies / responses / redirects you're seeing and why you think it's incorrect?  I bet we can build an automated test from there.

dbyron-sf:
More specifically, there are tests in [OAuth2IntegrationWithWireMockTest](https://github.com/spinnaker/spinnaker/blob/113494e3de0f32bb989d3fa3ae2f7fe862016b6b/gate/gate-web/src/test/java/com/netflix/spinnaker/gate/security/oauth/OAuth2IntegrationWithWireMockTest.java) that show the behavior of some flows.  Are there assertions there that are incorrect, or is there some other scenario that's not working?

utafrali:
Looking at those tests, this scenario isn't covered - I'll add a test that reproduces it.

jasonmcintosh:
@utafrali any chance of adding those tests?

jasonmcintosh:
Given integration tests that help us... closing this until we have more details/replication also tied to conversation via linked in :)

PR #7617, opened 2026-04-10
feat(echo): add trace hierarchy, stage names and customData promotion to OTLP CDEvents spans

tridentsx:
## What

Enhances the OTLP/gRPC CDEvents transport in Echo with producer-side trace correlation, stage name enrichment, and customData attribute promotion.

## Why

When CDEvents are delivered over the **HTTP transport**, the receiving broker (e.g. Keptn, Knative) correlates related events using the `subject.id` field in the CloudEvent payload — correlation is the consumer's responsibility.

**OTLP/OpenTelemetry backends** (Jaeger, SigNoz, Grafana Tempo) work differently: they expect the *producer* to establish trace correlation via shared trace/span IDs. Without this, every CDEvent appears as an independent root span, making it impossible to visualise a pipeline execution as a single trace.

### Trace hierarchy
- Pipeline start events (`pipelinerun.queued/started`) create a root span and store its `SpanContext` keyed by execution ID
- Stage events (`taskrun.started/finished`) and pipeline finish events become children of the root span
- Stored context is cleaned up when `pipelinerun.finished` is sent

### Stage name enrichment
- Span names include the stage name: `dev.cdevents.taskrun.started.Build` instead of `dev.cdevents.taskrun.started.0.1.1`
- `cdevents.stage.name` attribute added for filtering
- `cdevents.config.type` attribute indicates `pipeline` or `stage`

### customData attribute promotion
- Entries from the CDEvent `customData` map are promoted to individual `cdevents.custom.<key>` span attributes
- CI variables (commit SHA, repo, Helm chart version, Jenkins build number) become first-class searchable attributes in trace UIs
- Example: `customData: {"commit.sha": "abc123"}` → span attribute `cdevents.custom.commit.sha=abc123`

### No impact on HTTP transport
- `CDEventsNotificationAgent` only passes extra context inside the `if (otlp)` branch; the HTTP `else`-branch is unchanged
- `CDEventsOTelSender` is gated by `@ConditionalOnProperty(transport=otlp)` — never instantiated for HTTP
- Backwards-compatible `send(CloudEvent, String)` overload retained
- All 28 unit tests pass (27 existing + 1 new customData test)

## Trace visualisation

```
pipelinerun.started                (root span)
├── taskrun.started.Build          (child)
├── taskrun.finished.Build         (child)
├── taskrun.started.Helm Deploy    (child, with cdevents.custom.helm.*)
├── taskrun.finished.Helm Deploy   (child)
└── pipelinerun.finished           (child)
```


echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -40,16 +44,48 @@
...
 @ConditionalOnProperty(name = "cdevents.transport", havingValue = "otlp")
 public class CDEventsOTelSender {
 
   private static final String INSTRUMENTATION_NAME = "spinnaker-cdevents";
+  private static final ObjectMapper mapper = new ObjectMapper();
   // Unbounded cache keyed by endpoint URL. In practice, endpoints are static per-pipeline config,
   // so the number of entries is small and bounded by the number of distinct OTLP endpoints.
   private final Map<String, OpenTelemetrySdk> sdkCache = new ConcurrentHashMap<>();
+  // Tracks the root pipeline span context keyed by execution ID so stage spans become children.
+  private final Map<String, SpanContext> pipelineSpans = new ConcurrentHashMap<>();
```
xibz, line 88: Rather than storing SpanContext objects in a map, OTel's TextMapPropagator can encode the trace context as a plain W3C traceparent string. It's serializable, durable-store-friendly, and the idiomatic way to carry context across async boundaries:

```
// Instead of:
private final Map<String, SpanContext> pipelineSpans = new ConcurrentHashMap<>();

// Use:
private static final TextMapPropagator PROPAGATOR = W3CTraceContextPropagator.getInstance();
private final Map<String, String> pipelineTraceContext = new ConcurrentHashMap<>();

// On pipelinerun.started encode context into a traceparent string:
Span span = spanBuilder.startSpan();
Map<String, String> carrier = new HashMap<>();
PROPAGATOR.inject(Context.root().with(span), carrier, Map::put);
pipelineTraceContext.put(executionId, carrier.get("traceparent"));
span.end();

// On stage events restore context and parent the span:
String traceparent = pipelineTraceContext.get(executionId);
if (traceparent != null) {
Context parentCtx = PROPAGATOR.extract(
  Context.root(), Map.of("traceparent", traceparent), Map::get);
spanBuilder.setParent(parentCtx);
}

// On pipelinerun.finished:
pipelineTraceContext.remove(executionId);
```

The traceparent value (e.g. `00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01`) is just a string, so this map could trivially be backed by Redis or a DB if you want it to survive restarts. The SpanContext map can't be serialized at all.

xibz (changes requested):
Thanks for the PR. I had one major comment that I think we need to think about. The OTEL libraries have a lot of features that we can utilize to remove reimplementing the same thing that's already done in those libraries. Have you taken a look at the propagators in OTEL?

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Great catch — you're absolutely right that `SpanContext` is not serializable and using it here was a mistake.

I've switched to `W3CTraceContextPropagator` as you suggested. The `pipelineSpans` map is now `Map<String, String>` storing traceparent strings (`00-{traceId}-{spanId}-{flags}`). Pipeline start injects via `PROPAGATOR.inject()`, and child spans restore the parent via `PROPAGATOR.extract()`.

One thing worth noting: the `TextMapGetter` passed to `extract()` can't be a `Map::get` method reference since `TextMapGetter` has two abstract methods (`keys()` and `get()`), so it's not a functional interface. I added a small static `MAP_GETTER` instance to handle that.

As you mentioned, this also opens the door to backing the map with Redis/DB for restart resilience — the traceparent string is just plain text. I'll keep that as a follow-up if we want cross-instance or restart-safe correlation.

Also added `CDEventsOTelSenderSpec` with 8 tests covering trace correlation (parent-child linking, cleanup on finish), stage name enrichment, customData attribute promotion, and error status on failure outcomes. All 147 tests pass.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -33,23 +39,70 @@
...
+ * <p>Unlike the HTTP transport where the receiving broker (e.g. Knative, Keptn) correlates related
+ * CloudEvents using the {@code subject.id} field, OTLP/OpenTelemetry backends (e.g. Jaeger, SigNoz,
+ * Grafana Tempo) expect trace correlation to be established by the producer via shared trace/span
+ * IDs. This class therefore maintains a mapping of Spinnaker execution IDs to W3C {@code
+ * traceparent} strings so that all CDEvents belonging to the same pipeline execution appear as a
+ * single trace:
+ *
+ * <pre>
+ *   pipelinerun.started          (root span)
+ *   ├── taskrun.started.Build    (child)
```
xibz: The format is a little misleading at least from a CDEvents perspective. If this is an OTEL extension please let me know but `taskrun.started` does not have a `Build` component to it.

Additionally, `taskrun.started.Test` is describing what? CDEvents has test events so just want to make sure we are not describing something incorrectly.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -33,23 +39,70 @@
...
+ * in {@link com.netflix.spinnaker.echo.notification.CDEventsNotificationAgent} is completely
+ * unaffected.
+ */
 @Slf4j
 @Component
 @ConditionalOnProperty(name = "cdevents.transport", havingValue = "otlp")
 public class CDEventsOTelSender {
 
   private static final String INSTRUMENTATION_NAME = "spinnaker-cdevents";
+  private static final ObjectMapper mapper = new ObjectMapper();
```
xibz: Do we not have a bean somewhere for the object mapper?

tridentsx:
@xibz I've addressed your feedback by switching to W3CTraceContextPropagator — could you re-review when you get a chance? All CI checks are passing.

xibz:
@tridentsx - Yep, still reviewing. Hopped off of it last week, but should have it wrapped up sometime this week

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -75,39 +128,154 @@ public CDEventsOTelSender(CDEventsConfigProperties config) {
...
+    if (type != null && type.contains("finished")) {
       if (cdEvent.getData() != null) {
         String data = new String(cdEvent.getData().toBytes(), StandardCharsets.UTF_8);
         if (data.contains("\"outcome\":\"failure\"")) {
           span.setStatus(StatusCode.ERROR);
         }
       }
+      // Clean up stored context when pipeline finishes
+      if (isPipelineEnd && executionId != null) {
+        pipelineTraceContext.remove(executionId);
```
xibz: So if the pipeline gets lost or end is never received, I think this leaks?

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -75,39 +128,154 @@ public CDEventsOTelSender(CDEventsConfigProperties config) {
...
-            .setAllAttributes(attrs)
-            .startSpan();
+    var spanBuilder =
+        tracer.spanBuilder(spanName).setSpanKind(SpanKind.PRODUCER).setAllAttributes(attrs);
 
-    if (cdEvent.getType() != null && cdEvent.getType().contains("finished")) {
-      // Check data payload for outcome to set span status
+    // Parent child spans under the pipeline root span
+    boolean isPipelineStart =
+        type != null && type.contains("pipelinerun") && !type.contains("finished");
```
xibz: Would this match pipelinerun.queued?

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -75,39 +128,154 @@ public CDEventsOTelSender(CDEventsConfigProperties config) {
...
+    if (isPipelineStart && executionId != null) {
+      Map<String, String> carrier = new HashMap<>();
+      PROPAGATOR.inject(Context.root().with(span), carrier, Map::put);
+      pipelineTraceContext.put(executionId, carrier.get("traceparent"));
+    }
+
+    if (type != null && type.contains("finished")) {
       if (cdEvent.getData() != null) {
         String data = new String(cdEvent.getData().toBytes(), StandardCharsets.UTF_8);
         if (data.contains("\"outcome\":\"failure\"")) {
```
xibz: This is very fragile. You dont know the canonicalization, whitespace, formating. You're safer bet is parsing `JsonNode root = mapper.readTree(data);` then looking for the node directly.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -75,39 +128,154 @@ public CDEventsOTelSender(CDEventsConfigProperties config) {
...
+   * @param config Orca event config containing the event type ("pipeline" or "stage"), may be null
+   * @param event the original Orca event, used to extract stage name from {@code
+   *     content.context.stageDetails.name}, may be null
+   */
+  @SuppressWarnings("unchecked")
+  public void send(
+      CloudEvent cdEvent,
+      String otlpEndpoint,
+      Map<String, String> config,
+      com.netflix.spinnaker.echo.api.events.Event event) {
```
xibz, line 136: Why is this using the full path?

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -75,39 +128,154 @@ public CDEventsOTelSender(CDEventsConfigProperties config) {
...
       }
+      // Clean up stored context when pipeline finishes
+      if (isPipelineEnd && executionId != null) {
+        pipelineTraceContext.remove(executionId);
+      }
     }
 
     span.end();
-    log.info("Sent CDEvent {} as OTel span to {}", cdEvent.getType(), otlpEndpoint);
+    log.info("Sent CDEvent {} as OTel span to {}", type, otlpEndpoint);
```
xibz: This is going to be A LOT. I'd opt to move this to debug.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -33,23 +39,70 @@
...
 @ConditionalOnProperty(name = "cdevents.transport", havingValue = "otlp")
 public class CDEventsOTelSender {
 
   private static final String INSTRUMENTATION_NAME = "spinnaker-cdevents";
+  private static final ObjectMapper mapper = new ObjectMapper();
   // Unbounded cache keyed by endpoint URL. In practice, endpoints are static per-pipeline config,
   // so the number of entries is small and bounded by the number of distinct OTLP endpoints.
   private final Map<String, OpenTelemetrySdk> sdkCache = new ConcurrentHashMap<>();
+  // Tracks the root pipeline trace context keyed by execution ID so stage spans become children.
+  private static final TextMapPropagator PROPAGATOR = W3CTraceContextPropagator.getInstance();
```
xibz: I actually dont know if this solves my original concern. I thought OTEL handled state. By storing state we assume a single pod is going to map 1:1 to incoming requests. That wont be the case.

That means if we have 2 echo pods, and single orca pod that sends an event to pod 1, then a finished event to pod 2, the containing map will not have it?

xibz (changes requested):

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: You're right — storing state in a local map doesn't work across pods. I've completely removed the `ConcurrentHashMap`.

The new approach derives trace and span IDs **deterministically** from the execution ID using SHA-256:
- `traceId = SHA-256(executionId)[0:16]` → 32 hex chars
- `rootSpanId = SHA-256(executionId)[16:24]` → 16 hex chars

Every Echo pod produces the same trace hierarchy for the same execution without any shared state. No map, no Redis, no leak — fully stateless.

The execution ID is already present in every event Orca sends (`event.content.execution.id`), so this works regardless of which pod handles which event.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Fixed — there's no map at all now. Trace IDs are derived on-the-fly from the execution ID, so nothing is stored and nothing can leak.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: No longer relevant — with the deterministic approach, all spans (queued, started, finished, taskrun.*) get the same trace ID from the execution ID uniformly. No special-casing needed.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Fixed — now using `objectMapper.readTree(data)` → `root.path("outcome")` for proper JSON parsing. No more string contains.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Fixed — now injected via constructor: `CDEventsOTelSender(CDEventsConfigProperties config, ObjectMapper objectMapper)`. Uses the Spring-managed bean.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Fixed — now strips the version suffix. `cdevents.type` attribute uses the short form (e.g. `dev.cdevents.taskrun.started` instead of `dev.cdevents.taskrun.started.0.1.1`).

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Fixed — moved to `log.debug`.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
tridentsx: Good point — I've changed the format to use brackets: `taskrun.started [Build]` instead of `taskrun.started.Build`. This makes it clear the stage name is metadata and not part of the CDEvents type. The version suffix is also stripped from the span name now.

tridentsx:
You're right — storing state in a local map doesn't work across pods. I've completely removed the ConcurrentHashMap.

The new approach derives trace and span IDs deterministically from the execution ID using SHA-256:

traceId = SHA-256(executionId)[0:16] → 32 hex chars
rootSpanId = SHA-256(executionId)[16:24] → 16 hex chars
Every Echo pod produces the same trace hierarchy for the same execution without any shared state. No map, no Redis, no leak — fully stateless.

The execution ID is already present in every event Orca sends (event.content.execution.id), so this works regardless of which pod handles which event.

echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsOTelSender.java
```diff
@@ -75,39 +111,182 @@ public CDEventsOTelSender(CDEventsConfigProperties config) {
...
+    if (context == null) return null;
+    Map<String, Object> stageDetails = (Map<String, Object>) context.get("stageDetails");
+    if (stageDetails == null) return null;
+    return (String) stageDetails.get("name");
+  }
+
+  private String extractExecutionId(CloudEvent cdEvent) {
+    if (cdEvent.getData() == null) return null;
+    try {
+      JsonNode root = objectMapper.readTree(cdEvent.getData().toBytes());
```
xibz, line 257: Why are we converting this to JSON just to get the subject id?

xibz (dismissed):
I had one comment, but I think another thing we have to document is that we expect the subject.id to be stable across the pipeline for all derived events. Mind you, that is how CDEvents expects subject.id to behave, but others may not have that same understanding and reading the code may cause confusion without that information.

I also realized all the JSON indirection is due to the cdevents java sdk being hard to use. I am in the process of updating it so this will be easier to do in the future.

So other than that this PR looks good. Ill go ahead and approve, but please add that one doc comment explaining why we can use subject.id

tridentsx:
I added a comment about the  subject.id in the code I hope thats enough to satisfy.

tridentsx:
Sorry I am unsure on the status of the PR is it approved ?

xibz (approved):

xibz:
@tridentsx I thought I approved this, but it looks like I just commented T-T

I approved and merged!

PR #7624, opened 2026-04-10
fix(build/test): disable JVM class data sharing to remove warning on j

dbyron-sf:
For example:
```
> Task :gate:gate-web:test
OpenJDK 64-Bit Server VM warning: Sharing is only supported for boot loader classes because bootstrap classpath has been appended
```
but this occurs in test tasks across all microservices.

Adds -Xshare:off to all Test task JVM args via the shared SpinnakerBaseProjectConventionsPlugin.  See
https://stackoverflow.com/questions/54205486/how-to-avoid-sharing-is-only-supported-for-boot-loader-classes-because-bootstra for background.

[Benchmark](https://github.com/user-attachments/files/26641296/benchmark_cds.py) (100 iterations, alternating, Zulu JDK 17.0.11):
```
  With Class Data Sharing:  avg=252ms  p50=257ms  p95=281ms
  Without:                  avg=265ms  p50=267ms  p95=300ms
  Diff:                     ~13ms per JVM startup
```
This is potentially controversial since it is slows things down (slightly).  I think the reduction is noise on stdout, and generally getting closer to no warnings is worth it.  If we ever do get there, we can fail builds if warnings do get introduced.


jasonmcintosh (approved):
Meh I'm good with this ;) it should be minor enough at this point, and if we can do OTHER cleanup... well, I'm in favor of clean errors!

PR #7633, opened 2026-04-13
fix(pubsub/aws): support cross-account SQS queue lookup and make queue creation fallback configurable (#7609)

Steven1677:
Fixes https://github.com/spinnaker/spinnaker/issues/7609

`PubSubUtils.getQueueUrl()` fails for cross-account SQS queues because it only passes the queue name to the AWS `GetQueueUrl` API. AWS requires `QueueOwnerAWSAccountId` to look up a queue in a different account.

**Changes:**
- Pass `QueueOwnerAWSAccountId` (from the queue ARN) in both SDK v1 and v2 `getQueueUrl` calls
- Add `enableQueueCreationFallback` property (default `true` for backward compatibility) to guard the `createQueue` fallback, which silently creates a wrong-account queue for cross-account setups
- Existing 4-param `ensureQueueExists` signatures are preserved as backward-compatible delegates

**Configuration:**

No changes needed for existing same-account setups. For cross-account queues:

```yaml
pubsub:
  amazon:
    subscriptions:
      - name: my-subscription
        enableQueueCreationFallback: false
```

**Testing:**
- Updated `PubSubUtilsSpec` mocks to verify account ID is passed
- Added test for fallback=false throwing `QueueDoesNotExistException`
- Updated `SQSSubscriberTest` mock to match new request signature


jasonmcintosh:
Steven - if you can run `./gradlew spotlessApply` please!  There's a formatting issue!

jasonmcintosh:
Otherwise, looks good thank you!

Steven1677:
Fixed with a new commit. Thanks @jasonmcintosh for the review.

Let me know if there's anything else needed.

jasonmcintosh (approved):

PR #7643, opened 2026-04-15
feat(type): Move identifier queries to a set type which they were already to begin with...

jasonmcintosh:
Just makes it explicit these are SETS.  As I found code... basically duplicating the set creation because it wasn't explicit these were set types.


clouddriver/cats/cats-redis/src/main/java/com/netflix/spinnaker/cats/redis/cache/AbstractRedisCache.java
```diff
@@ -123,7 +123,7 @@ public Collection<String> existingIdentifiers(String type, Collection<String> id
     return responses.entrySet().stream()
```
jasonmcintosh, line 123: This in particular... I'm trying to understand if there's any reason FOR this...

jasonmcintosh: ``responses.keySet();``` aka just do this...

Issue #7463, opened 2026-02-12
Application discovery failure with Eureka v2 in Spinnaker 1.37.0


sudarshanagh:
**Label:**
[retrofit2migration](https://github.com/spinnaker/spinnaker/labels?q=retrofit)

**Issue Summary:**
Spinnaker 1.37.0 and above do not support Eureka with suffix v2

**Cloud Provider(s):**
AWS

**Environment:**
Spinnaker running on version 1.37.0 and deployed on EKS cluster.

**Feature Area:**
Clouddriver

Description:
Eureka v2 is used as the health provider for our account , with the URL:  _BASE_EUREKA_URL/eureka/v2_

 We discovered that in Spinnaker  1.37.0 , clouddriver tried to reach out to _BASE_EUREKA_URL/apps_ instead of _BASE_EUREKA_URL/eureka/v2/apps_, during application-discovery.

This was most likely the result of clouddriver-eureka being [updated](https://github.com/spinnaker/clouddriver/pull/6340) in Spinnaker 1.37.0, to use retrofit2 client instead of the previous retrofit client, since , as per Retrofit2 [documentation](https://square.github.io/retrofit/2.x/retrofit/retrofit2/Retrofit.Builder.html#baseUrl-okhttp3.HttpUrl-),  : 
```
Endpoint values which contain a leading / are absolute.

Absolute values retain only the host from baseUrl and ignore any specified path components.

Base URL: http://example.com/api/
Endpoint: /foo/bar/
Result: http://example.com/foo/bar/

Base URL: http://example.com/
Endpoint: /foo/bar/
Result: http://example.com/foo/bar/
```

Steps to Reproduce:
1. Configure Eureka as the health-provider for your AWS account, as described [here](https://github.com/spinnaker/spinnaker/tree/main/clouddriver/clouddriver-eureka#readme), 
2. Use the Pipeline stage [deploy](https://spinnaker.io/docs/reference/pipeline/stages/#deploy) to deploy your application in AWS.

The deploy stage will time-out even though instances have been deployed. 

Additional Details:

As observed in this  [PR](https://github.com/spinnaker/spinnaker/pull/7159), a number of Spinnaker API interfaces were updated to support the retrofit2 client after Spinnaker 1.37.0. 
However the API endpoints in clouddriver-eureka were missed.

[...]

PR #7646, opened 2026-04-16
fix(retrofit2): remove leading slashes from eureka retrofit2 api interfaces

sudarshanagh:
This is the fix for this [issue](https://github.com/spinnaker/spinnaker/issues/7463)
Without this patch, Spinnaker 1.37.0 and above are not able to use Eureka for application discovery if the base URL has the suffix ```/eureka/v2```

It seems like this was missed in this PR https://github.com/spinnaker/spinnaker/pull/7159 when all the retrofit2 API interfaces were being updated to support retrofit2


jasonmcintosh:
Any reason to close this?  Was llooking to review this ;)

jasonmcintosh:
Would have welcomed a few tests showing this too!

Issue #7339, opened 2025-11-14
2025.3.0 - Infinite loop to acquire redis lock causes high CPU usage.


cdufresne81:
Good morning Gentlemens,

Production environment:

- Deployment type: **Local Debian** on AWS EC2 instance.
- Services used: Deck, Gate, Echo, Front50, Orca, Fiat, Clouddriver and sprinkled with a bit of Rosco.
- Authentication: SAML with Okta.
- Version: **2025.3.0**
- Usage: Mostly deploy AWS ECS services.
- Noteworthy: all services that can use SQL are using it (MySQL 8).

Yesterday I've upgraded our production environment and found out that it has heavy CPU usage, something that wasn't there with the previous version (1.34.3). After a bit of investigation I was able to confirm that the issue comes from FIAT hammering REDIS to acquire lock **hundreds of time every second**. Setting fiat.writeMode.enabled to false fixes the issues and the CPU drop below 10% which is what I would expect and was the norm before the upgrade.

As can be seen in this image, Redis and FIAT and the two processes hugging the CPU
<img width="2115" height="222" alt="Image" src="https://github.com/user-attachments/assets/0866a1d3-d1f8-49d2-8645-73cd7c13ccde" />

Here's what I mean by hammering redis in a loop (found after setting Fiat log level to DEBUG) :
```Nov 14 17:10:31 prod-spinnaker fiat[82226]: 2025-11-14 17:10:31.210 DEBUG 82226 --- [    scheduler-5] c.n.s.kork.jedis.lock.RedisLockManager   : [] Could not acquire already taken lock Lock{name='fiat.userrolessyncer', ownerName='prod-spinnaker', leaseDurationMillis=10000, successIntervalMillis=3600000, failureIntervalMillis=600000, version=1763140070195, ownerSystemTimestamp=1763140070200, attributes=''}
Nov 14 17:10:31 prod-spinnaker fiat[82226]: 2025-11-14 17:10:31.210 DEBUG 82226 --- [    scheduler-5] c.n.spinnaker.fiat.roles.Synchronizer    : [] obtained the following lastSyncTime from redis: 1763140070465
[...]

PR #7647, opened 2026-04-17
Adopt latching for fiat role sync

stevenlandrydd:
taken from: https://github.com/spinnaker/fiat/compare/master...dbyron-sf:fiat:pr-4-synchronize-user-role-syncs

# Fiat User Roles Sync – Synchronization and Configuration

## Summary

This PR refactors Fiat's user-roles sync to use configuration properties, adds an optional cross-pod synchronization path (in-process latch + Redis lock), and preserves parallel Redis writes via `putAllById`. It also documents the Front50 option to avoid full sync locks when saving service accounts.

## What This PR Does

1. **Config-driven user roles syncer**
   - Sync is driven by `UserRolesSyncerConfig` (`fiat.write-mode.*`) instead of `@Value` and the previous `UserRolesSyncStrategy` abstraction.
   - The syncer now depends directly on `PermissionsRepository`, `PermissionsResolver`, `Synchronizer`, etc.
   - Uses `putAllById` for parallel Redis writes (preserves `syncThreads` behavior).

2. **Optional synchronization path (latch + Redis lock)**
   - When enabled:
     - **In-process CountDownLatch**: Only one thread per Fiat instance runs sync; others wait for that result.
     - **Redis lock** (`Fiat.UserRolesSyncer.Synchronize`): Only one Fiat pod runs a full sync at a time.
     - **Redis keys** for last sync time and sync count (`{prefix}:user_roles:last_sync_time`, `{prefix}:user_roles:count`) so pods that didn't run the sync can skip work and reuse the latest count.

3. **Backoff and health**
   - Sync uses `FixedBackOff` and respects `ResourceProvidersHealthIndicator`; failed syncs are retried until timeout/backoff exhaustion.

4. **Other changes**
   - Redis pool/config: client name, pool sizing, and logging (e.g. in `RedisConfig`).
   - `DefaultApplicationResourceProvider`: constructor parameter order fix and duplicate field cleanup.
   - Test updates for the new syncer constructor and behavior (`UserRolesSyncerSpec`, `RedisPermissionsRepositorySpec`).

---

## How to Enable the Synchronization (Latching) Path

The latch + Redis lock path is **off** by default. Enable it so only one pod runs a full sync and others wait or reuse the result.

### YAML (Fiat or shared config)

```yaml
fiat:
  write-mode:
    synchronization-config:
      enabled: true
    # optional overrides:
    # sync-delay-ms: 600000
    # sync-failure-delay-ms: 600000
    # sync-delay-timeout-ms: 30000
    # retry-interval-ms: 10000
```

### Properties

```properties
fiat.write-mode.synchronization-config.enabled=true
```

### Logs when enabled

- `since synchronization is enabled, forcing a sync of all user roles`
- `Not syncing since user roles lastSyncTime: ... is later than this thread's syncAttemptTime: ...` (when another pod/thread already synced)
- `user roles synced` after a successful run

---

## Avoiding Extra Locks When Saving Service Accounts (Front50)

When Front50 creates/updates a **service account**, it can either:

- Call Fiat's **full sync** for the affected roles (`sync(rolesToSync)`), which goes through the latch + Redis lock when synchronization is enabled, or
- Call Fiat's **service-account-only** sync (`syncServiceAccount(id, roles)`), which only updates that one service account and does **not** trigger the full sync path or those locks.

To avoid extra locks and full syncs on every service account save, use the service-account-only path by setting the following in **Front50** config:

### YAML

```yaml
fiat:
  disableRoleSyncWhenSavingServiceAccounts: true
```

### Properties

```properties
fiat.disableRoleSyncWhenSavingServiceAccounts=true
```

When `true`, Front50 uses `syncServiceAccount(serviceAccount.getId(), serviceAccount.getMemberOf())` on create/update instead of `sync(rolesToSync)`, so you get correct permissions without taking the full sync lock.

---

## Suggested Production Config (Multi-Pod Fiat + Front50)

```yaml
# Fiat – enable cross-pod sync coordination
fiat:
  write-mode:
    enabled: true
    synchronization-config:
      enabled: true
      prefix: "spinnaker:fiat"
    sync-delay-ms: 600000
    sync-failure-delay-ms: 600000
    sync-delay-timeout-ms: 30000
    retry-interval-ms: 10000

# Front50 – avoid full sync (and locks) when saving service accounts
fiat:
  disableRoleSyncWhenSavingServiceAccounts: true
```

This gives you the full PR behavior (latching + Redis lock + parallel writes), with synchronization enabled, and avoids extra locks when saving service accounts by using the dedicated `syncServiceAccount` path.

### DIFF
--- a/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
+++ b/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
@@ -175,7 +175,7 @@ public class UserRolesSyncer {
                         .getSynchronizationConfig()
                         .getSyncFailureDelayMs()));
 
-    long count;
+    long count = 0;
     while (true) {
       boolean isSyncAllowed = false; // flag to control which thread can attempt a sync
       // since countdown latches can't be reset, we keep a pointer to a global latch so that others
--- a/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
+++ b/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
@@ -19,19 +19,44 @@ package com.netflix.spinnaker.fiat.roles;
 import com.netflix.spectator.api.Gauge;
 import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
+import com.netflix.spinnaker.fiat.config.ResourceProvidersHealthIndicator;
+import com.netflix.spinnaker.fiat.config.UnrestrictedResourceConfig;
+import com.netflix.spinnaker.fiat.config.UserRolesSyncerConfig;
+import com.netflix.spinnaker.fiat.model.UserPermission;
+import com.netflix.spinnaker.fiat.model.resources.Role;
+import com.netflix.spinnaker.fiat.model.resources.ServiceAccount;
+import com.netflix.spinnaker.fiat.permissions.ExternalUser;
+import com.netflix.spinnaker.fiat.permissions.PermissionResolutionException;
+import com.netflix.spinnaker.fiat.permissions.PermissionsRepository;
+import com.netflix.spinnaker.fiat.permissions.PermissionsResolver;
+import com.netflix.spinnaker.fiat.providers.ProviderException;
+import com.netflix.spinnaker.fiat.providers.ResourceProvider;
 import com.netflix.spinnaker.kork.discovery.DiscoveryStatusListener;
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
 import com.netflix.spinnaker.kork.lock.LockManager;
 import java.time.Duration;
 import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
+import java.util.Set;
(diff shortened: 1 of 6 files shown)


fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/roles/UserRolesSyncer.java
```diff
@@ -104,11 +146,286 @@ public void schedule() {
...
+          globalLatch.set(new CountDownLatch(1));
+        }
+        localLatch = globalLatch.get();
+      }
+
+      if (isSyncAllowed) {
+        try {
+          // acquire redis lock to sync across fiat pods
+          LockManager.AcquireLockResponse<Long> acquireLockResponse =
+              lockManager.acquireLock(
```
jasonmcintosh, line 197: SO the one problem here right now looks like still exists via
https://github.com/spinnaker/spinnaker/issues/7339 and
this is very similar to https://github.com/spinnaker/spinnaker/pull/7115

Where... if a lock fails, the retry is IMMEDIATE.  That said, I'm ALSO seeing a new thing here the syncAttemptTime which MIGHT fix that issue...

jasonmcintosh (dismissed):
Overall think this fixes the previous latch attempt where concurrent redis was happening :) and matches previous attempts where we don't want multiple redis pods doing full syncs.  Looks good - though need to figure out test failure, I'm good with this code as a whole!

jasonmcintosh (approved):

Issue #95, opened 2014-09-16
Integrate PagerDuty API key into app creation/settings


PoorImpulseCtrl:
Input would be in "New Application" modal and subsequent settings screen.
Potentially block production pushes on apps that do not have pagerduty api key.

Fairly high priority, ties into availability initiative. Josh/Roy/Dianne driven, connect with Roy/Jeremy for more info.

PR #7648, opened 2026-04-17
Improve parallelization of redis fiat get and put

stevenlandrydd:
# Improve parallelization of Redis fiat `get` and `put`

## Summary

Parallelize Redis reads and writes in `RedisPermissionsRepository` and user permission resolution in `DefaultPermissionsResolver` to reduce the wall-clock time of fiat role syncs in environments with many users.

Ported from doordash/spinnaker PR #95 ("Improve parallelization of redis fiat get and put").

### `RedisPermissionsRepository`

- **Bulk `put(Map<String, UserPermission>)`**: replace the sequential `for` loop over `permissions.values()` with `permissions.values().parallelStream().forEach(this::put)` submitted to the existing `syncThreadPool` (`ForkJoinPool` sized via `configProps.getRepository().getSyncThreads()`).
- **`getAllByRoles` (bulk role reads)**: replace the sequential per-user loop calling `getUserResourceMapFromRedis` with a parallel stream over `userIds` submitted to `syncThreadPool`, accumulating into a `ConcurrentHashMap<String, Set<Role>>` instead of a plain `HashMap`.
- Wrap both parallel submissions with proper exception translation:
  - `ExecutionException` → unwrap and rethrow the original `RuntimeException` / `SpinnakerException`, or wrap in `PermissionReadException` / `IntegrationException`.
  - `InterruptedException` → restore the interrupt flag and surface as a `PermissionReadException` / `IntegrationException`.

### `DefaultPermissionsResolver`

- `resolveResources` now uses `parallelStream()` over the `userToRoles` entry set so that per-user `getResources` lookups run concurrently.
- Collect into a `ConcurrentMap` via `Collectors.toConcurrentMap(...)` to make the terminal operation thread-safe.

### Tests

- New `DefaultPermissionsResolverTest` covering the parallelized `resolveResources` path.
- New `RedisPermissionsRepositoryTest` covering parallel `put(Map)` and `getAllByRoles`, including exception propagation for interrupt and underlying Redis failures.

## Motivation

Fiat role syncs for large user sets were dominated by sequential Redis round trips. Using the already-configured `syncThreadPool` lets us fan out both reads and writes to the same bounded pool that the rest of the syncer uses, without introducing a new concurrency knob.

## Tuning Thread Parallelism

There are two pools involved, and they are controlled independently.

### 1. Redis read/write pool (`RedisPermissionsRepository`)

The parallel `put(Map)` and `getAllByRoles` operations run on a dedicated `ForkJoinPool` owned by `RedisPermissionsRepository`. Its size is controlled by the existing property:

```yaml
fiat:
  redis:
    repository:
      sync-threads: 16   # default: Runtime.getRuntime().availableProcessors()
```

Guidance:

- Start with the default (number of CPU cores). This is a good baseline because the work is a mix of Redis I/O and JSON (de)serialization.
- If fiat is I/O-bound against Redis (CPU idle, role sync still slow), increase `sync-threads` (e.g. 2–4x cores) to get more concurrent Redis round trips in flight.
- If Redis itself starts showing elevated latency, CPU saturation, or connection pool exhaustion, decrease `sync-threads`. Make sure the Jedis / Redis client pool (`max-total`) is ≥ `sync-threads`, otherwise threads will just block waiting for a connection.
- Very high values (hundreds) are rarely useful and can make Redis the bottleneck; prefer scaling Redis or sharding before pushing this too high.

### 2. Permission resolution pool (`DefaultPermissionsResolver`)

`resolveResources` uses `parallelStream()`, which runs on the JVM common `ForkJoinPool`. Its parallelism is controlled by the standard JVM system property:

```
-Djava.util.concurrent.ForkJoinPool.common.parallelism=16
```

Default is `Runtime.getRuntime().availableProcessors() - 1`. Raise this if permission resolution (not Redis) is the bottleneck — for example, when upstream role/account providers (Google Groups, LDAP, etc.) dominate sync time and can tolerate more concurrent calls. Note that this pool is shared JVM-wide, so tune with awareness of other `parallelStream()` users in fiat.

### Recommended approach

1. Leave both at defaults initially.
2. Measure: look at `UserRolesSyncer` sync duration, Redis latency/CPU, and fiat pod CPU.
3. Adjust `fiat.redis.repository.sync-threads` first — it has the most direct impact on the bulk Redis operations introduced here.
4. Only tune the common pool parallelism if provider-side resolution is clearly the bottleneck.

### DIFF
--- a/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
+++ b/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
@@ -206,10 +206,10 @@ public class DefaultPermissionsResolver implements PermissionsResolver {
   @Override
   public Map<String, UserPermission> resolveResources(
       @NonNull Map<String, Collection<Role>> userToRoles) {
-    return userToRoles.entrySet().stream()
+    return userToRoles.entrySet().parallelStream()
         .map(
             entry -> {
-              final String userId = entry.getKey();
+              final String userId = String.valueOf(entry.getKey());
               final Set<Role> userRoles = new HashSet<>(entry.getValue());
               final boolean isAdmin = hasAdminRole(userRoles);
 
@@ -220,7 +220,7 @@ public class DefaultPermissionsResolver implements PermissionsResolver {
                   .setAccountManager(hasAccountManagerRole(userRoles))
                   .addResources(getResources(userId, userRoles, isAdmin));
             })
-        .collect(Collectors.toMap(UserPermission::getId, Function.identity()));
+        .collect(Collectors.toConcurrentMap(UserPermission::getId, Function.identity()));
   }
 
   private Set<Resource> getResources(String userId, Set<Role> userRoles, boolean isAdmin) {
--- a/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
+++ b/fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/RedisPermissionsRepository.java
@@ -34,6 +34,7 @@ import java.time.Clock;
 import java.time.Duration;
 import java.time.Instant;
 import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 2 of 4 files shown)


fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -206,10 +206,10 @@ private Map<String, Collection<Role>> getAndMergeUserRoles(
   @Override
   public Map<String, UserPermission> resolveResources(
       @NonNull Map<String, Collection<Role>> userToRoles) {
-    return userToRoles.entrySet().stream()
+    return userToRoles.entrySet().parallelStream()
         .map(
             entry -> {
-              final String userId = entry.getKey();
+              final String userId = String.valueOf(entry.getKey());
```
jasonmcintosh, line 212: Hrmm, the key should already be a string?  Reason for the valueOf?

jasonmcintosh (approved):
Overall nice little improvement to move to parallel streams to handle concurrent loads of information.  I think we'll be able to improve the default permissions resolver a bit more... https://github.com/armory-io/fiat/pull/3/changes is one we were working for this that's VERY similar.

fiat/fiat-roles/src/main/java/com/netflix/spinnaker/fiat/permissions/DefaultPermissionsResolver.java
```diff
@@ -206,10 +206,10 @@ private Map<String, Collection<Role>> getAndMergeUserRoles(
   @Override
   public Map<String, UserPermission> resolveResources(
       @NonNull Map<String, Collection<Role>> userToRoles) {
-    return userToRoles.entrySet().stream()
+    return userToRoles.entrySet().parallelStream()
```
jasonmcintosh, line 209: NOTE:  We have an incoming PR that does much of this.  I'll look to merge this and we can enhance it with the additional config from our other PR...

jasonmcintosh (approved):

PR #7649, opened 2026-04-18
fix(fiat): Fiat perf fix from AI analysis for testing

jasonmcintosh:
This is an AI generated analysis of perf issue in fiat's sql implementation.  Putting it up for evaluation/testing.  Was experimenting with gemma4:31b models on a local machine and this is what it generated.  It's... interesting.  Will add some tests, then see how this compares before/after on operations...


jasonmcintosh:
Closing for now as hoping to redo fiat

PR #7655, opened 2026-04-24
feat(lambda): Make application names work with Lambda

jasonmcintosh:
Fixes a few long standing issues with Lambda.  The biggest impact is application name handling.


Jaimeoby (approved):

Jaimeoby (approved):

jasonmcintosh:
TO NOTE:  This ALWAYS will now use the LambdaTagNamer.  Other services (ECS/Google) enable a "default" strategy selection by account.  I opted NOT to do that in this PR at this time.  IT WILL default back to using the "frigga" naming strategy and will ONLY apply the tags if enabled via the config flag.  The change this makes is that if the tags are detected (e.g. someone set manually on the pipeline), it WILL use them for app naming in cache operations.

christosarvanitis (approved):
Nice implementation to fix all the lambda caching issues at scale 🚀

jasonmcintosh:
image
image

PR #7667, opened 2026-05-04
fix(ecs): Fix ecs on demand perf loading issue

jasonmcintosh:
Two issues with the pendingOnDemand cache handling for ECS:
* first an incorrect filter logic which returned ALL services... for ALL ACCOUNTS/REGIONS in the data return, but only the TASKS for the give account/region.  Mis-placed paranthesis.  Tests showing the filtering. NOTE Orca would then filter the payload on the remote side to the selected region in `ServerGroupCacheForceRefreshTask` code. E.g. all signs are this was an OLD bug on this method.
* The next problem... and this is as bad.  it loads ALL KNOWN on demand data into memory THEN filters.  This is a really bad implementation.  Instead, what this PR changes is to load the identifiers via the glob pattern on the onDemand tables.  This is the primary key, so does a wildcard.  This isn't "ideal" but MASSIVELY performs better than loading the entire table into memory THEN filtering
What's worse on the above, and I call it out in a comment, was... this was multiplicative.  IF You had 10000 ECS accounts, each account would load ALL data then ALL regions, so if you say have 1000 onDemand operations across various systems (including NON ecs on demands), you end up with 1000 * 10000 data returned AND processed.

### DIFF
--- a/clouddriver/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/cache/Cache.java
+++ b/clouddriver/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/cache/Cache.java
@@ -89,11 +89,14 @@ public interface Cache {
   Collection<String> filterIdentifiers(String type, String glob);
 
   /**
-   * Retrieves all the items for the specified type
+   * Retrieves all the items for the specified type.
+   *
+   * <p>"DO NOT USE THIS. It does really really really bad things to the cache"
    *
    * @param type the type for which to retrieve items
    * @return all the items for the type
    */
+  @Deprecated()
   Collection<CacheData> getAll(String type);
 
   Collection<CacheData> getAll(String type, CacheFilter cacheFilter);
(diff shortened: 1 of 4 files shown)


clouddriver/clouddriver-api/src/main/java/com/netflix/spinnaker/cats/cache/Cache.java
```diff
@@ -89,11 +89,14 @@ default Collection<String> existingIdentifiers(String type, String... identifier
...
   /**
-   * Retrieves all the items for the specified type
+   * Retrieves all the items for the specified type.
+   *
+   * <p>"DO NOT USE THIS. It does really really really bad things to the cache"
    *
    * @param type the type for which to retrieve items
    * @return all the items for the type
    */
+  @Deprecated()
```
jasonmcintosh, line 99: I've INTENTIONALLY marked this method as deprecated.  I plan to work on removing all usage of it as soon as I can for other options.

christosarvanitis (approved):
Looks good, the perf of ECS provider will be improved

PR #7672, opened 2026-05-07
fix(perf): ECS loaded ALL data for ALL accounts then filtered.  

jasonmcintosh:
Cleaned up & ready.  Should replace the getAll then filter pattern with a "filter for identifiers then get for those identifiers".  FUTURE enhancement is to support this more natively in the CATS system - SQL would handle this MUCH more efficiently for example by simple patterns on indexes instead of a "look up index entries first" situation.


clouddriver/clouddriver-cloudfoundry/src/main/java/com/netflix/spinnaker/clouddriver/cloudfoundry/controller/CloudFoundryImageController.java
```diff
@@ -43,10 +42,7 @@ public class CloudFoundryImageController {
    */
   @RequestMapping(value = "/find", method = RequestMethod.GET)
   public Collection<CloudFoundryCluster> list(@RequestParam(required = false) String account) {
-    Stream<CloudFoundryCluster> clusters =
-        account == null
-            ? clusterProvider.getClusters().values().stream().flatMap(Set::stream)
-            : clusterProvider.getClusters().get(account).stream();
+    Stream<CloudFoundryCluster> clusters = clusterProvider.getClusters("*", account).stream();
```
jasonmcintosh, line 45: I fixed this here as well, since was lookign for generic getALl behavior.  BUT... this is one of those I'm not sure this code is ACTUALLY accessed by anything given the code should be returning "images", and how the deck interaction is here.  BUT tests are happy with this change soooo refactor & let it go.

jasonmcintosh:
TO NOTE:  There's additional optimizations that could be done.  Currently the "find ids then load data for those ids" is two steps.  With SQL this could be a SINGLE step and MUCH more efficient.  Think that'll be an additional PR later.

christosarvanitis (approved):
Checking locally the code it seems that keyAccountRegionFilter is dead code now - Worth removing/commenting for removal - do not use. 
The PR looks good and will help massively

PR #7673, opened 2026-05-07
[Azure] Pass full-ARM-ID UAMIs through to ARM template verbatim

scuba10steve:
## Summary

When `description.userAssignedIdentities` is a full ARM resource ID (starts with `/subscriptions/`), emit it directly as the `identity.userAssignedIdentities` map key instead of wrapping with `resourceID('Microsoft.ManagedIdentity/userAssignedIdentities/', '<id>')`. ARM's `resourceID()` only resolves same-RG references, so a UAMI that lives in a different RG than the VMSS deployment cannot be referenced that way and produces:

> Unable to evaluate template language function 'resourceID': function requires exactly one multi-segmented argument which must be resource type including resource provider namespace.

Bare-name inputs continue to take the existing wrap path (no behavior change for the historical case). Detected by the unambiguous `/subscriptions/` prefix.

## Why we hit this

Concrete use case: tenant-shared UAMI in a `<tenant>-v2` RG, VMSS deployed into per-app `<app>-<region>` RGs (the convention `AzureUtilities.getResourceGroupName(application, region)` produces). Both ends are reasonable Azure topologies; `resourceID()` just can't bridge across RGs in a 2-arg form. Preserving the full ARM ID through to the template fixes it without changing anything else.

## Test plan

- [ ] Verified existing `'generate server group with userAssignedIdentities'` and `'... without useSystemManagedIdentity'` Spock tests still pass (bare-name path unchanged).
- [ ] Deploy attempt against devaz fails with the `resourceID` template error pre-patch.
- [ ] Deploy attempt against devaz succeeds post-patch (UAMI map key is the full ARM ID).


scuba10steve:
Mistargeted — re-creating against moderneinc/spinnaker (this is a fork-internal patch, not for upstream submission).

PR #7682, opened 2026-05-12
feat(upgrades): Spring boot on a supported version - 3.5.14

jasonmcintosh:
Fixes keel.  Removes overrides that are now handled or upgraded by spring boot (aka tomcat, liquibase, etc.).  Work in progress - there ARE a few areas of concern:
* The `RedBlackToBlueGreenK8sPipelinesMigrationSpec` file ... where in newer versions there's an empty object NOT being returned.  I removed it from the test... but I worry about the serialization changes in latest jackson libraries
* Task scheduler - the TaskExecutorBuilder.class is gone in spring, i've replaced it with a SimpleAsyncTaskExecutorBuilder.class which long term supports virtual threads and IMHO is an ideal replacement.  BUT there's another one that's available if any concern

The MAJORITY Of the work is keel was... kinda painful and jooq upgrades seemed to cause issues.  There may be some AI slop on fixing keel - but it's directly tied i think to the changelog behavior.  E.g.
```
-              .set(LIFECYCLE_MONITOR.IGNORE, true)
+              .set(LIFECYCLE_MONITOR.IGNORE, 1.toByte())
```
Where booleans didn't seem to work on the generated objects.  At this point, I think the above passes tests and should be equivalent.

More incoming as I fix each project.
==== Updates
* Spinnaker-gradle-project passes
* Echo builds/passes tests
* Keel builds/passes tests
* Kork builds/passes tests
* Igor builds/passes
* Rosco builds/passes
* Fiat builds/passes
* orca
* Kayenta
* front50

Remaining:
* clouddriver
* gate
* halyard (OR... do we disable halyard at this point for upgrades?)


clouddriver/build.gradle
```diff
@@ -95,15 +96,5 @@ subprojects {
     }
   }
 
-  // Avoid warnings like
```
jasonmcintosh, line 98: Moved into the spinnaker gradle plugin to apply to all java/groovy classes

ricbartm:
@jasonmcintosh I landed on this PR because we have found several unpatched HIGH severity and CRITICAL severity vulnerabilities in Apache Tomcat. How are those things connected, you may wonder.

I have learn today that Spring Boot version 3.5.14, the target version of this PR, brings Apache Tomcat 10.1.54 by default. Nevertheless it looks like [here](https://github.com/spinnaker/spinnaker/blob/main/kork/spinnaker-dependencies/spinnaker-dependencies.gradle#L29-L36) we are pinning to a specific Tomcat version because previous versions of Spring Boot were using outdated versions.

We would appreciate that as part of this effort to bring in a supported Spring Boot version we could either:
* remove this Apache Tomcat version pinning 🙏 so version 10.1.54 is installed by default
* pin Apache Tomcat version to 10.1.55 which is latest version as of today


I personally have preference for the later because I foresee other HIGH/CRITICAL severity CVEs to be discovered in Tomcat in the upcoming months, and it allows quick version bump and faster remediation without incurring into a Spring Boot version bumps that are harder to introduce and validate.

jasonmcintosh:
> @jasonmcintosh I landed on this PR because we have found several unpatched HIGH severity and CRITICAL severity vulnerabilities in Apache Tomcat. How are those things connected, you may wonder.
> 
> I have learn today that Spring Boot version 3.5.14, the target version of this PR, brings Apache Tomcat 10.1.54 by default. Nevertheless it looks like [here](https://github.com/spinnaker/spinnaker/blob/main/kork/spinnaker-dependencies/spinnaker-dependencies.gradle#L29-L36) we are pinning to a specific Tomcat version because previous versions of Spring Boot were using outdated versions.
> 
> We would appreciate that as part of this effort to bring in a supported Spring Boot version we could either:
> 
> * remove this Apache Tomcat version pinning 🙏 so version 10.1.54 is installed by default
> * pin Apache Tomcat version to 10.1.55 which is latest version as of today
> 
> I personally have preference for the later because I foresee other HIGH/CRITICAL severity CVEs to be discovered in Tomcat in the upcoming months, and it allows quick version bump and faster remediation without incurring into a Spring Boot version bumps that are harder to introduce and validate.

I'd removed the pinning already in this PR, but happy to add it back and point to the latest release - GENERALLY tomcat doesn't change THAT much on compatibility so I tend to upgrade independent of spring releases.  That said I'm HOPING to more "regularly" update this so we DON'T get back into the "we're way out of date situation".  I can't guarantee that.  BUT this is all tied to regular updates ;) and contributions helping fix these kinds of things.

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -10,30 +10,24 @@ ext {
...
-    springCloud      : "2022.0.5",
-    swagger          : "2.2.22",
-    // 3.1.12 brings in 10.1.24, whichs introduce CVEs. Continue to pin till
-    // >=3.1.11 which brings in 10.1.52 and currently (20-feb-26) has no
-    // reported vulnerabilities.
-    //
-    // See https://tomcat.apache.org/security-10.html for latest security fixes
-    // and https://tomcat.apache.org/tomcat-10.1-doc/changelog.html for release
-    // notes.
-    tomcat           : "10.1.52"
```
jasonmcintosh, line 36: NOTE:  Removed right now as Spring brings in a newer release.  We can leave the tomcat here if we need to regularly update tomcat independent of spring, but ideally... we regularly update spring boot ;)

gate/gate-web/src/test/java/com/netflix/spinnaker/gate/AbstractSAMLConfigurationIntegrationTest.java
```diff
@@ -311,6 +296,53 @@ private static void configureKeycloakRealm() throws Exception {
...
+         "protocol":"saml",
+         "redirectUris":["http://localhost:%s/saml/SSO"],
+         "attributes":{
+          "saml.authnstatement":"true",
+          "saml.server.signature":"true",
+          "saml.client.signature": "false"
+         }
+        }
+        """
+            .formatted(CLIENT_ID, port);
```
jasonmcintosh, line 332: PRIMARY change on this commit was to add port here so could explicitly validate the previous URLs match going forward.  The wildcard would have allowed ANY URL to match and I wanted to validate we're not having a change on the ACS url.  This is b/c INTERNALLY spring changed one of the internal redirect URLs per above.  That redirect URL should have 0 impact on SAML logins as it's the way spring does a ... "figure out which registrationId to use", but... good to make sure on these tests!

jasonmcintosh:
Status update:
* Found spock tests NOT running for some reason on latest release
* spin cli - swagger has some bugs on newer openapi spec... which generates stuff as "Object" which go has no concept of.  Seems an issue with swagger.  Needs investigating
* Found a few more failures due to some library mismatches.  NOT sure why compilation did NOT pick them up.  In progress now.
* Found DNS Rebinding zone which is used for URL validation tests is NOT giving out valid DNS entries at this time.  Going to comment out these tests for now

jasonmcintosh:
* Spock tests fixed.  LOOKS like they may not have been running or inconsistently running due to gradle config.  NOW we'll run ALL available engines on each test.  LOT more tests got run as a result
* Lib fixes done
* DNS is moved to a separate set of tests can re-enable if the rebinding gets fixed
LAST Is looking at the openapi spec generation for swagger.

jasonmcintosh:
```
06:28:11.279 [Thread-0] WARN  i.s.c.v.g.DefaultCodegenConfig - String to be sanitized is null. Default to Object
06:28:11.279 [Thread-0] WARN  i.s.c.v.g.DefaultCodegenConfig - String to be sanitized is null. Default to Object
06:28:11.279 [Thread-0] WARN  i.s.c.v.g.go.AbstractGoCodegen - error (reserved word) cannot be used as method name. Renamed to CallError
06:28:11.279 [Thread-0] WARN  i.s.c.v.g.DefaultCodegenConfig - String to be sanitized is null. Default to Object
06:28:11.279 [Thread-0] WARN  i.s.c.v.g.DefaultCodegenConfig - String to be sanitized is null. Default to Object
06:28:11.279 [Thread-0] WARN  i.s.c.v.g.DefaultCodegenConfig - String to be sanitized is null. Default to Object
```
Code gen seems to be changing allt he stuff to "Object" types incorrectly.

.github/actions/generic-build-publish/action.yml
```diff
@@ -83,6 +84,8 @@ runs:
       env:
         GRADLE_OPTS: ${{ inputs.gradle-opts }}
       run: |
+        ## GHA Blocks rebinding attack based which causes some tests to fail.  This changes this
```
jasonmcintosh, line 87: COULD revert this given how... it was an issue with the rebinding, but... leaving here for now as it shouldn't hurt

front50/front50-web/front50-web.gradle
```diff
@@ -53,8 +53,11 @@ dependencies {
     runtimeOnly project(it)
   }
 }
-
+tasks.withType(JavaCompile) {
```
jasonmcintosh, line 56: TECHNICALLY i think i SHOULDN'T need this since now the plugin has this...

gate/gate-web/src/test/java/com/netflix/spinnaker/gate/controllers/InvokePipelineConfigTest.java
```diff
@@ -359,7 +359,7 @@ void invokePipelineConfigFront50NonJsonRespone() throws Exception {
             status()
                 .reason(
                     "Unable to trigger pipeline (application: my-application, pipelineNameOrId: my-pipeline-name). Error: Failed to process response body: Unrecognized token 'this': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')\n"
-                        + " at [Source: (okhttp3.ResponseBody$BomAwareReader); line: 1, column: 5]"))
+                        + " at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 1]"))
```
jasonmcintosh, line 362: Vs. enabling this, I felt disablign the detailed response was fine ;) Open to changing this to enable this... but... I tend to prefer to NOT expose details unless needed.

igor/igor-web/src/main/java/com/netflix/spinnaker/igor/config/GoogleCloudBuildProperties.java
```diff
@@ -30,7 +30,6 @@
 import lombok.Data;
 import lombok.Value;
 import org.springframework.boot.context.properties.ConfigurationProperties;
-import org.springframework.boot.context.properties.bind.ConstructorBinding;
```
jasonmcintosh, line 33: NOTE I removed it but could add it back... these are not required given the args constructor handling spring has, and I like to only add where needed.  I think this was while i was doing cleanup.

gate/gate-web/src/test/java/com/netflix/spinnaker/gate/controllers/InvokePipelineConfigTest.java
dbyron-sf: If there's a change here, there's potentially a behavior change for users.

jasonmcintosh: So there's a change on the exception handling.  Specifically 
```
Key Behavior and Security
Purpose: When enabled, the input's type and a snippet of its content are included in the toString() method of a parse exception.
Security Risk: Including the source can leak sensitive or PII (Personally Identifiable Information) data into application logs via stack traces.
Default Change: Starting in Jackson 2.16, this feature defaults to false (disabled) to follow a "Secure by default" approach. In versions 2.15 and earlier, it typically defaulted to true
```
(note this is AI generated).  BUT if you look at the stuff aka https://github.com/FasterXML/jackson-core/issues/991 - it looks like a good change to NOT expose this information in the exception messages.  ONLY impacts exception messages.

gate/swagger/generate_swagger.sh
```diff
@@ -35,6 +35,56 @@ mv ./gate-web/swagger2.json ./gate-web/swagger.json
 cat ./gate-web/swagger.json | jq --argjson rbschema "$RB_OPT_SCHEMA" '.paths.["/pipelines/{application}/{pipelineNameOrId}"].post.requestBody = $rbschema' > ./gate-web/swagger2.json
 mv ./gate-web/swagger2.json ./gate-web/swagger.json
 
+# Fix CloudEventData type from null to object (swagger codegen doesn't handle "type": "null")
```
jasonmcintosh, line 38: THIS was a larger change.  Basically moving OFF of the swagger-cli generation to the openapi cli generation.  This handles a few of the type changes much cleaner, and seems to be more maintained.  There was an open bug on the swagger cli generation for the last 4 months on the Object type issue that never got reviewed nor fixed.  The code here feels a TOUCH hacky, but... DOES seem to work

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -154,7 +154,7 @@ class BuildController {
     return buildService.queuedBuild(master, item)
   }
 
-  @RequestMapping(value = '/builds/all/{master:.+}/**')
+  @RequestMapping(value = ['/builds/all/{master:.+}/**', '/builds/all/{master:.+}'])
```
jasonmcintosh, line 157: This change is tied to spring and / handling.  I'm honestly not sure we REALLY needed this EVERYWHERE.  basically spring no longer matches on a "something/" and "/something" - without this kind of change.

keel/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
```diff
@@ -105,7 +105,7 @@ class SqlArtifactRepository(
         .set(DELIVERY_ARTIFACT.TYPE, artifact.type)
         .set(DELIVERY_ARTIFACT.REFERENCE, artifact.reference)
         .set(DELIVERY_ARTIFACT.DELIVERY_CONFIG_NAME, artifact.deliveryConfigName)
-        .set(DELIVERY_ARTIFACT.IS_PREVIEW, artifact.isPreview)
+        .set(DELIVERY_ARTIFACT.IS_PREVIEW, if (artifact.isPreview) 1.toByte() else 0.toByte())
```
jasonmcintosh, line 108: This felt REALLY goofy but tied to keel's auto generated code stuff :( Still finally on current versions of liquibase and shouldn't actually CHANGE anything behavior wise from what I can see of the schemas.

gate/gate-web/src/test/java/com/netflix/spinnaker/gate/controllers/InvokePipelineConfigTest.java
dbyron-sf: Something for the release notes I guess.

jasonmcintosh: Yeah :) I plan to do a major release note update AND run this through all of the integration test systems I've got ;) e.g. once merged, until all of that runs I would NOT do a release on this.  MAYBE try an "ALPHA" release and see if a few of the key people I know in the community would try it out, see if they find any issues.

.github/actions/generic-build-publish/action.yml
```diff
@@ -117,7 +120,10 @@ runs:
         FULL_DOCKER_IMAGE_NAME: "${{ inputs.project }}-${{ steps.version.outputs.version }}-integration-test"
         GRADLE_OPTS: ${{ inputs.gradle-opts }}
       shell: bash
-      run: ./gradlew ${{ inputs.gradle-cmd-opts }} -Dorg.gradle.daemon=false :${{ inputs.project }}:${{ inputs.project }}-integration:test
+      run:
```
stevenlandrydd: This should be `run: |` otherwise the lines after are skipping the tests

keel/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/_executionModels.kt
```diff
@@ -101,18 +104,68 @@ data class OrcaContext(
...
+    @Suppress("UNCHECKED_CAST")
+    val stages: List<OrcaExecutionStage> = if (stagesNode.isMissingNode || stagesNode.isNull) {
+      emptyList()
+    } else {
+      try {
+        ctxt.mapper.readValue(ctxt.mapper.treeAsTokens(stagesNode), stageListType) as? List<OrcaExecutionStage> ?: emptyList()
+      } catch (e: Exception) {
+        emptyList()
+      }
+    }
```
stevenlandrydd, line 149: stageListType is TypeReference<ArrayList<LinkedHashMap<String, Any>>> 

so Jackson literally produces a List<LinkedHashMap>. The unchecked cast to List<OrcaExecutionStage> succeeds at runtime only because of JVM erasure, but the elements are still LinkedHashMaps. Any downstream code that does stages.first().id, etc., will throw ClassCastException. 

Same anti-pattern in ExecutionDetailResponseDeserializer for both stages and variables (well, variables is correct — it uses TypeReference<List<KeyValuePair>>).

The TypeReference for stages should be object : TypeReference<List<OrcaExecutionStage>>() {}, and the unsafe casts should be removed. Can you also add a test that asserts stages[0] is OrcaExecutionStage

front50/front50-core/src/main/java/com/netflix/spinnaker/front50/config/Front50CoreConfiguration.java
```diff
@@ -44,10 +45,11 @@ public Namer<?> namer() {
   }
 
   @Bean
-  Jackson2ObjectMapperBuilderCustomizer defaultObjectMapperCustomizer(List<Module> modules) {
+  public Jackson2ObjectMapperBuilderCustomizer defaultObjectMapperCustomizer(List<Module> modules) {
     return jacksonObjectMapperBuilder -> {
       modules.addAll(List.of(new Front50ApiModule()));
       jacksonObjectMapperBuilder.modules(modules);
+      jacksonObjectMapperBuilder.serializationInclusion(JsonInclude.Include.NON_NULL);
```
stevenlandrydd: This is going to be a behavior change for anything downstream that doesn't handle null/absent distinctions. I think you mentioned something about this in the PR description but this concerns me

clouddriver/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
```diff
@@ -274,12 +274,51 @@ class BasicAmazonDeployDescription extends AbstractAmazonCredentialsDescription
...
   }
 
-  @Canonical
   static class Source {
     String account
     String region
     String asgName
     Boolean useSourceCapacity
+
+    // Constructor that normalizes empty strings to null
```
stevenlandrydd, line 283: Similar concerns here, this is a behavior change in theory. I'm not as concerned about this as the front50 case.

existing pipelines that store "account": "" will now resolve to account=null during deserialization, which can change downstream lookups (account resolution, ASG cloning, region-from-source logic).

This could probably be moved to a new PR

spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/baseproject/SpinnakerBaseProjectConventionsPlugin.groovy
```diff
@@ -8,53 +8,72 @@ import org.gradle.api.java.archives.Manifest
...
+    })
+    
+    project.plugins.withType(JavaLibraryPlugin).configureEach {
+      JavaPluginExtension extension = project.extensions.getByType(JavaPluginExtension)
+      def sourceJar = project.tasks.create("sourceJar", Jar)
+      sourceJar.dependsOn("classes")
+      sourceJar.archiveClassifier.set('sources')
+      sourceJar.from(extension.sourceSets.getByName('main').allSource)
+      project.artifacts.add('archives', sourceJar)
+      extension    }
```
stevenlandrydd: is this a stray line?

kotlin.gradle
```diff
@@ -48,24 +48,31 @@ dependencies {
...
 compileTestKotlin {
   kotlinOptions {
-    languageVersion = "1.8"
+    // LIKELY hitting https://youtrack.jetbrains.com/projects/KT/issues/KT-78909/K2-Missing-diagnostics-CYCLICINHERITANCEHIERARCHY-for-recursive-class-inheritance-leads-to-StackOverflowError?_gl=1*1jq5htm*_gcl_au*NTY4NTA4OTk2LjE3NzczMDU5MjA.*_ga*MTk0Njk1MDEzNS4xNzc3MzA1OTIx*_ga_9J976DJZ68*czE3Nzg3MDg4NDIkbzQkZzEkdDE3Nzg3MDkwOTQkajU5JGwwJGgw
+    // on version 2.0 of the language.  Can test with 2.3 and fix later
+    languageVersion = "1.9"
     jvmTarget = "17"
+// Changes made also in SpinnakerBaseProjectConventionPlugin
+    freeCompilerArgs += "-java-parameters"
+    javaParameters = true
```
stevenlandrydd, line 75: are both 74 and 75 needed? I believe these are duplicate entries in practice.

kotlin.gradle
```diff
@@ -48,24 +48,31 @@ dependencies {
...
 compileKotlin {
   kotlinOptions {
-    languageVersion = "1.8"
+    // LIKELY hitting https://youtrack.jetbrains.com/projects/KT/issues/KT-78909/K2-Missing-diagnostics-CYCLICINHERITANCEHIERARCHY-for-recursive-class-inheritance-leads-to-StackOverflowError?_gl=1*1jq5htm*_gcl_au*NTY4NTA4OTk2LjE3NzczMDU5MjA.*_ga*MTk0Njk1MDEzNS4xNzc3MzA1OTIx*_ga_9J976DJZ68*czE3Nzg3MDg4NDIkbzQkZzEkdDE3Nzg3MDkwOTQkajU5JGwwJGgw
+    // on version 2.0 of the language.  Can test with 2.3 and fix later
+    languageVersion = "1.9"
     jvmTarget = "17"
+// Changes made also in SpinnakerBaseProjectConventionPlugin
+    freeCompilerArgs += "-java-parameters"
+    javaParameters = true
```
stevenlandrydd, line 62: are both 61 and 62 needed? I believe these are duplicate entries in practice.

gate/gate-basic/src/main/java/com/netflix/spinnaker/gate/security/basic/BasicAuthConfig.java
```diff
@@ -56,10 +57,17 @@ public BasicAuthConfig(
...
-  @Order(2)
+  @Order(3)
   public SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
     defaultCookieSerializer.setSameSite(null);
+
+    // Configure request cache to save original requests.  This enables the "continue" by doing a
+    // session lookup
+    // See: https://stackoverflow.com/a/75228661
+    HttpSessionRequestCache requestCache = new HttpSessionRequestCache();
+    requestCache.setMatchingRequestParameterName("continue");
```
stevenlandrydd: with this change saved-request restoration only kicks in when the post-auth callback has ?continue=.... This isn't a bad change, but it's a behavior change: existing deep-link → login → return-to-original flows will now land users on / instead of the deep-link, unless Deck (or the calling user) appends ?continue. I can't see that Deck does this though

front50/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
```diff
@@ -161,7 +161,9 @@ public Collection<Pipeline> list(
    */
   @PreAuthorize("#restricted ? @fiatPermissionEvaluator.storeWholePermission() : true")
   @PostFilter("#restricted ? hasPermission(filterObject.name, 'APPLICATION', 'READ') : true")
-  @RequestMapping(value = "triggeredBy/{id:.+}/{status}", method = RequestMethod.GET)
+  @RequestMapping(
+      value = {"triggeredBy/{id:.+}/{status}", "triggeredBy/{id:.+}/{status}/"},
```
stevenlandrydd, line 165: Is this needed as part of this PR? I think this is a good change due to the spring boot 3 update but I think this should be done for more endpoints than these two then

front50/front50-core/src/main/java/com/netflix/spinnaker/front50/config/Front50CoreConfiguration.java
jasonmcintosh: Yeah... let me pull this out & re-run.  I was trying to trace through what was a change - removing empty maps.  BUT... I'm inclined towards NOT having this and letting frameworks handle the defaults...

clouddriver/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/description/BasicAmazonDeployDescription.groovy
jasonmcintosh: What's interesting is this FIXES tests b/c of the serialization behavior chagnes.  AKA.. without this, the tests BREAK :(

kotlin.gradle
jasonmcintosh: SO from what I can tell, 75 doesn't actually WORK.  I had to do both.

kotlin.gradle
jasonmcintosh: TECHNICALLY but I found that the 62 didn't work :( that could have been just a weird state locally.  I'll try removing it and see, but in testing... only 61 worked

gate/gate-basic/src/main/java/com/netflix/spinnaker/gate/security/basic/BasicAuthConfig.java
jasonmcintosh: IF I recall right the ?continue was actually how spring did this to 'optimize' going FORWARD vs. hitting every request, but your'e right ;) will break RIGHT NOW.  

I can do the `null` WHICH will do what previous versions did.

front50/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
jasonmcintosh: so I MOSTLY was targeted on this change where it hits tests.  A lot of places did NOT use this and with retrofit 2 URL changes... and thus hit this.  https://github.com/square/retrofit/issues/3449#issuecomment-666391061 aka there's some conversation around the baseUrl always having a / and I think we added this in ONE of the PRs to set these.  BUT it's NOT universal.  As a result... this is some of the controllers MAY have a different experience.  My hope is we'll catch it more in unit tests or integration tests post this, but... I'm HOPIONG I don't have to UNIVERSALLY make this change.

spinnaker-gradle-project/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/baseproject/SpinnakerBaseProjectConventionsPlugin.groovy
jasonmcintosh: More I was missing a line break ;)

keel/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/_executionModels.kt
jasonmcintosh: Let me look more - I MOSTLY ignored keel if the tests passed... and threw some "AI" tooling to try to work around headaches with it ....

kotlin.gradle
jasonmcintosh: from AI:
```
2. Apply via kotlin-dsl (Kotlin Projects)In Kotlin projects, standard Java parameters might be lost or omitted during the build. You need to tell the Kotlin compiler to pass java parameters.Add this to your build.gradle.kts:kotlinkotlin {
    compilerOptions {
        javaParameters.set(true)
    }
}
Use code with caution.Note: For older Kotlin DSL versions, use kotlinOptions { javaParameters = true } instead.
```
well, that looks like part of it - maybe a change in the way kotlin handles this... with chase the freeCompilerArgs would handle it, and since upgrading kotlin... if i changed to the other path it might work instead.

kotlin.gradle
jasonmcintosh: Per up above... it seems like it's dependent upon the kotlin DSL where-as the freeCompilerArgs should always work.  I THINK I can remove the javaParameters as a result ;)

front50/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java
jasonmcintosh: ALSO: Yes it's needed BECAUSE... some of the baseURl calls use the / and some don't kinda situation.

gate/gate-basic/src/main/java/com/netflix/spinnaker/gate/security/basic/BasicAuthConfig.java
jasonmcintosh: https://github.com/spinnaker/spinnaker/commit/eb576df1f9454a4392a85c9fd5374fc4f8d8f3c9#diff-8fae666a8d9f957877e6b3df77601115d2601ad29b0cb7b9c77a65c04ca1151f interestingly enough... this came in on THIS PR :) AKA... setting the null SHOULD fix this for that as well.

clouddriver/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/HttpUrlRestrictions.java
```diff
@@ -155,8 +155,22 @@ boolean isValidHostname(String host) {
   }
 
   boolean isValidIpAddress(String host) {
+    // If the host is not an IP address, we need to resolve it first
+    // Spring Security 6.x IpAddressMatcher.matches() now validates that the input is an IP address
+    String ipToMatch = host;
+    if (!InetAddresses.isInetAddress(host)) {
```
christosarvanitis, line 161: Candidate for a different PR as it is unrelated to the Spring Boot upgrade?

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesAtomicOperationDescription.java
```diff
@@ -24,14 +24,20 @@
...
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import lombok.Setter;
 
 @Data
 @AllArgsConstructor
 @NoArgsConstructor
 public class KubernetesAtomicOperationDescription
     implements DeployDescription, CredentialsNameable {
+  @Setter private String account;
```
christosarvanitis, line 34: Probably cleaner with @Getter(onMethod_ = {@JsonProperty("account")})

echo/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/PipelineCache.java
```diff
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.echo.pipelinetriggers;
 
 import static java.time.Instant.now;
+import static java.util.stream.Collectors.*;
```
christosarvanitis: Explicit imports over .* to align with the rules

front50/front50-gcs/front50-gcs.gradle
```diff
@@ -21,7 +21,7 @@ dependencies {
   implementation project(":front50-api")
 
   implementation "com.fasterxml.jackson.core:jackson-databind"
-  implementation "com.google.apis:google-api-services-storage"
+//  implementation "com.google.apis:google-api-services-storage"
```
christosarvanitis, line 24: Is this coming as transient dependency?

keel/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
christosarvanitis: im not sure i understand the need for change here :)

christosarvanitis (commented):
Thanks @jasonmcintosh. I will build internally and run some tests in the next couple of days

clouddriver/clouddriver-artifacts/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/config/HttpUrlRestrictions.java
jasonmcintosh: SO it's tied to spring boot upgrades on IpAddressMatcher handling... :( AKA tied to the boot upgrades :(

front50/front50-gcs/front50-gcs.gradle
jasonmcintosh: SO looks like it why I removed it.

keel/keel-sql/src/main/kotlin/com/netflix/spinnaker/keel/sql/SqlArtifactRepository.kt
jasonmcintosh: So this unfortunately is tied to a delta in how keel operates - it dynamically generates the code based upon the liquibase schema stuff.  When we bumped liquibase versions... it changed ALL of the keel codebase as a result :( which changed all of these calls :(

christosarvanitis:
@jasonmcintosh @dbyron-sf @stevenlandrydd @xibz This is something that will affect this upgrade https://github.com/FasterXML/jackson-core/issues/1047 in all services. How do we want to address this. Keep sane defaults with configurable application properties per service?

keel/keel-orca/src/main/kotlin/com/netflix/spinnaker/keel/orca/_executionModels.kt
jasonmcintosh: Fixed in latest PR :) Claude helped a lot on this one... :)

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesAtomicOperationDescription.java
jasonmcintosh: Possibly.  this is one of those places that... irks me... as the parent credentials nameable:
```
public interface CredentialsNameable extends AccountNameable {
  AccountCredentials getCredentials();

  @JsonProperty("credentials")
  @Override
  default String getAccount() {
    return getCredentials().getName();
  ```
sets account to be named "credentials"... aka there's a mismatch here that's really really annoying.  I wasn't sure how much I wanted to fiddle with making it clean, as really feel like this needs to be redone or changed in a way to NOT have such a mismatch.

jasonmcintosh:
> Keep sane defaults with configurable application properties per service?

Basically... I think this which you've already done a PR to do ;) Though only customizable for Orca and not the others at this point, I'm ok with this since it's basically "super large" but it'd be nice LONG term if could shrink those defaults way down.

christosarvanitis (dismissed):
Tested this upgrade in an internal environment so im good merging. Potentially we need to make the Jackson defaults customizable for all services but that can be done in a follow up PR if needed. 
Thanks for the upgrade @jasonmcintosh 🚀

christosarvanitis (approved):

PR #7683, opened 2026-05-18
feat(api-tokens): add API token authentication support

stevenlandrydd:
# Spinnaker API Tokens

Adds first-class, revocable API tokens for authenticating programmatic
requests against Gate. Tokens are minted, listed, and revoked through a new
self-service UI in Deck, persisted in Redis by Gate, and resolved on every
inbound request by a dedicated stateless Spring Security chain.

The feature is **off by default** (`api-tokens.enabled=false`) — turning it on
registers all of the new beans, exposes the `/auth/apiTokens` endpoints, and
surfaces the "API Tokens" entry in the user menu.

## Why

Today, automation that talks to Spinnaker has to either ride a user's SSO
session or run as a service account behind some out-of-band auth (basic auth,
mTLS, IAP-issued JWT, etc.). None of those are revocable per-credential,
scoped to a single integration, or auditable from the UI.

API tokens give us:

- **Per-credential revocation.** Every token is independent; revoking one
  doesn't disturb other automation owned by the same user / SA.
- **Bounded lifetime.** Both user and service-account tokens have admin-tunable
  max lifetimes (90 / 365 days by default); user tokens always expire.
- **IAP-compatible transport.** `X-Spinnaker-Token: spk_…` flows cleanly
  through GCP IAP, which rejects non-JWT `Authorization: Bearer` values.
- **Operator observability.** A new `gate.requests` Spectator counter is
  tagged with `authType`, `principalKind`, `method`, `statusCode`, and
  `status` so we can answer "how much of our traffic is token-auth, and from
  which kind of principal?" without scraping logs.

## Token model

A token is a string of the form `spk_<240 bits of hex>`. Only the SHA-256
hash is persisted (in Redis); the plaintext is returned **once** at mint time
and never recoverable again.

Each token record carries:

| field | notes |
|-------|-------|
| `id` | UUID, stable identifier used by the UI and `DELETE /auth/apiTokens/{id}` |
| `name` | Human-readable; **unique per principal**, max 64 chars, no control chars |
| `principalType` | `USER` or `SERVICE_ACCOUNT` |
| `principalId` | Username for `USER` tokens; SA name for `SERVICE_ACCOUNT` |
| `createdByUserId` | The minting user (admins can mint SA tokens on behalf of an SA) |
| `expiresAt` | ISO-8601; required for `USER`, optional (forever) for `SERVICE_ACCOUNT` |
| `lastUsedAt` | Updated asynchronously off the auth hot path |
| `lastFiatCheckAt` | Throttles the optional "is this principal still in Fiat?" recheck |
| `hashRef` | SHA-256 hex of the plaintext; lets revoke-by-id find the hash key in O(1) |

## How the auth chain works

`api-tokens.enabled=true` registers a dedicated
`WebSecurityConfigurerAdapter` (`ApiTokenAuthConfigurerAdapter`, `@Order(50)`)
whose `RequestMatcher` only matches requests bearing `X-Spinnaker-Token:
spk_…` or `Authorization: Bearer spk_…`. That chain is configured
`STATELESS`, so:

- Token auth **never writes to `HttpSession`** and cannot leak into
  cookie-only requests.
- Non-token requests fall through to the existing browser/OAuth2 chain
  untouched.
- Unauthenticated token requests get a clean **`401`** (via
  `HttpStatusEntryPoint`) instead of a 302 to the SSO login page — what API
  clients actually want.

The `ApiTokenAuthenticationFilter` does, in order:

1. Strip `X-Spinnaker-Token` from MDC and wrap the request so kork's
   `AuthenticatedRequestFilter` can't re-emit it as an outbound header to
   downstream services. **This is a real leak source if you skip it** — kork
   auto-copies every `X-Spinnaker-*` header into MDC and then back into
   outbound calls.
2. SHA-256 the plaintext and look it up in Redis via
   `ApiTokenService.resolveByHash`. Unknown hashes → fall through (the
   request is left unauthenticated; downstream entry points decide what to
   do).
3. Optionally throttle-check Fiat (see below) and reject if the principal
   has been deprovisioned.
4. Read the cached `UserPermission.View` from `FiatPermissionEvaluator` —
   **not** `permissionService.login()`, because that triggers a Fiat
   role-provider re-resolve that returns an empty role set for OIDC users
   (whose roles come from the groups claim via `loginWithRoles`) and would
   silently demote the live session.
5. Build a `User` principal so `@SpinnakerUser` resolves correctly in
   controllers, populate `roles` / `allowedAccounts`, install a
   `PreAuthenticatedAuthenticationToken` on the `SecurityContext`, and
   set request attributes (`gate.isApiToken`, `gate.authType`,
   `gate.principalKind`) that `FiatSessionFilter`, future DPoP work, and the
   metrics filter all key off.
6. Fire-and-forget update `lastUsedAt` on a daemon executor (drained on
   `@PreDestroy`).

`FiatSessionFilter` short-circuits when `gate.isApiToken` is set, so the
stateless token path never invalidates a session it doesn't own.

## Redis layout

`RedisApiTokenRepository` owns all token I/O. Keys are namespaced by
`api-tokens.redis-key-prefix` (default `api-token`) so Gate can share a
Redis instance with other services:

```
{prefix}:hash:{sha256_hex}             STRING  → JSON TokenRecord  (EXPIREAT if token expires)
{prefix}:id:{uuid}                     STRING  → JSON TokenRecord  (EXPIREAT)
{prefix}:principal:{TYPE}:{id}         ZSET    → member UUIDs, score = expiry epoch seconds
                                                 (Long.MAX_VALUE for non-expiring tokens)
{prefix}:name:{TYPE}:{id}:{sha(name)}  STRING  → token UUID         (EXPIREAT)
```

Notable invariants:

- The principal ZSET is self-cleaning: `findByPrincipal` and
  `findAllByPrincipalType` both call `ZREMRANGEBYSCORE … 0 now` before
  reading.
- The name index hashes the user-supplied name (fixed-length suffix) so
  `:` characters in `principalId` can't make different (principal, name)
  pairs collide.
- Per-principal name uniqueness is enforced atomically by `SET NX` on the
  name key inside `save()`. The HTTP layer's `rejectDuplicateName` is just a
  fast-path UX optimisation; the repository's `SET NX` is the source of
  truth (and translates to `409 Conflict` via `DuplicateTokenNameException`).
- Revocation issues all four `DEL`s inside `MULTI/EXEC` so no Gate replica
  can observe a partial state.
- `updateLastUsed` / `updateLastFiatCheck` use `WATCH/MULTI/EXEC` with a
  bounded 3-attempt retry so concurrent writers don't clobber each other's
  sibling fields via stale `GET → mutate → SET` round-trips. TTLs are
  preserved across the rewrite.

## HTTP API (`/auth/apiTokens`)

Mounted on `ApiTokenController`, gated by `@ConditionalOnProperty("api-tokens.enabled")`.

| Method | Path | Who | Notes |
|--------|------|-----|-------|
| `GET`    | `/auth/apiTokens` | any authenticated user | Lists caller's `USER` tokens; admins additionally see all `SERVICE_ACCOUNT` tokens. |
| `POST`   | `/auth/apiTokens` | minting role (USER) or admin (SA) | Creates a token. Returns the plaintext **exactly once** under `token`. |
| `DELETE` | `/auth/apiTokens/{id}` | owner or admin | Atomic revocation. |
| `GET`    | `/auth/apiTokens/admin/users` | admin only | All `USER` tokens across all principals. |
| `GET`    | `/auth/apiTokens/serviceAccounts` | admin only | The subset of Front50 service accounts that are declared as token-eligible. |

`CreateApiTokenRequest` is a record (`name`, `principalType`, `principalId`,
`expiresAt`). `expiresAt` is optional; when omitted, user tokens default to
`maxUserTokenLifetimeDays` and SA tokens default to non-expiring. Names are
trimmed, length-checked, and rejected if they contain ISO control characters
(log-injection / display-spoofing risk once the name is echoed in API
responses and the UI).

## Front50 service account changes

Service-account API tokens may only target identities operators have
explicitly declared as "API-capable" in config:

```yaml
service-accounts:
  - name: deploy-bot
    memberOf:
      - deploy-team
```

`ServiceAccountsInitializer` is a Boot `ApplicationRunner` that **upserts**
each declared SA on every startup — config is authoritative; `memberOf` is
overwritten to match the file. A new
`GET /serviceAccounts/tokenEligible` Front50 endpoint returns the
intersection of the SA store and the configured names; Gate calls it both
to populate the create-SA-token dropdown and to validate the
`principalId` on `POST /auth/apiTokens`.

## Fiat changes

Two small, targeted additions:

1. **`PermissionsRepository.isEmpty()`** with efficient implementations:
   - `RedisPermissionsRepository` → `SCARD allUsersKey`.
   - `SqlPermissionsRepository` → `fetchExists(selectOne().from(USER))`.

   `AuthorizeController.getUserPermission` now returns **`503`** instead of
   `404` when the repository is empty (likely an accidental Redis flush or
   not-yet-warm SQL store), so token auth — and the existing session
   flow — fail open instead of treating every principal as departed.

2. **`PermissionService.getRolesForTokenAuth`** in `gate-core` calls Fiat
   directly (bypassing the local cache) inside `AuthenticatedRequest.allowAnonymous`
   and **propagates** `SpinnakerHttpException` so callers can distinguish a
   legitimate empty-roles response from a transient Fiat failure. This is
   what powers the token filter's optional
   `reject-if-no-principal-permissions` mode without breaking the live
   session of an in-flight user.

## Gate observability: `gate.requests`

`RequestMetricsFilter` (registered at `LOWEST_PRECEDENCE - 2`) emits one
Spectator counter per inbound HTTP request:

```
name: gate.requests
tags:
  authType:      api_token | session | oauth2 | x509 | pre_authenticated | anonymous | none
  principalKind: user | service_account | unknown
  method:        GET | POST | PUT | DELETE | …
  statusCode:    200 | 401 | 503 | …
  status:        2xx | 4xx | 5xx | …
```

`AuthTypeResolver` centralises classification: filters that know better
publish to request attributes (`gate.authType`, `gate.principalKind`);
otherwise we fall back to heuristics over `SecurityContextHolder`.
High-cardinality data (principal id, URI, query) is **deliberately not**
tagged.

## `GET /auth/user` change

The previous implementation returned a reflection-derived
`Map<String,Object>` over the Spring Security `User`, which silently
exposed every new field. It's been replaced with an explicit
`AuthUserResponse` record. The wire format is byte-for-byte compatible with
the legacy map for existing consumers, plus:

- `isAdmin: boolean`
- `canMintApiTokens: boolean`
- `maxUserTokenLifetimeDays: number?` (omitted when `api-tokens.enabled=false`)
- `maxServiceAccountTokenLifetimeDays: number?` (same)

`ApiTokenService.canMintApiTokens` is the single source of truth used by
both the UI flag (`AuthController.user`) and the POST enforcement path
(`ApiTokenController.create`) so the two can't drift. Resolution order:

1. Subsystem disabled → no.
2. Fiat admin → yes (bypass).
3. No `allowed-minting-roles` configured → no (for non-admins).
4. Otherwise the user must hold at least one configured minting role.

## Deck UI

A new `apitoken/` package under `packages/core/src`:

- **`ApiTokensPageContainer`** — fetches `/auth/user`, then renders…
- **`ApiTokensPage`** — sections for **My Tokens**, **Service Account
  Tokens** (admin), and **All User Tokens** (admin). Tables show name,
  expiry (red when past, "Never" for non-expiring), last-used, and a
  Revoke button.
- **`CreateApiTokenModal`** — name input with live max-length / duplicate
  detection, SA dropdown (admins, populated from
  `/auth/apiTokens/serviceAccounts`), expiry picker capped at the server's
  configured max. The "Maximum (N days, set by server policy)" radio
  intentionally surfaces that the cap is policy, not UI.
- **`RevokeApiTokenButton`** — confirmation modal; surfaces error state on
  network failure.
- **`ApiTokensPage.less`** — page-scoped styles, no global leakage.

## Configuration reference

```yaml
api-tokens:
  enabled: false
  token-prefix: spk_
  redis-key-prefix: api-token
  reject-check-interval-seconds: 60
  max-user-token-lifetime-days: 90
  max-service-account-token-lifetime-days: 365
  reject-if-no-principal-permissions: false
  allowed-minting-roles:
    - spinnaker-api-users
```

`token-random-bytes` (default 30 → 240 bits of entropy) is also tunable but
not commonly changed. Only affects newly minted tokens.

## Security notes

- **No plaintext is ever logged or persisted.** Only the SHA-256 hex hash
  is written to Redis, and the token's auth-time path
  (`extractToken → sha256Hex → findByHash`) never logs the plaintext.
- **`X-Spinnaker-Token` is stripped from MDC and from the wrapped request**
  before the filter chain continues, defending against kork's
  MDC → outbound-header propagation.
- **Per-principal name uniqueness is enforced server-side via `SET NX`**,
  not on the client. Concurrent creates with the same name resolve to
  `409 Conflict`.
- **Revocation is atomic across all four Redis keys** (hash, id, ZSET,
  name) inside `MULTI/EXEC`. There is no window where the hash key (the
  only key read on the auth hot path) outlives the rest.
- **Names are rejected if they contain ISO control characters** (`\0`,
  `\t`, `\n`, `\r`, ANSI escapes, …) — log-injection and display-spoofing
  risk once names are echoed in API responses and the UI.
- **Service-account principals are validated against the Front50
  token-eligible list** at create time, so admins can't mint tokens for
  arbitrary SAs that happen to exist in Front50.

## Operability

- `api-tokens.enabled=false` is the default. Flipping the flag is the only
  thing needed to roll the feature out; rolling it back leaves the Redis
  keys in place (they'll expire on their own).
- Fiat fail-open: if Fiat is unreachable on a token-auth request, the
  filter accepts the cached principal and logs a warning. Only an explicit
  `404` (principal definitively absent) — or a `503` once the repo is
  empty — causes a rejection.
- `lastUsedAt` and `lastFiatCheckAt` updates are best-effort; the auth hot
  path never blocks on them. The async pool is a single daemon thread,
  drained on `@PreDestroy` with a 2 s timeout.
- Admin-list endpoints (`/admin/users`, `/serviceAccounts`) use `SCAN`
  rather than `KEYS`. SCAN can return duplicates across iterations, so we
  dedupe principal keys via a `LinkedHashSet`.

### Manual verification before merge

- [x] Enable `api-tokens.enabled=true` in a dev Gate, log in, mint a user
      token, hit a Gate endpoint with `X-Spinnaker-Token`.
- [x] Mint the same token name twice → second request returns `409`.
- [x] Revoke a token, retry → `401` (or fall-through, depending on
      downstream entry point).
- [x] As admin, mint an SA token for a configured SA, list it in the
      Service Account Tokens table.
- [x] Confirm `gate.requests` counter has both `authType=api_token` and
      `authType=session` after a mixed-traffic burst.
- [x] Confirm `X-Spinnaker-Token` does **not** appear in downstream service
      logs (Fiat, Orca) after a token request.


gate/gate-core/src/main/java/com/netflix/spinnaker/gate/services/internal/Front50Service.java
```diff
@@ -165,4 +165,7 @@ Call<ResponseBody> saveNotificationConfig(
 
   @GET("installedPlugins")
   Call<List<SpinnakerPluginDescriptor>> getInstalledPlugins();
+
+  @GET("serviceAccounts")
+  Call<List<Map>> getApiEnabledServiceAccounts(@Query("apiEnabled") boolean apiEnabled);
```
jasonmcintosh: SOME DAY - not in this PR :P I'm going to want ServiceAccountDefinition to be moved to kork as a generic type ;) that way we can type these API calls....

jasonmcintosh: OR into fiat maybe...

front50/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/ServiceAccountsController.java
```diff
@@ -32,16 +34,36 @@
...
+   * configuration and are therefore eligible for API token minting.
+   */
+  @RequestMapping(method = RequestMethod.GET, value = "/tokenEligible")
+  public Set<ServiceAccount> getTokenEligibleServiceAccounts() {
+    Set<String> configuredNames =
+        apiServiceAccountsProperties.getServiceAccounts().stream()
+            .map(ServiceAccountsProperties.ServiceAccountDefinition::getName)
+            .collect(Collectors.toSet());
+
+    return serviceAccountService.getAllServiceAccounts().stream()
```
jasonmcintosh: NIT/CAUTION:  This loads ALL service accounts THEN filters.  WITH systems with LARGE numbers of SA's (pipeline permissions models), this is going to be a lot of data potentially (e.g. 5+ figures of these at worst).  NOMINALLY should have the DA accept a filter instead.

gate/gate-web/config/gate.yml
```diff
@@ -194,3 +194,16 @@ security:
       firstName:
       lastName: name
       username: login
+
+# API token configuration (disabled by default)
```
jasonmcintosh: IF POSSIBLE, it'd be nice if going forward we also add these to the spinnaker-kustomize files to document config options ;) Though LONG TERM I'd argue this should be "default on" some of the config properties be nice to document/expose via those files in the gate-local.yml files.  

Alternative, a good docs PR on this ;)

gate/gate-web/src/main/java/com/netflix/spinnaker/gate/security/apitoken/ApiTokenService.java
```diff
@@ -0,0 +1,217 @@
...
+          e.getResponseCode(),
+          principalId,
+          e.getMessage());
+      return true;
+    } catch (Exception e) {
+      log.warn(
+          "Could not reach Fiat for principal '{}' — failing open: {}",
+          principalId,
+          e.getMessage());
+      return true;
```
jasonmcintosh, line 135: DO we really want this as a fail open on a permissions check? I feel like this should fail closed kinda thing off hand.... for 128 as well....

gate/gate-web/src/main/java/com/netflix/spinnaker/gate/security/apitoken/ApiTokenService.java
```diff
@@ -0,0 +1,217 @@
...
+              record.getPrincipalId());
+          return Optional.empty();
+        }
+        Instant now = Instant.now();
+        redisRepo.updateLastFiatCheck(record.getId(), tokenHash, now);
+        record.setLastFiatCheckAt(now.toString());
+      }
+    }
+
+    return Optional.of(record);
```
jasonmcintosh: Should this just return opt?

gate/gate-web/src/main/java/com/netflix/spinnaker/gate/security/apitoken/ApiTokenService.java
stevenlandrydd: Yes, failing open here is the right thing. This is the principal check that validates if the user/account exists in fiat. If fiat was empty for some reason or temporarily down failing closed here would immediately revoke the access. In this case failing open doesn't mean the token is authorized just that we're assuming the principal still exists and then the next level of auth checks can still proceed.

gate/gate-web/src/main/java/com/netflix/spinnaker/gate/security/apitoken/ApiTokenService.java
stevenlandrydd: functionally the same, but cleaner, updated.

front50/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/ServiceAccountsController.java
stevenlandrydd: Good call, granted I think 5 figures would be fine here since the functional call here is O(n). I shifted this over to use the defined SA's from the config which is what were looking for anywhere which makes this O(k).

gate/gate-web/config/gate.yml
stevenlandrydd: added the gate-local doc. I agree on the docs PR, will submit one for this after merge

jasonmcintosh (approved):
I'm good with this ;) VERY nice feature been a LONG time waited for!

PR #7686, opened 2026-05-20
fix(provider/kubernetes): preserve '=' in runJobManifest property values

xudongli42:
## Issue summary

`PropertyParser.extractPropertiesFromLog` in clouddriver-kubernetes silently strips trailing `=` characters from `SPINNAKER_PROPERTY_*=value` lines emitted by `runJobManifest` Job pods. Any value that ends in `=` (or `==`, or contains `=` internally) is either corrupted or silently dropped.

The most user-visible impact: **base64-encoded payloads passed through this mechanism are corrupted whenever the encoded data length is not a multiple of 3 bytes** (i.e., when standard base64 padding produces a trailing `=` or `==`). Statistically, this affects roughly 2/3 of all base64 values.

## Code reference

[`clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/PropertyParser.java`](https://github.com/spinnaker/spinnaker/blob/main/clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/PropertyParser.java):

```java
private static final Splitter equalsSplitter =
    Splitter.on("=").omitEmptyStrings().trimResults();

// ...
List<String> splittedLine = equalsSplitter.splitToList(line);
if (splittedLine.size() != 2) {
    continue;
}
final String key = splittedLine.get(0).replaceFirst(MAGIC_SEARCH_STRING, "").toLowerCase();
final String value = splittedLine.get(1);
```

### Defect 1 — trailing `=` stripped from values

`Splitter.on("=")` splits at every `=`, including padding chars at the end of base64 values. `omitEmptyStrings()` then discards the empty strings produced by trailing `=`, so the line still passes the `splittedLine.size() != 2` check — but the captured `value` is missing its trailing `=` chars.

**Trace:**

Input line: `SPINNAKER_PROPERTY_payload=SGVsbG8gV29ybGQ=`

- `Splitter.on("=").splitToList` → `["SPINNAKER_PROPERTY_payload", "SGVsbG8gV29ybGQ", ""]`
- `.omitEmptyStrings()` → `["SPINNAKER_PROPERTY_payload", "SGVsbG8gV29ybGQ"]`
- Size 2 → passes the check
- `value` captured as `"SGVsbG8gV29ybGQ"` — the trailing `=` is GONE

Downstream consumers that base64-decode the value fail with `illegal base64 data at input byte N`, where N is the truncated value length.

### Defect 2 — values containing `=` are silently dropped

Same `Splitter.on("=")` splits every `=`, including ones inside the value. A line like:

`SPINNAKER_PROPERTY_token=abc=def`

splits to `["SPINNAKER_PROPERTY_token", "abc", "def"]` — 3 elements, fails the `size() != 2` check, and the loop `continue`s, silently skipping the property. The property never reaches the stage context. No log line about the drop.

## Fix

Use `Splitter.on("=").limit(2)` (split at most twice → at the first `=`, everything after is the value):

```java
private static final Splitter equalsSplitter =
    Splitter.on("=").limit(2).trimResults();
```

This:
- Preserves trailing `=` and `==` padding correctly
- Allows values containing `=` (no silent drop)
- Matches Java `Properties` file semantics (`key=value` splits at first `=`; value is everything after)

`omitEmptyStrings()` is removed because we now intentionally want to retain a possibly-empty (or padding-containing) trailing element.

## Tests added

Three test cases covering the previously-broken inputs:
- Trailing `=` padding preserved (e.g., `SGVsbG8gV29ybGQ=`)
- Trailing `==` padding preserved (e.g., `SGVsbG8gV29ybGQhIQ==`)
- `=` within value preserved (e.g., `foo=bar`)

## Backward compatibility

Existing tests pass unchanged. Behavioral changes:
- Values that previously came through correctly continue to do so
- Values that previously had trailing `=` stripped will now retain the `=` — this is the bug fix, expected behavior change
- Values previously silently dropped (containing `=`) will now be parsed — also the bug fix

No existing valid use case relies on stripping `=` from values.

## Severity / scope

- Hit by any pipeline that base64-encodes data through `runJobManifest` propertyFile capture
- Affects ~2/3 of all base64-encoded payloads (those whose decoded length isn't a multiple of 3)
- Failure mode is opaque: downstream consumers (Helm `--set`, K8s `binaryData`, SPEL `#fromBase64`) report `illegal base64 data at input byte N` with no obvious link back to property file parsing
- Reproducible trivially with any value ending in `=`, e.g. `SGVsbG8gV29ybGQ=`


jasonmcintosh (approved):
Nice simple fix with some good tests thank you!!

clouddriver/clouddriver-kubernetes/src/test/groovy/com/netflix/spinnaker/clouddriver/kubernetes/provider/view/PropertyParserSpec.groovy
jasonmcintosh, line 86: Course now... gotta figure out the right solution to THIS issue ;) An empty check should help :)

jasonmcintosh:
@xudongli42 can you fix this so we can merge it?

jasonmcintosh:
Closing in favor of recent fixes with this.

Issue #4083, opened 2019-03-07
CF - ensure that failing server groups are going to be visible in the cache


Jammy-Louie:
### Issue Summary:
If a server group fails to be deployed we should still see the server group added to the cache and visible in spinnaker with the appropriate details. 

currently failed deployments do not show up in spinnaker which will cause deployment issues because it will attempt to deploy another SG with with the same name as the failed deployment.

### Cloud Provider(s):
CF

### Feature Area (if this issue is UI/UX related, please tag `@spinnaker/ui-ux-team`):
Clouddriver

Jammy-Louie:
@spinnakerbot add-label provider/cf

Jammy-Louie:
- [ ] Caching is causing rolling red/black to fail because the new instance information is not cached (https://github.com/spinnaker/spinnaker/issues/3870)
- [ ] Failed deployments do not show up immediately in the app because they are not immediately cached
- [ ] Enabling a SG does not update the cache immediately with the latest information
- [ ] Disabling a SG does not update the cache immediately with the latest information

jkschneider:
> Failed deployments do not show up immediately in the app because they are not immediately cached

@Jammy-Louie on demand caching does not fix this problem. It is fixed by https://github.com/spinnaker/clouddriver/pull/3541

PR #7688, opened 2026-05-21
fix(orca): Pipeline bulk save breaks with managed service accounts

nicolasff:
* Support managed service accounts in pipeline bulk saves
* Update the tests, add a few more
* Document the expected format of the request body in Swagger

---

##  Issue

`POST /pipelines/bulksave` has been broken on our instance ever since we enabled managed service accounts. When enabled, `SavePipelineStage` runs `SaveServiceAccountTask` before `SavePipelineTask`. In a bulk save, the stage `context` provides the pipelines in a `pipelines` key (**plural**, base64) along with `isBulkSavingPipelines: true` ([here](https://github.com/spinnaker/spinnaker/blob/c69d00e4a43803757d4563b71551d042affa8328/gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/PipelineController.groovy#L148-L160)). There is no `pipeline` key (singular).

This first task `SaveServiceAccountTask` only ever supported the single-pipeline path, its very first input check is: https://github.com/spinnaker/spinnaker/blob/c69d00e4a43803757d4563b71551d042affa8328/orca/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java#L104-L106

→ the stage fails before `SavePipelineTask` ever runs.

## Fix

`SaveServiceAccountTask`: new `executeBulk` branch when `isBulkSavingPipelines` is set. Decodes `pipelines` (encoded by Gate), and for each entry:
- assigns `id` (+ `regenerateCronTriggerIds: true`) when missing; this is used by Front50
- handles service accounts if a `roles` list is present, just like in the single-pipeline path

`SavePipelineTask`: made two methods static to reuse them, both related to service accounts.

`SaveServiceAccountTaskSpec`: 3 new test cases for bulk saves (per-pipeline service account, cron trigger IDs with and without roles, and rejection when user lacks roles)

`PipelineController` in Gate: added a `RequestBody` Swagger annotation to describe what the body should contain on the Swagger page.

### DIFF
--- a/gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/PipelineController.groovy
+++ b/gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/PipelineController.groovy
@@ -144,6 +144,9 @@ class PipelineController {
     @Parameter(description = "Application in which to run the bulk save task",
       example = "bulk_save_placeholder_app",
       required = false) String application,
+    @io.swagger.v3.oas.annotations.parameters.RequestBody( // qualified due to name collision
+      required = true,
+      description = "JSON array of pipeline definitions to upsert")
     @RequestBody List<Map> pipelines) {
     def operation = [
       description: "Bulk save pipelines",
--- a/orca/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
+++ b/orca/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/tasks/SavePipelineTask.java
@@ -176,7 +176,7 @@ public class SavePipelineTask implements RetryableTask {
     return TimeUnit.SECONDS.toMillis(30);
   }
 
-  private void updateServiceAccount(Map<String, Object> pipeline, String serviceAccount) {
+  static void updateServiceAccount(Map<String, Object> pipeline, String serviceAccount) {
     if (StringUtils.isEmpty(serviceAccount) || !pipeline.containsKey("triggers")) {
       return;
     }
@@ -210,7 +210,7 @@ public class SavePipelineTask implements RetryableTask {
     return null;
   }
 
-  private boolean runAsUserIsNullOrManagedServiceAccount(String runAsUser) {
+  private static boolean runAsUserIsNullOrManagedServiceAccount(String runAsUser) {
     return runAsUser == null
         || runAsUser.endsWith(SavePipelineStage.SERVICE_ACCOUNT_SUFFIX)
         || runAsUser.endsWith(SavePipelineStage.SHARED_SERVICE_ACCOUNT_SUFFIX);
--- a/orca/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
+++ b/orca/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java
@@ -18,6 +18,8 @@ package com.netflix.spinnaker.orca.front50.tasks;
 
 import static java.lang.String.format;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.hash.Hashing;
 import com.netflix.spinnaker.fiat.model.UserPermission;
@@ -34,6 +36,7 @@ import com.netflix.spinnaker.orca.api.pipeline.models.StageExecution;
 import com.netflix.spinnaker.orca.front50.Front50Service;
 import com.netflix.spinnaker.orca.front50.pipeline.SavePipelineStage;
 import java.nio.charset.StandardCharsets;
+import java.util.Base64;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -63,6 +66,7 @@ public class SaveServiceAccountTask implements RetryableTask {
   private final FiatStatus fiatStatus;
   private final Front50Service front50Service;
   private final FiatPermissionEvaluator fiatPermissionEvaluator;
+  private final ObjectMapper objectMapper;
   private final boolean useSharedManagedServiceAccounts;
 
   @Autowired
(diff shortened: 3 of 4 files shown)


jasonmcintosh:
Overall looks like a good fix.  ONE question: Have you tested with shared pipeline permissions which does the dedupe?  I have NOT dug to look at the APIs on SA creation but there's... an interesting aspect I could see on bulk saves on this being potentially problematic.  I THINK it's ok to invoke and front50 would handle that case, and return the correct SA on the operation, just wondering...

nicolasff:
Hey Jason, thanks for this first look! I'm not sure what you mean by "shared pipeline permissions" here. I also don't follow the part about front50 "[returning] the correct SA". The managed service account names are computed in the same Task class, as a SHA-256 hash of the sorted, null-byte-separated list of roles. By _shared_, do you mean managed service accounts that might have some overlap in their associated roles? If that's what you have in mind, I don't think this really applies and there's no need for deduping of any kind.

Considering managed service accounts enabled, if a first pipeline was saved with a trigger and the following pipeline permissions:
```json
"permissions": [ "role-A", "role-B" ],
```
then on save orca will have computed `SHA256("role-a\0role-b")`, and set the trigger user to `ee225f6a12dc0999b75689660701d634f3e5c60abfabcf39768e52119e05153e@managed-service-account`.


<details>

<summary>How the hash is computed, from the list of roles</summary>

https://github.com/spinnaker/spinnaker/blob/c69d00e4a43803757d4563b71551d042affa8328/orca/orca-front50/src/main/java/com/netflix/spinnaker/orca/front50/tasks/SaveServiceAccountTask.java#L198-L206

</details>

Now if we save a second pipeline with `permissions` set to just `["role-A"]` or `["role-A", "role-B", "role-C"]`, an entirely new hash is computed, and the managed service account _will_ be different. Each one will be saved with its own list of associated permissions.

There is no deduping needed. I think of service accounts as a sort of "synthetic user" that would only have this very specific set of roles. Maybe your own account has `role-A` to `role-Z` and many more, but this one has only these few listed roles and nothing else.

From what I can tell, the managed service account is saved largely as a handle representing the identity of the pipeline runner in a trigger scenario. It's also useful to be able to [get back](https://github.com/spinnaker/spinnaker/blob/c69d00e4a43803757d4563b71551d042affa8328/gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/controllers/AuthController.groovy#L102-L105) to the list of roles from this placeholder name. The only difference between managed service accounts and regular service accounts is that now we don't need to come up with a name for each combination of roles, and most importantly we don't need an admin to `curl` front50 to [associate that name with a list of roles](https://spinnaker.io/docs/setup/other_config/security/authorization/service-accounts/#creating-service-accounts).

jasonmcintosh:
I've been using managed pipeline permissions for many, many years.  There's two gotchas with it:
* the default implementation generated them with a name associated to the pipeline.  This means it generates by default an SA PER pipeline.  
Example:
```
      "runAsUser": "6500b0c0-3fa8-4496-87cd-4f7a0262d0c0@managed-service-account"
```
Which is visible in the JSON and set POST save operation - the 6500b is the pipeline config id.  

What was added was this:
https://github.com/spinnaker/orca/pull/4083
Which enabled a feature to use "shared" service accounts deduping and using that hash you mentioned.  Frankly I see that as "should be a default" enabled setting at THIS point.  BUT It's not.  This is why i asked on bulk save... as there's an interesting aspect of "ok which particular user is it using for the operation".  AND I've NOT looked at much of this in a WHILE as I don't use the majority of the bulk operations at this point.

jasonmcintosh:
ALSO note: ee225f6a12dc0999b75689660701d634f3e5c60abfabcf39768e52119e05153e@managed-service-account with the hash isn't "managed-service-account" but normally "shared-managed-service-account" when enabling shared pipeline permissions.

nicolasff:
Oh, sorry about the confusion.

In the hash example I should indeed have used the `@shared-managed-service-account` suffix instead of `@managed-service-account`. The `hash(roles)` form is indeed what the shared mode produces.

The two SA naming modes are toggled by `tasks.use-shared-managed-service-accounts`:
  - `off` (by default): one SA per pipeline, named <pipelineId>@managed-service-account (your `6500b...` example).
  - `on`: this is where they are deduped, named `<sha256(sortedRoles)>@shared-managed-service-account` (the feature from #4083).

Both modes are produced by `generateSvcAcctName`, which already branches on the flag.

The `bulk-save` fix delegates to that same method per pipeline, so the dedupe behavior remains unchanged; essentially we're just looping over the existing code when an array of pipelines is sent instead of a single one:

```java
for (Map<String, Object> pipeline : pipelines) {
  ...
  String serviceAccountName = generateSvcAcctName(pipeline, roles);
  if (pipelineRolesChanged(serviceAccountName, roles)
      && !validateAndSaveServiceAccount(user, serviceAccountName, roles)) {
    return TaskResult.ofStatus(ExecutionStatus.TERMINAL);
  }
  SavePipelineTask.updateServiceAccount(pipeline, serviceAccountName);
}
```

In shared mode, two pipelines in the same bulk request with identical roles:
  - will hash to the same SA name → the value of `trigger.runAsUser` in _both_ pipelines gets rewritten to the same `<hash>@shared-managed-service-account`.
  - and will cause orca to call front50 at most twice with the same payload, those being idempotent updates.

So no duplicate SA records are produced, and the dedupe is preserved.

Re: "which particular user is it using for the operation": the bulk path uses the same value as the single-pipeline path, so the calling user is identical in both code paths.

As for how we've tested this so far, we are running it on Spinnaker instances where shared managed service accounts are enabled; in fact the vast majority of our service accounts are `<hash>@shared-managed-service-account`. I mentioned earlier that we had seen the bulk endpoint break when "we enabled managed service accounts"; what I actually meant was that we saw it break when we enabled _shared_ managed service accounts, since that's what we use. In any case, the bulk endpoint breaks whenever `use-managed-service-accounts` is on, independently of the shared flag.

nicolasff:
(rebased [without any changes](https://github.com/spinnaker/spinnaker/compare/4aeeba5a6a6cba9e98bc3140fc85f2c6f6d61b61..555810f9c90b2ff67071e66d0a39329b9314a134))

jasonmcintosh (approved):
Approving and good details on the state of bulk interaction with pipeline permissions thanks!!

jasonmcintosh:
Yeah I kinda wondered - i felt like the bulk operation might have been broken a while with pipeline permissions which is why I kinda asked - wasn't sure if this was fixing that breakage OR if this was ... really a separate issue with the shared pipeline permissions.  Overall though I think this looks good and approved.

PR #7692, opened 2026-05-26
fix(yaml): A few classes were not using safe constructors. Fix these …

shlomodaari:
…(#7669) (backport to 2025.3.x)

This backport also includes YamlHelper, YamlParserProperties, and YamlParserPropertiesCondition which were not present in 2025.3.x.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .gradle/
 .idea/
 build/
+.claude
 
 *.iml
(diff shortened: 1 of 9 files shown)


jasonmcintosh:
Closing for now as some of the classes in this don't exist at this release branch.  IF we want to backport more, can but... we'll see if needed at this point in time.

PR #7695, opened 2026-05-27
chore(api-token): hardening fixes

stevenlandrydd:
# API token: hardening fixes

Follow-ups to the upstream API-tokens feature (#185) addressing atomicity,
spec-compliance, and a couple of small correctness bugs found while reviewing
the save/auth/revocation paths.

## Changes

- **`RedisApiTokenRepository.save` — MULTI/EXEC instead of pipeline.**
  The hash key, id key, and principal ZSET now commit together. An aborted
  EXEC (returns `null`) is surfaced as an exception instead of silently
  succeeding, so a partial state can't leave a readable hash key with no id
  record.

- **Self-healing name reservation.**
  The `SET NX` on the name index key now carries a 60s TTL. If the JVM dies
  or the connection drops between the reservation and the MULTI commit (or
  the explicit cleanup itself fails), the orphan reservation is auto-evicted
  and the principal can retry the same name. On a successful commit the TTL
  is replaced with the token's real lifetime — `EXPIREAT` for expiring
  tokens, `PERSIST` for non-expiring ones (otherwise a non-expiring token's
  name would free up after 60s).

- **RFC 7235-compliant `Bearer` matching.**
  Both `ApiTokenRequestMatcher` and `ApiTokenAuthenticationFilter` now match
  the `Bearer` auth-scheme case-insensitively (`bearer …` / `BEARER …` are
  valid per RFC 7235 §2.1). The opaque `spk_…` token prefix stays
  case-sensitive.

- **`MDC.remove` locale fix.**
  Use `Locale.ROOT` for the `toUpperCase` of the header name so the MDC key
  is computed deterministically regardless of the JVM's default locale
  (avoids the Turkish-locale `i → İ` class of bugs that would prevent
  scrubbing the plaintext token from MDC).

- **`RedisApiTokenRepository.delete` — surface aborted EXEC.**
  A `null`/empty result from `tx.exec()` (WATCH conflict, connection drop)
  now throws. Without this the token would remain authenticatable until its
  next natural expiry — which for non-expiring tokens is never. Surfacing
  the failure lets the controller return 5xx and the caller retry.

## Test plan

Unit tests added/extended in `RedisApiTokenRepositoryTest` and
`ApiTokenAuthenticationFilterTest` for:

- atomic save commit + rollback on EXEC failure
- name reservation TTL behavior on success/failure paths
- case-insensitive `Bearer` token extraction
- delete throws on aborted EXEC

Run: `./gradlew :gate:gate-web:test --tests '*ApiToken*'`


jasonmcintosh (dismissed):
Case sensitivity *SIGH* so many times that has bitten and easy to miss.  NO major concerns on this looks like a nice little improvement ;)

jasonmcintosh (approved):

PR #7701, opened 2026-06-09
fix(echo): GitHub status conversion exception

### LINKED ISSUE #7700: echo: fix SpinnakerConversionException when sending GitHub status checks
## Problem

After upgrading to Spinnaker 2026.1.0, echo can log errors like:

> Failed to send github status for application: '****' pipeline: '****', com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerConversionException: Failed to process response body: Cannot construct instance of `retrofit2.Response` ...

This happens when echo sends GitHub status checks for pipelines/stages.

## Suspected root cause

This appears to be a Retrofit 1 -> Retrofit 2 migration issue. The current service method is declared as:

- `Call<Response<ResponseBody>> updateCheck(...)`

In Retrofit 2, `Call<T>` already returns a `retrofit2.Response<T>`, so the body type should be `ResponseBody` rather than another `retrofit2.Response`.

A similar incorrect pattern also exists in `CDEventsSenderClient.sendCDEvent(...)`.

## Impact

- GitHub status updates can fail with `SpinnakerConversionException`
- The HTTP request itself may still succeed, but response deserialization fails afterward
- Retry behavior may cause duplicate attempts

## Proposed fix

- Change `GithubService.updateCheck(...)` to return `Call<ResponseBody>`
- Change `CDEventsSenderClient.sendCDEvent(...)` to return `Call<ResponseBody>`
- Update `CDEventsSenderService` to use `Retrofit2SyncCall.executeCall(...)` so it still returns `Response<ResponseBody>` to callers

## Related

Potentially related to the Retrofit 1 -> Retrofit 2 migration in echo:
- https://github.com/spinnaker/echo/pull/1466

A similar Retrofit response-type issue was previously fixed in:
- https://github.com/spinnaker/echo/pull/1482

jgrumboe:
## Summary

Fix a Retrofit 2 migration bug in echo that causes `SpinnakerConversionException` when echo sends GitHub status checks, and the same issue for CDEvents responses.

After upgrading to Spinnaker 2026.1.0, echo can log errors like:

> Failed to send github status for application: '****' pipeline: '****', com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerConversionException: Failed to process response body: Cannot construct instance of `retrofit2.Response` ...

## Root cause

Two Retrofit service interfaces were migrated with the wrong return type:

- `GithubService.updateCheck(...)`
- `CDEventsSenderClient.sendCDEvent(...)`

Both were declared as `Call<Response<ResponseBody>>`, but in Retrofit 2 the generic parameter of `Call<T>` should be the actual response body type, not another `retrofit2.Response`.

This causes Jackson to try to deserialize the HTTP response into `retrofit2.Response<ResponseBody>`, which fails with `SpinnakerConversionException`.

## Changes

- Change `GithubService.updateCheck(...)` to return `Call<ResponseBody>`
- Change `CDEventsSenderClient.sendCDEvent(...)` to return `Call<ResponseBody>`
- Update `CDEventsSenderService` to use `Retrofit2SyncCall.executeCall(...)` so it still returns `Response<ResponseBody>` to callers

## Impact

- GitHub status updates no longer fail during response deserialization
- CDEvents response handling keeps the same caller contract
- The actual HTTP request behavior is unchanged

## Related

Fixes #7700.

### DIFF
--- a/echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsSenderClient.java
+++ b/echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsSenderClient.java
@@ -18,13 +18,12 @@ package com.netflix.spinnaker.echo.cdevents;
 
 import okhttp3.ResponseBody;
 import retrofit2.Call;
-import retrofit2.Response;
 import retrofit2.http.Body;
 import retrofit2.http.POST;
 import retrofit2.http.Path;
 
 public interface CDEventsSenderClient {
   @POST("{brokerUrl}")
-  Call<Response<ResponseBody>> sendCDEvent(
+  Call<ResponseBody> sendCDEvent(
       @Body String cdEvent, @Path(value = "brokerUrl", encoded = true) String brokerUrl);
 }
--- a/echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsSenderService.java
+++ b/echo/echo-notifications/src/main/java/com/netflix/spinnaker/echo/cdevents/CDEventsSenderService.java
@@ -59,7 +59,7 @@ public class CDEventsSenderService {
             .create(CDEventsSenderClient.class);
     String jsonEvent = converterFactory.convertCDEventToJson(cdEvent);
     log.info("Sending CDEvent Json {} ", jsonEvent);
-    return Retrofit2SyncCall.execute(
+    return Retrofit2SyncCall.executeCall(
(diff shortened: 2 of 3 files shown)


jgrumboe:
@jasonmcintosh, I've verified this fix in our setup.

jasonmcintosh (approved):
I'd normall prefer tests :) BUT I'm good with this without.  Can add tests as a later PR

jasonmcintosh:
@mergifyio backport release-2026.1.x release-2026.0.x release-2025.4.x release-2025.3.x

jasonmcintosh:
FYI got tests incoming for this PR :)

Issue #7699, opened 2026-06-09
Track migration from AWS SDK for Java v1 → v2 (v1 EoS Dec 31, 2025)


jgrumboe:
### Issue Summary
AWS has announced that **AWS SDK for Java v1.x reaches end-of-support on December 31, 2025** ([announcement](https://aws.amazon.com/blogs/developer/announcing-end-of-support-for-aws-sdk-for-java-v1-x-on-december-31-2025/)). After that date, v1.x will receive no security patches, bug fixes, or service updates. Spinnaker still depends on v1 in several services (pinned via `versions.aws` / `com.amazonaws:aws-java-sdk-bom` in [kork/spinnaker-dependencies/spinnaker-dependencies.gradle](https://github.com/spinnaker/kork/blob/master/spinnaker-dependencies/spinnaker-dependencies.gradle)) and needs to fully migrate to **`software.amazon.awssdk` (v2)**.

This issue is the umbrella tracker for that migration, so there's one place to follow progress, surface blockers, and call out user-visible behavior changes.

### Cloud Provider(s)
AWS

### Feature Area
Dependencies / cross-service (clouddriver, echo, front50, igor, kork, halyard, orca, rosco, kayenta)

### Why now
- v1.x EoS: **2025-12-31** — no further security or maintenance updates after that date.
- v1 and v2 can coexist on the classpath, so this can be done incrementally per service / per module.
- Some v2 APIs require behavior changes that need to be communicated (e.g. signer selection, region resolution).

### Progress so far
The following PRs are already merged (all credit to @kirangodishala):

- [x] echo → v2 — spinnaker/spinnaker#7243
- [x] front50 (incl. halyard's front50-s3) → v2 — spinnaker/spinnaker#7250
- [x] front50/s3: always set region for S3-compatible storage (follow-up) — spinnaker/spinnaker#7268
- [x] igor → v2 — spinnaker/spinnaker#7283
- [x] clouddriver-artifacts (S3) → v2 — spinnaker/spinnaker#7301

### Remaining work (rough checklist — please add/correct in comments)
- [ ] **kork** — still publishes `kork-aws`, `kork-secrets-aws`, `kork-pubsub-aws` etc. against v1; remove `com.amazonaws:aws-java-sdk-bom` once all consumers migrate
[...]

PR #7705, opened 2026-06-10
feat(aws): make AmazonClientProvider bilingual (v1 + v2 SDK)

jgrumboe:
Add AWS SDK v2 (software.amazon.awssdk) factory methods to AmazonClientProvider alongside the existing v1 ones. Purely additive — no consumer changes needed to merge. Unblocks clouddriver-ecs -> v2 (spinnaker/spinnaker#7699, Phase 4 step 1).

Changes:
- clouddriver-aws.gradle: add 7 v2 service deps + sts (applicationautoscaling, cloudwatch, ecr, ecs, iam, secretsmanager, servicediscovery, sts)
- AmazonCredentials: add getV2CredentialsProvider() returning DefaultCredentialsProvider for the default chain
- AssumeRoleAmazonCredentials: override getV2CredentialsProvider() with a lazily-initialised SpinnakerStsAssumeRoleCredentialsProviderV2
- SpinnakerStsAssumeRoleCredentialsProviderV2: v2 equivalent of NetflixSTSAssumeRoleSessionCredentialsProvider using StsAssumeRoleCredentialsProvider; handles standard/GovCloud/China
- AwsSdkV2ClientSupplier: Guava LoadingCache for v2 clients keyed by (builder class, credentials provider identity, region); 10-min expiry
- AmazonClientProvider: wire AwsSdkV2ClientSupplier and add 7 new getters with V2 suffix (no skipEdda param - Edda is v1-only because it uses a JDK dynamic proxy over v1 service interfaces): getAmazonEcsV2, getAmazonEcrV2, getIamV2, getAmazonCloudWatchV2, getAmazonSecretsManagerV2, getAmazonServiceDiscoveryV2, getAmazonApplicationAutoScalingV2

Tests: 18 new unit tests across 3 test classes; all pass. Spotless: clean.


clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
```diff
@@ -309,6 +317,7 @@ public AmazonClientProvider(
     this.awsSdkClientSupplier =
         new AwsSdkClientSupplier(
             rateLimiterSupplier, registry, retryPolicy, requestHandlers, proxy, useGzip);
+    this.awsSdkV2ClientSupplier = new AwsSdkV2ClientSupplier();
```
jasonmcintosh: For an initial implementation... I'm ok but long term we'd want:
* Rate limiting support
* Metrics somehow ;) though given where I'd LIKE to go with the MeterRegistry & OTEL... I'm not sure how critical THAT is
* Retry policy passing
* ETC
SOME of this I see... as somewhat native already, just looking at the options already passed to the base provider.  I'm not sure we need it in the INITIAL PR, but will want some of this soonish I'd think.  I believe a FEW things are already available with the new SDK but not looked.

jgrumboe: I'll check what's possible and a low hanging fruit.

jgrumboe: > 🤖 _AI-assisted analysis — sanity-check before relying on it._

Agreed on all of these, and the v2 supplier looks well-positioned. A few notes confirming the "somewhat native already" intuition:

- *Rate limiting:* RateLimiterSupplier / ServiceLimitConfiguration can be reused as-is; the only new piece is a v2 ExecutionInterceptor twin of RateLimitingRequestHandler (v2 has no RequestHandler2).

- *Retry policy:* native in v2, but the v1 com.amazonaws.retry.RetryPolicy isn't type-compatible with v2's software.amazon.awssdk.core.retry.RetryPolicy, so it needs translating rather than passing through.

- *Metrics:* v2's native MetricPublisher SPI is likely a better fit than the manual Spectator counter, and aligns with the MeterRegistry/OTEL direction — probably why this feels less critical now.

- *Proxy / user-agent / connections* are native too; gzip is handled differently in v2, so there may be no exact setUseGzip analog.

None of this looks blocking for the initial PR. A follow-up issue to track parity (rate limiting + retry first, metrics via MetricPublisher) would keep it from getting lost.

jgrumboe:
@jasonmcintosh 
Ratelimiter, Retrypolicy, metrics, proxy, and user-agent was added.

jasonmcintosh:
Looking now ;)

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AwsSdkV2ClientSupplier.java
```diff
@@ -0,0 +1,291 @@
+/*
+ * Copyright 2025 Netflix, Inc.
```
jasonmcintosh: DO we need to change the copyright? IF So whom?

jgrumboe: Good catch.
What about "The Linux Foundation"?

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/SpectatorMetricPublisher.java
```diff
@@ -0,0 +1,92 @@
...
+ * <p>Records:
+ *
+ * <ul>
+ *   <li>{@code aws.sdk.v2.apiCallDuration} — timer of overall API call duration
+ *   <li>{@code aws.sdk.v2.retryCount} — counter of retry attempts
+ * </ul>
+ *
+ * <p>Metrics are tagged with {@code serviceName} and {@code operationName} when available.
+ */
+public class SpectatorMetricPublisher implements MetricPublisher {
```
jasonmcintosh, line 45: SOoo this is some what of a side thing, and could be later.  WONDERING If it has to be "Spectator" or if could use plane ol "Micrometer" registry protocols.  NOT a huge deal as stated... just a long term be nice to get off of spectator except as an implementation detail.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/SpectatorMetricPublisher.java
```diff
@@ -0,0 +1,92 @@
...
+      if (retries > 0) {
+        registry
+            .counter(
+                "aws.sdk.v2.retryCount", "serviceName", serviceName, "operationName", operationName)
+            .increment(retries);
+      }
+    }
+
+    // Recurse into child collections (e.g. per-attempt metrics)
+    for (MetricCollection child : metricCollection.children()) {
```
jasonmcintosh, line 95: NOT dug but I DO worry about a... "unbounded recursion" situation here....

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
```diff
@@ -309,6 +317,13 @@ public AmazonClientProvider(
     this.awsSdkClientSupplier =
         new AwsSdkClientSupplier(
             rateLimiterSupplier, registry, retryPolicy, requestHandlers, proxy, useGzip);
+    software.amazon.awssdk.core.retry.RetryPolicy v2RetryPolicy = buildV2RetryPolicy(retryPolicy);
```
jasonmcintosh, line 358: I don't know if we have any OTHER requestHandlers, or if plugins add any - thi sis a lost functionality. E.g. if we need to support adding "interceptors" in other aspects.  I'm good with leaving that as a future PR to the v2 spec though

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AwsSdkV2ClientSupplier.java
jasonmcintosh: OR spinnaker.io ;) I'm good with that as well.

jasonmcintosh (dismissed):
Overall no major complaints.  JUST wonder about the "copyright".  THAT is almost an entirely separate discussion around should there be a "Contributor agreement" thing.  The project has NOT to this point required that.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AwsSdkV2ClientSupplier.java
jasonmcintosh: NORMALLY FYI this is the "company" or "entity" doing the contribution :) AKA harness or previously armory.

jgrumboe: I'm fine with "spinnaker.io".

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
jgrumboe: Good point. I've added a v2ExecutionInterceptor(ExecutionInterceptor) method to AmazonClientProvider.Builder — it's the v2 equivalent of the existing requestHandler(RequestHandler2). The v2 supplier now accepts a List<ExecutionInterceptor> that gets attached to every client it builds.
Commit 2d93263

No known consumer today, but the extension point is wired end-to-end so plugins can inject custom interceptors without touching internals when the time comes.

Or do you prefer not to add it and just add a comment about the missing functionality?

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/AwsSdkV2ClientSupplier.java
jgrumboe: Changed in [b8feb48](https://github.com/spinnaker/spinnaker/pull/7705/commits/b8feb488cb33169a14e93a86ca6dac2992afdb4f)

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/SpectatorMetricPublisher.java
jgrumboe: I've added a comment in 0da17e8, so we don't forget.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/SpectatorMetricPublisher.java
jgrumboe: Good call. Added a MAX_DEPTH = 5 guard — SDK metric collections are only ~2 levels deep in practice (parent + per-attempt children), but this caps it defensively so a hypothetical pathological collection can't blow the stack.

Commit 0da17e8

jgrumboe:
> Overall no major complaints. JUST wonder about the "copyright". THAT is almost an entirely separate discussion around should there be a "Contributor agreement" thing. The project has NOT to this point required that.

Thanks for the review.
I've addressed your comments (together with Kiro 😏 )

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
jasonmcintosh: I'd add it back - it's worth supporting in general.  WOULD NOT surprise me on this particular one if there were users using it for SOME odd thing.

jasonmcintosh (approved):
Good start towards AWS V2 sdk support with no impact to existing operations!

jasonmcintosh:
Plan to merge this tomorrow barring any last minute concerns FYI!

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonClientProvider.java
jgrumboe: Ok, we keep it there.

PR #7726, opened 2026-06-13
chore(tests): Adds integration tests for MTLS & OAuth2 flows with keycloak

jasonmcintosh:
This shows the full config working :) Emulates the SAML based integration tests


jasonmcintosh:
Adding x509 auth now... so some more tweaks incoming

christosarvanitis (approved):

PR #7741, opened 2026-06-18
feat(teams): Allow adaptive and custom templates for MS Teams messages using Jinja

jasonmcintosh:
Replaces the current implementation (which uses a hard coded POJO) with jinja template based POST body contents and should support more features around teams context.

* adds some watches for the deprecated URLs (which likely don't work)
* Adds a migration guide and some notes on the change (Claude helped with a lot of this... with some help of course)

ONE key change - this changes the default message!  I've NOT had a great test platform for this, so would welcome TEAMS users to validate!

Should get a lot of the README.md into a docs file.  I could see supporting this pattern for some of the other notifications longer term (Jinja the message, then jinja the actual payload contents with the message in the template) to allow dynamic configuration of payloads to platforms that support such a thing.


echo/echo-notifications/src/main/resources/templates/microsoftteams/pipeline-notification.jinja
```diff
@@ -0,0 +1,86 @@
...
+          },
+          {
+            "type": "FactSet",
+            "spacing": "Medium",
+            "facts": [
+              {% if applicationName %}
+              {
+                "title": "Application",
+                "value": "{{ applicationName }}"
+              },
```
christosarvanitis, line 41: The trailing commas will likely create an exception if one of the fields is not present. Could we switch to leading comma before each block to avoid this?

christosarvanitis (approved):
Approving with a concern on the jinja template

echo/echo-notifications/src/main/resources/templates/microsoftteams/pipeline-notification.jinja
jgrumboe: Just out of curiosity: should we use [Adaptive Cards Templating Language](https://learn.microsoft.com/en-us/adaptive-cards/templating/language) instead of Jinja? It's built-in, and you can manage field visibility dynamically based on the data.

jasonmcintosh: So the trailing comma shouldn't be an issue since it's skipping the applicationName kinda thing?  Maybe I'm missing something but not seeing a line where the trailing comma wouldn't get skipped by the jinja templating handling?  ALSO this is the "default" template - it CAN be overridden to any other format.

The adaptive card templating can do this it appears, but I don't know if it's AS flexible as jinja.  This can run on BOTH sides - spinnaker and remote this is literally JUST the payload to MS so can be changed to some OTHER format if needed/wanted.  I'm open to either path, but I thought this might be the "simplest default"...

jasonmcintosh: LAST note :) To get it into the payload, already using JINJA templating.  Chose this vs. the ObjectMapper path as it enables a bit more capabilities. EITHER way have to generate the payload somehow.  IF we're already templating, not sure matters which path used...

echo/echo-notifications/src/main/resources/templates/microsoftteams/pipeline-notification.jinja
```diff
@@ -0,0 +1,86 @@
...
+                "value": "{{ description }}"
+              },
+              {% endif %}
+              {% if customMessage %}
+              {
+                "title": "Message",
+                "value": "{{ customMessage }}"
+              },
+              {% endif %}
+              {% if status %}
```
jasonmcintosh, line 67: NOTE "status" becomes a required field in this PR...

echo/echo-notifications/src/main/resources/templates/microsoftteams/pipeline-notification.jinja
jasonmcintosh: Per slack :) It's status that becomes required with this, if status is missing, the JSON has an extra common on this list...

Issue #7740, opened 2026-06-17
CDEvents webhook returns HTTP 500 despite successful pipeline trigger


celiawa123:
### Summary

The /webhooks/cdevents/{source} endpoint returns HTTP 500 to the caller even when the pipeline is successfully triggered on 2025.3.2.  The same endpoint returns HTTP 200 in 1.35.4.

### Affected Versions
2025.3.2

### Working Version

- Gate: 6.63.0
- Echo: 2.42.0 (Spinnaker 1.35.4)

### Steps to Reproduce

```bash
curl -X POST \
  -H "Ce-Id: 1" \
  -H "Ce-Specversion: 1.0" \
  -H "Ce-Type: dev.cdevents.artifact.published.0.1.0" \
  -H "Ce-Source: test" \
  -H "Content-Type: application/json" \
  -d @example.json \
  https:/^[[A/<gate-url>/webhooks/cdevents/test
```

example.json:
```json
{
  "context": {
    "id": "test-id",
    "type": "dev.cdevents.artifact.published.0.1.0",
    "source": "https://ci-system/",
    "version": "0.1.2"
  },
  "subject": {
    "type": "ARTIFACT",
    "content": {
      "pipelineName": "my-pipeline",
      "outcome": "SUCCESS"
    }
  }
}
```

### Expected Behavior

HTTP 200 response (as in 1.35.4). Pipeline is triggered.

### Actual Behavior

HTTP 500 with:
* TLSv1.2 (IN), TLS header, Unknown (23):
* Connection #0 to host spinnaker-dev-api.rnd.gic.ericsson.se left intact
```{"error":"Internal Server Error","exception":"com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerConversionException","message":"Failed to process response body: No content to map due to end-of-input\n at [Source: (okhttp3.ResponseBody$BomAwareReader); line: 1, column: 0]","status":500,"timestamp":"2026-06-17T05:43:58.291+00:00"}```

**Pipeline is triggered successfully (confirmed via Echo logs: Successfully triggered Pipeline(...)).**

### Root Cause Analysis
With the help of kiro:

In Gate's WebhookService.groovy, the CDEvents webhook method calls:
groovy
Retrofit2SyncCall.execute(echoService.webhooks(source, cdEvent, ceDataJsonString, ...))

Echo's webhook controller processes the CDEvent and triggers the pipeline asynchronously, but returns an empty response body. With
[...]

PR #7747, opened 2026-06-23
Fix CDEvents webhook returns HTTP 500 despite successful pipeline trigger

anooptejt:
## Summary
description: The issue is caused by Retrofit attempting to deserialize an empty response body in the CDEvents webhook endpoint. This behavior was different in Retrofit 1.x where void/empty responses were handled transparently without attempting body deserialization.; impact: Users receive a HTTP 500 error when triggering a pipeline via the CDEvents webhook, even though the pipeline is successfully triggered.

## Issue
Fixes or relates to https://github.com/spinnaker/spinnaker/issues/7740

## Files Changed
- gate/gate-web/src/main/groovy/com/netflix/spinnaker/gate/services/WebhookService.groovy

## Validation
- Targeted tests/checks have not reported results yet.

## Security
- summary: No vulnerabilities were detected in the local scan. The code change does not introduce any new security risks.; details: The changes made are focused on handling empty response bodies gracefully, which is a common practice to avoid deserialization errors.

## Guardrails
- Draft PR created only after explicit user approval.
- Commit metadata uses the contributor GitHub identity.


anooptejt:
@Mergifyio queue

jasonmcintosh:
Need tests added to show the issues... also ... tagging @xibz who does a lot more of CDEvents stuff.  This... honestly feels like a different issue more on the remote side, but...

xibz:
@jasonmcintosh - yep, you are correct.

@anooptejt - We want to address the root cause, not treat the symptom. So this is an issue with the response that's returned in CDEvents notification flow. This could be due to async calls, etc. Try looking at where the empty response is coming from, and addressing what is returned. I imagine it is very straightforward once you can pinpoint where the empty body is being returned from.

jasonmcintosh:
Looking at this more... I'm starting to think this isn't a situation where gate should change, but ECHO should change to always return a valid object kinda thing.  Just digging on the root issue and the stuff reported there.

jasonmcintosh:
Closing this in favor of other PR that returns the correct responses.  NOTE:  gate does NOT send those back at this time last I looked - that'd be a separate PR for that.

PR #7748, opened 2026-06-23
fix(yaml): Fix a few places that do not handle large yamls with new snakeyaml

jasonmcintosh:
Fixes issues with newer snakeyaml procesing of those resources in those locations.  I've added tests in a few places around this - but this is more a "refactoring" situation.  I could see making a few more changes later on some of the classes if needed and/or get some additional tess.


igor/igor-web/src/main/java/com/netflix/spinnaker/igor/config/HelmConfig.java
```diff
@@ -37,6 +38,9 @@
 @Slf4j
 public class HelmConfig {
   @Bean
+  @DependsOn(
```
jasonmcintosh: NOT 100% sure this is really needed.  It's a 'safeguard' that potentially should be added elsewhere too.

igor/igor-web/src/main/java/com/netflix/spinnaker/igor/config/HelmConfig.java
```diff
@@ -37,6 +38,9 @@
...
+      "YamlHelper") // MAKE SURE the yamlHelper and it's properties are loaded FIRST... this enables
```
christosarvanitis: Can you please correct It to be @DependsOn("yamlHelper") instead of capital Y. Bean will be named yamlHelper :)

christosarvanitis:
wondering if its safer to pass the yamlParser to the constructors so we can have a guarantee that the yamlParser will be wired in the correct order. @jasonmcintosh what do you think?

jasonmcintosh:
Yeah... quite possibly.  This class was NOT written well to begin with, but... maybe time to refactor it a fair bit more.

christosarvanitis (approved):

jasonmcintosh:

PR #7751, opened 2026-06-26
fix(jooq): jOOQ 3.19.35 (Spring Boot 3.5.x) generates invalid MySQL syntax

christosarvanitis:
After upgrading to Spring Boot 3.5.x (which pulls in jOOQ 3.19.35), services using MySQL backend fail with BadSqlGrammarException on any upsert operation that uses InsertValuesStep.values() followed by onDuplicateKeyUpdate().

Example stack trace from clouddriver:
```
BadSqlGrammarException: You have an error in your SQL syntax;
... near 'as excluded on duplicate key update application = values(application)'
```
Root Cause

jOOQ 3.19.35 generates PostgreSQL-specific AS excluded syntax even when the configured dialect is MySQL, whenever onDuplicateKeyUpdate() is chained after .values():
```
insert into t (...) values (...) as excluded
on duplicate key update ...
```
MySQL rejects AS excluded as invalid syntax.


christosarvanitis:
Introduced in https://github.com/jOOQ/jOOQ/issues/16942 so we could just downgrade the jooq to `3.19.29` and avoid any code changes.

Issue #7750, opened 2026-06-24
Canceling execution failed to cancel a Jenkins job.


celiawa123:
We upgraded spinnaker version from 1.35.4 to 2025.3.2 found that cancelling an execution from Spinnaker didn't cancel the jenkins job.

Steps:
1. Run a pipeline with Jenkins job.
2. Cancel the pipeline.
3. The Jenkins job continue running.

We found this error in Orca pod:

`2026-06-24T20:08:37.188809478+02:00 stdout F 2026-06-24 18:08:37.188 ERROR 1 --- [       cancel-4] c.n.s.orca.igor.pipeline.CIStage         : [sa-ugpdupcsprx] Failed to cancel stage (stageId: 01KVXCTEJKYXMWXDT1NGDDNE4T, executionId: 01KVXCTEJJ3BX9ARYEY4AFT007), e: A @Path parameter must not come after a @Query. (parameter #3)`

celiawa123:
The issue seems in this function https://github.com/spinnaker/spinnaker/blob/71a7d490d2487a71a425b8fd557ce66bb583d781/orca/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/IgorService.java#L54
```
  @PUT("masters/{name}/jobs/stop/{queuedBuild}/{buildNumber}")
  Call<String> stopWithJobNameAsQueryParameter(
      @Path("name") String master,
      @Query(value = "jobName") String jobName,
      @Path(encoded = true, value = "queuedBuild") String queuedBuild,
      @Path(encoded = true, value = "buildNumber") Long buildNumber,
      @Body String ignored);
```

Retrofit parses method parameters in declaration order. Once it encounters a @Query, it expects all subsequent URL-related parameters to also be queries (or body/header/etc.). A later @Path is rejected.

Could someone fix the problem. Thank you.

jasonmcintosh:
We'd welcome PRs to fix this.  There's several of these that have been fixed in the past, looks like this one got missed.

celiawa123:
Thanks Jason. We'll submit a PR to fix this.

Luca-Mossa:
Branch including the fix has been merged. This issue can be closed. Thanks.

celiawa123:
Thank you Luca.

PR #7752, opened 2026-06-26
fix(jenkins): cancel Jenkins job when Spinnaker pipeline is cancelled

Luca-Mossa:
- Spinnaker pipeline cancellation does not cancel the underlying Jenkins job.
- Cancelling a Spinnaker pipeline leaves the associated Jenkins build running.
- Jenkins jobs continue running after the parent Spinnaker pipeline is cancelled.

The issue is that Retrofit requires @Path parameters to come before @Query parameters in the method signature. Currently, @Query comes before the @Path parameters, which violates Retrofit's parsing rules.

Fix: Reorder the parameters so @Path parameters come first:

```
@PUT("masters/{name}/jobs/stop/{queuedBuild}/{buildNumber}")
Call<String> stopWithJobNameAsQueryParameter(
    @Path("name") String master,
    @Path(encoded = true, value = "queuedBuild") String queuedBuild,
    @Path(encoded = true, value = "buildNumber") Long buildNumber,
    @Query(value = "jobName") String jobName,
    @Body String ignored);
```

The changes:
1. Move queuedBuild and buildNumber @Path parameters before jobName
2. Move jobName @Query parameter after all @Path parameters

This respects Retrofit's parameter order requirement: @Path → @Query → @Body/@Header/etc.

References: https://github.com/spinnaker/spinnaker/issues/7750

### DIFF
--- a/orca/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/BuildService.java
+++ b/orca/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/BuildService.java
@@ -51,7 +51,7 @@ public class BuildService {
     return this.igorFeatureFlagProperties.isJobNameAsQueryParameter()
         ? Retrofit2SyncCall.execute(
             igorService.stopWithJobNameAsQueryParameter(
-                master, jobName, queuedBuild, buildNumber, ""))
+                master, queuedBuild, buildNumber, jobName, ""))
         : Retrofit2SyncCall.execute(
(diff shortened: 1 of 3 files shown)


jasonmcintosh:
IF you could fix BuildServiceSpec tests that'd help!  We can get this in and thanks for the PR!

jasonmcintosh (approved):

jasonmcintosh (approved):

PR #7755, opened 2026-06-29
fix(mvc): UrlHandlerFilter to trailing slash handle route matching

christosarvanitis:
Spring Boot 3.x (Spring Framework 6.0+) removed lenient trailing-slash route matching, so for example PUT /artifacts/fetch/ no longer matches a controller mapped to /artifacts/fetch and returns 404. This broke inter-service calls such as Orca → Clouddriver fetchArtifact, where the Retrofit client emits a trailing slash. Because this is a cross-cutting regression affecting every service, the fix is implemented once in kork-web so all services inherit it with no per-service changes.

- New auto-configuration UrlHandlerFilterAutoConfiguration (kork-web) registers a UrlHandlerFilter as a FilterRegistrationBean at highest precedence, using wrapRequest() so the trailing slash is trimmed transparently (no redirect round-trip for internal calls). The setUseTrailingSlashMatch(true) — deprecated since Spring 6.0 so using the UrlHandlerFilterAutoConfiguration follows the recommended remediation. 
- Enabled by default; can be disabled with url-handler.trailing-slash.enabled=false or scoped via url-handler.trailing-slash.path-patterns.
```
url-handler:
  trailing-slash:
    enabled: true          # default; set false to opt out
    path-patterns:
      - "/**"              # default; narrow if desired
```

## Codebase Audit (trailing-slash controller mappings)
- **Core data-plane services are safe**: `clouddriver`, `orca`, `front50`, `igor`, `fiat`, `rosco`, `kayenta`, `keel` have **zero** trailing-slash mappings.
- `gate` (`RootController`) and `echo` (`HistoryController`) only map the **root `/`**, which the filter does **not** break (verified by `rootMappingIsNotBrokenWhenEnabled`).
- **`halyard` is the exception**: it depends on `kork-web` and maps many endpoints with explicit trailing slashes (`.../haServices/`, `/generate/`, `/clean/`, `/connect/`, `/deploy/`, `/version/`, etc.), which the `hal` CLI calls with the trailing slash. Halyard therefore opts out via `url-handler.trailing-slash.enabled: false`.

### DIFF
--- a/kork/kork-web/src/test/java/com/netflix/spinnaker/kork/web/trailingslash/FetchController.java
+++ b/kork/kork-web/src/test/java/com/netflix/spinnaker/kork/web/trailingslash/FetchController.java
@@ -15,6 +15,7 @@
  */
 package com.netflix.spinnaker.kork.web.trailingslash;
 
+import org.springframework.web.bind.annotation.PostMapping;
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
 import org.springframework.web.bind.annotation.RestController;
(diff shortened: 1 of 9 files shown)


kork/kork-web/src/test/java/com/netflix/spinnaker/kork/web/trailingslash/FetchController.java
```diff
@@ -0,0 +1,36 @@
...
+ * Test controller mapped to {@code /fetch} (no trailing slash), mirroring the real {@code
+ * ArtifactController.fetch} mapping that Orca calls with a trailing slash.
+ *
+ * <p>Lives in an isolated package so it is not picked up by other tests' component scans. It is
+ * registered explicitly via {@code @SpringBootTest(classes = ...)} rather than component scanning.
+ */
+@RestController
+public class FetchController {
+
+  @PutMapping("/fetch")
```
jasonmcintosh, line 33: What about controllers that have BOTH e.g. 
https://github.com/spinnaker/spinnaker/pull/7682/changes#diff-29762931835ddd5f19e509e306bc434d4811411532a1fb9e09868f82e8ce8bcaR290-R292
I fixed a few controllers here to set BOTH the / and non slash.  Want to make sure that's viable.  THEN there's a question of should we fix the controllers long term & remove this filter.

I think this will do it for now but... :) JUST want to make sure we get this tested so that if some OTHER change comes up we catch it.

christosarvanitis: Only halyard has trailing / in the endpoints. For halyard explicitly disabled the config.

jasonmcintosh (approved):

PR #7765, opened 2026-07-02
feat(deck/kubernetes): remove angular dependency

mattgogerly:
Removes the `angular` dependency from the `kubernetes` package by rewriting all Angular controllers and templates as React components.

The misc core changes are required to remove Angular's state management.


jasonmcintosh (approved):
i'm NOT the best at UI stuff, but... I'm approving it - and should get tested before the next release kinda thing.

mattgogerly:
> i'm NOT the best at UI stuff, but... I'm approving it - and should get tested before the next release kinda thing.

Absolutely, I'd encourage everyone to try out the main version. We've been running this in prod for a few months and haven't encountered any issues, but we don't use all of the functionality.

jgrumboe:
Also tried it in our staging environment, no issues so far.

PR #7766, opened 2026-07-03
fix(aws): bridge v1 credentials to v2 in AmazonCredentials

jgrumboe:
## Summary

`AmazonCredentials.getV2CredentialsProvider()` returned `DefaultCredentialsProvider` which cannot discover AWS credentials in non-AWS environments (e.g. GKE pods with static keys configured via Halyard). This caused 403 "security token invalid" errors for all AWS SDK v2 API calls on cross-account ECS caching agents.

## Root Cause

All Spinnaker accounts have a configured v1 `AWSCredentialsProvider` that handles STS assume-role with automatic token refresh. The v2 path (`getV2CredentialsProvider()`) ignored this entirely and relied on `DefaultCredentialsProvider.create()`, which has no way to discover the credentials on non-AWS infrastructure.

## Fix

Bridge the v1 provider to v2 by adapting each resolved credential (supporting both `AWSSessionCredentials` and basic credentials) on every `resolveCredentials()` call. This makes the v2 SDK path work everywhere the v1 path works, without requiring a separate v2 credential discovery chain.

## Testing

- 5 new unit tests covering basic creds, session creds, token refresh delegation, assume-role sessions, and null-provider fallback
- All `clouddriver-aws` tests pass (24 tests)
- All `clouddriver-ecs` tests pass
- Verified in production on GKE: all ECS caching agents (IAM, ServiceDiscovery, Secrets, CloudWatch, ELBv2 TargetHealth, ScalableTargets) working without errors

### DIFF
--- a/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentials.java
+++ b/clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/AmazonCredentials.java
@@ -20,6 +20,7 @@ import static java.util.Objects.requireNonNull;
 
 import com.amazonaws.auth.AWSCredentials;
 import com.amazonaws.auth.AWSCredentialsProvider;
+import com.amazonaws.auth.AWSSessionCredentials;
 import com.fasterxml.jackson.annotation.JsonIgnore;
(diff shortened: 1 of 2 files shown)


christosarvanitis (approved):
Thanks for the fix. Does make sense looking at the tests and the code

jasonmcintosh (approved):

jasonmcintosh:

PR #7770, opened 2026-07-07
feat(core): make pipeline-trigger execution dropdown limit configurable

shlomodaari:
The manual-trigger "Execution" dropdown (shown when the selected pipeline's trigger type is Pipeline) hardcoded a limit of 20 recent executions fetched via getExecutionsForConfigIds. This adds SETTINGS.maxPipelineTriggerExecutionOptions (default 20, same as before) so operators can override it via settings-local.js, mirroring the existing maxFetchHistoryOnEvaluateVariables pattern.

- settings.ts: add optional maxPipelineTriggerExecutionOptions field + default
- settings.js: document the default value (20)
- PipelineTriggerTemplate.tsx: read the limit from SETTINGS instead of a hardcoded literal
- PipelineTriggerTemplate.spec.tsx: add regression tests covering the default, an overridden limit, rendered option count, and re-initialization on source-pipeline change


jasonmcintosh (dismissed):

jasonmcintosh:
The ONLY concern I have is ... https://github.com/spinnaker/spinnaker/blob/main/spinnaker-kustomize/base/deck/files/settings.js this file doesn't quite match this settings.js file.  @mattgogerly not sure if any concerns here... or if they should always be in sync kinda deal?

jasonmcintosh:
NOTE :) My concern is NOT directly related to this PR - why I approve this ;)

mattgogerly:
> The ONLY concern I have is ... https://github.com/spinnaker/spinnaker/blob/main/spinnaker-kustomize/base/deck/files/settings.js this file doesn't quite match this settings.js file. @mattgogerly not sure if any concerns here... or if they should always be in sync kinda deal?

We probably want to keep them in sync.

jasonmcintosh (approved):

jasonmcintosh:
Merging can do a follow-on PR around the whole 'mismatch' of settings.js file

PR #7771, opened 2026-07-08
Remove Angular dependency from Docker/CF/Huawei/Tencent

mattgogerly:
Removes the angular dependency from Docker, CloudFoundry, Huawei & Tencent packages.


jasonmcintosh (approved):
YAY! :)

PR #7772, opened 2026-07-08
refactor(deck): remove angular dependency from dcos and cloudrun

mattgogerly:
Remove the angular dependency from `dcos` and `cloudrun` packages.


jasonmcintosh (approved):

jasonmcintosh:
UNFORTUNATELY... dcos is STILL heavily used from what I can see... so yeah still gotta migrate :(

jasonmcintosh (approved):

PR #7773, opened 2026-07-08
refactor(deck/appengine+oracle): remove angular from appengine and oracle

mattgogerly:
Remove the angular dependency from `appengine` and `oracle`.


deck/packages/app/src/app.ts
```diff
@@ -6,12 +6,12 @@ import './strictDi';
 import { CORE_MODULE } from '@spinnaker/core';
 import '@spinnaker/docker';
 import { AMAZON_MODULE } from '@spinnaker/amazon';
-import { APPENGINE_MODULE } from '@spinnaker/appengine';
+import '@spinnaker/appengine';
```
jasonmcintosh, line 9: This is perhaps a side question :) Any reason for this vs... 
```
import './appengine'
```
kinda syntax?

jasonmcintosh (approved):

deck/packages/appengine/src/domain/index.ts
```diff
@@ -2,5 +2,4 @@ export * from './IAppengineAccount';
 export * from './IAppengineInstance';
 export * from './IAppengineLoadBalancer';
 export * from './IAppengineServerGroup';
-export * from './IAppengineStageScope';
```
jasonmcintosh, line 5: I ASSUME no issues b/c of the way we're changing to TSX on this ;) aka... this should be a no impact - but what about the other IAppengine types?

jasonmcintosh: I see... https://github.com/spinnaker/spinnaker/pull/7773/changes#diff-41e9a37a83fa8b493d8a4065b9cc37dfa0f4d109919d7a9f0feda8fad2ce46f6R1 which looks like has some of this?

deck/packages/app/src/app.ts
mattgogerly: They're separate packages, so `webpack` won't find the files. I'm not sure why it was set up this way historically.

jasonmcintosh: JUST something to consider - does this make sense long term or would it make it easier if you redid some of this?  :) I'm open to whatever on that point!

deck/packages/appengine/src/domain/index.ts
mattgogerly: It was only imported by the angular controllers, isn't imported anymore.

Issue #7017, opened 2025-03-06
Orca: sql execution compression errors out against postgresql


anindyameister:
I am trying to set up spinnaker v1.37.4 on kubernetes with postgresql db. When I enable pipeline execution compression in orca, which was merged in [4620](https://github.com/spinnaker/orca/pull/4620), I'm seeing this exception in orca logs.

```
2025-03-06 03:03:51.465  INFO 1 --- [0.0-8083-exec-2] c.n.s.o.s.p.p.SqlExecutionRepository     : [anonymous] Body length 10458 for execution of 01JNMPHKG27W21BP7KW5R9GK7Z breaches the compression threshold 1024
2025-03-06 03:03:51.465  INFO 1 --- [0.0-8083-exec-2] c.n.s.o.s.p.p.SqlExecutionRepository     : [anonymous] Compression complete for 01JNMPHKG27W21BP7KW5R9GK7Z. Uncompressed body length: 10458 Compressed body length: 1453 Compression ratio: 7.197522367515485 Compression time(ms): 0 
2025-03-06 03:03:51.465  INFO 1 --- [0.0-8083-exec-2] c.n.s.o.s.p.p.SqlExecutionRepository     : [anonymous] Upserting execution into the pipelines table with id 01JNMPHKG27W21BP7KW5R9GK7Z
2025-03-06 03:03:51.467  INFO 1 --- [0.0-8083-exec-2] c.n.s.o.s.p.p.SqlExecutionRepository     : [anonymous] Upserting compressed execution into the pipelines_compressed_executions table with id 01JNMPHKG27W21BP7KW5R9GK7Z
2025-03-06 03:03:51.468  WARN 1 --- [0.0-8083-exec-2] c.n.s.k.w.e.GenericExceptionHandlers     : [anonymous] Handled error in generic exception handler
org.springframework.jdbc.BadSqlGrammarException: jOOQ; bad SQL grammar [insert into pipelines_compressed_executions (id, compressed_body, compression_type) values (?, ?, ?) on conflict (id) do update set id = ?, compressed_body = ?, compression_type = ? -- user: anonymous origin: api]; nested exception is org.postgresql.util.PSQLException: ERROR: column "compression_type" is of type compression_type_enum but expression is of type character varying
  Hint: You will need to rewrite or cast the expression.
  Position: 101
	at org.jooq_3.14.16.POSTGRES.debug(Unknown Source)
[...]

PR #7778, opened 2026-07-10
fix(sql): use inlined literal for compression_type to fix Postgres enum cast failure

shlomodaari:
**Bug**
SqlExecutionRepository fails to persist compressed pipeline executions against PostgreSQL, throwing BadSqlGrammarException on the pipelines_compressed_executions (and equivalent orchestration_stages_compressed_executions) INSERT/ON CONFLICT statement. MySQL/MariaDB are unaffected. Matches [spinnaker/spinnaker#7017](https://github.com/spinnaker/spinnaker/issues/7017).

**Root Cause**
compression_type was bound as an untyped jOOQ bind parameter. PostgreSQL allows an implicit cast from a string literal to a native enum column (compression_type_enum), but rejects an implicit cast from an untyped bind parameter of type character varying.

**Fix**
orca-sql/.../SqlExecutionRepository.kt:1223 — wrap the value with DSL.inline(...) so jOOQ renders it as an inlined SQL literal rather than a ? bind placeholder. Since it's a single combined INSERT ... ON CONFLICT (id) DO UPDATE SET ... statement, Postgres type-checks the entire statement (including the DO UPDATE SET clause) up front — so this one change fixes both the INSERT and the ON CONFLICT UPDATE paths in the same statement.

**Test Environment**
Local Kubernetes (Rancher Desktop/k3s), Spinnaker OSS deployed via Armory Operator + kustomize
PostgreSQL 16 (postgres:16-alpine), fresh database, Orca's execution repository fully switched to SQL (execution-repository.sql.enabled: true, execution-repository.redis.enabled: false), dialect: POSTGRES
Compression forced via bodyCompressionThreshold: 1 to guarantee every execution triggers the compressed-write path
**Verification Steps**
Before fix (armory/orca:2.38.0-rc1, unpatched): triggered pipeline execution → reproduced exact error:
org.postgresql.util.PSQLException: ERROR: column "compression_type" is of type compression_type_enum 
but expression is of type character varying
**Hint**: You will need to rewrite or cast the expression.
After fix (armory/spinnaker-orca:2026.07.10.16.28.53.fix-postgres-compression-type-enum-cast, built from this branch): confirmed bytecode contains the DSL.inline call in SqlExecutionRepository.class
Triggered pipeline execution twice independently → both succeeded with no exception:
sql
SELECT id, compression_type, LENGTH(compressed_body) FROM pipelines_compressed_executions;
-- 01KX6X4KVT7FCRDWTC3N9K5WWQ | ZLIB | 449
-- 01KX6X7DHWJFQ63XFZ59ZCHRMY | ZLIB | 450

### DIFF
--- a/orca/orca-sql/src/test/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositoryPostgresCompressionTest.kt
+++ b/orca/orca-sql/src/test/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositoryPostgresCompressionTest.kt
@@ -21,12 +21,15 @@ import com.netflix.spinnaker.kork.sql.config.RetryProperties
 import com.netflix.spinnaker.kork.sql.test.SqlTestUtil
 import com.netflix.spinnaker.orca.api.pipeline.models.ExecutionType
 import com.netflix.spinnaker.orca.jackson.OrcaObjectMapper
-import dev.minutest.junit.JUnit5Minutests
-import dev.minutest.rootContext
 import org.assertj.core.api.Assertions.assertThat
 import org.assertj.core.api.Assertions.assertThatCode
 import org.jooq.impl.DSL.field
+import org.junit.jupiter.api.AfterEach
 import org.junit.jupiter.api.Assumptions.assumeTrue
+import org.junit.jupiter.api.BeforeAll
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.junit.jupiter.api.TestInstance
 import org.mockito.kotlin.mock
 import org.testcontainers.DockerClientFactory
 import java.lang.System.currentTimeMillis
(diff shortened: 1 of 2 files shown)


orca/orca-sql/src/test/kotlin/com/netflix/spinnaker/orca/sql/pipeline/persistence/SqlExecutionRepositoryPostgresCompressionTest.kt
```diff
@@ -0,0 +1,165 @@
...
+ *   nested exception is org.postgresql.util.PSQLException: ERROR: column "compression_type"
+ *   is of type compression_type_enum but expression is of type character varying
+ *
+ * Root cause: `compression_type` was bound as an untyped jOOQ bind parameter (a plain Kotlin
+ * String). PostgreSQL refuses to implicitly cast an untyped bind parameter to a native enum
+ * column type, though it will implicitly cast an inlined string literal. MySQL/MariaDB never
+ * exhibited this bug (no native enum type strictness), which is why the existing, MySQL-only
+ * SqlExecutionRepositoryTest suite never caught it.
+ */
+class SqlExecutionRepositoryPostgresCompressionTest : JUnit5Minutests {
```
jasonmcintosh: "minor" but... I'm kinda trying to rip out "Minutetests" stuff - it's not really getting any maintenacne at this point.

jasonmcintosh (dismissed):
Moving from minutetests to regular tests is preferred, but otherwise I think this'll solve the issue posted in https://github.com/spinnaker/spinnaker/issues/7017

jasonmcintosh (approved):

jasonmcintosh:

PR #7787, opened 2026-07-14
chore(deps): bump base images and tool versions to remediate CVEs

jasonmcintosh:
- Ubuntu: jammy → resolute across all service Dockerfiles
- Alpine: 3.20/3.18 → 3.24; Python base: 3.12-alpine3.20 → 3.14-alpine3.24
- kubectl default: 1.22→1.30.14; drop pre-1.30 versions; add 1.31.14/1.32.13/1.33.13
- aws-iam-authenticator: 0.6.14→0.7.18; fix halyard to use GitHub releases URL
- Google Cloud SDK (the CLI): 476.0.0→575.0.1
- AWS CLI: 2.15.57→2.35.22; replace deprecated pip-installed v1 in halyard slim
- Packer: 1.10.1→1.15.4; helmfile: 1.1.7→1.7.0
- kustomize: add KUSTOMIZE5_VERSION=5.8.1 (binary: kustomize5).  Mark kustomize3 as "deprecated".
- Rosco: replace KUSTOMIZE TemplateRenderer with KUSTOMIZE5 in service/config/tests
- kayenta/Dockerfile: gradle 7.6.1→8.14.5 (matches project wrapper); alpine 3.20→3.24
- spin: golang 1.17→1.26.5; pin alpine:3.24
- echo/orca compile: openjdk11→openjdk17
- deck compile: fix broken mixed apk/apt-get Dockerfile


christosarvanitis:
Agreed on dropping the kustomize3 but this change requires Deck changes as well. How about this? 

Today:
- KUSTOMIZE -> v3
- KUSTOMIZE4 -> v4

After this PR:
- KUSTOMIZE -> v4
- KUSTOMIZE5 -> v5? 

Also we need to consider that for people using v3 still that they wont have to change all their pipeline across the board so for Deck i would say lets drop the KUSTOMIZE4 and add KUSTOMIZE5.

jasonmcintosh:
Kustomize I'm ... honestly not sure of.  I LEAN towards keeping it explicit.  OR say "KUSTOMIZE" is deprecated, we keep it, add kustomize5, and then remove kustomize this fall kinda thing. That's probably "safer", but kustomize itself now has known CVE library deps, which is why I'd love to remove it.

jasonmcintosh:
Restored kustomize.  Will add to upcomign release notes that ti'll be removed in 2027.0.0 - and users should migrate off of it to kustomize4/5.

christosarvanitis (approved):

PR #7791, opened 2026-07-15
fix(accounts): Fix a race condition on dynamic accounts loading

jasonmcintosh:
I had theories, but gotta give Anthropic credit - basically where I KINDA thought it might be, but it did a good analysis and found it quick.  Course I POINTED it at it, which i think helped, but... yeah interesting issue here.  I see this MOST likely when the parse tries to do some validation on accounts OR a few other potential issues on operations that'd trigger this.

NOTE:  I'm not sure this is the ONLY fault, just a POTENTIAL fault condition.


christosarvanitis (approved):
Looks solid. Had some failures where AWS account was not present but ECS account was added causing silent exception and issues

PR #7796, opened 2026-07-15
refactor(deck): remove Angular from titus

mattgogerly:
Removes Angular wiring from the Titus Deck package.


jasonmcintosh:
Interesting question:  should we remove titus from the project entirely?  Netflix has their fork, so not sure this would be used by anyone?

mattgogerly:
> Interesting question: should we remove titus from the project entirely? Netflix has their fork, so not sure this would be used by anyone?

I wouldn't object 😄

jasonmcintosh (dismissed):

jasonmcintosh:
https://github.com/spinnaker/spinnaker.io/pull/626/changes#diff-da149601addb46d3ccb7e84e9a3f7ef7b7375976b363b892e682180db7239bd1R53 ;) that said, fine to merge this and will just remove it when get to that point...

jasonmcintosh (approved):

PR #7798, opened 2026-07-15
refactor(deck): remove Angular from app canary

mattgogerly:
## Summary
Replaces app canary and ACA Task Angular UI with React registrations.


deck/packages/core/src/reactShims/react.injector.ts
```diff
@@ -58,6 +58,8 @@ export class CoreReactInject extends ReactInject {
   public get pageTitleService() { return this.$injector.get('pageTitleService') as PageTitleService; }
   public get providerServiceDelegate() { return this.$injector.get('providerServiceDelegate') as ProviderServiceDelegate; }
   public get securityGroupReader() { return this.$injector.get('securityGroupReader') as SecurityGroupReader; }
+  public get serverGroupCommandBuilder() { return this.$injector.get('serverGroupCommandBuilder') as any; }
```
jasonmcintosh, line 61: Is this really related to this?  seemed like a different area...

jasonmcintosh (approved):
No major concerns - we'll have to test like the other refactors, as it's hard to evaluate truly but nothing "immediately broken obviously" from what my limited UI skills can see ;)

deck/packages/core/src/reactShims/react.injector.ts
mattgogerly: Probably not, but this file is removed in a couple of PRs time anyway so..

PR #7799, opened 2026-07-15
feat(accounts): Add a UI to help with adding/removing accounts

jasonmcintosh:
ALSO fully enables the account APIs by default, removes beta tags.


christosarvanitis (approved):
Useful for account management

PR #7802, opened 2026-07-15
feat(kubernetes): Bump SDK and tests to remove OLD kubernetes support

jasonmcintosh:
Removes some long dead API response handling and manifest handling.  Reduces the matrix range of supported versions as well.


christosarvanitis (approved):
Good upgrade 🚀

PR #7805, opened 2026-07-16
fix(libdiffs): order versions of differing component counts correctly

semx:
### What

`DefaultComparableLooseVersion` (`orca/orca-core`) mis-orders versions that have different numbers of components, and violates the comparator's antisymmetry contract.

`parse()` pads each version into a fixed `Integer[4]`, so `"1.2"` becomes `[1, 2, null, null]`. `compareTo` compares components with `>` / `<`; unboxing a `null` pad throws an NPE that is swallowed by `catch (Exception e) { return -1; }`. As a result:

- `compare("1.2.3", "1.2")` returns `-1` (should be positive)
- `compare("1.2", "1.2.3")` also returns `-1`

So a genuine library **upgrade** `1.2 → 1.2.3` is reported as a **downgrade** in the JarDiffs report (`LibraryDiffTool.calculateLibraryDiffs` maps `1` → upgraded, `-1` → downgraded), and the comparator is not antisymmetric.

### Why it's wrong

The class documents itself as a *"Groovy implementation of python LooseVersion class"* — in Python `LooseVersion("1.2.3") > LooseVersion("1.2")`. The existing `LibraryDiffSpec` never caught this because every case it covers compares equal-length versions, which never hit the `null` pad.

### Fix

Treat a missing component as `0` (Python pads absent trailing components), so a shorter version orders below its longer prefix, instead of letting a swallowed NPE force `-1`.

### Testing

Added `DefaultComparableLooseVersionTest` covering equal-length comparisons (unchanged), differing component counts, and antisymmetry. The new cases fail on `main` and pass with the fix; the existing `LibraryDiffSpec` still passes.


jasonmcintosh (approved):
Looks good thank you!

jasonmcintosh:
@Mergifyio backport release-2026.2.x release-2026.1.x release-2026.0.x

PR #7806, opened 2026-07-16
fix(packer): pass through Boolean false and numeric 0 -var values

semx:
### What

`LocalJobFriendlyPackerCommandFactory.buildPackerCommand` (`rosco/rosco-core`) silently omits any Packer `-var` whose value is Boolean `false` or numeric `0`, because it gates emission on Groovy truthiness:

```groovy
parameterMap.each { key, value ->
  if (key && value) {   // false, 0 and "" are all Groovy-falsy → dropped
    ...
  }
}
```

### Why it matters

Provider handlers deliberately place raw Booleans into the map, guarded by `!= null`, to override the Packer template default. For example `AWSBakeHandler` sets `aws_associate_public_ip_address` from `awsBakeryDefaults.awsAssociatePublicIpAddress` when it is non-null, and the template default (`halconfig/packer/aws-ebs.json`) is `"true"`.

So an operator who sets `awsAssociatePublicIpAddress: false` to **disable** public IPs has the `-var` dropped; Packer falls back to the template default `true` and the instance gets a public IP — the **opposite** of the configured intent. Any `-var` whose intended value is `false` or `0` is affected.

### Fix

Drop only `null` and `""` (the original intent of skipping unset values), so `false` and `0` are passed through as `-var key=false` / `-var key=0`.

### Testing

Added `PackerBooleanFalseVarBugSpec` asserting that a Boolean `false` and a numeric `0` value each produce a `-var` entry. Both fail on `main` (the entries are absent) and pass with the fix; `LocalJobFriendlyPackerCommandFactorySpec`, `AWSBakeHandlerSpec` and `GCEBakeHandlerSpec` still pass.


rosco/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
```diff
@@ -36,7 +36,7 @@ class LocalJobFriendlyPackerCommandFactory implements PackerCommandFactory {
     packerCommand.addAll(roscoPackerConfigurationProperties.additionalParameters)
 
     parameterMap.each { key, value ->
-      if (key && value) {
+      if (key && value != null && value != "") {
```
jasonmcintosh: NIT:  Probably should instead have "!value.isEmpty()" as string comparison on this though it works... it's better to do actual comparison vs. object ref comparison (risk on some of these).

jasonmcintosh (dismissed):

rosco/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
semx: Good call on making it explicit — done in 7516aff.

One caveat on the exact form: `value` here is polymorphic (Boolean / Number / String — preserving `false` and `0` is the whole point of the fix), so a bare `!value.isEmpty()` would throw `MissingMethodException` on the Boolean/Integer cases, since `isEmpty()` only exists on String/collections. So I guarded it with an instanceof:

```groovy
if (key && value != null && !(value instanceof String && value.isEmpty())) {
```

That keeps the intent obvious (we only ever strip empty *strings*) and short-circuits before calling `isEmpty()` on a non-String. Also worth noting: in Groovy `!=`/`==` dispatch to `.equals()` rather than identity, so the previous `value != ""` was already a value comparison — but the instanceof form is clearer about what we mean. Thanks for the review!

jasonmcintosh (approved):

rosco/rosco-core/src/main/groovy/com/netflix/spinnaker/rosco/providers/util/LocalJobFriendlyPackerCommandFactory.groovy
jasonmcintosh: Yeah USUALLY I'm using ObjectUtils or similar vs String.  Meant that more of an example ;)

jasonmcintosh: And already know that does that on groovy - but one of those headaches I've seen so I try to guard for it... as I've been working to migrate from groovy to java on all of this code.

jasonmcintosh: Actually interesting tidbit even with java it still works, but there's some interesting edge cases I'd seen in JVM handling of internalized strings on this ;) Either way this looks good thank you!

jasonmcintosh:
@Mergifyio backport release-2026.2.x release-2026.1.x release-2026.0.x

Issue #6998, opened 2024-12-03
FR: Deck - Refresh button for console log modal viewing JobManifest console output


jgrumboe:
### Issue Summary:
I'd love to see a "Refresh" button next to the "Close" button of the "Console output" modal window.

### Cloud Provider(s):
K8s

### Feature Area:
UI

### Description:
We have some long running jobs I want to follow the console output.
At the moment I need to close and reopen the "Console output" modal to get fresh logs.
I'd like to click a "Refresh" button instead of reopening the modal.

### Additional Details:
I'd be willing to create a PR for this.

jgrumboe:
A quick solution could be to just duplicate the "Close" button and call the `onClick` function of packages/core/src/manifest/stage/JobManifestPodLogs.tsx.
The changes could look like this: https://github.com/jgrumboe/deck/commit/f1951e8c5b05e1df50e3d6f8d5e63b769fe19f94

jasonmcintosh:
hrmm I thought one of these did an automatic refresh... but... I may be thinking of something else...  or one of these was literally a stream poll....

jgrumboe:
So, at least the JobManifestPodLogs doesn't do an automatic refresh.
Would an automatic refresh be preferred by the team? I can try to prepare a PR for this.
If you remember where the stream poll is already implemented I can have a look at the implementation and mimick it.

jgrumboe:
@jasonmcintosh I searched a little bit and found that the instance console output modal has a refresh button: https://github.com/spinnaker/deck/blob/404046190620d8c1fae68715815adf69e4a5cc76/packages/core/src/instance/details/console/ConsoleOutputModal.tsx#L83 
I'll try to adapt the JobManifestPodLogs to be similar.

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

jgrumboe:
@spinnakerbot remove-label stale

spinnakerbot:
This issue hasn't been updated in 45 days, so we are tagging it as 'stale'. If you want to remove this label, comment:

> @spinnakerbot remove-label stale

iniinikoski:
@spinnakerbot remove-label stale

jgrumboe:
@spinnakerbot remove-label stale

spinnakerbot:
[...]

PR #7820, opened 2026-07-16
feat(logs): Add an autorefresh logs button thing to do logs

jasonmcintosh:
Support for https://github.com/spinnaker/spinnaker/issues/6998 :)


christosarvanitis (approved):
Looks good

PR #7823, opened 2026-07-20
fix(spinnaker-kustomize): Move default kustomize install to mysql due to compatibility issues with mariadb latest

jasonmcintosh:


christosarvanitis (approved):
Makes sense to have mySQL as the default

PR #7825, opened 2026-07-20
refactor(deck): remove React Angular bridges

mattgogerly:
- replace global React and modal injectors with directly initialized Angular services
- replace angular2react and react2angular bridge usage with explicit adapters and component registration
- preserve current upstream behavior while removing the direct bridge dependencies
- retain the existing React error boundary behavior and cover adapter bindings with regression tests


jasonmcintosh (approved):
Basically looks straightforward - moving from the library react2angular to the core AngularServices stuff :)

PR #7826, opened 2026-07-21
chore(cleanup): Migrate custom saml to spring

jasonmcintosh:
Basically rips out a ton of the custom saml implementation, has a migration doc on moving to spring saml properties instead.  Keeps a few custom things e.g. the translation for a saml user to a spinnaker user (which someday hope to remove)


christosarvanitis (approved):
We had a lot of issues with the custom implementation in the past so any change towards a native supported functionality is great. 
Note: We will need a Breaking change note in the release notes

PR #7827, opened 2026-07-21
chore(lambda): Migrate from v1 to v2 aws sdk

christosarvanitis:
## Summary

Migrates the clouddriver-lambda module from AWS SDK v1 to v2, restores the client-level tuning (retry, socket timeout, TCP keep-alive) that the migration initially dropped, and fixes a timeout conflict in the Lambda invoke path — without changing timeout behavior
for any other Lambda operation.

## Changes

1. AWS SDK v2 migration (clouddriver-lambda)

  • LambdaClientProvider now returns a v2 LambdaClient via AmazonClientProvider.getLambdaV2(...).
  • All atomic operations, the caching agent, and cache clients updated to v2 request/response/paginator types and exceptions.
  • LambdaFunction cache model decoupled from v1 SDK types.

2. Restored client-level tuning (clouddriver-aws)

  • Added AwsSdkV2ClientConfiguration (retry count, socket timeout, TCP keep-alive) — the v2 equivalent of the v1 ClientConfiguration.
  • AwsSdkV2ClientSupplier applies it when building/caching v2 clients.
  • AmazonClientProvider.getLambdaV2(...) overload accepts the config.

3. Fixed invoke timeout conflict (dedicated invoke client)

### Fix (scoped to invoke operation only):
**Problem:** the per-request apiCallTimeout used by InvokeLambdaAtomicOperation was silently capped by the client's Apache socketTimeout (default 50s). A synchronous RequestResponse invoke holds the socket idle for the whole function run, so the socket timeout fired before the intended per-request timeout.

  • LambdaClientProvider now exposes a separate getInvokeLambdaClient() whose socketTimeout is raised to the Lambda maximum (15 min) as a hard ceiling; the per-request apiCallTimeout is now the authoritative bound.
  • The shared getLambdaClient() used by all other operations is unchanged (socketTimeout = invokeTimeoutMs, no new total-call cap).
  • To make the two clients distinct, AwsSdkV2ClientSupplier's cache key (V2ClientKey) now includes AwsSdkV2ClientConfiguration by value equality: different tuning → distinct cached clients, equal tuning → same cached instance (caching preserved).


clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/jackson/AwsSdkV2JacksonConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+
+package com.netflix.spinnaker.clouddriver.aws.jackson;
+
+import com.fasterxml.jackson.databind.Module;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+public class AwsSdkV2JacksonConfiguration {
```
jasonmcintosh, line 24: To be clear... I see why this is happening... looks like tied to https://github.com/aws/aws-sdk-java-v2/issues/5251 and https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/migration-serialization-changes.html - but... i wonder if this is the right way vs. that 
```java
   HeadBucketResponse v2Deserialized = mapper.readValue(
            v2Serialized, HeadBucketResponse.serializableBuilderClass())
            .build();
```
type approach.  This is just an observation vs. "not approved" at this point.

jasonmcintosh: I'm not sure i'm COMFORTABLE with this not having issues is mostly the concern here.

clouddriver/clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/cache/model/LambdaFunction.java
```diff
@@ -16,25 +16,67 @@
...
+public class LambdaFunction implements Function {
   private String cloudProvider;
   private String account;
   private String region;
 
-  private Map<String, String> revisions;
-  private List<AliasConfiguration> aliasConfigurations;
-  private List<EventSourceMappingConfiguration> eventSourceMappings;
-  private FunctionCodeLocation code;
-  private Map<String, String> tags;
```
jasonmcintosh, line 38: hrmm, are we not losing field data here?  AKA tags as a result of the above change? Not to mention the other fields in caching...

clouddriver/clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/description/CreateLambdaFunctionConfigurationDescription.java
```diff
@@ -40,9 +38,9 @@ public class CreateLambdaFunctionConfigurationDescription extends AbstractLambda
   List<String> layers;
   Map<String, String> envVariables;
   Map<String, String> tags;
-  DeadLetterConfig deadLetterConfig;
```
jasonmcintosh, line 43: Any reason for this vs https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/scheduler/model/DeadLetterConfig.html ?

clouddriver/clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/ops/UpdateLambdaConfigurationAtomicOperation.java
```diff
@@ -45,44 +58,62 @@ public UpdateLambdaConfigurationAtomicOperation(
...
+            .memorySize(description.getMemorySize())
+            .role(description.getRole())
+            .timeout(description.getTimeout())
+            .layers(description.getLayers())
+            .kmsKeyArn(description.getKmskeyArn())
+            .runtime(description.getRuntime());
+
+    if (description.getDeadLetterConfig() != null) {
+      requestBuilder.deadLetterConfig(
+          software.amazon.awssdk.services.lambda.model.DeadLetterConfig.builder()
```
jasonmcintosh: Any reason we need full path here vs. import?

clouddriver/clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/deploy/description/CreateLambdaFunctionConfigurationDescription.java
christosarvanitis: This is the DeadLetterConfig in v2 https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/lambda/model/DeadLetterConfig.html

Implemented this local POJO in order to translate the v2 builder-only/immutable model and at the same time keep parity with the v1 output.  Alternative approach would be to deseriaze with an object mapper. will check

clouddriver/clouddriver-lambda/src/main/java/com/netflix/spinnaker/clouddriver/lambda/cache/model/LambdaFunction.java
christosarvanitis: Good call, added them back through attributes. The v2 aws sdk model is builder-only so getting the fields through the attributes

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/jackson/AwsSdkV2JacksonConfiguration.java
christosarvanitis: Both options work. This approach is reusable and global. I can explore the above and see what it takes

jasonmcintosh: Long as it works... it's one of those I wasn't sure how well it'd work for ALL Usages ;) but... I'm flexible.

jasonmcintosh (approved):

christosarvanitis:
Work towards https://github.com/spinnaker/spinnaker/issues/7699

PR #7834, opened 2026-07-23
fix(build): Set group name on all root projects

jcavanagh:
The Nexus plugin generates a staging repo description of the form: `$group:$name:$version`, in the context of the current `rootProject`

When running `findNexusStagingRepository`, it looks for repos by description. In the case of most of the root projects, there was no group name set - it was set only on subprojects.

This caused a problem with the Nexus plugin - its regex to find the description is the following: `descriptionRegex = "\\b" + Regex.escape(repositoryDescription) + "(\\s|$)"`

The `\b` boundary selector will not match the start of a string unless the first character is a word - in the case of the missing group, the first character was a colon. This caused the generated repository to be unfindable by the Nexus plugin.

This wasn't an issue when running publish/close/release in the same Gradle context, since the Nexus plugin would read its own internal state to find the staging repository ID, and didn't use the description matching.

I think having the publish -> find/close/release as separate Gradle tasks is still the right approach - the Nexus plugin doesn't fit monorepo setups very well (see all the task shimming we have to do in there), and I don't have a good answer for the flaky empty repository errors we were originally seeing.


jasonmcintosh (approved):
OK... :) Nice find/fix!  Seeing similar headaches with some other plugins not handling composite builds or other quirks based upon how we've got things setup, fixing too as I find them.  Thank you for this!!

jasonmcintosh:
@mergifyio backport release-2026.2.x release-2026.1.x release-2026.0.x

PR #7850, opened 2026-07-26
fix(orca): honor the page parameter when paging pipeline executions by config id

### LINKED ISSUE #6786: [kayenta] Paging seems broken
### Issue Summary:
Kayenta's canary/execution-API seems to ignore the page-parameter.

### Cloud Provider(s):
Not relevant.

### Environment:
Kayenta only, running in Kubernetes or locally in a Docker container

### Feature Area:
Kayenta

### Description:
We are using Kayenta in standalone-mode and we noticed that paging of executions seems to be broken:
e.g. http://localhost:8090/canary/executions?application=myApp1&limit=2&page=X returns the same 2 results no matter what number I put for X

Currently we can work around it by adjusting the limit parameter to the right size for a lookback, but there may be scenarios when one page would become too big where we would need paging.

### Steps to Reproduce:

1. Start Kayenta
2. Do 2 analysises
3. Call http://localhost:8090/canary/executions?application=myApp1&limit=1&page=1
4. Call http://localhost:8090/canary/executions?application=myApp1&limit=1&page=2
5. Results are equal

### Additional Details:
This can be reproduced using S3 repository and local in-memory-store.

jasonmcintosh:
## Summary
- Fixes [spinnaker/spinnaker#6786](https://github.com/spinnaker/spinnaker/issues/6786): kayenta's `GET /canary/executions?page=N` always returned the same results, regardless of `page`.
- The bug wasn't in kayenta — `CanaryController` correctly builds and passes an `ExecutionRepository.ExecutionCriteria` with `page` set, but all three `ExecutionRepository` backends mishandled it in `retrievePipelinesForPipelineConfigId`:
  - **SQL**: only applied `.limit(pageSize)`, never an `.offset(...)` for `page` — every page returned the same top-N rows. The sibling method `retrieveOrchestrationsForApplication` in the same file already did this correctly; this brings `retrievePipelinesForPipelineConfigId` in line with that existing pattern.
  - **Redis**: same issue — always took indices `0..pageSize-1`, ignoring `page` entirely.
  - **InMemory**: did use `page`, but was off-by-one (`chunked(pageSize)[page]` against a 1-based page and 0-based chunk index), and threw `IndexOutOfBoundsException` once `page` exceeded the available data instead of returning empty.

## Test plan
- [x] Added a shared test to `PipelineExecutionRepositoryTck.groovy` (exercised by both `SqlPipelineExecutionRepositorySpec` — MySQL & Postgres via testcontainers — and `JedisPipelineExecutionRepositorySpec` via embedded Redis) asserting distinct, newest-first pages and an empty page past the end of the data.
- [x] Added `InMemoryExecutionRepositoryTest.java` covering the same paging contract for the in-memory backend (previously untested).
- [x] `./gradlew :orca-core:test :orca-sql:test :orca-redis:test` all green.


jasonmcintosh:
I THINK this'll fix a gate API issue that I'd known about, same exact issue there as well.

christosarvanitis (approved):

PR #7852, opened 2026-07-26
feat(aws): Warm pool support

### LINKED ISSUE #6399: Support Warm Pools for AWS AutoScalingGroups
### Issue Summary:

Support enabling and setting the desired size/state of the Warm Pool on an AWS AutoScalingGroup.

### Description:

Warm Pools are a new feature that have been added to ASGs to improve responsiveness of auto scaling. It allows the ASG to keep a specified number of instances pre-initialized in a stopped state (or running) that aren't actively in use, these are then able to start up significantly faster. It would be ideal if this could be configured in the settings for a server group.

Available settings for the warm pool are:

- Min Size: Number of instances to keep in the warm pool
- Instance State: What state the instances will be kept in i.e Stopped or Running
- Max Group Prepared Capacity: The maximum capacity of the ASG including the warm pool

AWS feature documentation: https://aws.amazon.com/blogs/compute/scaling-your-applications-faster-with-ec2-auto-scaling-warm-pools/

jasonmcintosh:
Fixes #6399 

NOTE:  HEAVILY AI generated, but also mostly copy/paste of existing stuff.


orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/model/ServerGroup.java
```diff
@@ -85,6 +85,7 @@ public static class Asg {
     public Object minSize;
     public Object maxSize;
     public List<Process> suspendedProcesses;
+    public Map<String, Object> warmPoolConfiguration;
```
jasonmcintosh: This is one... not sure we NEED this year in the "generic" ServerGroup section.

jasonmcintosh: Just removed this... didn't like how this worked.

jasonmcintosh:
Fixes https://github.com/spinnaker/spinnaker/issues/6399

christosarvanitis (changes requested):
Would it be possible to convert everything to Java as this is a new functionality added?

jasonmcintosh:
It's new, but a good chunk of the existing code for this is already groovy.  I'd like to keep "like-for-like" as I'm already hitting situations where "oh crud that one file is in groovy other in java" and change whole packages from groovy to java instead as a later fix.  That said ... this package half is already in java, and half in groovy so yeah :) I was just looking at the groovy base... same sorta "ugh" moment of half these were in groovy and I missed the java half.  Good call out :)

christosarvanitis (approved):
Awesome! looks good and follows the established patterns for aws operations/stages. Thanks a lot

PR #7863, opened 2026-07-28
feat(authz): Remove Fiat in favor of token-based service-to-service authorization

stevenlandrydd:


gate/gate-core/src/main/java/com/netflix/spinnaker/gate/security/token/GateAuthzProperties.java
```diff
@@ -0,0 +1,91 @@
...
+   * Role names that grant Spinnaker account-manager status. When a caller holds any of these roles
+   * the minted token carries the {@code account_manager} claim.
+   */
+  private List<String> accountManagerRoles = new ArrayList<>();
+
+  public Duration getRoleCacheTtl() {
+    return roleCacheTtl;
+  }
+
+  public void setRoleCacheTtl(Duration roleCacheTtl) {
```
jasonmcintosh, line 63: NIT:  This looks ideal for a lombok `@Getter` and similar annotations.

kork/kork-security/src/main/java/com/netflix/spinnaker/security/AuthenticatedRequest.java
```diff
@@ -277,6 +277,10 @@ private static <V> Callable<V> wrapCallableForPrincipal(
     String requestId = getSpinnakerRequestId().orElse(null);
     String spinnakerAccounts = getSpinnakerAccounts(principal).orElse(null);
     String spinnakerApp = getSpinnakerApplication().orElse(null);
+    // The signed identity token must cross the thread boundary too, otherwise work dispatched onto
+    // another thread (e.g. Echo's async event processing) loses it and downstream services that
+    // authorize against the verified token (authz.enabled=true) treat the call as anonymous.
+    String identityToken = get(Header.IDENTITY_TOKEN).orElse(null);
```
jasonmcintosh, line 283: I'll admit, I KINDA was hoping to get rid of this class and avoid the usage of the MDC at all... :( still probably needed at this time given the above, just hoped we could simplify this more.

kork/kork-security/src/main/java/com/netflix/spinnaker/security/token/SpinnakerTokenSettings.java
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * Configuration for minting and verifying Spinnaker identity tokens (claims and validation
+ * tolerances), bound from the {@code authz.token} property prefix.
+ *
+ * <p>The authorization master switch is <em>not</em> here — it lives at the top-level {@code
+ * authz.enabled} ({@link AuthorizationProperties}). This class only carries token-specific claim
+ * and validation settings.
+ */
+@ConfigurationProperties("authz.token")
+public class SpinnakerTokenSettings {
```
jasonmcintosh, line 33: PLEASE for these - lombok!  No manual getters/setters!

kork/kork-security/src/main/java/com/netflix/spinnaker/security/token/SpinnakerTokenClaims.java
```diff
@@ -0,0 +1,234 @@
...
+/**
+ * Immutable representation of the identity claims carried by a signed Spinnaker identity token.
+ *
+ * <p>The token carries the caller's subject ({@code sub}), their resolved {@code roles}, and the
+ * Spinnaker-specific {@code admin}/{@code accountManager} flags, along with the standard JWT
+ * registered claims ({@code iss}, {@code aud}, {@code iat}, {@code exp}). Services verify the
+ * signature and the registered claims, then populate Spring authorities from {@link #getRoles()}
+ * plus the admin/account-manager flags.
+ */
+public final class SpinnakerTokenClaims {
```
jasonmcintosh, line 38: Another place - lombok can reduce this code a lot!

jcavanagh:
I think this change is worthy of an RFC and/or some more detailed technical design.  

Removal of existing critical functionality should be the last step of the process.  Internals are tightly coupled to Fiat, but those can be decoupled and new methodologies built in parallel, until they can be proven to work reliably and fit the required use cases.  The decoupling itself provides value, even if a specific approach might fail.  

Replacing Fiat creates opportunity to use existing authz engines like SpiceDB/OPA/etc, and there should be analysis as to why those options should or should not be considered.  Evaluating alternatives for service-to-service should also be done, e.g. istio mTLS instead of self-issued tokens.  There is significant prior art in both of these domains.  

I also think Front50 is a poor choice for a replacement, since it's on the brink of being a library as well.  Its caching functionality creates staleness and replica disagreement potential, and frankly puts more load on the database than normal use with stock DB memory caching or a fronting LRU cache would.  It does not strongly own any data, just validators and DAOs, and Fiat retrieves all its objects already for computing permissions today.

nandak522:
+1 to @jcavanagh comment @stevenlandrydd. A seriously-long single-commit rewrite of the authz model is not reviewable to the standard this code needs. Recommended: RFC covering the alternatives analysis (mesh mTLS, OPA/SpiceDB/OpenFGA) and the Front50-as-authority choice, then split the whole code into a 6-step series with Fiat removal last 👇:
1. PEP/PDP seam, Fiat behind a PolicyDecisionPoint impl
2. signed tokens alongside legacy headers
3. s2s
4. new PDP
5. deprecate
6. remove Fiat. 

Each can be landed and rolledback independently.

nandak522:
Reviewed this PR through Claude and it gave some more feedback:

Right 👍 
---

- PEP/PDP seam. PolicyDecisionPointPermissionEvaluator registered as spinnakerPermissionEvaluator means existing `@PreAuthorize("hasPermission(...)")` / `@PostFilter` call sites stay untouched, and OPA/SpiceDB/OpenFGA can slot in behind PolicyDecisionPoint later. Correct seam, correct place.
- Signed identity instead of unsigned X-SPINNAKER-USER / -ACCOUNTS. Closes a real long-standing hole. RS256-pinned key selector, kid-based rotation, no alg confusion.
- Explicit s2s auth with codified @AllowServiceCallers beats implicit network trust. K8sServiceAccountTokenResolver is genuinely careful — audience required or resolver stays disabled, namespace matched, offline-verification residual risk documented, JWKS fetch bounded by timeouts + size cap.
- AUTHZ_MIGRATION.md is unusually good for a PR this size.

Need to revisit 👀 
---
- Hand-rolled JWT/JWKS/rotation stack (~100 new security-critical classes) with no written analysis of why not mesh mTLS + an existing authz engine. That is a permanent maintenance burden the project takes on undiscussed.
- Front50 as token authority and ACL owner is the weakest pick. It is now both the minting authority and on the per-check hot path. Compromise Front50 → mint any subject including admin=true. Concentrates blast radius in the service with known staleness/replica-disagreement issues.
- fiat-api deleted with no deprecation cycle — breaks every third-party plugin compiling against Fiat interfaces.

---

Security findings, for you to verify and come back
---
1. Identity token leaks to third-party hosts. Header.IDENTITY_TOKEN("X-SPINNAKER-IDENTITY-TOKEN") is set into the MDC by IdentityTokenPropagationFilter. AuthenticatedRequest.getAuthenticationHeaders() (AuthenticatedRequest.java:309-334) copies every MDC key matching the X-SPINNAKER prefix, and SpinnakerRequestHeaderInterceptor:79 is a single global bean applied to every kork OkHttp client, with propagateSpinnakerHeaders = true by default (OkHttpClientConfigurationProperties.java:46) and no per-destination allowlist. So the signed bearer token rides along on Igor→Jenkins/GitHub/GitLab, Clouddriver→Docker registries and artifact stores, webhook targets. The PR's own s2s/README.md explicitly refuses a global OkHttp customizer for the s2s token for exactly this reason — the identity token doesn't get the same treatment.

2. Gate accepts the identity token as authentication at the internet edge. GateIdentityTokenInboundFilter authenticates any request carrying a valid X-SPINNAKER-IDENTITY-TOKEN. No jti/replay guard, no cnf/DPoP, no mTLS or source binding. So a token leaked via (1) is a working impersonation credential from the internet for validity + clockSkew = 5m + 60s by default. Fiat's model exposed no comparable bearer credential. Fix: at the edge, only accept the token from an s2s-authenticated peer, or bind it.

3. Front50 `/auth/runAsToken` privilege escalation. `authorizeInitialMint` accepts `hasTrustedServiceCaller() || hasVerifiedIdentityToken()` — the OR holds even when s2s is enabled, so any valid identity token satisfies the check. The only other gate is isConfiguredRunAsUser(pipelineId, sa); there is no check that the token's subject has any permission on the pipeline or its application. Any authenticated user who can reach Front50 and knows a pipelineId mints that pipeline's run-as SA token and inherits its roles. The javadoc's claim — "no proof can mint an arbitrary (e.g. more-privileged) subject" — is true about arbitrary subjects, not about subjects the caller is unauthorized for.

4. Migration step 2 opens an unauthenticated mint oracle. AUTHZ_MIGRATION.md step 2 instructs operators to configure signing keys while authz.enabled=false, before s2s (step 3). In that window authorizeInitialMint returns early with no checks, and issueExecutionToken's ORCA guard is gated on authz.isEnabled() — so Front50 will mint a token for an arbitrary subject with admin=true for anyone who can reach it. Rollouts are per-service, so any service already flipped to authz.enabled=true honours those tokens. Gate both endpoints on caller proof independently of authz.enabled.

---
Correctness / performance

5. The headline claim is inverted for Clouddriver. Commit message: "removing the availability and latency coupling that Fiat introduced on every authorized request." But `ApplicationsController.list` keeps `@PostFilter hasPermission(filterObject.name, 'APPLICATION', 'READ')`, and `RemoteApplicationAclResolver` memoizes only for the current request — explicitly "not a cache." So one synchronous `Front50 GET /permissions/applications/{app}?effective=true` per distinct application per request. N applications = N serial HTTP calls. Fiat served this from a materialized, Redis-cached view. Coupling moved from Fiat to Front50 with strictly less caching, and Front50's DB is now on the authz hot path.

6. Front50 outage can fail open. In `RemoteApplicationAclResolver.lookupOwnAcl` and `Front50ApplicationAclResolver`, a 404 ("no record" → unrestricted-or-defaults) and a transient error both return null, and the evaluator then applies allowAccessToUnknownApplications. With that flag true (the legacy-compat setting), a Front50 outage becomes allow-all on every application. Both cases are logged at `debug` only, so it's invisible. Separate the two; fail closed on error.

7. ApplicationSearchProvider:92 flips a default from deny to allow. `permissionEvaluator != null && hasPermission(...)` became `permissionEvaluator == null || hasPermission(...)`. A missing evaluator bean used to exclude the app from search results; now it includes it.

8. SpringAclPolicyDecisionPoint.decide allocates per decision. New EmbeddedPermissionsAclService + AclPermissionEvaluator on every call, plus an ObjectIdentityImpl that is built and never used (AclPermissionEvaluator resolves by name). Dead line on a hot path.

9. ServiceCallerEnforcementAspect — `@within(allow) && execution(* *(..))` fires on every method of an annotated class, not only request handlers.

10. Leftover release tooling. fiat/ is fully deleted but .github/workflows/fiat.yml and .github/actions/spinnaker-release/src/bom/services/fiat.ts still reference it. Release/BOM pipeline will break.

11. SpinnakerPermissionFactory.toPermission is unused. Lombok nits from @jasonmcintosh are valid and consistent with repo style.

qunayu:
Hi! Jason pointed me to this PR as I've also recently removed fiat from our fork of Spinnaker. We've faced similar performance issues over the last couple of years and finally decided to replace it with an internal policy enforcer.

Couple of thoughts after briefly going over the RFC + parts of the code:
- I doubt we're unique in this but our users typically have thousands of roles attached, with shared prefixes only being a small part of the actual role name. I can see we've tried to address that with DEFLATE compression, but I suspect we'll still run into this problem with larger orgs. 
- Echoing some concerns above - this puts front50 on the critical path for almost all key operations. Lots of performance implications (both on the application and db layers) for spinnakers with high levels of concurrent executions / operations.
- WRT more "safely" rolling out the change - we leveraged dynamic config a lot during our own migration. It looks like this change can also lend itself quite well to that approach (e.g. both initially and also gate the hybrid mode where fiat is used as a fallback).

PR #7864, opened 2026-07-29
fix: semgrep-run-as-non-root

j-sandy:
### Pull Request — Semgrep Rule Fix

- Rule ID: run-as-non-root
- Rule Message: When running containers in Kubernetes, it's important to ensure that they  are properly secured to prevent privilege escalation attacks.  One potential vulnerability is when a container is allowed to run  applications as the root user, which could allow an attacker to gain  access to sensitive resources. To mitigate this risk, it's recommended to  add a `securityContext` to the container, with the parameter `runAsNonRoot`  set to `true`. This will ensure that the container runs as a non-root user,  limiting the damage that could be caused by any potential attacks. By  adding a `securityContext` to the container in your Kubernetes pod, you can  help to ensure that your containerized applications are more secure and  less vulnerable to privilege escalation attacks.
- File Path: https://github.com/spinnaker/spinnaker/blob/main/clouddriver/clouddriver-kubernetes/src/integration/resources/manifests/cronJob.yml#L10
- Line: 10


jasonmcintosh:
yeah only used in integration tests, but good to have it :) no impact to running spinnaker though.

jasonmcintosh (approved):

PR #7866, opened 2026-07-29
chore(metrics): Remove spectator for micrometer

jasonmcintosh:
UNFORTUNATELY this isn't exactly a "small" pr, given this is a core library... but it should be VERY straight forward. 

## Why?
Spectator is a netflix specific implementation not used by the wider community.  It uses custom frameworks, and though it itself registers and acts AS a spectator registry, this removes the implementation in favor of the abstraction enabling easier configuration of alternative metric providers (aka prometheus).  Further, this is yet another custom dependency that we can remove in favor of standard off-the-shelf component libraries.

Further... this ALSO works much better with OTHER standard observability tooling that can autodetect micrometer registries and engage with them.  Example:  
* https://opentelemetry.io/docs/zero-code/java/agent/supported-libraries/ - note that micrometer itself is supported, but not spectator
* https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/supported-libraries.md#disabled-instrumentations when using the instrumentation here

Though this isn't MANDATORY to use OTEL, since spectator itself IS a Micrometer registry at this point, this should reduce the library usage in the project and required upgrade headaches.


keel/keel-core/src/main/kotlin/com/netflix/spinnaker/keel/telemetry/TelemetryListener.kt
```diff
@@ -29,7 +28,7 @@ import java.util.concurrent.atomic.AtomicReference
 
 @Component
 class TelemetryListener(
-  private val spectator: Registry,
+  private val spectator: MeterRegistry,
```
christosarvanitis, line 31: rename to registry to avoid spectator references?

clouddriver/clouddriver-dcos/src/main/groovy/com/netflix/spinnaker/clouddriver/dcos/security/DcosAccountCredentials.groovy
```diff
@@ -135,7 +135,7 @@ class DcosAccountCredentials extends AbstractAccountCredentials<DcosCredentialMa
       return this
     }
 
-    Builder spectatorRegistry(Registry spectatorRegistry) {
+    Builder spectatorRegistry(MeterRegistry spectatorRegistry) {
```
christosarvanitis, line 138: Rename builder/registry to not include spectator?

christosarvanitis (commented):
Its a big one, didnt see anything that stands out in the migration. Only thing is that DCOS, keel and probably somewhere else it is still naming some objects with spectator and for the sake of not confusing people maybe its better to have them as generic names. registry or meterRegistry etc

PR #7877, opened 2026-08-04
Fix(CVE): bump versions

shlomodaari:
## Summary

Bumps several transitive dependencies in `spinnaker-dependencies` to clear known High CVE findings in published OSS jars, and fixes a BOM import-order issue that caused Netty / Log4j pins to be ignored by Maven consumers of the published platform POM.

### Dependency bumps
| Component | From | To |
|---|---|---|
| `org.apache.commons:commons-compress` | 1.21 (local hardcodes / unconstrained) | **1.27.1** |
| `ch.qos.logback:logback-core` / `logback-classic` | Spring Boot managed | **1.5.37** |
| `org.apache.httpcomponents.core5:httpcore5` / `httpcore5-h2` | Spring Boot managed | **5.4.3** (via `httpcore5-parent` platform) |
| `org.apache.tomcat.embed:tomcat-embed-*` | Spring Boot managed | **10.1.56** |

Also removes hardcoded `commons-compress:1.21` versions in `rosco-manifests` / `rosco-web` so they pick up the central pin.

### Netty / Log4j BOM ordering
`netty-bom` (`4.1.136.Final`) and `log4j-bom` were already declared in `spinnaker-dependencies`, but **after** `spring-boot-dependencies`. That works for build-from-source Gradle conflict resolution (highest version wins), but when `spinnaker-dependencies` is published as a Maven POM and consumed via `platform()` / `enforcedPlatform()`, Maven's `<dependencyManagement>` import rule is **first import wins**. Spring's nested Netty/Log4j BOMs therefore silently won, leaving consumers on older vulnerable versions.

This PR moves `log4j-bom`, `netty-bom`, and `httpcore5-parent` **above** `spring-boot-dependencies` so the published platform POM honors the intended pins for both Gradle and Maven consumers.

### Why some pins stay as constraints
- **logback** / **tomcat-embed**: no official consumer-facing BOM is published, so explicit version constraints are the correct approach.
- **httpcore5**: uses the official `httpcore5-parent` platform instead of a bare constraint.

## Test plan
- [x] Resolve `clouddriver` / `rosco` runtimeClasspath and confirm bumped versions
- [x] Confirm Netty resolves to `4.1.136.Final` and Log4j to the pinned BOM version when consuming the published `spinnaker-dependencies` POM via `enforcedPlatform`
- [ ] CI green on this PR (note: known flaky `front50` `multi-threaded cache refresh` test is unrelated)

### DIFF
--- a/kork/spinnaker-dependencies/spinnaker-dependencies.gradle
+++ b/kork/spinnaker-dependencies/spinnaker-dependencies.gradle
@@ -29,6 +29,10 @@ ext {
     // https://github.com/springdoc/springdoc-openapi/blob/v2.8.17/springdoc-openapi-starter-common/src/main/java/org/springdoc/core/providers/ObjectMapperProvider.java#L85
     // referencing a field that didn't come in... until 2.2.30
     swagger          : "2.2.49",
+    httpcore5        : "5.4.3",
+    logback          : "1.5.37",
+    tomcat           : "10.1.56",
+    commonsCompress  : "1.27.1",
   ]
 }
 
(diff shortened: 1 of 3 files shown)


clouddriver/gradle.properties
```diff
@@ -1,6 +1,6 @@
 org.gradle.parallel=true
 artifactTypes=bitbucket,custom,docker,embedded,front50,gcs,github,gitlab,gitrepo,helm,http,ivy,jenkins,kubernetes,maven,oracle,s3
-commonsCompressVersion=1.21
+commonsCompressVersion=1.27.1
```
jasonmcintosh: Not needed here with change below

jasonmcintosh (approved):

jasonmcintosh:

PR #7883, opened 2026-08-12
fix(docs): Publish API docs as part of release based upon the openapi generated specs

jasonmcintosh:
Fixes also a dependabot upgrade issue on uuid.  Uses https://redocly.com/docs/redoc/quickstart to provide the UI around the openapi spec.  Open to other paths (that's in the incoming spinnaker.io PR)


jasonmcintosh:
image
This adds/changes the visualization to look like this.  I'd imagine there ARE Other solutions/options on this but... kept it simple at the moment.

christosarvanitis (approved):
Useful! Thanks

PR #7885, opened 2026-08-13
feat(gitlab-ci): Add pipeline trigger, cancel, and StoppableBuildService interface

eeigjon:
## Summary

This PR adds native pipeline trigger and cancel support for GitLab CI in Igor, and introduces a `StoppableBuildService` interface to decouple build cancellation from Jenkins-specific code.

Currently, cancelling a running CI build in Spinnaker only works for Jenkins (`BuildController.stopJob()` checks `instanceof JenkinsService`). This change makes cancellation extensible to any CI provider by introducing a common interface.

### New: `StoppableBuildService` interface (`igor-core`)

```java
public interface StoppableBuildService {
    void stopRunningBuild(String jobName, long buildNumber);
    default void stopQueuedBuild(String queuedId) { /* throws UnsupportedOperationException */ }
}
```

Any CI service that supports stopping/cancelling builds implements this interface. The `stopQueuedBuild` method has a default implementation since not all providers support queue cancellation.

### GitLab CI: Trigger and Cancel (`igor-web`)

- **`GitlabCiClient.java`** — Added `triggerPipeline()` and `cancelPipeline()` Retrofit endpoints
- **`GitlabCiProperties.java`** — Added `triggerToken` configuration field to `GitlabCiHost`
- **`GitlabCiService.java`** — Implemented `triggerBuildWithParameters()` (was `UnsupportedOperationException`) and added `cancelPipeline()`. Implements `StoppableBuildService`. Overrides `queuedBuild()` to return pipeline ID directly (GitLab CI has no queue concept).
- **`GitlabCiResultConverter.java`** — Added handling for all GitLab pipeline statuses (`created`, `waitingForResource`, `preparing`, `scheduled`, `manual`). Previously these threw `IllegalArgumentException` when the monitor polled a newly-triggered pipeline.

### BuildController: Generic stop logic

- **`BuildController.groovy`** — `stopJob()` now uses `instanceof StoppableBuildService` instead of `instanceof JenkinsService`. Removed Groovy `metaClass.respondsTo` reflection in favor of direct interface method calls.

### JenkinsService: Backward compatibility

- **`JenkinsService.java`** — Implements `StoppableBuildService` with bridge methods to existing `stopRunningBuild(String, Long)` and `stopQueuedBuild(String)` methods. No behavior change for Jenkins users.

## Configuration

To enable GitLab CI pipeline triggering, add `triggerToken` to the master config:

```yaml
gitlab-ci:
  enabled: true
  masters:
    - name: my-gitlab
      address: https://gitlab.example.com
      privateToken: <api-token>
      triggerToken: <pipeline-trigger-token>
      permissions:
        READ:
        - my-team
        WRITE:
        - my-team
```

The `triggerToken` is a [GitLab Pipeline Trigger Token](https://docs.gitlab.com/ee/ci/triggers/) created in the GitLab project settings. It is stored server-side in Igor's configuration — never exposed in pipeline JSON.

The `triggerToken` is optional — if not configured, pipeline monitoring and status polling still work (using `privateToken`), but triggering will fail with a clear error message.

Fiat `permissions` must be configured on the master to control access (same as Jenkins masters).

## Testing

- **`GitlabCiServiceSpec`** — 5 new test cases: trigger happy path, default ref, missing token validation, cancel, stopRunningBuild delegation
- **`BuildControllerStopSpec`** — 7 new test cases: stopRunningBuild routing, stopQueuedBuild routing, 404 handling, error propagation, UnsupportedOperationException handling, non-stoppable service rejection, unknown master handling
- **`JenkinsServiceSpec`** — 3 new test cases: bridge method delegation for stopRunningBuild, stopQueuedBuild, and interface implementation check
- All existing tests pass (no regressions)
- Manually tested end-to-end on a live Spinnaker deployment:
  - Trigger: Spinnaker UI → Orca → Igor → GitLab CI pipeline triggered
  - Monitor: Stage polls until GitLab pipeline completes, marks SUCCEEDED
  - Cancel: Spinnaker stage cancels correctly from UI (stops monitoring)
    - Note: Cancel does not yet propagate to GitLab CI (requires Orca-side CancelStageHandler wiring — documented as follow-up)
  - Status handling: Pipeline in "created" state handled gracefully (no crash)

## Follow-up Work

- **Orca**: Create `GitlabCiStage.java` for a dedicated stage type (currently uses the Jenkins stage type via `CIStage`)
- **Orca**: Wire cancel propagation via `CancelStageHandler` to call Igor's stop route for CI stages
- **Deck**: Add a GitLab CI stage selector with project/ref/variables form fields
- **Igor**: Implement `CiBuildService` for the CI view in the UI and Managed Delivery artifact metadata
- **Task renaming**: Rename `StartJenkinsJobTask` → `StartBuildJobTask`, `StopJenkinsJobTask` → `StopBuildJobTask` (requires backward compatibility validation with stored pipeline executions)
- **Fiat permissions**: Document configuration for GitLab CI masters

## Related Discussion

Community discussion on task naming and `StoppableBuildService` approach: https://spinnakerteam.slack.com/archives/C091CCWRJ/p1782487022482969

### DIFF
--- a/igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
+++ b/igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2026 LM Ericsson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.netflix.spinnaker.igor.service;
+
+/**
+ * Interface for CI build services that support stopping/cancelling running or queued builds.
+ *
+ * <p>Implement this interface alongside {@link BuildOperations} to enable Spinnaker's pipeline
+ * cancel functionality for a given CI provider.
+ */
+public interface StoppableBuildService {
+
+  /**
+   * Stop a running build.
+   *
+   * @param jobName The name or identifier of the job (e.g., Jenkins job path, GitLab project ID)
+   * @param buildNumber The build number to stop
+   */
+  void stopRunningBuild(String jobName, long buildNumber);
+
+  /**
+   * Stop a queued build that has not yet started execution.
+   *
+   * <p>Not all CI providers support this operation. The default implementation throws {@link
+   * UnsupportedOperationException}.
+   *
+   * @param queuedId The identifier of the queued item
+   */
+  default void stopQueuedBuild(String queuedId) {
+    throw new UnsupportedOperationException("Stopping queued builds is not supported");
+  }
+}
--- a/igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
+++ b/igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.igor.service.ArtifactDecorator
 import com.netflix.spinnaker.igor.service.BuildOperations
 import com.netflix.spinnaker.igor.service.BuildProperties
 import com.netflix.spinnaker.igor.service.BuildServices
+import com.netflix.spinnaker.igor.service.StoppableBuildService
 import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException
 import com.netflix.spinnaker.kork.web.exceptions.InvalidRequestException
@@ -189,28 +190,28 @@ class BuildController {
 
   void stopJob(String master, long buildNumber, String jobName, String queuedBuild) {
     def buildService = getBuildService(master)
-    if (buildService instanceof JenkinsService) {
+    if (buildService instanceof StoppableBuildService) {
+      StoppableBuildService stoppable = (StoppableBuildService) buildService
       // Jobs that haven't been started yet won't have a buildNumber
       // (They're still in the queue). We use 0 to denote that case
-      if (buildNumber != 0 &&
-        buildService.metaClass.respondsTo(buildService, 'stopRunningBuild')) {
-        buildService.stopRunningBuild(jobName, buildNumber)
+      if (buildNumber != 0) {
+        stoppable.stopRunningBuild(jobName, buildNumber)
       } else {
-        // The jenkins api for removing a job from the queue (http://<Jenkins_URL>/queue/cancelItem?id=<queuedBuild>)
-        // always returns a 404. This try catch block insures that the exception is eaten instead
-        // of being handled by the handleOtherException handler and returning a 500 to orca
+        // Some CI providers (e.g., Jenkins) return a 404 when cancelling a queued build.
+        // We catch that to avoid returning a 500 to Orca.
         try {
-          if (buildService.metaClass.respondsTo(buildService, 'stopQueuedBuild')) {
-            buildService.stopQueuedBuild(queuedBuild)
-          }
+          stoppable.stopQueuedBuild(queuedBuild)
         } catch (SpinnakerHttpException e) {
           if (e.getResponseCode() != NOT_FOUND.value()) {
             throw e
           }
+        } catch (UnsupportedOperationException e) {
+          log.debug("Queued build stop not supported for master '{}', ignoring", master)
         }
       }
+    } else {
+      throw new InvalidRequestException("Build service '${master}' does not support stopping builds")
     }
-
   }
 
   @RequestMapping(value = "/masters/{name}/jobs/**/update/{buildNumber}", method = RequestMethod.PATCH)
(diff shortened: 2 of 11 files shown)


igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright 2025 Netflix, Inc.
```
jasonmcintosh: Is this netflix?  OR some other org?

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public interface StoppableBuildService {
+
+  /**
+   * Stop a running build.
+   *
+   * @param jobName The name or identifier of the job (e.g., Jenkins job path, GitLab project ID)
+   * @param buildNumber The build number to stop
+   */
+  void stopRunningBuild(String jobName, long buildNumber);
```
jasonmcintosh, line 32: ONE thing and not checked the system today - buildNumber is a long here.  I"d wonder if it shouldn't be a string - e.g. if someone uses a UUID for a build number.

jasonmcintosh: This would also match the queueId & then implementations could parse it if needed.  Again NOT looked at all the platforms out there - but wouldn't surprise me if ONE of htem uses a string UUID for build numbers for some reason.

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -189,28 +190,28 @@ class BuildController {
 
   void stopJob(String master, long buildNumber, String jobName, String queuedBuild) {
     def buildService = getBuildService(master)
-    if (buildService instanceof JenkinsService) {
+    if (buildService instanceof StoppableBuildService) {
+      StoppableBuildService stoppable = (StoppableBuildService) buildService
```
jasonmcintosh, line 194: NIT:  Java 17 syntax lets you do
```
    if (buildService instanceof StoppableBuildService stoppable) {
```
Removing the need for the separate cast.

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
```diff
@@ -83,6 +84,14 @@ public void setPrivateToken(String privateToken) {
       this.privateToken = privateToken;
     }
 
+    public String getTriggerToken() {
```
jasonmcintosh, line 87: NIT:  NOMINALLY could use Lombok getters/setters.  Would simplify this class a moderate amount, but not really needed for this initial PR, just be a "nice to have"

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
jasonmcintosh: I'll note:  This is again a "nice-to-have" long term, I think for now long is OK.  Just mentioning it here as a potential future thing... so going to resolve this.

jasonmcintosh (dismissed):
Overall nice little enhancement and makes stopping re-usable across multiple CI providers :)

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
eeigjon: Ah ok. Sorry new to this Jason so i didnt know what the legalities are here :).  Do i just add my own org (Ericsson) here instead?

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
eeigjon: Good point! I just tried to match the current  pattern but i guess i can change to a string to support UUID. want me to do that now?

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
eeigjon: Perfect. i just kept same as existing fields but that makes sense!!

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
jasonmcintosh: :) I think if possible that'd be great? I'd check with your manager to confirm.  IF it's not ok, then can use your name

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
jasonmcintosh: Na don't bother for now since all the API contracts are long based... that's a future state.

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
eeigjon: getting an error in the compile here. it seems groovy only added that alignment with Java 17 recently enough. woulf there be in issue with the build config? I might keep it as is for now anyway!

igor/igor-core/src/main/java/com/netflix/spinnaker/igor/service/StoppableBuildService.java
eeigjon: Changed org to Ericsson now

igor/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
jasonmcintosh: Ahh dang it yeah - groovy vs. java :) Keep it as it is

jasonmcintosh (approved):

PR #7886, opened 2026-08-13
feat(front50): deprecate non-SQL metadata storage backends

shlomodaari:
Warn at startup and mark S3/GCS/Redis/Azure/Oracle/Swift metadata storage deprecated so operators can migrate to SQL before removal after Spinnaker 2027.0.0. S3 plugin-binary storage is left alone.

https://spinnaker.io/docs/releases/roadmap/


jasonmcintosh (approved):
Won't hurt to add the logs so people see this and update the README with this notice.

PR #7888, opened 2026-08-13
fix(tests): Running tests as non root.  Previous attempt missed a few aspects to make this work

jasonmcintosh:
E.g.: https://github.com/spinnaker/spinnaker/pull/7872
Misses the root issue with the images & the security context stuff to make non-root stuff work correctly.


christosarvanitis (approved):
If we see that we spent too much on the fail-fast: false we can always revert

PR #7889, opened 2026-08-13
feat(helmfile): Adds the ability to use artifacts that allow overrides

### LINKED ISSUE #6925: support state-values-file for helmfile render engine
### Issue Summary:
add support for `state-values-file` as part of helmfile

### Cloud Provider(s):
all

### Environment:
any

### Feature Area:
bake manifest stage

### Description:
currently, helmfile supports passing in the values file (values.yaml) for the helm template, but not for the helmfile template itself. note that they are 2 different kinds of values files.

if one uses templating in the helmfile itself, the helmfile values cannot be overridden currently. 

### Steps to Reproduce:
deploy any helmfile (helmfile.yaml) manifest that requires a values.yaml file to override the helmfile itself.

example `helmfile.yaml`:
```helmfile
environments:
  foo:
  bar:

repositories:
- name: real-release
  url: https://someurl/helm-release/

releases:
- name: foo
  namespace: foo
  chart: ./base
  values:
  - ./environments/{{ .Environment.Name }}/{{ .Values.jobName }}.yaml
  - ../../metadata/environments/{{ .Environment.Name }}.yaml.gotmpl
```

example `values.yaml`:
```yaml
jobName: thejob
```

### Additional Details:

jasonmcintosh:
Should solve https://github.com/spinnaker/spinnaker/issues/6925


rosco/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helmfile/HelmfileTemplateUtils.java
```diff
@@ -129,6 +159,14 @@ public BakeRecipe buildCommand(
       command.add("--include-crds");
     }
 
+    if (!stateValuePaths.isEmpty()) {
```
christosarvanitis: Probably safeguard with a null check as well?

christosarvanitis (commented):
One comment on the missing null check

rosco/rosco-manifests/src/main/java/com/netflix/spinnaker/rosco/manifests/helmfile/HelmfileTemplateUtils.java
jasonmcintosh: TECHNICALLY b/c of the way the HelmTemplateUtils stuff calls this - it is never null.  
https://github.com/spinnaker/spinnaker/pull/7889/changes#diff-1725d28932f726e1657b0334ae90b1532395dc4f8387e3f207c3785370c9ba16R89
I don't MIND adding a null check just in case though to be defensive.

jasonmcintosh: Cause... arguably didn't need the isEmpty givent he way foreach operates :)

christosarvanitis (approved):
Looks good, thank you!

PR #7890, opened 2026-08-13
feat(helmfile): Expose environment and namespace arguments.  Add some validation as well 

jasonmcintosh:
Exposes the helm file environment/namespace arguments, and adds validation on those fields as well as UI helper information.


christosarvanitis (approved):
UI support for namepsace and environment arg in helmfile. Validation is a nice addon

PR #7892, opened 2026-08-14
feat(artifacts): add Fiat permissions to artifact accounts

### LINKED ISSUE #6592: Allow artifacts configuration to have permissions enabled
### Issue Summary:
Allow artefacts configuration to have permissions enabled like it's done in clusters, docker registries, jenkins ci so that the view can be filtered and only show the authorised artefacts

### Cloud Provider(s):
AWS

### Environment:
Running on halyard
Spinnaker v1.26.6
Running in HA with external Redis and Aurora database

### Description:
Would like to have enabled a feature to allow artefacts to be filtered on the UI based on the permissions set, like it's done with cluster, jenkins ci, etc.
For example for an helm chart type artifact we would have:
```
    helm:
      enabled: true
      accounts:
      - name: helm_chart_role_1
        permissions:
          READ:
          - role1
          WRITE:
          - role1
        repository: https://path_to_charts
        username: encrypted:secrets-manager!r:<region>!s:<id>!k:<key>
        password: encrypted:secrets-manager!r:<region>!s:<id>!k:<key>
```

That would allow to view and select certain repos from the Manifest artifact selection list:
image

Thanks

jasonmcintosh:
Marking as draft, as I'd like to dot his AFTER the https://github.com/spinnaker/spinnaker/pull/7863 PR work is evaluated and hopefully merged.  This will need updating post that work.

Will fix https://github.com/spinnaker/spinnaker/issues/6592


clouddriver/clouddriver-web/src/main/java/com/netflix/spinnaker/clouddriver/controllers/ArtifactController.java
jasonmcintosh, line 92: NOTE:  The ArtifactAccountAuthorizer system... bypasses fiat's permissions resolver.  ALSO we're NOT putting PreAuthorize stuff here... so it's not FULLY there.

jasonmcintosh:
I'll ALSO note I've NOT added this to the fetch operations.  It's not using PreAuthorize hooks either.  SO there IS a bit more work likely needed on this PR to make it "secure".  This is more... an "introduction" to the concepts.

jasonmcintosh:
FYI on the which provider to use...
---
2. On the FiatPermissionsResolver/@PreAuthorize question — evaluated, not yet changed. ArtifactAccountAuthorizer reimplements permission evaluation locally: it pulls AccountSecurityPolicy.getRoles(username)/isAdmin(username) and intersects them against the artifact account's own Permissions object by hand. That's a third pattern in the codebase, different from both existing ones:

- Cloud-provider accounts: DefaultAccountSecurityPolicy (the real, non-test AccountSecurityPolicy impl) delegates straight to FiatPermissionEvaluator.hasPermission(username, account, ResourceType.ACCOUNT.getName(), WRITE) — checked against Fiat's synced UserPermission.View. That requires the account to be a Fiat-known resource, populated by DefaultAccountResourceProvider polling clouddriver's /credentials endpoint. Artifact accounts aren't exposed there, so this route isn't available without new sync work on the Fiat side.
- Locally-permissioned objects (e.g. kork-secrets' UserSecret): implement AccessControlled.isAuthorized(Authentication, Object) directly. FiatPermissionEvaluator/AbstractPermissionEvaluator's object-form hasPermission(Authentication, Object, Object) already handles this generically — no Fiat resource registration needed at all, since the object carries its own permissions. ArtifactAccount already extends AccessControlledAccountDefinition (which declares getPermissions()), but ArtifactCredentials/ArtifactAccount don't implement AccessControlled — that's the missing link.

The second pattern is the natural fit here (artifact accounts already carry embedded Permissions, same shape as UserSecret, and aren't the kind of resource Fiat syncs today). The concrete move would be: make ArtifactCredentials implement AccessControlled.isAuthorized() with the same admin+role logic ArtifactAccountAuthorizer has now, then either annotate the controller methods with @PreAuthorize("hasPermission(#credentials, 'WRITE')") or inject Spring's PermissionEvaluator (of which FiatPermissionEvaluator is the runtime impl) at the getCredentialsForType()e bespoke authorizer class — same underlying mechanism the rest of the codebase uses, no parallel

---
Going to move this over to the AccessControlled path now ;)

jasonmcintosh:
Acutally... let me delay -that SECOND path... was never fully completed it looks like.  I'll go back to the Fiat path FOR NOW.

PR #7893, opened 2026-08-16
fix(signalfx): Move from signalfx library to standard retrofit to remove an old dep issue.

jasonmcintosh:
The signalfx java library is dead with a recommendation to move to OTEL.  HOWEVER... OTEL is for sending metrics, not querying them. Thus this migration which replaces the signalfx-java with simple parsers/converters.  The primary driver for this is that the signalfx-java library includes some CVE code that needs to be removed.  BUT this also further removes external custom deps that can polluate the system AND makes the tests more reliable.  This still does integration with a server but we mock the data vs. needing a real signalfx account to make this work.

### DIFF
--- a/kayenta/kayenta-signalfx/kayenta-signalfx.gradle
+++ b/kayenta/kayenta-signalfx/kayenta-signalfx.gradle
@@ -86,9 +86,9 @@ String getRequiredSystemProp(String key) {
 
 //noinspection GroovyAssignabilityCheck
 task integrationTest(type: Test) {
+  useJUnitPlatform()
   doFirst {
     systemProperty("redis.port", startEmbeddedRedis.ext.redisPort)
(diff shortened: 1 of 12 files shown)


kayenta/kayenta-signalfx/kayenta-signalfx.gradle
```diff
@@ -114,14 +114,16 @@ configurations {
 dependencies {
   implementation project(":kayenta-core")
 
-  api 'com.signalfx.public:signalfx-java:1.0.49'
```
jasonmcintosh, line 117: This here was the primary goal - to remove this library dep (has a databind CVE).

christosarvanitis (approved):
Successful integration tests 👍 and the rest look good

PR #7894, opened 2026-08-18
chore(groovy): Migrate pipeline triggers tests to groovy

jasonmcintosh:
Trying to start getting all the groovy code gone... starting with echo :)


christosarvanitis (approved):
Approving as tests pass and its 1to1 mapping

PR #7897, opened 2026-08-18
feat(artifacts): support GitHub App authentication for git/repo and github/file artifact accounts

christosarvanitis:
Adds GitHub App authentication to clouddriver's git/repo and github/file artifact accounts, as an alternative to personal access tokens, basic auth and SSH keys. GitHub Apps give better rate limits, fine-grained per-repository
permissions and clearer audit trails than PATs, and avoid long-lived credentials in Spinnaker config.

Accounts opt in with a nested githubApp block. Two modes:

Pinned — one installation serves the account:
```
artifacts:
  git-repo:
    enabled: true
    accounts:
      - name: my-repos
        githubApp:
          appId: "123456"
          appPrivateKeyPath: /secrets/gh-app-key.pem   # supports encrypted: / encryptedFile: URIs
          appInstallationId: "789012"
          # apiBaseUrl: https://ghe.example.com/api/v3   # optional, GitHub Enterprise
```
Derived — omit appInstallationId and the installation is resolved from the repository being accessed, so one account can serve several organizations:
```
        githubApp:
          appId: "123456"
          appPrivateKeyPath: /secrets/gh-app-key.pem
          allowedOrganizations:      # optional; omitting it allows any org where the app is installed
            - my-org
            - my-other-org
```
The same block works under artifacts.github.accounts. When present, GitHub App auth takes precedence over the other auth methods on the account.

### Changes

kork-github (reusable across services; fiat's existing usage is unchanged)

  • GitHubAppCredentials — bindable config model, validation (appId/appPrivateKeyPath required, appInstallationId numeric), and toAuthenticator(accountDescription) which is the single place that warns about unrestricted configs and
  wraps key-load failures.
  • GitHubAppAuthenticator — reworked: tokens cached per installation, plus an owner→installation map so cache hits make zero GitHub API calls; installations resolved via GET /repos/{owner}/{repo}/installation, which works for
  organization and user-owned repositories; installation resolution and token minting share one API round trip; app-authentication failures are reported distinctly from missing installations; mint locks are striped (fixed 32) rather
  than per-key, since derive-mode keys come from user-supplied artifact references.
  • GitHubRepoRef — conservative owner/repo parser shared by both artifact modules. It deliberately refuses ambiguous URLs rather than guessing.

clouddriver-artifacts-gitrepo

  • New GITHUB_APP auth type: clones over HTTPS as x-access-token:<token>, resolving the token once per command chain.
  • git pull on a retained clone re-points origin at a fresh token first, so clone retention works with expiring installation tokens.

clouddriver-artifacts-github

  • Installation token attached per request, so the 1-hour expiry is transparent.

clouddriver-artifacts (shared)

  • BaseHttpArtifactCredentials.getHeaders(T) now declares throws IOException, and a getHeaders(T, HttpUrl) overload was added for credentials whose auth material depends on the URL being fetched. See compatibility below.

### Notes

  • In derive mode the repository — and therefore the organization — comes from the artifact reference, which is pipeline-author input. An account without allowedOrganizations can reach every organization where the app is installed.
  The allowlist is enforced inside the authenticator (before any API call, so rejection is free and leaks nothing), and accounts that derive without one log a warning at startup. Default is permissive, matching the existing allowedH
  osts behaviour on git/repo accounts.
  • appPrivateKeyPath supports kork-secrets encrypted: / encryptedFile: URIs with no extra code, via the existing SecretPropertyProcessor file-property handling. Covered by tests.
  • Tokens never appear in process arguments or logs — git commands reference $GIT_TOKEN from the environment, as the existing token auth type does.


clouddriver/clouddriver-artifacts/clouddriver-artifacts-github/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/github/GitHubArtifactCredentials.java
```diff
@@ -77,7 +107,36 @@ private HttpUrl getMetadataUrl(Artifact artifact) {
...
+    return metadataUrl;
+  }
+
+  /**
+   * The installation token is minted against githubApp.apiBaseUrl but used against the artifact
+   * reference's host, so both must belong to the same GitHub instance. The reference is already a
+   * GitHub API URL (e.g. https://api.github.com/repos/... or https://ghe.example.com/api/v3/...),
+   * so the hosts must match exactly.
+   */
+  private void validateGitHubAppApiBaseUrlHost(HttpUrl reference) {
```
jasonmcintosh: This logic... seems fairly duplicated - at least on the apiHost validation stuff below.  Just wondering if that should be extracted so it's NOT duplicated between this and the JobExecutor.  OR if we REALLY need both?

kork/kork-github/src/main/java/com/netflix/spinnaker/kork/github/GitHubAppCredentials.java
```diff
@@ -0,0 +1,78 @@
...
+ * <pre>
+ * githubApp:
+ *   appId: "123456"
+ *   appPrivateKeyPath: /secrets/gh-app-key.pem
+ *   appInstallationId: "789012"
+ *   # apiBaseUrl: https://ghe.example.com/api/v3  # optional, for GitHub Enterprise
+ * </pre>
+ */
+@Value
+public class GitHubAppCredentials {
```
jasonmcintosh, line 51: This seems like a good candidate to move into kork-github & re-use overall.  Not REQUIRED but seems like some crossover pieces to;
https://github.com/spinnaker/spinnaker/blob/main/kork/kork-github/src/main/java/com/netflix/spinnaker/kork/github/GitHubAppAuthenticator.java#L75
that could be merged together into a better solution (e.g. base base url handling)

christosarvanitis: This is mainly a shared model (under kork-github) to be used by the clouddriver/clouddriver-artifacts/clouddriver-artifacts-github/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/github/GitHubArtifactAccount.java and clouddriver/clouddriver-artifacts/clouddriver-artifacts-gitrepo/src/main/java/com/netflix/spinnaker/clouddriver/artifacts/gitRepo/GitRepoArtifactAccount.java
The GithubAppAuthenticator provides the functionality for generating and caching the installation token. Both are in kork-github so can be shared by any service

jasonmcintosh:
Per slack, need more thought given that GHA are tied on installation token to a specific org vs PATs which can cross orgs.

jasonmcintosh (approved):
Approving :) As looked good to me.  Claude did call out two things:

```
Minor findings (non-blocking)

1. GitHubRepoRef.fromPathSegments edge case (kork/kork-github/.../GitHubRepoRef.java:202): a 2-segment path where the first segment is literally repos (e.g. an incomplete API path like .../repos/my-org missing the repo name) falls through to the "clone URL shape" branch and gets misparsed as owner="repos", repo="my-org" instead of being rejected as ambiguous. Low severity — it can't grant access beyond what the GitHub API itself allows (a bogus owner just fails upstream), but a one-line guard (skip the 2-segment fallback when segments.get(0).equals("repos")) would close it cleanly. Not covered by the current parameterized test set.
2. Unnecessary throws IOException on BitbucketArtifactCredentials.getHeaders (clouddriver-artifacts-bitbucket/.../BitbucketArtifactCredentials.java:46): the method body is unchanged and doesn't throw; Java allows overrides to narrow checked exceptions, so this addition isn't required by the BaseHttpArtifactCredentials signature change. Harmless, just a bit of drive-by noise worth a second look in case it was meant to signal something.

Not an issue (verified, initially looked suspicious)

- In derive mode, the raw download_url returned by the GitHub contents API (different host/path layout, e.g. raw.githubusercontent.com) doesn't get a derived Authorization header — this is deliberate and tested (gitHubAppDeriveModeOmitsCredentialsWhenRepositoryIsNotDerivableFromUrl), matching GitHub's actual behavior of returning pre-signed download URLs from the contents API.

```

PR #7759, opened 2026-07-01
feat(ecs): migrate ECS core agents to AWS SDK v2 EcsClient

jgrumboe:
Migrate all ECS core caching agents from v1 AmazonECS to v2 EcsClient.

## Changes (17 source files + 35 test files)

**Base classes:**

- AbstractEcsCachingAgent — v2 EcsClient, remove AWSCredentialsProvider
- AbstractEcsOnDemandAgent — same
- AbstractEcsAwsAwareCachingAgent — same

**Caching agents:**

- EcsClusterCachingAgent, ServiceCachingAgent, TaskCachingAgent
- TaskDefinitionCachingAgent, ContainerInstanceCachingAgent
- TaskHealthCachingAgent, ApplicationCachingAgent, TargetHealthCachingAgent

**Supporting files:**

- EcsCredentialsLifeCycleHandler — remove awsCredentialsProvider field
- EcsResource, EcsResourceService, EcsTagNamer — v2 Tag model
- EcsServerGroupNameResolver — v2 EcsClient
- CreateServerGroupAtomicOperation — v2 tags for EcsResource

**Integration tests:**

- CreateServerGroupSpec, CreateServerGroupForExistingServiceSpec, CreateServerGroupWithAppAutoScalingSpec, CreateServerGroupWithArtifactsSpec, CreateServerGroupWithMonikerSpec — add v2 EcsClient mock for EcsServerGroupNameResolver

## Design decisions

- AWSCredentialsProvider removed from agent constructors — resolved internally via AmazonClientProvider.getAmazonEcsV2()
- TargetHealthCachingAgent: ELBv2 code kept on v1 (no v2 getter yet, separate PR)
- Cache write uses v2 model accessors directly — JSON shape unchanged

## Cache compatibility

No impact on existing entries. Field names are identical between v1 and v2.

## What was tested

- compileJava — clean
- spotlessCheck — clean
- `./gradlew :clouddriver:clouddriver-ecs:test` — all tests pass (unit + Groovy specs)
- `./gradlew :clouddriver:clouddriver-ecs:integrationTest` — all 33 tests pass

Part of spinnaker/spinnaker#7699.

PR #7898, opened 2026-08-18
fix(ecs): read cached task containers with the SDK v2 model

jgrumboe:
## Problem

Since #7759 migrated the core ECS caching agents to AWS SDK v2, `TaskCachingAgent` writes v2 `software.amazon.awssdk.services.ecs.model` objects into the `TASKS` namespace, but `TaskCacheClient` still deserialized `containers` into the v1 `com.amazonaws.services.ecs.model.Container`.

Where a container reports managed agents (ECS Exec), the cached `managedAgents[].lastStartedAt` is a floating-point epoch, because the v2 model uses `Instant`. The v1 `java.util.Date` binding rejects it:

```
java.lang.IllegalArgumentException: Cannot deserialize value of type `java.util.Date` from
  Floating-point value (token `JsonToken.VALUE_NUMBER_FLOAT`) (through reference chain:
  com.amazonaws.services.ecs.model.Container["managedAgents"]->java.util.ArrayList[0]
  ->com.amazonaws.services.ecs.model.ManagedAgent["lastStartedAt"])
```

Every task conversion throws, which means:

- `GET /applications/{app}/serverGroups` answers `400`, so the Deck cluster view shows no ECS server groups and no task instances
- `TaskHealthCachingAgent` fails on every run

Reproduced on a staging cluster running a build of `main`: the endpoint returns the error above, and the caching pod logs the same stack trace once per agent cycle.

## Fix

Move the task read path to the v2 model: the `Task` cache model, `TaskCacheClient`, and the `Container` / `NetworkBinding` / `NetworkInterface` call sites in `TaskHealthCachingAgent`, `ContainerInformationService`, `EcsServerClusterProvider`, `EcsInstanceProvider` and `EcsTask`.

`ContainerDefinition`, `PortMapping`, `LoadBalancer` and `TaskDefinition` stay on v1, matching the caches that still write them.

No writer changes, so the cached format is untouched and rolling deploys are unaffected in both directions — this only aligns readers with a format that already changed in #7759. The v2 `NetworkInterface` serializes to the same JSON field names, so the `EcsTask` payload Deck receives is unchanged.

## Tests

- New `TaskCacheClientSpec` covers the agent write path, the JSON round trip the cache performs, and a container carrying managed agents (the case that failed).
- The affected specs now use an `ObjectMapper` with `AwsSdkV2Module` registered, as clouddriver does. A bare `new ObjectMapper()` hides this class of bug, since v2 objects only serialize correctly with that module.
- `:clouddriver:clouddriver-ecs:test`, `:integrationTest` and `spotlessCheck` pass.

## Follow-up (not in this PR)

`ServiceCacheClient` (v1 `LoadBalancer`) and `TaskDefinitionCacheClient` (v1 `ContainerDefinition`) also read v2-written attributes. Neither type has a date field, so they round-trip by field name today and do not crash — but they are the same latent mismatch and will break as soon as one of those shapes gains an `Instant` or a differently-serialized enum.

### DIFF
--- a/clouddriver/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/TaskCacheClient.java
+++ b/clouddriver/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/cache/client/TaskCacheClient.java
@@ -18,7 +18,6 @@ package com.netflix.spinnaker.clouddriver.ecs.cache.client;
 
 import static com.netflix.spinnaker.clouddriver.ecs.cache.Keys.Namespace.TASKS;
 
-import com.amazonaws.services.ecs.model.Container;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spinnaker.cats.cache.Cache;
 import com.netflix.spinnaker.cats.cache.CacheData;
@@ -29,6 +28,7 @@ import java.util.List;
 import java.util.Map;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
+import software.amazon.awssdk.services.ecs.model.Container;
 
 @Component
 public class TaskCacheClient extends AbstractCacheClient<Task> {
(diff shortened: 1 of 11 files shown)


jasonmcintosh (approved):
Looks like should help fix with some of the stuff on migration!  Thanks!

PR #7901, opened 2026-08-19
fix(ecs): re-describe task definitions cached in a degraded form

jgrumboe:
Follow-up to #7899, and an answer to the operational caveat that PR left behind:

> Cached entries already stripped by an affected build stay stripped after this fix, since they are still never re-described. Those cache entries need to be evicted once for the data to come back.

## Problem

`TaskDefinitionCachingAgent` skips `describeTaskDefinition` whenever an entry is already cached. That is sound on its own — task definitions are immutable — but it means a cached entry is *never* described again, so an entry cached in a degraded form stays that way indefinitely. #7899 fixed a read path that dropped `portMappings`, `environment` and `healthCheck`; entries already written by an affected build keep those gaps, and rolling back does not repair them either, because the reuse shortcut exists in every version.

Today the only remedy is deleting the affected records straight from the cache backend. Nothing in cats can prompt a re-fetch:

- the hash/delta write path only compares freshly computed data against what is stored, to skip redundant writes — and here the computed data *is* the degraded data, so there is no delta to notice;
- a forced refresh via `POST /cache/ecs/serverGroups` reaches the agent (it is an `AbstractEcsOnDemandAgent`) but re-enters this same `getItems`, so it takes the reuse branch as well;
- there is no TTL on these records and no evict endpoint.

Concretely, on a staging instance every cached task definition had `"portMappings":[]` and no `containerPort` or `healthCheck` keys at all. `TaskHealthCachingAgent` logged `Container does not contain a port mapping with load balanced container port: 7007` and cached 0 health entries per cycle, so Deck showed `Unknown` for load balancer health on every ECS instance in the account. Deleting the rows was the only thing that brought it back.

## Change

Make the reuse conditional on the cached entry still carrying what the ECS provider reads from it. An entry is reused when:

- it has container definitions, and
- for each container port the services using it load balance on, it has a matching port mapping — exactly what `TaskHealthCachingAgent` needs to resolve task health.

Anything else is described again, so degraded entries repair themselves on the next caching cycle with no operator action.

Notes on cost and edge cases:

- The load balanced ports come from the services the agent already reads to collect task definition ARNs, so this adds no API calls.
- A task definition whose services have no load balancer is only checked for container definitions.
- If a task definition genuinely never satisfies the check, the cost is one describe per caching cycle for that one task definition — and describing it is the correct behaviour regardless.

## Tests

Three new cases in `TaskDefinitionCachingAgentTest`: a cache entry missing the load balanced port mapping is described again, an entry without container definitions is described again, and a complete entry with the load balanced port mapping is still reused. The two "described again" cases fail against `main` and pass with this change.

`shouldRetainCachedTaskDefinitions` previously cached nothing but an ARN, which no longer counts as complete, so its fixture now caches a container definition to keep expressing reuse.

All the tests here distinguish reuse from a describe call by the image they assert on, rather than by verifying against the `EcsClient` mock — that mock is `static` in `CommonCachingAgent` and therefore shared across tests, so stubs leak between them and `verify()` counts accumulate. Worth flagging separately; not touched here.

`:clouddriver-ecs:test` and `spotlessCheck` pass. `integrationTest` passes; `CreateServerGroupWithAppAutoScalingSpec` is intermittently flaky on `main` independently of this change (reproduced on the unmodified base, roughly 1 run in 3).

## Alternative considered

Versioning the cache key instead — following `CachingSchema` in the Titus provider and the `kubernetes.v2` key prefix — would let existing eviction logic clear outdated records automatically, and would generalise to future serialization changes. It also changes the key format everywhere task definition keys are built or parsed (Titus needed a `removeSchemaVersion` helper at API boundaries), so I went with the narrower fix here. Happy to switch approach if you would prefer the general mechanism.

### DIFF
--- a/clouddriver/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskDefinitionCachingAgent.java
+++ b/clouddriver/clouddriver-ecs/src/main/java/com/netflix/spinnaker/clouddriver/ecs/provider/agent/TaskDefinitionCachingAgent.java
@@ -19,6 +19,7 @@ package com.netflix.spinnaker.clouddriver.ecs.provider.agent;
 import static com.netflix.spinnaker.cats.agent.AgentDataType.Authority.AUTHORITATIVE;
 import static com.netflix.spinnaker.clouddriver.ecs.cache.Keys.Namespace.TASK_DEFINITIONS;
 
+import com.amazonaws.services.ecs.model.LoadBalancer;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.cats.agent.AgentDataType;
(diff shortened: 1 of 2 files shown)


jasonmcintosh (approved):
I'm inclined to accept this - though this is a bug no one SHOULD hit... validating the cache on this isn't a bad idea - and we've improved ECS in general enough lately that I think the perf hit is also acceptable.

PR #7902, opened 2026-08-19
fix(sql): qualify GROUP BY columns so getAllByApplication works on Postgres

christosarvanitis:
## Problem

With `sql.cache` backed by **PostgreSQL**, `GET /applications/{application}/serverGroups`
returns an empty list. The same request scoped to a cluster works and returns data:

```
GET /applications/myapp/serverGroups                        -> 200 []          # wrong
GET /applications/myapp/serverGroups?clusters=acct:myapp-x  -> 200 [ ... ]     # correct
```
## Root cause

`SqlCache.getDataWithRelationshipsByApp` builds a `UNION ALL` whose second branch joins
the resource table (aliased `r`) to the relationship table (aliased `rel`), then groups by
**unqualified** column names:

```kotlin
.from(table(...).as("r"))
.innerJoin(table(...).as("rel"))
.on(sql("rel.id=r.id"))
.where(relWhere)
.groupBy(field("rel_id"), field("id"), field("rel_type"))   // <-- unqualified
```

Both `r.id` and `rel.id` are in scope, so `group by ... id ...` is ambiguous:

```
ERROR:  column reference "id" is ambiguous
LINE 9:  group by rel_id, id, rel_type;
                          ^
```

The query throws, and the surrounding `catch (e: Exception)` converts the failure into an
**empty result** — so the API returns `200 []` instead of surfacing an error.

### Why only Postgres

The two engines resolve `GROUP BY` names in opposite orders. MySQL prefers the *output*
column aliases (`rel.id as id`), so the query is unambiguous and succeeds. PostgreSQL
resolves against *input* columns first, where `id` matches both joined tables.

This is not a dialect-configuration problem. `field(String)` is jOOQ's plain-SQL overload,
so the fragment is passed through verbatim; rendering the query with `SQLDialect.POSTGRES`
emits the identical `group by rel_id, id, rel_type`. No configuration avoids it.

Unfortunately `getAllByApplication` is only covered by `SqlProviderCacheSpec`, which ran against MySQL
only. All of its existing cases also pass `RelationshipCacheFilter.none()`, which routes to
the simpler `getDataWithoutRelationshipsByApp` query — so even running that spec against
Postgres as written would not have caught this.

## Fix

Qualify the `GROUP BY` columns with the `rel` alias, matching the style already used in the
`SELECT` list immediately above it. Renders identically on both engines; MySQL behaviour is
unchanged.

## Scope

Affects only code paths reaching `ClusterProvider.getClusterDetails()`, and only the AWS
provider (`AmazonClusterProvider` is the sole provider that branches on
`supportsGetAllByApplication()`):

- `GET /applications/{application}/serverGroups` (with and without `expand`)
- `GET /serverGroups?applications=...`

Not affected: `GET /applications/{app}` (uses `getClusterSummaries`, `includeDetails=false`),
all `ClusterController` endpoints, `GET /serverGroups?ids=...`, and non-AWS providers.

## Testing

- `SqlProviderCacheSpec` is now abstract with `initDatabase()`/`getDialect()` hooks; added
  `MySqlProviderCacheSpec` and `PostgreSqlProviderCacheSpec`. All existing test bodies are
  unchanged and now run against both engines. This mirrors the existing
  `SqlCacheSpec` / `MySqlCacheSpec` / `PostgreSqlCacheSpec` split.
- Added `can retrieve by application with relationships`, covering the previously untested
  relationship-bearing branch. Parameterised over `RelationshipCacheFilter.include(...)` and
  `null` — the latter resolves to the `"ALL"` prefix, which is what `AmazonClusterProvider`
  relies on for `CLUSTERS.ns`.

Confirmed the new tests fail without the fix:

```
PostgreSqlProviderCacheSpec > can retrieve by application with relationships FAILED
    fooData['serverGroup'].collect { it.id } == ['fooSg1']
    |      |               |
    |      []              []
    [serverGroup:[]]
```

And pass with it:

| Spec | Tests | Failures |
|---|---|---|
| `MySqlProviderCacheSpec` | 28 | 0 |
| `PostgreSqlProviderCacheSpec` | 28 | 0 |

### DIFF
--- a/clouddriver/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
+++ b/clouddriver/cats/cats-sql/src/main/kotlin/com/netflix/spinnaker/cats/sql/cache/SqlCache.kt
@@ -1143,10 +1143,13 @@ class SqlCache(
               .innerJoin(table(sqlNames.relTableName(type)).`as`("rel"))
               .on(sql("rel.id=r.id"))
               .where(relWhere)
+              // Columns must be qualified with the `rel` alias: both `r` and `rel` expose an `id`
+              // column, and postgres resolves group by names against input columns before output
+              // aliases, so unqualified names fail with "column reference is ambiguous".
               .groupBy(
-                field("rel_id"),
-                field("id"),
-                field("rel_type")
+                field("rel.rel_id"),
+                field("rel.id"),
+                field("rel.rel_type")
               )
           )
           .fetch()
--- a/clouddriver/cats/cats-sql/src/test/groovy/com/netflix/spinnaker/cats/sql/MySqlProviderCacheSpec.groovy
+++ b/clouddriver/cats/cats-sql/src/test/groovy/com/netflix/spinnaker/cats/sql/MySqlProviderCacheSpec.groovy
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2026 Harness, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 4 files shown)


jasonmcintosh (approved):

christosarvanitis:
Backport candidate for the active releases. Seems this bug has been there for a while now. But deck can reconstruct the UI view from the clusters api call

jasonmcintosh:
@mergifyio backport release-2026.2.x release-2026.1.x

PR #7896, opened 2026-08-18
feat(ecs): migrate atomic operations and scalable targets to AWS SDK v2

jgrumboe:
Migrate all ECS atomic operations and ScalableTargetsCachingAgent from AWS SDK v1 to v2.

> **Depends on:** spinnaker/spinnaker#7895 (TargetHealth/ELBv2 migration) — this PR includes those commits for CI purposes; after #7895 merges, this branch will be rebased to contain only the atomic-ops/scalable-targets work.

## Changes

**Atomic Operations (9 files):**
- CreateServerGroupAtomicOperation — full v2 (ECS, IAM, ELBv2, AppAutoScaling)
- DestroyServiceAtomicOperation, DisableServiceAtomicOperation
- EnableServiceAtomicOperation, ResizeServiceAtomicOperation
- TerminateInstancesAtomicOperation, AbstractEcsAtomicOperation
- CreateServerGroupDescription (v2 placement/capacity provider types)
- EcsCreateServerGroupDescriptionValidator (v2 PlacementStrategyType)

**ScalableTargets (3 files):**
- ScalableTargetsCachingAgent — migrated to v2 ApplicationAutoScalingClient
- ScalableTargetCacheClient — v2 deserialization via serializableBuilderClass()
- EcsServerClusterProvider — updated to consume v2 ScalableTarget

**SDK v2 Deserialization (1 new file):**
- AwsSdkV2JacksonModule — deserializes SDK v2 model types via builder + sdkFields() metadata (fixes artifact-based task definition deserialization)

**Tests (15 files):**
- All atomic operation Spock specs updated to v2 mocks and builders
- ScalableTargetCachingAgentSpec + EcsServerClusterProviderSpec migrated to v2
- All 5 integration test specs migrated to full v2 mocks (EcsClient, ElasticLoadBalancingV2Client, ApplicationAutoScalingClient)

Zero `com.amazonaws` imports remain in the deploy/ directory.

## What was tested
- All 391 clouddriver-ecs unit tests pass, 0 failures
- Integration tests compile cleanly
- spotlessCheck clean

Part of spinnaker/spinnaker#7699.

PR #7903, opened 2026-08-19
feat(clouddriver-aws): migrate SNS, SQS, SWF, Support, CloudFormation to AWS SDK v2

jgrumboe:
Phase 1 of the AWS SDK v1 → v2 migration for clouddriver-aws. Part of spinnaker/spinnaker#7699

Services migrated:
- AmazonSNS → SnsClient
- AmazonSQS → SqsClient
- AmazonSimpleWorkflow (SWF) → SwfClient
- AWSSupport → SupportClient
- AmazonCloudFormation → CloudFormationClient

Changes:
- Add v2 SDK dependencies (cloudformation, sns, sqs, support, swf, iam-policy-builder)
- Add v2 client getters to AmazonClientProvider
- Migrate InstanceTerminationLifecycleWorker and LaunchFailureNotificationAgent to v2 SNS/SQS clients and v2 IAM policy builder
- Migrate all CloudFormation atomic operations and caching agent to v2
- Convert CloudFormation Change objects to plain maps for Jackson serialization (v2 SDK models are not Jackson-serializable)
- Remove v1 getters and imports for migrated services
- Rewrite all affected test specs to use v2 client mocks

Note: Shield migration deferred to Phase 3 (callers are in ELB upsert operations, which are Phase 3 scope).

Part of #7699


clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/model/AmazonCloudFormationChangeSet.java
```diff
@@ -15,17 +15,17 @@
...
+import java.util.Map;
 
 @JsonInclude(JsonInclude.Include.NON_EMPTY)
 public class AmazonCloudFormationChangeSet {
 
   private String name;
   private String status;
   private String statusReason;
-  private List<Change> changes;
+  private List<Map<String, Object>> changes;
```
jasonmcintosh: No immediate concerns but commenting to mention - this is a change in structure POTENTIALLY where there's less... "structure" on the changes as a result.  NOT sure if V2 has an equivalent cloudformation change.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCloudFormationCachingAgent.java
```diff
@@ -263,78 +275,108 @@ public CacheResult queryStacks(
...
+    stackAttributes.put("stackName", stack.stackName());
     stackAttributes.put("region", region);
     stackAttributes.put("accountName", account.getName());
     stackAttributes.put("accountId", account.getAccountId());
-    stackAttributes.put("stackStatus", stack.getStackStatus());
-    stackAttributes.put("creationTime", stack.getCreationTime());
-    stackAttributes.put("changeSets", getChangeSets(stack, cloudformation));
-    getStackStatusReason(stack, cloudformation)
+    stackAttributes.put("stackStatus", stack.stackStatusAsString());
+    stackAttributes.put(
```
jasonmcintosh, line 292: KINDA wonder if we have other places with an issue on this time stamp handling that we could make more generic... @christosarvanitis any thoughts on the jacksonmodule handling this kinda thing?  I'm not sure how universal this really is though...

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCloudFormationCachingAgent.java
```diff
@@ -263,78 +275,108 @@ public CacheResult queryStacks(
...
-                    changeSetAttributes.put("changes", describeChangeSetResult.getChanges());
+                        DescribeChangeSetRequest.builder()
+                            .changeSetName(summary.changeSetName())
+                            .stackName(stack.stackName())
+                            .build();
+                    DescribeChangeSetResponse describeChangeSetResponse =
+                        cloudFormationClient.describeChangeSet(describeChangeSetRequest);
+                    changeSetAttributes.put(
+                        "changes",
+                        describeChangeSetResponse.changes().stream()
```
jasonmcintosh: KINDA wondering about this tied to the above Change object replacement situation.  Overalll... I THINK this is "ok" but... still interesing to note how much changes here.

E.g. there's a lot of new things LIKE the physicalResourceId.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/model/AmazonCloudFormationChangeSet.java
jasonmcintosh: Thinking of this: https://docs.aws.amazon.com/java/api/latest/software/amazon/awssdk/services/cloudformation/model/Change.html for reference

jasonmcintosh (dismissed):
This is a TENTATIVE approval pending responses - a few minor concerns around the object type changes.

jgrumboe:
@jasonmcintosh Thanks for the review! Addressed the feedback:

**Re: `List<Change>` → `List<Map<String, Object>>`** — Reverted this. The `Change` type is now kept as-is (strongly typed). Serialization is handled by the existing `AwsSdkV2Module` (registered globally via `AwsSdkV2JacksonConfiguration`), same module referenced in #7896. No need for manual map conversion.

**Re: timestamp handling** — Added a null-guard for `stack.creationTime()` since v2 SDK models return `null` for unset fields (unlike v1 which returned a `Date` object). If we see this pattern repeat in later phases we can extract a utility, but for now it's a one-liner.

Also merged upstream/main to resolve the conflict (new `ElasticLoadBalancingV2Client` getter was added at the same location).

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/provider/agent/AmazonCloudFormationCachingAgent.java
christosarvanitis: Cant recall any other situations that we are caching the timestamp. I would assume cloudformation is unique in this. Also not really sure why we are caching the creation timestamp (but thats a completely different topic - what clouddriver needs vs what is really doing). 

I would say lets go with this and once the migration to v2 is over we will have the opportunity to optimize stuff?

jasonmcintosh (approved):

PR #7910, opened 2026-08-20
feat(mcp): Add a spinnaker native MCP server

jasonmcintosh:
NOTE:  This PR heavily AI generated with input, directions, etc.  That said... this stuff is FAIRLY standard to implement.

----
Adds gate-mcp, a new Gradle module exposing Spinnaker CD operations as MCP tools/resources/prompts (applications, deploys, pipeline configs and executions, manual judgments, ad-hoc tasks, Kayenta canary analysis, Managed Delivery/Keel, and infrastructure search), so an MCP client can drive Spinnaker directly.

Disabled by default (mcp.server.enabled=false) and read-only by default once enabled (mcp.server.read-only=true), so operators opt in explicitly to both exposing the endpoint and allowing mutating tools to execute. Kayenta/Keel-backed tools additionally require their respective services.*.enabled properties.

VERY preliminary ...


jasonmcintosh:
NOTE:  There's a bug in the search controller system fixing as part of this.  UI worked around it but... trying to fix it fully here.

gate/gate-mcp/src/main/java/com/netflix/spinnaker/gate/mcp/tools/ApplicationTools.java
```diff
@@ -0,0 +1,140 @@
...
+      description =
+          "List Spinnaker applications visible to the caller, optionally filtered by owner email.")
+  @SuppressWarnings("unchecked")
+  public List<Map<String, Object>> listApplications(
+      @McpToolParam(
+              description = "Only include applications owned by this email address",
+              required = false)
+          String owner) {
+    List<Map> applications =
+        Retrofit2SyncCall.execute(front50Service.getAllApplicationsUnrestricted());
```
jasonmcintosh, line 70: Reviewing... this MIGHT be unintentionally bypassing front50 auth here...

jasonmcintosh: This is why we always review AI prs! :)

gate/gate-mcp/src/main/java/com/netflix/spinnaker/gate/mcp/tools/PipelineTools.java
christosarvanitis, line 1: @GetMapping("{id}/evaluateExpression") would be useful as well

christosarvanitis:
Nice! Admin tools would be a nice to have? Cancel zombies, publish the global banners etc. It can be a follow up of course

christosarvanitis:
Could be useful https://docs.spring.io/spring-ai/reference/api/prompt.html#_using_resources_instead_of_raw_strings Some predefined Prompts but also provide a way for users to mount their own

jasonmcintosh:
What would be interesting on the prompts is the ability to load - either from a configmap or similar a set of prompts to make this more dynamic/configurable for users :) I think that maybe another PR if this one is good - as it's already a touch big.
Just add the expressions endpoint & admin endpoints.  AND Audit log feeds.

christosarvanitis (approved):
Approving as an ALPHA functionality

PR #7916, opened 2026-08-23
feat(deck): Add org/repo/path URL builder for git file artifacts

jasonmcintosh:
This fixes a LONG standing headache of mine - having to CAREFULLY type in (via copy/paste or otherwise) the "api.github.com/content/org/repo" path structure by providing some helper boxes instead.  SHOULD make loading a single file from git a touch easier :)  The URL in the pipeline is NOT changing - just a UI feature here.

Introduces a toggle that builds the URL from separate base URL, org/project, repo, and file path fields, previewing the resolved URL instead of requiring users to hand-assemble it. 

Also fixes Gitlab's default editor reusing GitHub's path regex, which never matched Gitlab URLs and left artifact.name set to the whole URL instead of the extracted file path.


christosarvanitis (approved):
Agreed

PR #7918, opened 2026-08-24
fix(spin): send Bearer token on all API requests when using oauth2 auth

### LINKED ISSUE #7917: spin CLI with oauth2 auth cannot call any protected Gate API - Bearer token only sent on login, never on API requests
### Issue Summary:

With `auth.oauth2` configured, spin completes the authorization-code+PKCE login and caches the token, but every subsequent command fails with `Could not reach Gate, please ensure it is running. Failing.` / `undefined response type`. The cause: spin sends the Bearer token ONLY on its initial login request - every actual API call goes out with NO Authorization header at all.

Captured by pointing spin at a header-logging fake Gate:

```
REQ /login/oauth2/code/other | Authorization: Bearer eyJ... (present)
REQ /version                 | Authorization: NONE
REQ /applications            | Authorization: NONE
```

### Cloud Provider(s):

n/a (spin CLI / Gate authentication)

### Environment:

- Gate 2026.2.x (Spring Security oauth2Login() stack, OIDC provider: Okta), deployed via halyard
- Affected spin builds: official release binary 2026.2.3 AND a fresh build from main @ 3a5bc5f (2026-08-24) - identical captures from both

### Feature Area:

spin CLI, authentication (oauth2)

### Description:

Expected: API requests authenticate with the cached OAuth2 access token as a Bearer header. Gate supports this server-side - `ExternalAuthTokenFilter` validates bearer tokens per-request against the provider's userinfo endpoint, and `OAuth2IntegrationWithWireMockTest.bearerTokenAuthenticatesViaUserInfoEndpoint` asserts bearer + `/credentials` returns 200.

Actual: only the `login()` call (`GET /login/oauth2/code/{provider}`) carries the Bearer header. All API requests rely on a session cookie - but a Gate running the Spring Security `oauth2Login()` stack never issues a session to non-browser clients (the login callback endpoint requires a real authorization-code exchange), so every protected endpoint responds 302-to-login, which the generated client surfaces as `undefined response type`.

Root cause chain:

1. #7612 ("update gateclient to always set Bearer token for oauth2") stores the token in `gate.ContextAccessToken` (`spin/cmd/gateclient/client.go`) - but nothing consumes that value.
2. Gate's `swagger.json` defines no `securitySchemes`, so openapi-generator emits no auth-injection code in the generated gateapi client. The `ContextAccessToken` constant only exists because `spin/swagger/generate_swagger.sh` sed-injects it into the generated `configuration.go` - the generated `client.go` never reads it.
3. The one helper that would map the context value to a header (`AddAuthHeaders` in `spin/cmd/gateclient/client.go`) has zero callers.
4. The test added with #7612 asserts the context value is SET, and although it captures the login request's Authorization header into a channel, it never reads that channel - so no test ever asserted the header actually reaches Gate on an API request.

Predecessor issue: #5913 (fixed for the login flow by #7391/#7392; this issue is about the API requests after login).

### Steps to Reproduce:

1. Configure `~/.spin/config` with a working `auth.oauth2` block (PKCE public client) against any Gate running the oauth2Login() stack.
2. Run `spin application list` - login succeeds ("Caching oauth2 token."), then the command fails.
3. Optionally point `--gate-endpoint` at a small HTTP server that logs request headers to observe the missing Authorization header on `/version` and `/applications`.

### Additional Details:

Fix incoming as a linked PR: transport-level Bearer injection (host-scoped so redirects to the IdP don't leak the token) plus a regression test that asserts the header is received by a fake Gate on an API request. Longer-term, declaring `securitySchemes` in Gate's OpenAPI spec would let openapi-generator emit native auth handling in the generated client, making the sed-injection and manual plumbing unnecessary - considered out of scope for the minimal fix.

iniinikoski:
Fixes #7917.

### Summary

With `auth.oauth2` configured, spin sends the Bearer token only on its initial `login()` request; every actual API call goes out with **no Authorization header**, relying on a session cookie that a Gate running the Spring Security `oauth2Login()` stack never issues to non-browser clients. Result: oauth2-configured spin cannot call any protected Gate endpoint (`Could not reach Gate... undefined response type`). Broken in the 2026.2.3 release binary and at current `main` (verified with a header-logging fake Gate - captures in the linked issue).

### Root cause

#7612 stores the token in `gate.ContextAccessToken`, but nothing consumes it:
- Gate's `swagger.json` declares no `securitySchemes`, so openapi-generator emits no auth-injection code in the generated gateapi client (the `ContextAccessToken` constant is only sed-injected into the generated `configuration.go` by `spin/swagger/generate_swagger.sh`).
- The `AddAuthHeaders` helper that would map the context value to a header has no callers.

### Fix

A `bearerAuthTransport` (`http.RoundTripper`) wired into the gate HTTP client when an oauth2 cached token exists:
- **Host-scoped** to the gate host, so the token is not replayed to other hosts on redirects (e.g. the identity provider).
- Skips requests that already carry an Authorization header.
- Reads the token through a closure, so tokens refreshed by `authenticateOAuth2` are picked up.

The existing `ContextAccessToken` line is kept for API parity.

### Test

`TestNewGateClientSendsBearerTokenOnApiRequests` asserts the Authorization header is actually **received** by a fake Gate on an API request (`/version`). This completes the assertion the #7612 test set up but never made (it fills a channel with the login request's Authorization header and never reads it). Verified the new test fails without the fix and passes with it. Also verified end-to-end against a real Okta-backed Gate deployment (oauth2Login() stack): `spin application list` works with only an OIDC bearer token.

### Future work (out of scope)

Declaring `securitySchemes` in Gate's OpenAPI spec would make openapi-generator emit native auth handling in the generated client, removing the need for the sed-injection and this transport. Larger blast radius (full client regeneration), left for a follow-up.

### DIFF
--- a/spin/cmd/gateclient/client.go
+++ b/spin/cmd/gateclient/client.go
@@ -187,6 +187,25 @@ func NewGateClient(ui output.Ui, gateEndpoint, defaultHeaders, configLocation st
 		// on every API request, rather than relying solely on session cookies.
 		if gateClient.Config.Auth.OAuth2.CachedToken != nil {
 			gateClient.Context = context.WithValue(gateClient.Context, gate.ContextAccessToken, gateClient.Config.Auth.OAuth2.CachedToken.AccessToken)
+
+			// The generated gate client never consumes ContextAccessToken, so
+			// inject the Bearer header at the transport level for requests to
+			// the gate host. Without this, API calls carry no Authorization
+			// header at all and depend on a session cookie that a gate running
+			// Spring Security oauth2Login() never issues to non-browser clients.
+			if gateURL, err := url.Parse(gateClient.GateEndpoint()); err == nil {
+				oauth2Config := gateClient.Config.Auth.OAuth2
+				httpClient.Transport = &bearerAuthTransport{
+					next:     httpClient.Transport,
+					gateHost: gateURL.Host,
+					tokenFn: func() string {
+						if oauth2Config.CachedToken != nil {
+							return oauth2Config.CachedToken.AccessToken
+						}
+						return ""
+					},
+				}
+			}
 		}
 
 		updatedMessage = "Caching oauth2 token."
@@ -309,6 +328,26 @@ func userConfig(gateClient *GatewayClient, configLocation string) error {
 		gateClient.Config = config.Config{}
 	}
 	return nil
+}
+
+// bearerAuthTransport injects the cached OAuth2 access token as a Bearer
+// Authorization header on requests to the gate host that do not already
+// carry one. Scoped to the gate host so the token is not replayed to other
+// hosts on redirects (e.g. the identity provider).
+type bearerAuthTransport struct {
+	next     http.RoundTripper
+	gateHost string
+	tokenFn  func() string
+}
+
+func (b *bearerAuthTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Header.Get("Authorization") == "" && req.URL.Host == b.gateHost {
+		if token := b.tokenFn(); token != "" {
+			req = req.Clone(req.Context())
+			req.Header.Set("Authorization", "Bearer "+token)
+		}
+	}
+	return b.next.RoundTrip(req)
 }
 
 // InitializeHTTPClient will return an *http.Client configured with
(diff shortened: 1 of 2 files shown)


jasonmcintosh (approved):
Good tests too on this let's hope this fixes these issues!!  Thank you!

PR #7919, opened 2026-08-24
fix(orca): ack queue messages atomically so a concurrent push survives for stages that have looping tasks

### LINKED ISSUE #5952: Pipeline executions are occasionally stuck
### Issue Summary:
Aloha! 

We are occasionally seeing that our pipeline executions get stuck, where none of the pipeline stages actually run.

### Cloud Provider(s):
GCP

### Environment:
Spinnaker deployed into GKE cluster.

### Feature Area:
Pipelines

### Description:
**Behavior that we're seeing:**
- The pipeline's status says that it is `RUNNING`, but the very first stage is `NOT_STARTED`. The very first stage has no preconditions.
- It stays stuck indefinitely like this for hours.
- We encounter this once in a while, but not all the time. Anecdotally, we would say ~1-2% of pipeline executions.

pipeline_stuck

**How we've tried to remedy this:**

- Pressing the cancel button (the red X in the web UI) will show a confirmation dialog, but accepting the dialog causes the dialog to be stuck on the browser window forever.
- Using the `spin` CLI to call `spin pipeline execution cancel ...` will say that the pipeline execution is `successfully canceled`, but nothing actually changes. The UI is still in this state, and subsequent pipeline executions cannot proceed.
- When we see this, the current remedy we have is to delete the entire application and re-install pipelines.
- As suggested by @kskewes (thank you!), we could try attempting to delete the associated row in SQL -- though would prefer not to have to do this. We're interested in understanding and fixing the root cause! 🍍 🙂  🍍 
- We tried searching through logs for messages containing the associated execution ID, but did not see anything of note.

### Steps to Reproduce:
This is hard to reproduce, but anecdotally, it happens in around ~1-2% of our pipeline executions. Our pipeline configuration has not been changing -- the same configuration will get stuck sometimes, but run just fine in other times.

### Additional Details:
- We're currently running on Spinnaker `1.20.4` on GKE. Our Front50 and Orca are using SQL. Our `spin` CLI is version `1.17.0`.
- We're using the webhook API to trigger pipelines, but unsure if this is a contributing factor.

apoorvmahajandd:
## Problem

- With the way orca acknowledges a message, there is a small window where orca can delete a fingerprint of a task message from the queue because it thinks it is safe to delete. But when you have stages that can redirect between tasks, this can lead to zombie executions. This can happen when such a message arrives between the commands where orca checks if it exists in the queue, and then deletes it.  This happens because these steps aren't atomic.
- Note: usually that window is very small, so this isn't something that happens frequently. And it only directly impacts those stages that have tasks that can redirect or loop between tasks.

### Queue Messages

* Every message on the Redis Queue has a **fingerprint** which is a hash of its content. `RunTask`'s fingerprint derives from `(executionType, executionId, application, stageId, taskId, taskType)`. 
* Two messages with identical content do not exist as two entries in Redis, they are one entry. This is deliberate; it is how the queue deduplicates.

### Redis layout

`RedisQueue` (`orca/keiko-redis/src/main/kotlin/com/netflix/spinnaker/q/redis/RedisQueue.kt`) uses:

| key | contents |
|---|---|
| `<q>.queue` (`queueKey`) | sorted set of fingerprints awaiting delivery, scored by delivery time |
| `<q>.unacked` (`unackedKey`) | sorted set of fingerprints being handled, scored by ack deadline |
| `<q>.messages` (`messagesKey`) | hash of fingerprint → JSON payload |
| `<q>.attempts`, `<q>.locks` | retry counters and per-fingerprint locks |

### Message lifecycle

`poll()` moves a fingerprint from `queue` to `unacked` and returns its payload. This step is atomic: it runs as the Lua script `READ_MESSAGE_WITH_LOCK_SRC`, executed server-side by Redis.

`QueueProcessor.callback` (`QueueProcessor.kt:82-99`) then runs the handler (RunTaskHandler for a RunTask message etc) and acknowledges afterwards:

```kotlin
handler.invoke(message)
ack.invoke()
```

### Acknowledging a message

Acknowledging has two branches, because a handler may legitimately re-push its own message.

* The if branch -  A `RunTask` for a task still `RUNNING` re-pushes its own fingerprint before returning. That is how polling tasks re-check every 15 seconds. So the ack must not delete a fingerprint that has been re-queued:

```kotlin
// RedisQueue.kt:284-298
private fun ackMessage(fingerprint: String) {
  pool.resource.use { redis ->
    if (redis.zismember(queueKey, fingerprint)) {   // CHECK — round-trip 1
      redis.multi {                                 // re-queued: clear only this delivery
        zrem(unackedKey, fingerprint)
        del("$locksKey:$fingerprint")
      }
    } else {
      redis.removeMessage(fingerprint)              // ACT — round-trip 2
    }
    fire(MessageAcknowledged)                       // a counter, not a log
  }
}
```

* Else branch: `removeMessage` (`:321-329`) is unconditional and total:

```kotlin
multi {
  zrem(queueKey, fingerprint)     // removes a queue entry it never examined
  zrem(unackedKey, fingerprint)
  hdel(messagesKey, fingerprint)  // and the payload, making the entry unrecoverable
  del("$locksKey:$fingerprint")
  hdel(attemptsKey, fingerprint)
}
```

Neither branch logs. `fire(MessageAcknowledged)` increments a counter, so a message removed this way leaves no trace in the logs at all.

### Looping Tasks or the ones with the REDIRECT status

`REDIRECT` means "loop back and re-run this stage from an earlier task". `CompleteTaskHandler.kt:103-104` routes it to `handleRedirect()` (`:179-190`):

```kotlin
private fun StageExecution.handleRedirect() {
  tasks.let { tasks ->
    val start = tasks.indexOfFirst { it.isLoopStart }
    val end = tasks.indexOfLast { it.isLoopEnd }
    tasks[start..end].forEach { it.endTime = null; it.status = NOT_STARTED }
    repository.storeStage(this)
    queue.push(StartTask(execution.type, execution.id, execution.application, id, tasks[start].id))
  }
}
```

The cascade is deterministic, and its final step regenerates the fingerprint that just retired:

```
CompleteTask(t3, REDIRECT)
  -> handleRedirect(): reset t2..t3 to NOT_STARTED, push StartTask(t2)
     -> StartTask(t2)  -> push RunTask(t2)
        -> RunTask(t2) -> push CompleteTask(t2, SUCCEEDED)
           -> CompleteTask(t2) -> nextTask(t2) == t3 -> push StartTask(t3)
              -> StartTask(t3) -> push RunTask(t3)      <-- StartTaskHandler.kt:76
```

Because `RunTask`'s fingerprint covers only `(executionType, executionId, application, stageId, taskId, taskType)`, that final `RunTask(t3)` is byte-identical to the `RunTask(t3)` that returned `REDIRECT`. Same content, same fingerprint, F.

`REDIRECT` is also the **only** `RunTask` outcome in which the handler does not re-push its own fingerprint - it pushes `CompleteTask` instead. So it is the only case where a `RunTask`'s a**ck takes the destructive branch**. That combination is what makes the defect reachable.

##  Why is this defect rare?

* Given that the destructive branch runs on nearly every message, the collision alone is not sufficient. The replacement message must arrive inside a window that is normally microseconds wide, whereas the restart that produces it takes around 370 ms on average - **so the acknowledgement has to stall for roughly as long as the restart takes**.  Four conditions must coincide:

1. **A new message must reuse a retired fingerprint.** Normally impossible: a pipeline moves forward, so the next message has a different `taskId` or `stageId`. `REDIRECT` is the exception because it loops backwards, regenerating a fingerprint that has just been retired.
2. **The push must land inside the check→act window.** Normally two consecutive round-trips on an already-borrowed connection — microseconds. The redirect cascade takes ~370 ms (six queue hops, each gated by the 50 ms poll interval), so the ack normally finishes with a very large margin. A fast cascade cannot cause this; the cascade cannot approach a microsecond window.
3. **Therefore the ack must stall by roughly the cascade duration**, precisely between its read and its write.
4. **The lost message must be unrecoverable.** A `RunTask` for a task the database calls `RUNNING` is terminal — nothing regenerates it. Other message types may be re-pushed by later events, so the race can fire harmlessly.

## Fix

Move both the decision and the removal into `ACK_MESSAGE_SRC`, executed server-side by Redis as one indivisible step.

**Why Lua and not Kotlin.** This is about *where* the code runs, not language preference — the Kotlin method remains and delegates via `evalsha`. Alternatives considered:

- **`MULTI`/`EXEC`** cannot express this. It queues commands and cannot branch on a value read mid-transaction — which is exactly why the `zismember` sat *outside* the existing `multi` block.
- **`WATCH`/`EXEC`** could, but `queueKey` is the hottest key in the system (every push and poll from every pod writes it), so an optimistic retry loop there would thrash.

Decisively, **`poll()` has always worked this way** via `READ_MESSAGE_WITH_LOCK_SRC` — a Lua script embedded as a `const val` in the same file, loaded by the same `cacheScript()`, invoked with the same `NOSCRIPT` re-cache fallback. The poll side of the message lifecycle was already atomic; the ack side wasn't, and that asymmetry was the defect.

Two details:
- The script tests presence with `ZRANK` rather than rank — it returns `0` for the first member, which is truthy in Lua, while a missing member comes back as `false`.
- The extension is named `ackMessageAtomically`, not an `ackMessage` overload: inside a member extension, an unqualified recursive `ackMessage(fingerprint)` can bind to the private member instead of the extension receiver, recursing through `pool.resource`.

## Tests

Commit 1 is the failing reproduction, commit 2 makes it pass.

Three tests model the incident's two pods against a real Redis — `podA` polls and acks (the `REDIRECT` ack), `podB` pushes the identical successor:

| test | before | after |
|---|---|---|
| `ack of a finished message must not delete an identical message pushed during the ack` | **FAIL** (actual `[]`) | PASS |
| `ack keeps a message the handler re-pushed before acking` | PASS | PASS |
| `ack of a finished message cleans up completely` | PASS | PASS |

Two design choices that matter:
- The colliding push is injected after whichever command the ack issues **first**, not after `ZRANK` specifically. Anchoring on `ZRANK` would make the test pass vacuously once that call disappeared — green for the wrong reason.
- Assertions are on **observable queue state**, not Redis commands, so they survived the implementation change without edits. An interaction-based test would have broken on the fix rather than passing.
- The third test guards the opposite direction: it stops a "fix" that simply stops deleting and leaks every completed payload.

## Known limitation

`RedisClusterQueue.ackMessage` (`RedisClusterQueue.kt:374-385`) has the **identical** defect. Not fixed here: a Lua fix needs all five keys in one hash slot, which is a real design question rather than a mechanical port. Prod uses `RedisQueue`.


jasonmcintosh:
OH MAN.  I FIGURED It was someplace in here but could NEVER find it - something on the transition handling where the message was checked out moving to the next location.  THIS IS AWESOME!!!!

jasonmcintosh:
Just gotta fix imports looks like ;) 
```  
e: file:///home/runner/work/spinnaker/spinnaker/orca/keiko-redis/src/main/kotlin/com/netflix/spinnaker/q/redis/RedisQueue.kt:319:16 Unresolved reference: ScriptingCommands

e: file:///home/runner/work/spinnaker/spinnaker/orca/keiko-redis/src/main/kotlin/com/netflix/spinnaker/q/redis/RedisQueue.kt:327:15 Unresolved reference: ScriptingCommands
Task :orca:keiko-redis:compileKotlin FAILED
```

jasonmcintosh (approved):

PR #7922, opened 2026-08-25
chore(aws): Migrate secrets and config server off of v1 sdk.  Also remove "dead" bastion configurations

jasonmcintosh:
Bastion config though interesting used IDMSv1 based operations, binaries not avialable by default and appears to be mostly dead code.  This removes this.  I can restore if we think it's REALLY requested by end users, but there are other alternative options including plugins that would better serve this need IMHO.


christosarvanitis (approved):
Nice cleanup of dead code

PR #7933, opened 2026-08-28
chore(aws): Migrate rest of services to SDKv2

jasonmcintosh:
Migrates AsgService, SecurityGroupService, NetworkInterfaceService, and LaunchTemplateService along with every write-path consumer across deploy/asg/, deploy/ops/ (including securitygroup/, discovery/, actions/), and deploy/handlers/BasicAmazonDeployHandler.groovy off AWS SDK v1 onto v2. Originally scoped as separate PR3-PR6 but combined into one PR since these services return v1 model types directly to 100%-unmigrated consumers, making them structurally inseparable.

Also fixes three latent bugs surfaced only when tests were run (Groovy's dynamic typing let them silently compile): a LifecycleState enum-vs-String comparison in AbstractEnableDisableAtomicOperation, a nonexistent v1-style .getAutoScalingGroups() call in ResizeAsgAtomicOperation, and rewires BasicAmazonDeployHandlerUnitSpec (missed from earlier tracking) to v2 types.

---
NOTE, once this is merged along with the other PRs, there'll be some cleanup to do - e..g remove edda and some various libs that are no longer used, but really want all of them merged before the final cleanup


clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/deploy/ops/loadbalancer/IngressLoadBalancerBuilder.java
```diff
@@ -119,26 +120,32 @@ private SecurityGroupLookupFactory.SecurityGroupUpdater getOrCreateSecurityGroup
 
   private void filterOutExistingPermissions(
       List<IpPermission> permissionsToAdd, List<IpPermission> existingPermissions) {
-    permissionsToAdd.forEach(
```
jasonmcintosh, line 122: Double check this spot for sure

jasonmcintosh:
Last note - there ARE more changes, this is one of the biggest ones, but there's some additional cleanup incoming.

clouddriver/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/deploy/ops/securitygroup/UpsertSecurityGroupAtomicOperation.groovy
```diff
@@ -135,8 +135,8 @@ class UpsertSecurityGroupAtomicOperation implements AtomicOperation<Void> {
         //Update tags to ensure they are consistent with rule changes
         securityGroupUpdater.updateTags(description, dynamicConfigService)
         task.updateStatus BASE_PHASE, status
-      } catch (AmazonServiceException e) {
-        task.updateStatus BASE_PHASE, "Error updating ingress to '${description.name}' - ${e.errorMessage}"
+      } catch (Ec2Exception e) {
```
christosarvanitis: This is not the equivalent of AmazonServiceException in v2. Wondering if it should be left to the base class AmazonServiceException

christosarvanitis (dismissed):
Big one but its direct migration of v1 to v2. Minor comment - not blocking as the ec2exception extents the base class.

christosarvanitis (approved):
Looks good!

PR #7943, opened 2026-08-31
feat(azure): add instance actions and fix server group action menu

shanman190:
Azure instances had no **Instance Actions** menu in Deck, and `clouddriver-azure` had no instance-level atomic operations at all. This adds the three operations orca's provider-agnostic instance stages already expect, plus the Deck wiring to invoke them. Orca needs no changes.

## clouddriver-azure

New `resources/instance` package with `rebootInstances`, `terminateInstances`, and `terminateInstanceAndDecrementServerGroup`, each with a converter and validator.

`AzureComputeClient` resolves scale set VM names to the instance indexes the VMSS APIs require — orca addresses instances by name, every Azure API by index. An unresolved name aborts the operation rather than being skipped, so a terminate cannot report success having done nothing.

**Terminate uses VMSS `reimage()`, not delete.** Azure has no "terminate but keep capacity" primitive: deleting a VMSS instance always decrements capacity. Reimage replaces the VM in place in one call with no race. This mirrors `clouddriver-google`, which implements the same action via a managed instance group's `recreateInstances`; orca detects completion by comparing `launchTime`, a path that already exists for that model. `terminateInstanceAndDecrementServerGroup` deletes and lets capacity decrement, which is Azure's native behavior.

### Behavior change worth reviewer attention

`AzureInstanceDescription` implements `ServerGroupsNameable`, so Fiat now performs the account and `APPLICATION/WRITE` checks on these operations. Without it `DescriptionAuthorizerService` skips both checks entirely, leaving a per-instance reimage/delete path unauthorized. Callers lacking `APPLICATION/WRITE` on the target application will now be denied where they previously were not checked. Follows the existing precedent in `ResizeAzureServerGroupDescription`.

## deck

New `AzureInstanceWriter`, because core's `InstanceWriter` cannot be used as-is:

- `terminateInstance` never sets `cloudProvider`, so orca defaults it to `aws`
- neither `terminateInstance` nor `rebootInstance` sends `serverGroupName`; clouddriver cannot locate the scale set without it, and Azure instances are not searchable, so orca's fallback would report success having done nothing
- `terminateInstanceAndShrinkServerGroup` reads `serverGroup.asg.minSize`/`maxSize`, and Azure server groups have no `asg` field — they expose `capacity`

The instance details controller populates `id` and `serverGroup`, which `AzureInstance` does not carry (its `Instance` model has no `getId()`), and drops the unreachable load balancer and discovery actions no Azure operation can serve — VMSS backend-pool membership lives on the scale set's NIC config, not per instance, and discovery is Eureka-specific.

## Server group action menu

Brings the Azure menu in line with AWS's guards: a divider after Rollback, the Rollback condition moved from the `<a>` onto the `<li>` so a suppressed entry leaves no blank row, Disable only when enabled, and Enable only when there are disabled instances and no resize is in flight. Rollback now also appears on a disabled server group when an enabled sibling exists in the same cluster/region/account, matching AWS.

## Testing

- `clouddriver-azure`: 11 new tests, all passing. Suite at 209 tests; the 26 failures are pre-existing and confined to six unrelated template/description specs.
- `clouddriver-web:installDist` succeeds, confirming the six new Spring beans register without collision.
- deck: 15 new tests across three suites, all passing (full deck suite 1761 passed, 0 failures).


shanman190:
Opened against the wrong repository by mistake; this belongs on an internal fork. Closing immediately.

PR #7942, opened 2026-08-31
chore(aws): migrate remaining v1 EC2/AutoScaling client calls to SDK v2

jasonmcintosh:
## Summary
- Migrates the last real callers of `AmazonClientProvider`'s v1 EC2 and AutoScaling accessors to their v2 equivalents: `DefaultAWSAccountInfoLookup` (account/region bootstrapping), `AmazonHealthIndicator`, `CleanupDetachedInstancesAgent`, `ReconcileClassicLinkSecurityGroupsAgent` (the full ClassicLink API surface — verified present in the v2 EC2 SDK, this is a real migration, not a feature-removal decision), `AmazonInstanceProvider`'s console-output call, and the ASG "scaling activities" feature (`LaunchFailureNotificationCleanupAgent`, `AmazonClusterController`) that an earlier migration PR explicitly deferred.
- Adds a `getAmazonEC2V2(AWSCredentialsProvider, region)` overload to `AmazonClientProvider` for `DefaultAWSAccountInfoLookup`, which doesn't have a `NetflixAmazonCredentials` to build against; deletes the now-fully-dead v1 `getAmazonEC2`/`getAutoScaling` accessor families and the classic `getAmazonElasticLoadBalancing` accessor (confirmed zero real callers anywhere, including tests).
- Drops the `com.amazonaws:aws-java-sdk-sns`/`-sqs`/`-sts` declarations from `kork-aws` — confirmed via a full per-file check that every SNS/SQS consumer (`kork-pubsub-aws`, `echo-pubsub-aws`, clouddriver-aws's own lifecycle workers) is already 100% v2, so these had zero real consumers regardless of anything else in this change.

**Note on scope**: `AwsSdkClientSupplier`, the v1 `RequestHandler2` impls, `AWSProxy`'s v1 method, the v1 `RetryPolicy` plumbing, and kork-aws's retry/backoff/metrics classes remain in place — they're still load-bearing for several other v1 accessors (CloudWatch, IAM, Shield, ELBv2, etc.) that are provably dead but whose removal is already owned by the separate, still-open Edda removal PR (#7941). Deleting them here would either break compilation or duplicate that PR's work on a second branch. Expect a normal reconciling merge conflict on `AmazonClientProvider.java` whichever of the two PRs lands second.

## Test plan
[...]

PR #7941, opened 2026-08-31
chore(aws): remove Edda dynamic-proxy client and credential/config plumbing

jasonmcintosh:
## Summary
- Removes Edda (Netflix's AWS resource read-through cache) support from clouddriver-aws entirely, per the decision to drop it for the upcoming release. Every non-Edda account already took the direct-AWS-SDK-v2 code path in the load-balancer/security-group/ASG caching agents, so this deletes the now-permanently-dead `useEdda`/`eddaEnabled` branches and the Edda-only files (`EddaApi`, `EddaApiFactory`, `EddaConfiguration`, `EddaLoadBalancerCachingAgent`, Edda-specific model DTOs) they backed.
- Rewrites `AmazonClientProvider`'s remaining v1 accessors to call the SDK directly instead of routing through the JDK dynamic-proxy layer (`ProxyHandlerBuilder`/`AmazonClientInvocationHandler`), then deletes that proxy machinery along with `EddaTemplater`/`EddaTimeoutConfig`/`EddaException`, and drops the `edda`/`eddaEnabled` fields and templating config from `NetflixAmazonCredentials`, `AccountsConfiguration`, `CredentialsConfig`, and `AmazonCredentialsParser`.
- Cleans up now-fully-dead `getAmazonEcs`/`getIam`/`getAmazonCloudWatch`/`getAmazonShield`/etc. v1 accessors that lost their only real callers earlier in the broader v1→v2 migration, the now-unused `httpClient`/`maxConnections` Builder settings, and the vestigial `skipEdda` naming in `SecurityGroupLookupFactory`.
- Drops the dead `com.amazonaws:aws-java-sdk` (v1) dependency declaration from `clouddriver-titus` and `clouddriver-ecs` — neither module had any real usage left (the one reference in `clouddriver-ecs` was dead test setup, never wired into the code under test).
- Small follow-up touches outside `clouddriver-aws` with no other owner: `clouddriver-ecs`'s `EcsAccountBuilder`, a stale Edda comment in orca's `WaitForUpInstancesTask`, deck Cypress fixtures, and the `clouddriver.yml` sample config.

Builds on the four independent SDK v2 migration PRs (#7936, #7937, #7938, #7939) that have already merged.

## Test plan
- [x] `./gradlew :clouddriver:clouddriver-aws:compileJava :clouddriver:clouddriver-aws:compileGroovy :clouddriver:clouddriver-aws:compileTestGroovy :clouddriver:clouddriver-aws:compileIntegrationJava`
[...]

PR #7944, opened 2026-09-01
chore(aws): Flip credentials to AWS SDK v2 and remove aws-java-sdk (v1) entirely

jasonmcintosh:
## Summary

Final PR in the AWS SDK v1→v2 migration series. Depends on #7941 and #7942 landing first (this branch is built on a local merge of both, since neither has merged yet) — do not merge before those two.

- Flips `AmazonCredentials`' stored credentials provider from v1 `AWSCredentialsProvider` to v2 `AwsCredentialsProvider`, propagating through `NetflixAmazonCredentials`, `AssumeRoleAmazonCredentials`, the credentials parser/loader/factories, and `DefaultAWSAccountInfoLookup`. Fixes a double STS-assume-role hop along the way (`AssumeRoleAmazonCredentials` previously bridged v1→v2 lazily on top of an already-v1-assumed session).
- Migrates the remaining v1-typed view providers (`AmazonSecurityGroupProvider`, `AmazonSubnetProvider`, `AmazonVpcProvider`) and model classes (`AmazonSecurityGroup`, `AmazonServerGroup`, `AmazonClusterProvider`, `AmazonImage`) off v1 EC2/AutoScaling types.
- Deletes `AwsSdkClientSupplier` and the v1 `RequestHandler2` family it fed, now that every `AmazonClientProvider` accessor builds v2 clients directly. Deletes kork-aws's `InstrumentedRetryCondition`/`InstrumentedBackoffStrategy`/`AwsMetricsSupport` (these only ever added Spectator metrics on top of default v1 retry behavior, never customized it).
- Removes the last `com.amazonaws`/`aws-java-sdk` references from the build entirely (verified via repo-wide grep sweep).

### Real bugs found and fixed along the way
- `amazonObjectMapper` was never wired to serialize AWS SDK v2 model objects, silently breaking Jackson support for several already-migrated caching agents.
- `AmazonClientProvider.DEFAULT_REGION` (null, meant to resolve dynamically) would NPE against the v2 client supplier; now falls back through the SDK's default region provider chain.
- v2's NTLM proxy support (`ntlmDomain`/`ntlmWorkstation`) was never wired despite the SDK supporting it — v1's `AWSProxy#apply()` set these but the v2 proxy-translation code never carried them over.
- AWS SDK v2 client metrics (call latency, retry counts) are now wired automatically for every v2 client via the SDK's classpath-based global-interceptor discovery mechanism (`software/amazon/awssdk/global/handlers/execution.interceptors`), matching v1's `AwsSdkMetrics.setMetricCollector` global hook — previously this required per-client opt-in.

### Restored rather than dropped
- `LogEndpointRequestHandler` (v1) → `LogEndpointExecutionInterceptor` (v2), wired back through `AmazonClientProvider.Builder#logEndpoints()`.

## Test plan
- [x] `./gradlew :clouddriver:clouddriver-aws:test` (1458/1458)
- [x] `./gradlew :clouddriver:clouddriver-aws:integrationTest` (30/30)
- [x] `./gradlew :kork:kork-aws:test`, `:front50:front50-s3:test`, `:echo:echo-pubsub-aws:test`, `:kork:kork-pubsub-aws:test` all green
- [x] `:clouddriver:clouddriver-artifacts:clouddriver-artifacts-s3:compileTestJava` green (its own test suite requires Docker/LocalStack, unavailable in this environment — pre-existing, unrelated to this change)
- [x] Repo-wide grep sweep for `com.amazonaws`/`aws-java-sdk` returns zero real code/build references
- [x] New unit test proves the classpath global-interceptor discovery actually works (`GlobalExecutionInterceptorTest`), not just that it compiles


christosarvanitis:
First pass looks ok, will wait for the merge of the depended PRs for another review

clouddriver/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonServerGroup.groovy
```diff
@@ -126,22 +125,23 @@ class AmazonServerGroup implements ServerGroup, Serializable {
       securityGroups = (Set<String>) launchConfig.securityGroups
     }
 
-    RequestLaunchTemplateData launchTemplateData = null
+    Map<String, Object> launchTemplateData = null
```
jasonmcintosh, line 128: I'll note the drop of the specific type.  I can try to find the concrete data type... but wasn't sure the "right" way to handle this based upon the objects...

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/RateLimitingRequestHandler.java
```diff
@@ -1,57 +0,0 @@
...
-
-import static java.util.Objects.requireNonNull;
-
-import com.amazonaws.Request;
-import com.amazonaws.handlers.RequestHandler2;
-import com.google.common.util.concurrent.RateLimiter;
-import com.netflix.spectator.api.Counter;
-import java.util.Objects;
-
-/** A RequestHandler that will throttle requests via the supplied RateLimiter. */
```
jasonmcintosh, line 27: Checking this one... worried this may be... an unintentional removal

jasonmcintosh: Ok RateLimitingExecutionInterceptor should handle this now.

clouddriver/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
```diff
@@ -44,10 +43,10 @@ class AmazonImageProviderSpec extends Specification {
     then:
     AmazonImage expectedImage = new AmazonImage()
     expectedImage.setRegion("eu-west-1")
-    expectedImage.setImage(new Image())
-    expectedImage.image.setImageId("ami-123321")
-    expectedImage.image.setName("some_ami")
-    expectedImage.image.setOwnerId("1233211233231")
+    expectedImage.setImageAttribute("imageId", "ami-123321")
```
jasonmcintosh, line 46: To note, there's... some concern on this one from a migration perspective.  Confirming the behavior on this change.

clouddriver/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/model/AmazonServerGroup.groovy
jasonmcintosh: FYI from AI:
---
I'd recommend keeping the map-based design. Here's why:

This predates the SDK v2 migration by years — it's not something we introduced. AmazonServerGroup.groovy has stored image, launchConfig, launchTemplate, and asg as Map<String, Object> since 2014 (image) and 2020 (launchTemplate, added in #7852, git log confirms). The v1→v2 migration only changed what populates the map (SdkPojoSerializer converting v2 SdkPojo objects instead of default Jackson bean introspection converting v1 objects) — it didn't change the shape of the model.

What's cached and read back is ResponseLaunchTemplateData — the type LaunchTemplateVersion.launchTemplateData() actually returns (see CopyLastAsgAtomicOperation.groovy:172, PrepareModifyServerGroupLaunchTemplate.java:106). Even a "go back to a concrete class" change would need to target the read-side type.

And even ResponseLaunchTemplateData wouldn't be a 1:1 fit. AmazonServerGroup.launchTemplate isn't actually a serialized LaunchTemplateData at all — AmazonLaunchTemplateCachingAgent.java:115-121 builds it from a v2 LaunchTemplate object (the template metadata — id, name, latestVersionNumber — a different type than LaunchTemplateData), then bolts on synthetic keys (application, latestVersion, versions) that don't exist on any single AWS SDK class. There's no concrete type that matches what's actually stored — it's a hand-assembled composite by design.

Why the map design exists at all: cats caching agents store data as generic JSON documents in Redis/SQL, and the REST-facing model classes (AmazonServerGroup, AmazonImage) intentionally decouple the cache/API schema from any specific AWS SDK version's type shape. Pinning AmazonServerGroup.launchTemplate to a concrete SDK type would re-couple the cache format to that SDK's class shape — exactly the kind of coupling that made this v1→v2 migration painful in the places it did exist (e.g. AmazonImage's old @JsonUnwrapped com.amazonaws.services.ec2.model.Image, which we just removed for this same reason). It would also mean a future v3 SDK migration (or a field rename within v2) breaks REST API consumers and cache deserialization directly, instead of being absorbed by the flat-map translation layer.

So: nothing to fix here — the map design is intentional, older than this PR, and going back to a concrete class would reintroduce the coupling problem this pattern is meant to avoid.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/config/AmazonCredentialsParser.java
```diff
@@ -122,7 +122,7 @@ public AmazonCredentialsParser(
     this.awsConfigurationProperties = Objects.requireNonNull(awsConfigurationProperties);
 
     // look in the credentials config to find default region names
-    if (!CollectionUtils.isNullOrEmpty(credentialsConfig.getDefaultRegions())) {
+    if (!CollectionUtils.isEmpty(credentialsConfig.getDefaultRegions())) {
```
christosarvanitis, line 125: Reason for removing the null check?

clouddriver/clouddriver-aws/src/test/groovy/com/netflix/spinnaker/clouddriver/aws/provider/view/AmazonImageProviderSpec.groovy
jasonmcintosh: Similar to other analysis... using a map is recommended here due to the type of object.

clouddriver/clouddriver-aws/src/main/java/com/netflix/spinnaker/clouddriver/aws/security/config/AmazonCredentialsParser.java
jasonmcintosh: Different collection utils - spring vs. the aws.  isEmpty does a nullcheck internally as part of how it works.

clouddriver/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/config/AwsConfiguration.groovy
```diff
@@ -85,14 +84,10 @@ import org.springframework.core.annotation.Order
 class AwsConfiguration {
 
   @Bean
-  AmazonClientProvider amazonClientProvider(AwsConfigurationProperties awsConfigurationProperties, RetryCondition instrumentedRetryCondition, BackoffStrategy instrumentedBackoffStrategy, AWSProxy proxy, ServiceLimitConfiguration serviceLimitConfiguration, Registry registry) {
+  AmazonClientProvider amazonClientProvider(AwsConfigurationProperties awsConfigurationProperties, AWSProxy proxy, ServiceLimitConfiguration serviceLimitConfiguration, Registry registry) {
```
christosarvanitis, line 87: Trying to follow where the Retry config has landed :)

jasonmcintosh: https://github.com/spinnaker/spinnaker/pull/7944/changes/BASE..e131c8d0397b60abc6db5e1fe3f3a1e4ea81eca1#diff-72cf769b69da98d9c05ecf538ab71a666f5351b79b57441f9032f69c7e0bd4d8R113 here ;)

jasonmcintosh: TO NOTE< and I recall this when I last was fiddling with this on lambda, the retry config was always... not entirely "accurate":
---
What's genuinely different, not just relocated: v1's RetryPolicy also carried a custom RetryCondition/BackoffStrategy (InstrumentedRetryCondition/InstrumentedBackoffStrategy, deleted from kork-aws in this PR). We traced that earlier in this migration and confirmed AmazonClientProvider's old buildV2RetryPolicy() only ever read .getMaxErrorRetry() off that v1 object — the condition/backoff strategy never actually influenced retry decisions, they only added Spectator metrics recording around whatever decision the v1 SDK's default retry logic made. So no retry-decision behavior was lost; the v2 clients now just use the SDK's own default RetryCondition/BackoffStrategy for numRetries beyond what's overridden, same as they always effectively did, and the metrics side of it is now covered by the global SpectatorExecutionInterceptor classpath wiring we discussed earlier instead of a bespoke retry-condition wrapper.

jasonmcintosh: TO NOTE... this is one of those... the SDK operates just a BIT differently on some of these... so I'd suspect the spectator v2 metrics stuff may show some of these metrics a bit differently as a result.

christosarvanitis (approved):
🚀 lets go for it

PR #7945, opened 2026-09-01
fix(front50): filter pipelines by application, not pipeline name, in PostFilter

### LINKED ISSUE #7940: Front50 PipelineController restricted filter checks pipeline name instead of application
> *This was generated by AI during triage.*

## Summary

`front50-web` `PipelineController.list()` and `getTriggeredPipelines()` use `@PostFilter("hasPermission(filterObject.name, 'APPLICATION', 'READ')")` when `restricted=true` (the default). `filterObject` is a `Pipeline` whose `.name` is the pipeline name, not the owning application (`Pipeline.application`). This filters by the wrong principal and can expose or hide pipelines incorrectly when Fiat is enabled.

Related Deck work discovered this while auditing project pipeline tag discovery (PR #7934, issue #7935): the Deck fallback was routed to the RBAC-filtered `/executions` endpoint as a minimal fix, but the Front50 restricted list remains incorrect.

## Evidence

- `front50/front50-web/src/main/java/com/netflix/spinnaker/front50/controllers/PipelineController.java:107-108`:
  ```java
  @PostFilter("#restricted ? hasPermission(filterObject.name, 'APPLICATION', 'READ') : true")
  @RequestMapping(value = "", method = RequestMethod.GET)
  public Collection<Pipeline> list(@RequestParam(value = "restricted", defaultValue = "true") boolean restricted, ...)
  ```
- Same pattern at `162-163` for `getTriggeredPipelines()`.
- `Pipeline` model (`front50/front50-api/.../Pipeline.java`) has `String name` (pipeline name) and `String application` (owning app); the filter should use `filterObject.application`.
- Other controllers in the same module correctly filter on `filterObject.application` (e.g., `StrategyController`, `DeliveryController`, `SnapshotsController`).

## Proposed fix

- Change both `@PostFilter` expressions to `hasPermission(filterObject.application, 'APPLICATION', 'READ')`.
- Add a method-security test covering `restricted=true` vs `restricted=false` (e.g., extend `PipelineControllerSpec`/`PipelineControllerTck`) asserting that a pipeline whose `application` the user cannot READ is filtered out, even if its `name` coincidentally matches an allowed application.
- Consider whether `/pipelineConfigs` (Gate `PipelineConfigController.getAllPipelineConfigs()`) and project pipeline endpoints (`Gate ProjectController`, `Orca ProjectController`) need explicit Fiat checks or should continue relying on Front50's filtering; track separately if out of scope.

## Scope

Backend only, no Deck changes. Minimal diff: two annotation fixes + test. No schema/persistence/dependency changes.

## Related

- PR #7934 (Deck project pipeline tag discovery)
- Issue #7935

*AI disclosure: this issue was drafted with LLM assistance (muse-spark-1.2 / opencode) and reviewed by the contributor.*

KaioAntonio:
Fixes #7940

## Root cause

`PipelineController.list()` and `getTriggeredPipelines()` use:

```java
@PostFilter("#restricted ? hasPermission(filterObject.name, 'APPLICATION', 'READ') : true")
```

Permission in Spinnaker is granted per-application via Fiat, but `filterObject` here is a `Pipeline`, and `Pipeline.name` is the pipeline's own name, not the owning application (`Pipeline.application`). Checking permission against the wrong field means the restricted list filters on the wrong principal: it can expose pipelines the caller has no `READ` access to, or hide ones they do, depending on how pipeline names happen to collide with application names the caller can or cannot see.

Other controllers in the same module already filter correctly, e.g. `StrategyController`:

```java
@PostFilter("hasPermission(filterObject.application, 'APPLICATION', 'READ')")
```

## Fix

Changed both `@PostFilter` expressions in `PipelineController` to use `filterObject.application` instead of `filterObject.name`, matching the existing correct pattern elsewhere in the module. Kept the `#restricted ?` ternary as-is since only these two endpoints expose that toggle.

## Testing

- Added a test asserting the `@PostFilter` SpEL on both `list()` and `getTriggeredPipelines()` references `filterObject.application` and not `filterObject.name`. This was needed because the existing `PipelineControllerSpec` runs with `@AutoConfigureMockMvc(addFilters = false)` and a test context that doesn't enable Spring method security (no `@EnableMethodSecurity` / Fiat authentication config wired in), so the `@PostFilter` annotation itself was previously unexercised by any test — a typo in the SpEL string like this one would not have been caught by the existing suite regardless of `restricted` value.
- `./gradlew :front50:front50-web:test` passes (all existing `PipelineControllerSpec` tests plus the new one).


front50/front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerSpec.groovy
```diff
@@ -334,6 +334,27 @@ class PipelineControllerSpec extends Specification {
     (all.get(0).get().status == 200 && all.get(1).get().status == 400) || (all.get(0).get().status == 400 && all.get(1).get().status == 200)
   }
 
+  @Unroll
```
js4683, line 337: **Nit comment:** The test only reflects the annotation and checks string contents. It does not execute either endpoint through Spring method security or verify that an unauthorized application is filtered out. A runtime security-wiring or SpEL problem could therefore pass.

christosarvanitis: I will echo @js4683 NIT. ideally some mock tests that demonstrate the response from the endpoints

christosarvanitis:
Nice catch and wondering how it got missed so far :) Possible candidate for backports on the active releases

KaioAntonio:
@js4683 @christosarvanitis Fair point, thanks for catching that. Pushed `83d34f4`: `PipelineControllerSecuritySpec`, a separate Spring context with `@EnableMethodSecurity` actually enabled and a `MethodSecurityExpressionHandler` wired to a mocked `FiatPermissionEvaluator`, driving `list()` through the real container-managed, AOP-proxied controller via `MockMvc`. It sets up two pipelines with swapped `name`/`application` values so filtering on the wrong field would let the wrong one through, and asserts only the permitted application's pipeline appears in the response.

Kept it as its own spec/context rather than folding into `PipelineControllerSpec`, since that spec's autowired `MockMvc` also exercises `save()`, whose `@PreAuthorize` would newly be enforced once method security is turned on there, and those tests don't stub `fiatPermissionEvaluator` or set up an `Authentication` for that.

`./gradlew :front50:front50-web:test` passes, including all of the existing `PipelineControllerSpec` tests unaffected.

front50/front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerSecuritySpec.groovy
jasonmcintosh, line 1: NOTE:  It'd be better if these were NOT in groovy.  Given other tests are in groovy... I'd be "ok" for now, but soonish going to start working on mass migrations to java and be nice to have it in java initially.

front50/front50-web/src/test/groovy/com/netflix/spinnaker/front50/controllers/PipelineControllerSecuritySpec.groovy
```diff
@@ -0,0 +1,146 @@
+/*
+ * Copyright 2026 Netflix, Inc.
```
jasonmcintosh, line 2: Not sure this is the right license ;) It's a "default" due to hooks but... if wanted to change it can

jasonmcintosh (approved):
This has my approval with some "nice-to-haves"

christosarvanitis (approved):
Thanks for the tests! Looks good

jasonmcintosh:
Waiting on antoher big PR to merge, then will get this in & backported

jasonmcintosh:
@mergifyio backport release-2026.2.x release-2026.1.x

PR #7950, opened 2026-09-01
chore(builds): Add a build cache to reduce some of the build times

jasonmcintosh:
Also reduces the CPUs from 4 to whatever the OS has available.  SHOULD actually increase on local tests if you're running a bigger mac.


christosarvanitis (approved):
🤞

PR #7951, opened 2026-09-01
feat(builds): Parallelize the tests to run faster

jasonmcintosh:


jasonmcintosh:
SHOULD have realized - *-web tests... are heavily dependent upon the various resources in the cloud providers to work.  Looking at options - possibly on some of these adding dependson or similar to only load if specific cloud providers are enabled.

PR #7953, opened 2026-09-02
feat(orca): inline Armory Run Script preconfigured job

shlomodaari:
## Summary
- Inline the Armory Run Script plugin (`runContainerScript`) into `orca-clouddriver` as a PreconfiguredJobConfigurationProvider, matching prior donations such as Run Multiple Pipelines and Evaluate Artifacts.
- Kubernetes Job YAML plus optional overrides via `job.preconfigured.run-script.*`.
- Unit tests cover YAML load and config overrides.

## Test plan
- [x] `:orca:orca-clouddriver:test --tests com.netflix.spinnaker.orca.clouddriver.config.RunScriptPreconfiguredJobProviderTest`
- [ ] Confirm Deck still shows the `runContainerScript` preconfigured job
- [ ] Confirm `job.preconfigured.run-script.enabled=false` hides the stage


jasonmcintosh:
I'm... somewhat iffy on this..  It's using an init container built off of:
https://github.com/ethanfrogers/summit-2019-demo/tree/master/jobs/fetch-artifact
which is ANCIENT and no longer supported.  It also is set to try to use as examples the "artifacts" endpoint - which isn't a bad thing (see https://github.com/ethanfrogers/summit-2019-demo/blob/master/jobs/fetch-artifact/fetch) to pull down a TGZ via this path.  That's essentially how bake operations work.

BUT this is very old docker images, using an endpoint that requires auth OR an internal API (not good).  It's a TOUCH more convoluted to really use/support this IMHO as a result.  I'd like to rethink the way this works


Some ideas:
* API token - gate already provides this.  Generate an API token & inject it into the job for a given account.  THEN can use the gate API endpoint to allow fetching this data from clouddriver's cache
* Docker:  For SURE would need to NOT support the current one but maybe an inline script based upon a busybox:latest or similar.  Shouldn't be TOO hard. 

I'd also want to think a TOUCH more about security aspects - arbitrary scripts should NEVER run from the internal - so even on the API token type approach to allow ad-hoc jobs to communicate with the spinnaker APIs, how do you restrict the permissions?  Probably define the service account as well for these operations?  ETC.

shlomodaari:
Thanks - the fetch image, Clouddriver-internal call, and permissions are the original Armory plugin. Agreed that’s a weak fit for core as-is. We’ll decide internally whether to redesign or close this out, and follow up here.

PR #7964, opened 2026-09-04
fix(ecs): populate imageId when scoping docker image find by account

jasonmcintosh:
## Summary
- The account-scoped docker image lookup added by #7746 (`updateDockerRegistryAccount` in `Container.tsx`/`TaskDefinition.tsx`) cast the raw `/images/find` response directly to `IEcsDockerImage`, but that response never included an `imageId` field (only `account`/`registry`/`repository`/`tag`/`artifact`).
- Every dropdown option ended up with the same `undefined` label/value, so the ECS create/clone server group "Container Image" picker showed only `(undefined)` entries and selecting one had no effect.
- Extracted the existing `imageId` derivation logic from `EcsCloneServerGroupModal` into a shared `dockerImage.util.ts` and reused it in both `Container.tsx` and `TaskDefinition.tsx` so images fetched via the account picker get a proper `imageId`.

## Test plan
- [x] `./gradlew :deck:test` — BUILD SUCCESSFUL, includes updated `Container.spec.tsx` / `TaskDefinition.spec.tsx` unit tests asserting the normalized image shape (with `imageId` populated) after selecting a docker registry account.

https://claude.ai/code/session_016hcp89wpNQVdMi9dmUifXQ

### DIFF
--- a/deck/packages/ecs/src/serverGroup/configure/dockerImage.util.ts
+++ b/deck/packages/ecs/src/serverGroup/configure/dockerImage.util.ts
@@ -1,0 +1,29 @@
+import type { IEcsDockerImage } from './serverGroupConfiguration.service';
+
+export function buildEcsImageId(image: any): string {
+  if (image.imageId) {
+    return image.imageId;
+  }
+  if (image.fromContext) {
+    return image.imageLabelOrSha;
+  }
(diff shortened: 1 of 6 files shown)


cristhian-castaneda (approved):
neat, loved the utility extraction, looks fancy

PR #7965, opened 2026-09-04
fix(amazon): fix AMI selection in Create Server Group image picker

jasonmcintosh:
## Summary
- The AMI picker in the AWS Create/Clone Server Group wizard shows images in the dropdown, but clicking one didn't set the value, and "Search All Images" had no way back to the plain dropdown.
- Root cause: `AmazonImageSelectInput`'s custom `menuRenderer` re-wrapped its options in a second `.Select-menu-outer > .Select-menu` pair on top of react-select's own wrapper. `TetheredSelect` only overrides the *outermost* wrapper's position to `static` so Tether can position it; the duplicate inner wrapper kept its default `position: absolute`, pulling it out of flow and collapsing the real, Tether-positioned menu's hit-testable area. The list rendered, but clicks landed on nothing. This is the only place in the app that supplies a custom `menuRenderer` to `TetheredSelect`, which is why every other select was unaffected.
- Also switched the option's selection handler from `onClick` to `onMouseDown` with `preventDefault`/`stopPropagation`, matching react-select's own `Option` component, for defense in depth.
- Added a "Back to Package Images" control so `selectionMode` can return from `searchAllImages` to the normal dropdown, which previously had no way back.

## Test plan
- [x] Added `AmazonImageSelectInput.spec.tsx` covering: no duplicate menu wrapper, selecting an image from the package-images dropdown, switching to/back from "Search All Images", and selecting an image while searching all images.
- [x] `./gradlew :deck:karma` — 3478/3478 passing.
- [ ] Manual smoke test of the Create Server Group wizard in a running app (not done in this session - no local backend was available).

### DIFF
--- a/deck/packages/amazon/src/serverGroup/configure/AmazonImageSelectInput.spec.tsx
+++ b/deck/packages/amazon/src/serverGroup/configure/AmazonImageSelectInput.spec.tsx
@@ -1,0 +1,125 @@
+import type { ReactWrapper } from 'enzyme';
+import { mount } from 'enzyme';
+import React from 'react';
+import { act } from 'react-dom/test-utils';
+
+import type { Application, TetheredSelect as TetheredSelectType } from '@spinnaker/core';
+import { TetheredSelect } from '@spinnaker/core';
+
+import { AmazonImageSelectInput } from './AmazonImageSelectInput';
+import type { IAmazonImage } from '../../image';
+import { AwsImageReader } from '../../image';
+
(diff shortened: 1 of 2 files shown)


cristhian-castaneda (approved):
LGTM 🌮

cristhian-castaneda:
good work @jasonmcintosh, as we talked would love to have a design system like for the ui components down the road, that would help simplify a lot of things when it comes to implement a UI component

PR #7967, opened 2026-09-04
fix(rosco): create writable home dir for spinnaker system user in Dockerfile.ubuntu

jasonmcintosh:
## Summary
- The `release-2026.3.x` initial branch build failed on `rosco` build: https://github.com/spinnaker/spinnaker/actions/runs/33917959467/job/101169420566
- Root cause: #7787 bumped rosco's Ubuntu base image from `jammy` to `resolute`. Ubuntu resolute's `adduser` defaults `--system` accounts to `HOME=/nonexistent` and does not create the directory, whereas jammy created and chowned `/home/<user>`.
- `rosco/Dockerfile.ubuntu` is the only service Dockerfile that runs a build-time step as the `spinnaker` user (`packer plugins install`), which needs to write to `$HOME/.config/packer/plugins`. With `HOME=/nonexistent` that fails:
  ```
  could not create plugin folder "/nonexistent/.config/packer/plugins/github.com/hashicorp/amazon": mkdir /nonexistent: permission denied
  ```
- Fix: give the `spinnaker` system user an explicit home directory (`--home /home/spinnaker`), matching jammy's implicit behavior. Verified via local repro in both `ubuntu:jammy` and `ubuntu:resolute` containers.
- Checked the other nine `Dockerfile.ubuntu` files (keel, front50, echo, clouddriver, gate, kayenta, igor, fiat, orca) — none run build-time steps as `spinnaker`, so they're unaffected by this same bug.

## Test plan
- [x] Reproduced the failure locally by running `adduser --system --uid 10111 --group spinnaker` in `ubuntu:jammy` vs `ubuntu:resolute` and comparing `/etc/passwd` + home dir permissions
- [x] Verified the fix: with `--home /home/spinnaker`, the directory is created and owned by `spinnaker`, and `mkdir -p $HOME/.config/packer/plugins` succeeds as that user in `ubuntu:resolute`
- [ ] CI: rosco Docker build passes

### DIFF
--- a/clouddriver/Dockerfile.ubuntu
+++ b/clouddriver/Dockerfile.ubuntu
@@ -8,10 +8,7 @@ ENV KUBECTL_RELEASES="${KUBECTL_DEFAULT_RELEASE} 1.31.14 1.32.13 1.33.13"
 ENV AWS_CLI_VERSION=2.35.22
 ENV AWS_AIM_AUTHENTICATOR_VERSION=0.7.18
 
-RUN apt-get update && apt-get install -y curl gnupg && \
-  curl -s https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key add - && \
-  echo "deb https://packages.cloud.google.com/apt cloud-sdk  main" > /etc/apt/sources.list.d/cloud-sdk.list && \
-  apt-get update && \
+RUN apt-get update && \
   apt-get upgrade -y && \
   apt-get install -y \
   curl \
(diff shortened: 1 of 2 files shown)


jasonmcintosh:
SHOULD fix build issues due to https://github.com/spinnaker/spinnaker/actions/runs/33917959467/job/101169420645 and past error on the upgrades.  Unfortunately updating the docker files did NOT validate the images until merge, so we missed this... :(

cristhian-castaneda (approved):
LGTM

jasonmcintosh:

PR #7969, opened 2026-09-05
fix(rosco,clouddriver): use bsdtar to unblock arm64 Dockerfile.ubuntu builds

jasonmcintosh:
## Summary

`rosco` and `clouddriver`'s `Dockerfile.ubuntu` (`ubuntu:resolute`) arm64 builds are broken: `ubuntu:resolute` ships a GNU tar patched for CVE-2025-45582, which now extracts via `openat2()`. The QEMU version pinned in `.github/actions/publish-docker/action.yml` (`qemu-v7.0.0-28`, pinned as a workaround for an unrelated segfault bug, `docker/setup-qemu-action#198`) predates QEMU's `openat2` support, so every emulated `linux/arm64` build fails during helm/kustomize/helmfile/gcloud-sdk extraction with:

```
tar: ...: Cannot open: Function not implemented
```

## Fix

Swap `bsdtar` (via `libarchive-tools`) in for GNU tar in the two affected Dockerfiles. `bsdtar` doesn't use `openat2`, so it sidesteps the incompatibility without touching CI topology or the QEMU pin.

Verified locally against the exact pinned `qemu-v7.0.0-28` image: reproduced the failure with GNU tar, then confirmed `bsdtar` succeeds for `get-helm-3`, the piped kustomize/helmfile extraction, and the gcloud-sdk tarball extraction.

## Status

This is a backup/fallback fix. The preferred fix (bumping the QEMU pin itself to `qemu-v10.2.3-68`, which appears to resolve both this bug and the original segfault bug it was pinned to avoid) is being pursued separately off `main`. Opening this as a draft in case that path doesn't pan out.

## Test plan
- [ ] CI run on this branch builds rosco and clouddriver's ubuntu-variant arm64 images successfully

### DIFF
--- a/clouddriver/Dockerfile.ubuntu
+++ b/clouddriver/Dockerfile.ubuntu
@@ -8,6 +8,11 @@ ENV KUBECTL_RELEASES="${KUBECTL_DEFAULT_RELEASE} 1.31.14 1.32.13 1.33.13"
 ENV AWS_CLI_VERSION=2.35.22
 ENV AWS_AIM_AUTHENTICATOR_VERSION=0.7.18
 
+# libarchive-tools' bsdtar replaces GNU tar as /usr/bin/tar below: GNU tar (patched for CVE-2025-45582) now
+# extracts via openat2(), which QEMU 7.0.0 (pinned in .github/actions/publish-docker/action.yml as a workaround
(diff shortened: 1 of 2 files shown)


jasonmcintosh:
THIS IS A BACKUP plan - not really want to merge this one vs. just bumping QEMU which should REALLY fix several issues

PR #7972, opened 2026-09-06
fix(deck): scope rxjs override so lerna publish doesn't crash

jasonmcintosh:
## Summary
- The workspace-wide `rxjs: 6.6.7` pnpm override in `deck/pnpm-workspace.yaml` (added to keep the deck app bundle on a single rxjs version) was also forcing `lerna`'s own `inquirer@12.9.6` dependency down to rxjs 6.6.7.
- `inquirer@12` requires `rxjs ^7.8.2` and its ESM build does `import { concatMap } from 'rxjs'` — an export rxjs 6 only exposes via `rxjs/operators`, not the package root. Loading inquirer therefore crashed with `SyntaxError: The requested module 'rxjs' does not provide an export named 'concatMap'`, which killed the `npx lerna publish` step before any prompt logic ran.
- Observed in the deck Lerna Publish job: https://github.com/spinnaker/spinnaker/actions/runs/33989236425
- Scoped the override to `inquirer@12>rxjs: ^7.8.2` so only lerna's inquirer gets the modern rxjs; every other consumer (core, kayenta, `@uirouter/rx`, `redux-observable`, the `inquirer@7.3.3` used by graphql-codegen, etc.) stays pinned to 6.6.7.

## Test plan
- [x] `pnpm install` in `deck/` regenerates the lockfile cleanly with a single new `rxjs@7.8.2` entry scoped to inquirer's tree, no other resolution changes
- [x] Loaded the exact file that crashed CI (`inquirer/dist/esm/ui/prompt.js`) directly in Node — no error
- [x] `npx lerna --version` runs successfully
- [x] Re-running `pnpm install` shows no further drift

### DIFF
--- a/deck/pnpm-lock.yaml
+++ b/deck/pnpm-lock.yaml
@@ -13,6 +13,7 @@ overrides:
   react: ^16.14.0
   react-dom: ^16.14.0
   rxjs: 6.6.7
+  inquirer@12>rxjs: ^7.8.2
   typescript: 5.0.4
 
 importers:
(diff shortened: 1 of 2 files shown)


christosarvanitis (approved):

jasonmcintosh:

PR #7973, opened 2026-09-07
fix(deck): Case-insensitve on virtualizationType comparison

christosarvanitis:
Follow up fix of  https://github.com/spinnaker/spinnaker/pull/7963 it turns out that the virtualization type had a similar issue in some architecture selection.  Tested and verified that the instanceTypes are now populated per api response


jasonmcintosh (approved):

jasonmcintosh:

PR #7981, opened 2026-09-07
fix(deck): bump React from 16.14 to 17.0.2 across the workspace

jasonmcintosh:
## Summary

**Stacked on #7980** (rxjs 6->7) — this branch is built on top of it since react-redux 7 needs kayenta's redux 4.2.1 (from that PR). The diff here will shrink to just the React-related changes once #7980 merges and this gets rebased onto `main`; until then it includes #7980's commits too.

Stage 1 of the React modernization: **16 -> 17 only**. Going further (18/19) requires replacing `enzyme` across ~257 spec files first (enzyme never shipped an official adapter past React 16) — that's its own migration, intentionally out of scope here.

- Bumped `react`/`react-dom`/`@types/react`/`@types/react-dom` to `17.0.2`/`17.0.93`/`17.0.26` across all packages, and `react-redux` 5.x -> `7.2.9` in core and kayenta (react-redux 7 still supports the `connect()` HOC used throughout, and its peer range covers 16/17/18, so this also pre-positions for a future React 18 step). Swapped the enzyme adapter from `enzyme-adapter-react-16` to the community-maintained `@wojtekmaj/enzyme-adapter-react-17` in the single place it's configured (`karma-shim.js`) — none of the 257 spec files reference the adapter directly.
- react-redux 7 requires an explicit `@types/react-redux` devDependency (it depends on the DefinitelyTyped package rather than bundling types) for pnpm to hoist it somewhere TypeScript can resolve; added it back to kayenta and moved 9 stray `Dispatch`/`Store` type imports from `'react-redux'` (which no longer re-exports them) to `'redux'`.
- Fixed real type breaks react-redux 7's stricter `connect()` typings surfaced: kayenta's generic `disableable()` HOC couldn't unify react-redux's `Matching<>` helper against an unresolved generic, and `metricStoreConfig.service.ts`'s `metricConfigurer: React.ComponentClass` field was too narrow for what `connect()` now returns (widened to `React.ComponentType<any>`).
- Fixed a genuine **React 16->17 regression**: React 17's dev-mode `forwardRef()` defines `displayName` as a non-enumerable getter/setter (for its own warnings), which `CloudProviderRegistry.getProvider()`'s `lodash.cloneDeep` silently dropped when copying registered component references. Switched to `cloneDeepWith` with a customizer that passes functions/React-element-shaped values through by reference instead of cloning them — the correct fix regardless of React version, since component references were never meant to be deep-cloned.
- Fixed a real **functional-test regression**: `react-bootstrap@0.32.1`'s and `react-bootstrap-typeahead@3.4.7`'s `react-overlays` dependency (`^0.8.0`, capped by real semver-caret rules at 0.8.x) hit a known-and-fixed upstream bug where `RootCloseWrapper`'s "click outside" handler fires immediately under React 17's changed event delegation, closing every dropdown the instant it opens ([react-bootstrap/react-overlays#889](https://github.com/react-bootstrap/react-overlays/pull/889), fixed in 0.9.x). Overrode `react-overlays` to `^0.9.3` — confirmed via two independent Cypress functional runs that consistently failed on the same 3 dropdown-driven specs (`kubernetes/clusters`, `kubernetes/firewalls`, `core/projects/project`) before the override, and pass cleanly after.
- Two test-only React 17 adjustments: `ReactModal.spec.tsx` wraps its `dismissAll()` calls in `act()` so the modal's `useEffect` cleanup (which calls `monitor.onModalClose()`) flushes synchronously, and `infrastructure.states.spec.ts` compares the routed component's stable `render` function reference instead of `displayName`, for the same non-enumerable-getter reason as the `CloudProviderRegistry` fix — this time inside `@uirouter/react`'s own internals.

## Test plan
- [x] `pnpm install` in `deck/` resolves cleanly to a single react/react-dom/react-redux/react-overlays version across the workspace
- [x] `./gradlew :deck:test` — full workspace TS build, `eslint --max-warnings=0`, `prettier --check`, the karma suite (3493 tests), and all 26 Cypress functional specs are green
- [x] Reproduced the dropdown regression twice before the `react-overlays` override (deterministic, not flaky) and confirmed it's gone after


jgrumboe:
@jasonmcintosh 
I tried out the new code style for a review. Is that helpful?

## Review — React 16→17 / rxjs 6→7 upgrade

Reviewed against [`CODE_STYLE.md`](../blob/main/CODE_STYLE.md) §1–16. Severity per that guide's ladder. Nothing blocking.

Reviewed locally off `refs/pull/7981/head` (diffed against merge-base `83d59408`), so the three `Merge branch 'main'` commits are excluded and this covers only the two substantive commits.

### Verified, not taken on faith

- **The redux-observable 0.14→1.2 epic rewrite handles the ordering trap.** In 1.x, epics don't run until `.run()`. `epicMiddleware.run()` is called inside `initializeCanaryStore` (`deck/packages/kayenta/src/kayenta/canary.tsx:40`) and the only `INITIALIZE` dispatch is in the `Canary` constructor (`canary.tsx:60`), which runs later — so `loadKayentaAccountsEpic` (`epics.ts:216`) still fires. `store.getState()` → `state$.value` is a faithful translation at every call site.
- **Dropping the `configValidationErrors: () => null` reducer is safe** — zero references anywhere in the tree at this head.
- **`RegionFilter.tsx:27`** is the same hazard class as the `react-overlays` bug fixed here by override (document-level outside-click listener + React 17 root-level delegation), but the `rootRef.current.contains(event.target)` guard makes it immune.
- **`Overridable.tsx:147`'s `.render`-sniffing `isClassComponent` check** has a latent hole, but forwardRef objects carried `.render` under React 16 too — pre-existing, out of scope.

### 1. Should fix — §13 (tests must exercise runtime behavior)

`deck/packages/core/src/search/infrastructure/infrastructure.states.spec.ts:36,57`

```ts
expect((errorBoundary.prop('children').type as any).render).toBe((SearchV1 as any).render);
```

Both operands are `any`-cast reads of an optional property. Today it's a real assertion — `SearchV1` is `withDeckRuntimeServices(withRouter(SearchV1Component))` and both wrappers return `React.forwardRef` objects, so `.render` is a defined function. But if either wrapper ever stops being a forwardRef, both sides become `undefined`, `undefined === undefined` passes, and the test silently stops verifying that the route registers the right component. One line closes it:

```ts
expect((SearchV1 as any).render).toEqual(jasmine.any(Function));
```

The move off `displayName` itself is well-justified — React 17 dev-mode defines it as a non-enumerable accessor on forwardRef objects, which `@uirouter/react`'s internal enumerable-keys copy drops.

### 2. Should fix — §6 (call out intentional behavioral change) + release process

`deck/packages/core/src/plugins/sharedLibraries.ts:44-48`, `deck/packages/scripts/config/rollup.config.base.plugin.js:17-32`

Deck hands plugins `react`, `react-dom`, `react-redux` and `rxjs` as runtime globals, and plugins compile against `pluginsdk-peerdeps`. This PR moves three of those across major versions at once (React 16→17, react-redux 5→7, rxjs 6→7).

`rxjs/operators` is not in the shared-globals list, so an already-published plugin ships its own bundled rxjs-6 operators and will now apply them to rxjs-7 `Observable`s from the host — that works only through rxjs 7's `lift`/`SafeSubscriber` compat path. react-redux 5→7 drops the legacy-context `connect()` plumbing outright. Existing compiled plugins need a rebuild.

`pluginsdk-peerdeps/package.json` gets its dependency bumps but no version change, and nothing in the diff flags this for plugin authors. Suggest an explicit "breaking for Deck plugins" callout in the PR description plus a peerdeps major bump.

### 3. Question — §10 (leave a trail for deferred work)

`deck/pnpm-workspace.yaml:24-33`

`'@spinnaker/kayenta>rxjs': 6.6.7` keeps kayenta on rxjs 6 while everything else moves to 7. The override comment is genuinely good on *why* (redux-observable 2.x/3.x needs redux ≥5 + react-redux 8/9 + React 18). Two loose ends:

1. It ships two rxjs majors in one bundle, with a live cross-major seam: `canary.states.ts:80` resolves `uiRouter.globals.params$` from `@uirouter/rx@1.0.0` (rxjs 7) and `detailLoader.tsx:32` subscribes to it while typing it as an rxjs-6 `Observable`. It works today (plain `.subscribe`, no rxjs-6 operators applied to it), but it's a fragile seam worth naming.
2. §10 asks for a `TODO` plus a tracking issue for deferred work; right now the deferral lives only in a YAML comment. Is there an issue for finishing kayenta → rxjs 7 / React 18?

### 4. Nit — §6

`deck/packages/core/src/cloudProvider/CloudProviderRegistry.ts:33-43`

The fix is right — component references were never meant to be deep-cloned. Three small points:

- The comment's stated mechanism ("lodash's `cloneDeep` silently drops the outer object's own properties") isn't quite what happens: `cloneDeep` copies own *enumerable* properties, and React 17 dev-mode makes `displayName` a non-enumerable accessor on forwardRef/memo objects, so it gets skipped. The commit message states this correctly; worth matching here.
- `isFunction(value) || (value && typeof value === 'object' && '$$typeof' in value)` re-implements `react-is`'s `isValidElementType` (already available transitively via react-redux) — less brittle than an ad-hoc `$$typeof` sniff.
- This narrows the mutation-safety contract: callers now share `$$typeof`-tagged values with the registry rather than getting copies. I checked every `getProvider`/`getValue` call site and none mutate, so no live defect.

### 5. Question — §13

`SearchV1.spec.tsx:27-31`, `DockerTriggerTemplate.spec.tsx:18-24`

Trading `jasmine.clock()` for real `setTimeout` waits swaps determinism for a workaround. `SearchV1` now has eight real 301ms waits against a 300ms `debounceTime` — a 1ms margin. Timer due-time ordering keeps that from being an actual race (the debounce timer is scheduled first and is due first), and `DockerTriggerTemplate`'s 250-vs-200ms margin is comfortable, so I'm not asserting a defect. Two questions:

- Was `jasmine.clock().mockDate()` tried? Jasmine fakes `setTimeout`/`setInterval` but not `Date.now`, which rxjs 7's `AsyncScheduler` reads via `dateTimestampProvider`.
- Was rxjs 7's `timeoutProvider` / `intervalProvider` `delegate` hook — added in v7 specifically for test control — considered?

If neither works, widening SearchV1's margin to ~350ms costs nothing and removes the 1ms cliff.

### Not applicable

§15 (no auth or URL-fetching surfaces touched), §16 (this PR removes legacy tech rather than adding any), §1/§11 (no new plumbing).

One process note: the commit messages carry a real test plan (full TS build, lint, prettier, karma 3493 tests, all 26 Cypress specs, plus before/after functional runs isolating the `react-overlays` dropdown regression). §PR-description conventions wants that in the PR description too — I couldn't read the description from my environment, so ignore this if it's already there.

---
_Generated by [Claude Code](https://claude.ai/code)_

jgrumboe:
The review comment from before was too verbose (the code style guide needs some modifications on this).

This is a hopefully better one.

## Review — React 16→17 / rxjs 6→7

Reviewed against `CODE_STYLE.md`, off `refs/pull/7981/head` vs merge-base `83d59408`. Nothing blocking; the migration looks careful and the commit messages carry real validation evidence.

Spot-checked and clean: `epicMiddleware.run()` lands before the only `INITIALIZE` dispatch (`canary.tsx:40` vs `:60`), so kayenta's epics still fire under redux-observable 1.x; the dropped `configValidationErrors` reducer has zero references; `RegionFilter.tsx:27`'s document-level click listener survives React 17's delegation change thanks to its `contains()` guard.

**1. Should fix (§13) — `infrastructure.states.spec.ts:36,57`.** `expect((x as any).render).toBe((SearchV1 as any).render)` is a real assertion today (both are forwardRefs), but goes vacuous — `undefined === undefined` — if either wrapper stops being one. Add `expect((SearchV1 as any).render).toEqual(jasmine.any(Function))`.

**2. Should fix (§6) — plugin-facing break.** `sharedLibraries.ts:44-48` hands plugins `react`, `react-dom`, `react-redux` and `rxjs` as runtime globals; this bumps three across majors at once. `rxjs/operators` isn't in that list, so published plugins apply their own bundled rxjs-6 operators to rxjs-7 Observables (works only via v7's `lift` compat path), and react-redux 5→7 drops legacy-context `connect()`. Existing plugins need a rebuild. Worth a "breaking for Deck plugins" callout in the description plus a `pluginsdk-peerdeps` major bump.

**3. Question (§10) — `pnpm-workspace.yaml:24-33`.** The kayenta rxjs-6 pin is well explained, but it ships two rxjs majors and leaves a cross-major seam (`canary.states.ts:80` → `detailLoader.tsx:32` subscribes to an `@uirouter/rx` 1.0.0 / rxjs-7 stream typed as rxjs 6 — fine today, no operators applied). Is there a tracking issue for finishing kayenta → rxjs 7?

**4. Nit — `CloudProviderRegistry.ts:33-43`.** Right fix. The comment's mechanism is slightly off: `cloneDeep` copies own *enumerable* props, and React 17 makes `displayName` a non-enumerable accessor — the commit message states it correctly. Also, `react-is`'s `isValidElementType` would beat the ad-hoc `$$typeof` sniff.

**5. Question — `SearchV1.spec.tsx`.** Eight real 301ms waits against a 300ms debounce. Not a race (due-time ordering holds), but was `jasmine.clock().mockDate()` or rxjs 7's `timeoutProvider.delegate` tried? If not, widening to ~350ms removes the 1ms cliff for free.

---
_Generated by [Claude Code](https://claude.ai/code)_

jasonmcintosh:
Overall good feedback/reviews ;) there's a few things on this - I'm trusting a LOT to UI & actual real sanity checks on things ... particularly since I'm not as good at UI code ;) I don't mind additional UI validations of this post sync so definitely useful and called out a couple of good things!

PR #7983, opened 2026-09-08
feat(orca): Add opt-in feature to rewrite dashed SpeL variables

christosarvanitis:
## Problem

SpEL tokenizes `-` as the arithmetic minus operator, not a valid identifier character. As a
result, an expression like:

```
${my-container-name}
```

is not parsed as "the variable named `my-container-name`" — it's parsed as
`my - container - name` (property `my` minus property `container` minus property `name`). This
breaks any manifest, parameter, or context value that legitimately contains a hyphen (e.g.
Kubernetes container/label names), and is a long-standing, well-known limitation of Spinnaker's
expression evaluator (`DeployManifestStage`, `SavePipelineTemplateTask`, and Deck's
"Evaluate Variables" stage all separately guard against or are affected by this).

## Solution

An **opt-in, backward-compatible** rewrite: when a `${...}` expression fails to evaluate because
its entire body is a hyphen-joined bareword (e.g. `${my-container-name}`), retry it as a literal
key lookup (`${#root['my-container-name']}`) before giving up. SpEL's bracket/indexer syntax is
immune to this problem because the hyphenated text is a quoted string literal, never re-tokenized
as an operator.

This is disabled by default (`expression.dashed-identifiers.enabled: false`), so no existing
pipeline behavior changes unless an operator explicitly opts in. The rewrite only triggers **after**
normal evaluation fails, so it never changes the result of an expression that already succeeds
(e.g. real arithmetic like `${trigger.buildNumber - 1}`).

### `kork-expressions`
- **`DashedIdentifierResolver`** (new): pure utility that recognizes a `${...}` expression whose
  entire body is a hyphen-joined bareword and rewrites it to `${#root['...']}`. Deliberately leaves
  real arithmetic, dotted paths (`${metadata.labels.app-name}` — known limitation, fast-follow), and
  partial/composite strings untouched.
- **`ExpressionTransform`**: new constructor overload taking a `dashedIdentifierSupportEnabled`
  flag (default `false`, existing callers unaffected). On evaluation failure, retries with the
  resolver's rewritten expression before recording the failure.
- **`ExpressionProperties`**: new `dashedIdentifiers` feature flag (`expression.dashed-identifiers.enabled`, default `false`), following the existing `doNotEvalSpel` convention.

### `orca-core`
- **`PipelineExpressionEvaluator`**: reads `ExpressionProperties.getDashedIdentifiers()` and threads
  it into `ExpressionTransform`.
- **`ExpressionCapabilityResult` / `CapabilitiesService`**: `GET /capabilities/expressions` now also
  returns `dashedIdentifiersEnabled`, so clients (Deck) can discover whether the feature is active
  without hardcoding it. No `gate` change needed — `/capabilities/expressions` is already proxied
  through as a generic `Map`.

### `deck`
- **`ExpressionCapabilitiesReader`** (new): fetches/caches `/capabilities/expressions`, mirroring the
  existing `DeploymentMonitorReader` pattern.
- **`EvaluateVariablesStageConfig`**: the "Evaluate Variables" stage's variable-name validator
  (previously a hardcoded `^[a-zA-Z_][a-zA-Z0-9_]*$`, rejecting hyphens unconditionally) now accepts
  hyphens when the orca capability reports the feature is enabled — fetched once per component mount
  via `useEffect`, defaulting to today's strict behavior while loading/unavailable.

## Testing

Built test-first throughout:
- `DashedIdentifierResolverTest` (kork-expressions, JUnit5): rewrite rules for barewords, real
  arithmetic, dotted paths, digit-leading identifiers, blank/null input.
- `ExpressionTransformTest` (kork-expressions, JUnit5): enabled/disabled resolution, real arithmetic
  still works with the flag on, unresolved-key degrades to INFO (not ERROR).
- `PipelineExpressionEvaluatorSpec` (orca-core, Spock): end-to-end evaluation with the flag on/off.
- `CapabilitiesServiceSpec` (orca-core, Spock): `dashedIdentifiersEnabled` reflects the injected
  config in both states.
- `EvaluateVariablesStageConfig.spec.ts` (deck, Karma/Jasmine): validator behavior on/off, still
  rejects other invalid characters/leading digits when enabled, duplicate-key validator unaffected.
- Regression-checked: full `kork-expressions`, `orca-core` suites, the pre-existing
  `DeployManifestExpressionEvaluationTest` integration test, and the full Deck Karma suite
  (3499 tests) — all green.
- `spotlessCheck`/`spotlessApply` (backend) and `prettier`/`eslint` (deck) clean.

## Known limitations / follow-ups
- Compound paths mixing `.` and `-` (e.g. `${metadata.labels.app-name}`) are not yet rewritten —
  explicitly locked in by a test as a documented limitation, left for a potential follow-up.
- Pipeline-template variable names (`SavePipelineTemplateTask`) still hard-reject dashes
  server-side and have no client-side validator in Deck at all; out of scope for this PR.

## Rollout
- Fully backward compatible: default is off, and even when enabled, the rewrite only engages on
  evaluation failure, so no currently-passing expression changes behavior.
- Deck honors the same server-side flag via `/capabilities/expressions`, so the UI and runtime
  evaluation can never disagree about whether hyphenated identifiers are supported.


kork/kork-expressions/src/main/java/com/netflix/spinnaker/kork/expressions/config/ExpressionProperties.java
```diff
@@ -30,6 +30,14 @@ public class ExpressionProperties {
   /** Flag to determine if SpEL evaluation to be skipped. */
   private final FeatureFlag doNotEvalSpel = new FeatureFlag().setEnabled(true);
 
+  /**
+   * Flag to enable resolving {@code ${...}} expressions whose entire body is a hyphen-joined
+   * bareword (e.g. {@code ${my-container-name}}) as a literal key lookup, instead of failing
+   * because SpEL interprets the hyphen(s) as subtraction operators. Disabled by default so that
+   * pipelines relying on real SpEL subtraction inside {@code ${...}} expressions are unaffected.
+   */
+  private final FeatureFlag dashedIdentifiers = new FeatureFlag().setEnabled(false);
```
jasonmcintosh, line 39: This isn't a "approve/disapprove" comment, but... these aren't really FeatureFlags in the traditional sense.  BE NICE If there really WAS a "FF" concept in here that used a TRUE FF system to dynamically load/set these kinds of configuration properties.  This is a simple wrapper class around booleans... which IMHO just doesn't make a LOT of sense.  Separate issue from today.

jasonmcintosh (approved):
Interesting enhancement.  SHOULD make one specific edge case on expression resolution much easier... though not supporting it via ${something.parameter-dashed-name}... makes it LESS usable, but I don't know that supporting that case would be all that viable.  Given this is a configurable enhancement, I'm also good with backporting.

christosarvanitis:
@mergifyio backport release-2026.3.x release-2026.2.x

PR #7996, opened 2026-09-11
chore: migrate to Spring Boot 4.1.1

js4683:
## Summary
Upgrades this repo to Spring Boot 4.1.1 on Java 17 (Spring Cloud 2025.1.3, Kotlin 2.3.21, Gradle 9.6.1, jOOQ 3.19.35), preserving existing behavior and public contracts.

Scope (291 files vs main):
- Framework 7 / Boot 4 API and dependency migrations across services (Jakarta, Security, OpenSAML, health endpoints, Redis, JDBC, Quartz, RestClient, MVC, WireMock, Jackson, Kubernetes, integrations).
- Echo: preserve Boot MVC defaults while registering the CloudEvent custom converter; drop obsolete Spring Data REST/HATEOAS deps.
- Orca Webhook: replace removed Spring 6 OkHttp request-factory support with a local bridge (3 new files).
- Tests updated only where Boot 4 APIs/behavior require it.
- Pre-existing Tomcat bug fix (approved in review): relaxed-query-characters branch now calls setRelaxedQueryChars(getRelaxedQueryCharacters()) instead of resetting path chars, plus a focused regression test (RelaxedUriPropertiesTest, 2 tests).

Base: main at cfdcefbf79. Migration-only history, no unrelated feature commits.

## Validation
- no-mistakes gate passed-with-skips on this head (rebase/review/test/document/lint/push green; PR+CI skipped locally as provider unknown).
- :kork:kork-tomcat:spotlessCheck and :kork:kork-tomcat:test pass under Java 17.
- New regression test fails with the old copy/paste bug and passes with the fix (verified live against a real Tomcat protocol handler).
- Full multi-service suite + Docker/MySQL-backed tests (EmbeddedRedis, Keel MySQL) remain for remote CI.

## Risk / rollback
Medium (broad framework major upgrade, disciplined and reviewed). Rollback: close this PR; no data migrations.

/cc @spinnaker/reviewers


clouddriver/clouddriver-kubernetes/clouddriver-kubernetes.gradle
```diff
@@ -76,11 +76,19 @@ dependencies {
...
   implementation "io.spinnaker.kork:kork-security"
-  implementation "io.kubernetes:client-java"
-  // client-java-api-fluent is not managed by the spring-cloud BOM (which supplies the
-  // client-java version above), so keep this pinned to the same version client-java
-  // resolves to.
-  implementation "io.kubernetes:client-java-api-fluent:19.0.3"
+  // Keep the API and fluent artifacts on the same version. Spring Cloud 2025.1 manages
+  // client-java at 24.x, whose generated models reject fields accepted by our existing
+  // Kubernetes manifests; this migration does not require upgrading the Kubernetes SDK.
+  implementation("io.kubernetes:client-java") {
```
jasonmcintosh, line 82: Dang, we really DO need to get this updated sometime SOONISH... 19 is OLD on this :( 

NOTE We need to do this more holistically in spinnaker-dependencies which sets overall project versions rather than here!

clouddriver/build.gradle
```diff
@@ -25,6 +25,19 @@ plugins {
 allprojects {
   apply plugin: 'io.spinnaker.project'
   group = "io.spinnaker.clouddriver"
+
+  // WireMock 3.10.0's default Jetty 11 server (and the jetty12 variant's Jetty
```
jasonmcintosh, line 29: Good idea wondering if this shouldn't be more "global" for other wiremock usages... and then set the overall version constraints again in spinnaker-dependencies ;)

jasonmcintosh: ALSO wondering if there isn't a newer wiremock that MIGHT resolve this issue without needing to use different versions... ?  NOT looked

echo/echo-notifications/echo-notifications.gradle
```diff
@@ -36,15 +36,18 @@ dependencies {
...
+    exclude group: "io.opentelemetry", module: "opentelemetry-exporter-sender-okhttp"
+  }
   testImplementation project(":echo-test")
   testImplementation("com.icegreen:greenmail:2.1.8")
   testImplementation "org.apache.httpcomponents.client5:httpclient5"
   testImplementation "org.spockframework:spock-spring"
   testImplementation "org.springframework:spring-test"
   testImplementation "com.squareup.retrofit2:retrofit-mock"
-  testImplementation "com.github.tomakehurst:wiremock-jre8-standalone"
+  testImplementation "org.wiremock:wiremock"
```
jasonmcintosh, line 50: Example here from above... if we used standalone would this be better?

echo/echo-notifications/echo-notifications.gradle
```diff
@@ -36,15 +36,18 @@ dependencies {
   implementation "io.cloudevents:cloudevents-json-jackson:3.0.0"
   implementation ("dev.cdevents:cdevents-sdk-java:0.3.1")
   implementation "io.opentelemetry:opentelemetry-api"
+  implementation "io.opentelemetry:opentelemetry-api-incubator:1.62.0-alpha"
   implementation "io.opentelemetry:opentelemetry-sdk"
-  implementation "io.opentelemetry:opentelemetry-exporter-otlp"
+  implementation("io.opentelemetry:opentelemetry-exporter-otlp") {
```
jasonmcintosh, line 41: I'm assuming there isn't an updated otel sdk/exporter that works with spring boot 4 then??  I REALLY would like to avoid overrides/exclusions like this if at all possible

echo/build.gradle
```diff
@@ -23,6 +23,19 @@ plugins {
 allprojects {
   apply plugin: 'io.spinnaker.project'
   group = "io.spinnaker.echo"
+
+  // WireMock 3.10.0's default Jetty 11 server (and the jetty12 variant's Jetty
+  // 12.0.x) conflicts with Boot 4.1's managed Jetty 12.1.x on test classpaths
+  // (HttpServerFactoryLoader finds no loadable factory). The shaded standalone
+  // artifact is self-contained, so substitute it for all WireMock test usage.
+  configurations.all {
```
jasonmcintosh, line 31: Ditto above comments - individual overrides like this are to be avoided as much as possible.  Either upgrading a library OR setting restrictions at the spinnaker-dependencies layer.

gate/gate-web/gate-web.gradle
```diff
@@ -40,15 +40,15 @@ dependencies {
 
   implementation "org.springframework:spring-web"
   implementation "org.springframework.boot:spring-boot-starter-actuator"
-  implementation "org.springframework.boot:spring-boot-starter-aop"
+  implementation "org.springframework.boot:spring-boot-starter-aspectj"
   implementation "io.github.resilience4j:resilience4j-spring-boot2"
 
-  implementation "org.springdoc:springdoc-openapi-starter-webmvc-ui:2.2.0"
+  implementation "org.springdoc:springdoc-openapi-starter-webmvc-ui:3.1.1"
```
jasonmcintosh, line 46: This isn't a bad place to put this as only place openapi stuff comes in, but ... I'd honestly prefer if we could move the deps on cdevents & some of this into spinnaker-dependencies.  OpenAPI one is LESS of a risk as gate is only place doing that... but cdevents is duplicated in echo I believe now...

gate/build.gradle
```diff
@@ -11,6 +11,19 @@ allprojects {
 
   group = "io.spinnaker.gate"
 
+  // WireMock 3.10.0's default Jetty 11 server (and the jetty12 variant's Jetty
+  // 12.0.x) conflicts with Boot 4.1's managed Jetty 12.1.x on test classpaths
+  // (HttpServerFactoryLoader finds no loadable factory). The shaded standalone
+  // artifact is self-contained, so substitute it for all WireMock test usage.
+  configurations.all {
+    resolutionStrategy.dependencySubstitution {
```
jasonmcintosh, line 19: Same duplication issue on these everywhere...

kayenta/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaJackson3SubtypeConfiguration.java
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2026 Harness, Inc.
```
jasonmcintosh, line 2: Not sure this is a "harness" copyright ;)

keel/keel-sql/keel-sql.gradle
```diff
@@ -50,11 +50,35 @@ dependencies {
     // avoid circular dependency which breaks Liquibase
     exclude(module: "keel-sql")
   }
-  testImplementation("org.testcontainers:mysql")
+  testImplementation("org.testcontainers:testcontainers-mysql")
 
   jooqGenerator(platform("io.spinnaker.kork:kork-bom"))
   jooqGenerator("com.mysql:mysql-connector-j")
-  jooqGenerator("org.jooq:jooq-meta-extensions:3.19.35")
+  // Pin the whole codegen family to the last Java 17-compatible jOOQ line: the
```
jasonmcintosh, line 57: This shouldn't be needed since i merged java 25 in now ;)

kork/kork-eureka/src/main/java/com/netflix/spinnaker/kork/eureka/BootHealthCheckHandler.java
```diff
@@ -22,9 +22,10 @@
 import com.netflix.appinfo.InstanceInfo;
 import java.util.Map;
 import java.util.Objects;
-import org.springframework.boot.actuate.health.*;
-import org.springframework.boot.actuate.health.HealthIndicator;
-import org.springframework.boot.actuate.health.Status;
+import org.springframework.boot.health.actuate.endpoint.StatusAggregator;
+import org.springframework.boot.health.contributor.*;
```
jasonmcintosh, line 26: NIT: IF we're redoing imports... shouldn't use wildcards please ;)

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -10,7 +10,10 @@ ext {
     awsv2            : "2.23.7",
     bouncycastle     : "1.84", // fixes CVE-2024-34447, CVE-2025-14813
     gcp              : "26.34.0",
-    groovy           : "4.0.15",
+    // Aligned with Spock 2.4-groovy-4.0's required Groovy (4.0.29): keeps the
+    // Groovy 4 line Boot 3 used while staying off Boot 4.1's Groovy 5.0.8,
+    // which Spock 2.4 does not support. Enforced via strictly below.
```
jasonmcintosh, line 15: This... kinda has me wondering if could upgrade spock... or should... I'll say "ok" for now... but would love a PR to fix this so we can go to the current groovy.  That said what I'd love even MORE is to get off of groovy but... that's more work than we can do here for now ;)

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
```diff
@@ -36,6 +40,10 @@ ext {
     // GenericContainer wrapper, so the emulator version comes from the image tag
     // (MINISTACK_IMAGE_TAG in the tests), not from here.
     testcontainersMiniStack : "0.1.5",
+    kubernetes       : "19.0.3",
+    // Boot 4.1 manages jOOQ 3.21.7 (Java 21 bytecode, major 65). Keep Java 17 by
```
jasonmcintosh, line 44: Not needed since I merged java 25 ;)

orca/orca-webhook/src/test/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfigurationTest.java
```diff
@@ -262,14 +261,14 @@ void testEventLoggingVerbosity(boolean verbose) {
             });
   }
 
-  /** Retrieve the client member from the OkHttp3ClientHttpRequestFactory bean */
+  /** Retrieve the client member from the OkHttpClientHttpRequestFactory bean */
   private static OkHttpClient getOkHttpClient(AssertableApplicationContext ctx) {
-    OkHttp3ClientHttpRequestFactory requestFactory =
-        ctx.getBean(OkHttp3ClientHttpRequestFactory.class);
+    OkHttpClientHttpRequestFactory requestFactory =
```
jasonmcintosh, line 266: SO I'd REALLY like to name this something OTHER than "OkHttpClientHttpRequestFactory" - "SpinnakerOkHttpClientRequestFactory" or similar so we don't get into confusion on legacy classes/names.  (same for the class itself that's new here)

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
```diff
@@ -171,7 +170,7 @@ public ClientHttpRequestFactory webhookRequestFactory(
     }
 
     var client = builder.build();
-    var requestFactory = new OkHttp3ClientHttpRequestFactory(client);
+    var requestFactory = new OkHttpClientHttpRequestFactory(client);
```
jasonmcintosh, line 173: Same as others OkHttpClientHttpRequestFactory is... a confusing name that could be assumed to be a non-spinnaker related factory.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/OkHttpClientHttpResponse.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatusCode;
+import org.springframework.lang.Nullable;
+import org.springframework.util.Assert;
+
+/**
+ * Vendored from Spring Framework 6.2 {@code OkHttp3ClientHttpResponse} (removed in Spring 7). See
+ * {@link OkHttpClientHttpRequestFactory} for rationale.
+ */
+class OkHttpClientHttpResponse implements org.springframework.http.client.ClientHttpResponse {
```
jasonmcintosh, line 32: I'd REALLY like to avoid this... wondering if there isn't another solution here that'd... be more viable...

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/OkHttpClientHttpRequestFactory.java
```diff
@@ -0,0 +1,104 @@
...
+import org.springframework.http.client.ClientHttpRequestFactory;
+import org.springframework.util.Assert;
+
+/**
+ * Vendored from Spring Framework 6.2 {@code OkHttp3ClientHttpRequestFactory} (removed in Spring 7).
+ * Temporary bridge to preserve orca-webhook's OkHttp-based {@code RestTemplate} behavior (custom
+ * SSL, interceptors, timeouts) after Spring dropped OkHttp support. Revisit if webhooks migrate to
+ * JDK/Jetty/Reactor clients.
+ */
+public class OkHttpClientHttpRequestFactory implements ClientHttpRequestFactory, DisposableBean {
```
jasonmcintosh, line 37: Per other conversation... "ok" but... probably Spinnaker named, and maybe... another immediate PR post removal to switch or revisit this as the comment suggests... OkHttpClient is still perfectly valid as a framework!  

Also spring may have revmoeved this, but wondering if another solution here for this to bring it in.  OR move to an alternative RestTemplate approach.

clouddriver/clouddriver-kubernetes/clouddriver-kubernetes.gradle
js4683: Done - both client-java and client-java-api-fluent are now constrained once in spinnaker-dependencies (kubernetes: 19.0.3), and the module uses plain coordinates with a comment explaining why we stay below Boot's managed 24.x.

clouddriver/build.gradle
js4683: Done - extracted wiremock-substitution.gradle at the repo root (single home for the version) and all 11 composites now just apply it. kork-bom still constrains the versions for resolution.

js4683: Checked - WireMock 3.x (up to 3.13.x) still ships Jetty 11 by default and the jetty12 variant is Jetty 12.0.x, so a newer 3.x would not resolve the clash with Boot's Jetty 12.1.x (Jetty 12.1 default only lands in WireMock 4). Kept 3.10.0 with the standalone substitution.

echo/echo-notifications/echo-notifications.gradle
js4683: Yes, good call - echo-notifications test dep now declares wiremock-standalone directly instead of relying on the substitution.

echo/echo-notifications/echo-notifications.gradle
js4683: Kept but documented: sender-okhttp 1.62 requires okhttp 5.x while the tree resolves okhttp 4.12.0, and the CDEvents sender here uses the gRPC exporter so the okhttp sender is unneeded. A newer OTel would not help while the okhttp majors diverge. The api-incubator version moved to spinnaker-dependencies.

echo/build.gradle
js4683: Done - same shared substitution script as above; no per-module versions remain.

gate/gate-web/gate-web.gradle
js4683: Done - springdoc version dropped (platform already constrains it, gate is the only consumer) and cdevents-sdk is now pinned in spinnaker-dependencies. The io.cloudevents artifacts stay local: echo is on 3.x (matching cdevents-sdk 0.3.1) while gate is on 5.x, so they cannot share one constraint.

gate/build.gradle
js4683: Done - gate uses the same shared wiremock-substitution.gradle now.

kayenta/kayenta-core/src/main/java/com/netflix/kayenta/config/KayentaJackson3SubtypeConfiguration.java
js4683: Fixed - switched to a Netflix header to match the kayenta module convention. (And noted on the Atlassian suggestion - staying neutral here.)

keel/keel-sql/keel-sql.gradle
js4683: Done - removed the local strictly pins; codegen now follows the Boot-managed 3.21.7. Verified MySQLDSL still exists in 3.21 and the sql modules compile.

kork/kork-eureka/src/main/java/com/netflix/spinnaker/kork/eureka/BootHealthCheckHandler.java
js4683: Done - expanded to explicit imports.

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
js4683: Acknowledged - leaving Spock 2.4 / Groovy 4 as-is here; happy to do the Spock/Groovy follow-up as a separate PR.

kork/spinnaker-dependencies/spinnaker-dependencies.gradle
js4683: Done - removed the jOOQ version and constraints globally now that Java 25 is the toolchain. Kept the kubernetes pin: that one is about 24.x model incompatibility with existing manifests, unrelated to the Java version.

orca/orca-webhook/src/test/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfigurationTest.java
js4683: Renamed - the whole family is now SpinnakerOkHttpClientRequest/RequestFactory/Response.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/WebhookConfiguration.java
js4683: Renamed as above.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/OkHttpClientHttpResponse.java
js4683: Kept the vendored class but renamed (see above). Rationale: Spring 7 removed OkHttp support entirely, and webhooks rely on OkHttp-specific behavior (custom SSL socket factory, network interceptors, timeouts). Moving to the JDK or Jetty client would change that behavior, so a bridge is needed; the javadoc now calls out revisiting if webhooks migrate clients.

orca/orca-webhook/src/main/java/com/netflix/spinnaker/orca/webhook/config/OkHttpClientHttpRequestFactory.java
js4683: Renamed to Spinnaker-prefixed classes, and the factory javadoc now records the revisit path. Agree on a follow-up to evaluate non-OkHttp options once this is in.

js4683:
All 17 review comments are addressed in 83c58e6 (plus a main sync in 43c26d0/fc4369e). Summary of the resolutions:

- **WireMock versions everywhere**: new `wiremock-substitution.gradle` at the repo root is the single home for the version; all 11 composites just `apply from` it. kork-bom still constrains the versions for resolution. Newer WireMock 3.x would not help (still Jetty 11 by default; Jetty 12.1 default only lands in v4), so 3.10.0 + standalone stays.
- **OTel exclusion stays** (now documented): sender-okhttp 1.62 needs okhttp 5.x while the tree is on 4.12.0, and the CDEvents sender uses gRPC. The incubator version moved to spinnaker-dependencies.
- **jOOQ pins dropped** (runtime + keel-sql codegen now follow Boot's 3.21.7) now that Java 25 is the toolchain. k8s 19.x pin stays centrally since 24.x models reject existing manifests.
- **OkHttp bridge renamed** to `SpinnakerOkHttpClient*`; kept as a bridge because Spring 7 removed OkHttp support and webhooks need its SSL/interceptor/timeout behavior.
- Nits (wildcard import, Kayenta header) fixed; Spock stays as-is for now.

Note: Jackson stays out of this PR per your message - it lives in #8033.

One CI note: the single `build` failure is `HttpCloudFoundryClientTest.createRetryInterceptorShouldRetryOnInternalServerErrorsThenTimeOut`, which passes locally 3/3 on this branch - looks like a timing-sensitive flake rather than something from this change.

echo/echo-notifications/echo-notifications.gradle
jasonmcintosh: I'm resolving but ONE caution... on gradle deps if I recall right... echo-web MIGHT still get some... alternatives on okhttp resolution that COULD conflict with this.  SO definitely want to test OTEL for end users on this release.

gate/gate-saml/src/main/java/com/netflix/spinnaker/gate/security/saml/Saml2RelyingPartyProperties.java
```diff
@@ -0,0 +1,362 @@
...
+/**
+ * SAML2 relying party properties, mirroring Spring Boot 3's {@code
+ * spring.security.saml2.relyingparty} namespace.
+ *
+ * <p>Spring Boot 4 removed its SAML relying-party auto-configuration, so gate binds and builds
+ * {@code RelyingPartyRegistration}s itself to preserve the Boot 3 operator contract. Shape and
+ * defaults intentionally match Boot 3.5's {@code Saml2RelyingPartyProperties}.
+ */
+@ConfigurationProperties("spring.security.saml2.relyingparty")
+public class Saml2RelyingPartyProperties {
```
jasonmcintosh: IF we bring in the library spring-boot-starter-security-saml2 I think this goes away, as it should still match standard spring configuration properties.  E.g. we should NOT need to do this in this way.  This was handled on the last upgrade on moving everything but a few SELECT configurations over to native spring properties.

gate/gate-saml/src/main/java/com/netflix/spinnaker/gate/security/saml/Saml2RelyingPartyRegistrationConfiguration.java
```diff
@@ -0,0 +1,239 @@
...
+ *
+ * <p>Faithful port of Spring Boot 3's {@code Saml2RelyingPartyRegistrationConfiguration}, which
+ * Boot 4 removed along with its SAML auto-configuration. Preserves the Boot 3 operator contract
+ * ({@code spring.security.saml2.relyingparty.registration.*}) on Boot 4 / Spring Security 7.
+ */
+@Configuration(proxyBeanMethods = false)
+@Conditional(Saml2RelyingPartyRegistrationConfiguration.RegistrationConfiguredCondition.class)
+@ConditionalOnMissingBean(RelyingPartyRegistrationRepository.class)
+@EnableConfigurationProperties(Saml2RelyingPartyProperties.class)
+class Saml2RelyingPartyRegistrationConfiguration {
```
jasonmcintosh: Similar to above - we should NOT need these classes as they are imported in this way.

echo/echo-notifications/echo-notifications.gradle
js4683: Good caution - verified: the OTel path is covered by CDEventsOTelSenderSpec (no okhttp sender on the classpath since the module excludes it, and the tree stays on okhttp 4.12.0 everywhere else). I ran the echo-notifications test compile green locally; the full OTel end-user path is worth an eye in CI on this release as you say.

gate/gate-saml/src/main/java/com/netflix/spinnaker/gate/security/saml/Saml2RelyingPartyProperties.java
js4683: Done - removed the hand-ported Saml2RelyingPartyProperties; gate-saml now uses spring-boot-starter-security-saml2 (which carries Boot's Saml2RelyingPartyAutoConfiguration + standard properties). Only SecuritySamlProperties (custom attribute extraction) and the Boot-necessary OpenSaml5/cookie changes remain. Validated with the Keycloak-backed SAML integration tests, 5/5 green locally.

gate/gate-saml/src/main/java/com/netflix/spinnaker/gate/security/saml/Saml2RelyingPartyRegistrationConfiguration.java
js4683: Done - Saml2RelyingPartyRegistrationConfiguration (plus its test and cert fixtures) is deleted with the above; nothing else referenced it.

jasonmcintosh (approved):

PR #7998, opened 2026-09-11
fix(kubernetes): validate namespace allow-list when no namespace is set

### LINKED ISSUE #5992: Kubernetes deploy stage namespaceOverride allows bypassing account namespace restrictions.
### Issue Summary: 
If you use the namespace override functionality in the Kubernetes deploy stage, or rely on the default namespace and submit manifests with no namespace defined, the Kubernete's account namespace restrictions can be bypassed.

### Cloud Provider(s): 
Kubernetes

### Feature Area: 
Kubernetes Accounts

### Description:

By using the namespace override functionality in the deploy manifest stage, the namespace gets set after validation checks occur, in [KubernetesDeployManifestOperation](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/manifest/KubernetesDeployManifestOperation.java#L105-L108). Validation checks are done on the description when it comes in, on the original manifest namespace (via the [KubernetesValidationUtil](https://github.com/spinnaker/clouddriver/blob/master/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/validator/KubernetesValidationUtil.java#L118-L120)).

This can be mitigated by using appropriate RBAC in the kubernetes account used by the kubeconfig for the Spinnaker account - however, the `namespaces` field as described in the halyard docs is `A list of namespaces this Spinnaker account can deploy to and will cache.`, which indicates it is meant to be a restriction of namespaces to be deployed to, and not just for caching purposes.

Additionally, when you have multiple accounts for the same cluster that differ only in the permissions set and namespaces list (e.g. multiple logical teams/multitenancy in a single kubernetes cluster), it can be attractive to use the same kubeconfig with wider RBAC permissions, especially when the number of teams starts scaling up.

This also applies for the default namespace scenario. If you supply manifests with no namespace defined, kubectl will be executed with no namespace option and will deploy to the default namespace, even though it may not be in the account namespace list.

### Steps to Reproduce:
* Define a Kubernetes account with a limited set of namespaces allowed.
* Use a Deploy manifest stage with an override namespace that isn't in the set.

jasonmcintosh:
## Summary

Kubernetes accounts can restrict which namespaces they're allowed to deploy to or operate on via the `namespaces` config field. That restriction was bypassable whenever a request omitted a namespace: `KubernetesValidationUtil#validateCredentials` treated a blank namespace as automatically valid, so the allow-list check never ran. kubectl still resolved a real target underneath — the kubeconfig context's default namespace (typically `default`) — so the request could land somewhere the account was never authorized for, without Spinnaker ever inspecting where it actually went.

This closes the second, previously-unaddressed half of spinnaker/spinnaker#5992 (the `namespaceOverride` half was already fixed in [clouddriver#5106](https://github.com/spinnaker/clouddriver/pull/5106)).

## Root causes & fixes

The gap lived in `KubernetesValidationUtil.validateCredentials`, which short-circuited to `true` whenever the incoming namespace string was blank — before ever consulting `credentials.getNamespaces()`. Every manifest-based operation validator (deploy, delete, patch, scale, enable/disable, rollout pause/resume/undo, rolling restart, resize-server-group) funnels through this same method, so the gap was shared across all of them whenever their coordinate's namespace was unset.

Two fix strategies were considered:
- **Reject** a blank namespace outright on restricted accounts.
- **Default** a blank namespace to `"default"` — the same value kubectl itself falls back to — before validation, so it gets checked like any other namespace.

Went with the latter: it preserves the existing "deploy with no namespace" workflow for accounts whose allow-list includes `"default"` (the common case), and correctly rejects it for accounts that don't.

Concretely: `KubernetesCoordinates.withDefaultedNamespace(KubernetesCredentials)` resolves a blank namespace to `"default"` when the resource kind is namespace-scoped, leaving cluster-scoped kinds and already-explicit namespaces untouched. Applied everywhere a coordinate is built ahead of validation/execution:
- `KubernetesDeployManifestConverter#updateNamespace` (deploy)
- `KubernetesManifestOperationDescription#getPointCoordinates` (scale, enable/disable, rollout pause/resume/undo, rolling restart)
- `KubernetesPatchManifestDescription#getPointCoordinates` (patch)
- `KubernetesDeleteManifestDescription#getPointCoordinates` **and** `#getAllCoordinates` (delete — both the single-target and label-selector/"dynamic" paths; both ultimately call kubectl with no `--namespace` flag when blank)
- `KubernetesServerGroupOperationDescription#getCoordinates` (resize server group)

`KubernetesValidationUtil` itself needed no change — its "blank namespace → valid" short-circuit is now correctly reached only for genuinely cluster-scoped resources, since namespaced-kind blanks are resolved upstream before it ever sees them.

**Sibling audit:** label-selector/"dynamic" delete (`KubernetesDeleteManifestDescription#getAllCoordinates`) was initially assumed safe (cache-scoped); tracing it through `CanDelete#delete` → `KubernetesCredentials#delete` → `KubectlJobExecutor#delete` showed it hits kubectl directly with the same blank-namespace gap, so it's fixed here too rather than left for a follow-up.

## Severity

| Component | Severity |
|---|---|
| clouddriver (Kubernetes provider) | Medium — requires a configured namespace allow-list and a request that omits a namespace; mitigated in practice by Kubernetes RBAC scoping the underlying credential, per the original issue discussion |

## Behavioral change (intentional)

Accounts with a `namespaces` allow-list that does **not** include `"default"` will now reject a manifest/operation that omits a namespace, where before it silently succeeded against whatever the kubeconfig context resolved to. Accounts whose allow-list includes `"default"`, or that don't restrict namespaces at all, see no change.

## Test plan

- [x] `./gradlew :clouddriver:clouddriver-kubernetes:test` — BUILD SUCCESSFUL (770 tests)
- [x] `./gradlew :clouddriver:clouddriver-kubernetes:compileIntegrationJava` — BUILD SUCCESSFUL
- [x] `./gradlew :clouddriver:clouddriver-kubernetes:spotlessApply` — clean
- [x] New unit tests: `KubernetesCoordinatesTest`, `KubernetesManifestOperationDescriptionTest`, `KubernetesPatchManifestDescriptionTest`, `KubernetesDeleteManifestDescriptionTest`, `KubernetesServerGroupOperationDescriptionTest` — each proves blank+namespaced → `"default"`, blank+cluster-scoped → unchanged, explicit namespace → unchanged
- [x] New integration test `DeployManifestIT#shouldNotDeployToDefaultNsWhenNotListed` — proves the previously-unauthorized case (deploy with no namespace to an account whose allow-list excludes `"default"`) is actually rejected (400 + `wrongNamespace`), not just that validation code exists
- [ ] `DeployManifestIT` / `DeleteManifestIT` / `PatchManifestIT` full integration suite (requires a local kind cluster + kubectl; not run in this environment — flagging for CI/maintainer verification)

Part of spinnaker/spinnaker#5992


christosarvanitis (approved):
Agree with this fix but also need to note that the best prevention is properly configured kubeconfigs matching the Spinnaker restrictions

jasonmcintosh:
Agree on prevention... AND we'll want to call this out in release notes that this CAN break some deployments.

PR #7913, opened 2026-08-21
fix(entity-store): Fix a UI bug with entity store references and a few stages which didn't work with it

jasonmcintosh:
image
Should resolve this rather annoying UI message when entity store is enabled.  AND this also enables entity store for a few stages that used a singular manifest in the context (Run Job stages and the Find Resource stages)

PR #7999, opened 2026-09-11
fix(entity-store): restore artifact name/kind/namespace on stored manifest references

jasonmcintosh:
## Summary

The entity-storage feature (#7072) swaps large deploy-manifest payloads in the pipeline
execution context for a small `ref://...` pointer to shrink context size. Replacing the
inline manifest with a bare pointer cost deck a real capability: it can no longer tell
the user what was actually deployed.

## Root cause & fix

`EntityHelper.toArtifact` hardcoded `name: "stored-entity"` and dropped `kind` /
`metadata.name` / `metadata.namespace` from the manifest it was storing. `DeployStatus`
in deck's `kubernetes`/`cloudrun` packages can't determine kind/name/namespace for a
stored reference without fetching its full content, so it fell back to an opaque
"Load Manifest Details" link only (added as a stopgap in #7913/#7914).

`EntityHelper.toArtifact` now best-effort extracts `kind`/`name`/`namespace` from the
manifest map before storing it, and surfaces them via `Artifact`'s existing `name`/
`metadata` fields — no model changes needed, since those fields already round-trip
into the placeholder map that lands back in the stage context. Both callers
(`ManifestMapStorageHandler`, `ManifestStorageCollectionHandler`) get this for free
with no code changes, since they both go through `EntityHelper.toArtifact`. Maps that
aren't manifest-shaped (no `kind`/`metadata.name`) keep the previous `"stored-entity"`
fallback, so this is backward compatible for any other future callers.

On the deck side, `IStoredManifestReference` (kubernetes + cloudrun, identical shape in
both packages) gains optional `name`/`metadata` fields, and `DeployStatus` renders a
"Kind namespace/name" label next to the existing link when present. References stored
before this change carry no metadata and render exactly as they do today (link only).

## Test plan

- [x] `./gradlew :kork:kork-artifacts:test` — BUILD SUCCESSFUL, including new
      `EntityHelperTest` cases covering manifest-identity extraction and the
      non-manifest fallback.
- [x] `./gradlew :kork:kork-artifacts:spotlessCheck` — BUILD SUCCESSFUL.
- [x] `pnpm --filter @spinnaker/kubernetes build` and
      `pnpm --filter @spinnaker/cloudrun build` — type checking passed. (Workspace-wide
      `pnpm test`/`:deck:test` currently fails on `main` due to an unrelated pre-existing
      TypeScript error in `kayenta/middleware/epics.ts`; confirmed on a clean `main`
      checkout before making any changes here.)
- [ ] Manual verification of `DeployStatus` in a running deploy-manifest pipeline with
      entity storage enabled — not done in this environment; would appreciate a
      maintainer/reviewer smoke test before merge.


christosarvanitis (approved):
Good fix for the entity store, thanks!

PR #8000, opened 2026-09-11
feat(entity-store): add a SQL-backed artifact/entity store

jasonmcintosh:
## Summary

The entity-storage feature (#7072) only ships an S3-compatible backend
(`S3ArtifactStoreStorer`/`S3ArtifactStoreGetter`), which blocks operators on
Azure, or anyone without an S3-compatible object store, from using it at all.
This adds a SQL-backed `ArtifactStoreGetter`/`ArtifactStoreStorer` pair behind
the same interfaces, so any operator with a database can use the feature. The
goal is still to shrink the *pipeline execution context* per request - not
necessarily total DB footprint - so content is addressed by `(application,
hash)`, the same identity `ArtifactStoreURISHA256Builder` already builds for
S3, meaning identical content stored more than once collapses to a single row.

New module: `kork-artifacts-sql`, kept separate from `kork-artifacts` itself
so the jOOQ/SQL dependency doesn't get pulled into every service that merely
uses the artifact model (igor, rosco, keel, ...). It's wired into
`clouddriver-web` and `orca-web` - the two runnable modules that already
`@Import` `ArtifactStoreConfiguration` - the same way `clouddriver-sql`/
`orca-sql` themselves are wired in (unconditional at the `-web` level,
activated purely by runtime config).

## Design notes

- **Connection**: reuses the host service's `sql.connection-pools` rather than
  inventing a separate connection block. By default it's the same database as
  the service's own SQL persistence; `artifact-store.sql.connectionPool` can
  point it at a different named pool instead.
- **Every participating service must point at the same physical database** -
  entity storage is only useful cross-service (clouddriver stores a manifest,
  orca reads it back later in the same pipeline), exactly like the S3 backend
  already requires every service to share one bucket. Documented on both the
  config property and `SqlArtifactStoreConfiguration`'s class javadoc.
- **Migration is self-managed**, not layered onto the host's own changelog:
  each connection pool gets its own Liquibase run against a uniquely-named
  changelog (`db/artifact-store-changelog-master.yml`, not the conventional
  `db/changelog-master.yml` every kork-sql-based service already uses for its
  own schema - a real collision I hit and fixed once this module ended up on
  the same runtime classpath as clouddriver/orca's own changelogs).
- **Not registered as a Spring-managed `SpringLiquibase` bean.**
  `DefaultSqlConfiguration`'s own liquibase bean is
  `@ConditionalOnMissingBean(SpringLiquibase.class)`; a second bean of that
  type here, even for an unrelated changelog, could suppress the host
  service's own migration depending on configuration-processing order. The
  migration runs directly inside the `artifactStoreDataSource` bean method
  instead. Caught by actually booting the Spring context in
  `SqlArtifactStoreConfigurationTest`, not just constructing the storer/getter
  classes directly.
- **Dialect-portable upsert-or-ignore**: MySQL's `onDuplicateKeyIgnore()` and
  Postgres's `onConflictDoNothing()` aren't portable jOOQ calls (each throws
  `SQLDialectNotSupportedException` on the other dialect), so the storer tries
  one and falls back to the other - the same dialect-fallback dance
  front50-sql's `SqlStorageService` already uses for the same reason.
- **One intentional behavioral difference from the S3 getter**: it denies
  access when the S3 object has no application tag at all (an S3-only failure
  mode). The SQL schema has `application` as `NOT NULL`, so that state can't
  occur here.

## Verified: shared-database safety

Since clouddriver and orca both wire this in and must point at the same
database for entity storage to actually work, `SqlArtifactStoreSharedDatabaseTest`
simulates two independent services (two separate `HikariDataSource` pools) against
one shared MySQL container:
- Two independent pools can each run/re-run the migration against the shared
  database - the second is a safe no-op via Liquibase's own
  `DATABASECHANGELOG` tracking.
- An artifact stored through one pool is readable through a completely
  separate pool pointed at the same database.
- Two pools concurrently storing identical content race safely down to one
  row under real concurrent threads, not just sequential calls.

(Also found, and left a comment about: Liquibase's own `Scope` mechanism isn't
safe to invoke concurrently from two threads in one JVM - a Liquibase-internal
quirk, not a bug here, and not a real concern in production since
clouddriver/orca are always separate processes.)

## Test plan

- [x] `./gradlew :kork:kork-artifacts-sql:test` - BUILD SUCCESSFUL: unit tests
      (Mockito + a real in-memory H2 for the getter's query path, since
      jOOQ's fluent generics don't play well with Mockito deep stubs), a
      Spring-context wiring test, real MySQL + Postgres integration tests
      (run as separate JUnit classes so only one container is alive at a
      time), and the shared-database tests above.
- [x] `./gradlew :kork:kork-artifacts-sql:spotlessCheck` - BUILD SUCCESSFUL.
- [x] `./gradlew :clouddriver:clouddriver-web:compileJava` and
      `./gradlew :orca:orca-web:compileJava` - BUILD SUCCESSFUL, confirming
      the composite-build dependency substitution resolves
      `io.spinnaker.kork:kork-artifacts-sql` correctly.
- [ ] Manual end-to-end verification (deploy a manifest large enough to
      trigger entity storage with `artifact-store.type: sql` enabled on a
      real clouddriver+orca pair pointed at a shared database) - not done in
      this environment; would appreciate a maintainer/reviewer smoke test
      before merge.

Independent of #7999 (the entity-store metadata/UI fix) - no overlapping files.


kork/kork-artifacts-sql/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/sql/ArtifactStoreTable.java
```diff
@@ -0,0 +1,36 @@
+/*
+ * Copyright 2026 Apple Inc.
```
christosarvanitis: Update needed for the License headers

kork/kork-artifacts-sql/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/sql/SqlArtifactStoreConfiguration.java
```diff
@@ -0,0 +1,166 @@
...
+ * even pointed at a different changelog -- could suppress the host service's own migration
+ * depending on configuration-processing order.
+ */
+@Configuration
+@Import({DefaultSqlConfiguration.class, EntityStoreConfiguration.class})
+@ConditionalOnProperty(name = "artifact-store.type", havingValue = "sql")
+public class SqlArtifactStoreConfiguration {
+
+  @Bean(name = "artifactStoreDataSource")
+  public DataSource artifactStoreDataSource(
```
christosarvanitis, line 104: My only objection here is that the artifactStoreDataSource user has to have permissions to run the liquibase migrations as well. This contradicts the usual permissions we have documented where the migration user and the RW user have different permissions.

christosarvanitis (commented):
Want to discuss a bit the migration pool and permissions option as we are opening up the permissions on the SQL db post migration scripts

kork/kork-artifacts-sql/src/main/java/com/netflix/spinnaker/kork/artifacts/artifactstore/sql/SqlArtifactStoreConfiguration.java
jasonmcintosh: Let me look at adding a migration data source... it's already a shared database (sorta like S3 is a shared external storage)...

christosarvanitis (approved):
Thanks for addressing the comments

PR #8005, opened 2026-09-12
feat(orca): support additional metric tags on stage execution context

apoorvmahajandd:
Extract an optional additionalMetricTags map from the stage context during StageExecutionImpl construction, mirroring how other reserved context keys (e.g. requisiteStageRefIds) are already extracted.

We can now inject custom metric tags from stages into existing stage metrics

## Testing
- unit tests
- verified that the metric had the service tag showing up.
```
stage_invocations_duration_seconds_count{...,service="foo", stageType="deployManifest", status="SUCCEEDED"}
```

after adding this:
```
  "additionalMetricTags": {
    "service": "foo"
  },
  ```
  to the stage context

### DIFF
--- a/orca/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
+++ b/orca/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/model/StageExecutionImpl.java
@@ -139,6 +139,7 @@ public class StageExecutionImpl implements StageExecution, Serializable {
     this.requisiteStageRefIds =
         Optional.ofNullable((Collection<String>) context.remove("requisiteStageRefIds"))
             .orElse(emptySet());
+    this.additionalMetricTags = (Map<String, String>) context.remove("additionalMetricTags");
 
     this.context.putAll(context);
   }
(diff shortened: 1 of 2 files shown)


jasonmcintosh (approved):

jasonmcintosh:
I'll note ONE Risk:  It's user input being added to a system that sends data to a remote system.  I'm still in general in FAVOR of this... but I could also see this used to make an remote type attack with ... bad label inputs kinda things.  I'd think this is a LOW possibility overall, why I'm approving it, just a "caution"

jasonmcintosh:
SO looking at this more, what I THINK we need is some validation on these:
* Specifically... maybe either a whitelist or similar to prevent "let's add timestamp as a metric label!" kinda things which don't make sense
* Possibly a regex of approve names
* BLOCK used of EXISTING tags to prevent override on things like "status" which this would allow as a result

Overall no MAJOR concerns and I think the libraries handle TYPE and RCE issues aren't there, but making sure the tags & tag values are "sane" would help IMHO

christosarvanitis (commented):
Nice, useful for teams looking to monitor their pipelines better. 
Maybe we could add a Deck component to this instead of relying on the JSON for better user experience? 
Proposing on the pipeline config tab to have an optional under metadata to have a Metrics Tags component that will reflect this. If we want to keep the Deck part out of scope lets raise an Issue to track the deck work.

PR #8019, opened 2026-09-15
feat(sbom): generate a per-service CycloneDX SBOM, starting with clouddriver

jasonmcintosh:
## Summary
- Adds `SpinnakerSbomPlugin` (wraps the CycloneDX Gradle plugin) into the shared `SpinnakerBaseProjectPlugin`, so every service's root project gets a `cyclonedxBom` task that aggregates all of its subprojects into a single SBOM — matching the unit Spinnaker actually ships (one container image / deb per service) rather than one SBOM per Gradle module or one for the whole monorepo.
- Threads an opt-in `generate-sbom` flag through `generic-build-publish` → `publish-docker`, setting `sbom: true` on the `docker/build-push-action` steps to attach a buildx/Syft SBOM attestation to published container images. Enabled for `clouddriver` only, as a first rollout; every other service defaults to `false` with no behavior change.

## Why
New requirements call for an SBOM per OSS project artifact. A per-service SBOM (matching each independently-versioned release pipeline already in `.github/workflows/<service>.yml`) is more useful than one monorepo-wide document, since consumers deploy specific services rather than "all of Spinnaker."

## Test plan
- [x] `./gradlew -p spinnaker-gradle-project :spinnaker-project-plugin:compileGroovy` — plugin compiles
- [x] `./gradlew :clouddriver:cyclonedxBom --dry-run` — task graph builds correctly, one `cyclonedxDirectBom` per clouddriver subproject feeding the aggregate
- [x] `./gradlew :clouddriver:cyclonedxBom` (real run) — produces `clouddriver/build/reports/cyclonedx/{bom.json,bom.xml}`, valid CycloneDX 1.6 document with 892 components and license metadata
- [x] `actionlint` clean on `clouddriver.yml`
- [ ] Live CI run on a clouddriver PR/release to confirm the buildx SBOM attestation attaches to the pushed image (not exercised locally — needs a real `docker buildx build --push`)

https://claude.ai/code/session_012rSDN6izyGfcVcZSsFfY8B


jasonmcintosh:
The ONE piece not working FULLY here is deck yet - b/c deck uses a dispatch event.  Given recent fixes on npmjs.org on trusted publishing to allow multiple workspace files, I think we can go BACK to using workflow files vs. dispatches and get better results on this.

christosarvanitis (approved):

PR #8022, opened 2026-09-16
fix(front50): repair dormant pipeline staleCheck optimistic-concurrency check

### LINKED ISSUE #2966: Spin CLI: Fingerprint resources to protect against concurrent writes

jasonmcintosh:
## Summary

Fixes #2966. `PipelineController`'s `?staleCheck=true` optimistic-concurrency check has been a silent no-op since a 2022 refactor (front50#1142/#1159), for every caller including Deck (which has sent `staleCheck=true` on every pipeline save for years with no effect). Two independent gaps combined to cause this:

- `Pipeline.updateTs` — the wire-format fingerprint clients receive — had no setter, so Jackson silently dropped it on deserialize. `Pipeline.lastModified` (the `@JsonIgnore`d backing field the check actually compares) was therefore always `null` on an HTTP-submitted pipeline, so `PipelineController.validatePipeline`'s guard (`pipeline.getLastModified() != null`) never let the check run.
- `SqlStorageService.loadObject` (the single-row read `checkForStalePipeline` uses via `findById`) never selected `last_modified_at`, unlike `loadObjects`/`listObjectVersions`. Fixing only the client side would have made every save "stale" instead.

## Changes

- `Pipeline.java` — add `setUpdateTs(String)`, so the fingerprint round-trips. Null/blank/unparseable values are treated as "no fingerprint supplied" rather than an error.
- `SqlStorageService.kt` — `loadObject` now selects and assigns `last_modified_at`.
- `PipelineControllerConfig.java` — new `controller.pipeline.save.stale-check-enabled` flag, **default `false`**.
- `PipelineController.java` — `checkForStalePipeline` also requires the new flag, so this fix is inert by default. Existing behavior (check never fires) is preserved until the flag is explicitly turned on for a later release, per the plan discussed on the issue.

Out of scope, left for follow-up per the issue discussion: opting the `spin` CLI into `staleCheck`, the `spin` `go test` CI gap, and non-SQL storage backends (Azure/Oracle/etc.), which have the same single-row `lastModified` gap but are unverified here.

## Test plan

- [x] `PipelineSpec.groovy` — `updateTs` → `lastModified` deserialization round trip, plus null/blank/unparseable edge cases
- [x] `PipelineControllerSpec.groovy` — existing stale-check test updated to explicitly enable the flag; new test confirms `staleCheck=true` is a no-op with the flag at its default (off); new test exercises the real HTTP JSON round trip end-to-end (accept + reject) through the actual mixin-registered `ObjectMapper`
- [x] `SqlStorageServiceTests.kt` — `loadObject` now agrees with `loadObjects` on `lastModified`, verified across both H2 and MySQL dialects
- [x] `./gradlew :front50:front50-web:test :front50:front50-core:test :front50:front50-sql:test` all pass
- [x] `./gradlew spotlessCheck` passes for all touched modules


christosarvanitis (approved):

jasonmcintosh:
@mergifyio backport release-2026.3.x release-2026.2.x release-2026.1.x

PR #8029, opened 2026-09-18
build(test): switch embedded/testcontainer Redis image to Valkey 8

jasonmcintosh:
## Summary
- Standardizes all Testcontainers-based Redis test fixtures on `valkey/valkey:8`, removing Redis as the project's default test dependency in favor of its actively maintained, license-compatible fork. Matches the pattern already adopted in `echo-integration/StandaloneContainerTest`.
- Updates `kork-jedis-test`'s shared `EmbeddedRedis` (covering ~25 downstream test modules transitively: front50-redis, clouddriver cats-redis/core, gate-web/core, kayenta-core, igor-web, orca-*-redis, keiko-redis, kork-jedis itself, etc.) plus 11 independently hardcoded `GenericContainer` image pins across clouddriver, orca, fiat, gate, igor, kayenta, and rosco integration tests.
- Adds `plans/2026/pub-sub-support.md`, a tracked, multi-phase plan for adding Broadcast + Single-delivery pub/sub semantics to kork (this migration is Phase 0 of that plan — a decoupled prerequisite, not part of the pub/sub feature work itself).

## Test plan
- [x] `RedisLockManagerSpec` (kork-jedis), `RedisPermissionsRepositorySpec`/`RedisPermissionsRepositoryTest`, and `UserRolesSyncerSpec` (fiat-roles) all pass against Valkey (37 tests, 0 failures).
- [x] All touched `*-integration` modules (clouddriver, orca, fiat, gate, igor, kayenta, rosco) compile cleanly.
- [ ] The `*-integration` modules' `StandaloneContainerTest`/`RedisContainerTest` suites spin up full built service Docker images and are best verified via CI rather than locally.
- [x] `spotlessApply` run on all touched modules.


plans/2026/pub-sub-support.md
```diff
@@ -0,0 +1,284 @@
+# Kork Pub/Sub: Broadcast + Single-delivery semantics, backed by Redis
```
cristhian-castaneda: Really like where this is headed. Broadcast in `kork` closes a gap we've worked around for a long time (the `AccountDefinitionConfiguration` TODO being the obvious one), and I'm glad someone's picking it up.

What's your take on the doc going to [`spinnaker/governance/rfc/` ](https://github.com/spinnaker/governance/tree/master/rfc) instead? That's where the other cross-cutting designs live, and given the shape of this one it feels like the place people will look for it later.

Either way I'd lean toward pulling the doc out of this PR so the Valkey change can merge on its own, and carrying the design as an RFC or alongside #8030 where Phase 1 actually lands.

And if the intent is that `plans/` becomes an in-repo convention for implementation tracking, I'm not against it, it's just worth its own small PR with a `plans/README.md` so it's a deliberate call rather than something that arrives with a test-image bump 🙂.

kayenta/kayenta-integration/src/test/java/com/netflix/spinnaker/kayenta/StandaloneContainerTest.java
```diff
@@ -52,7 +52,7 @@ class StandaloneContainerTest {
   private static final Network network = Network.newNetwork();
 
   private static final GenericContainer redis =
-      new GenericContainer(DockerImageName.parse("library/redis:5-alpine"))
+      new GenericContainer(DockerImageName.parse("valkey/valkey:8"))
```
cristhian-castaneda, line 55: Small scope thing on the summary: it reads as "all Testcontainers based Redis fixtures", but there's one left. 

`kayenta-integration-tests.gradle:9`:
https://github.com/spinnaker/spinnaker/blob/2f9a8c864c73a4bd0be972f60024ab01b355f32e/kayenta/kayenta-integration-tests/kayenta-integration-tests.gradle#L9


It pulls `com.playtika.testcontainers:embedded-redis:3.1.24`, which is Testcontainers under the hood and resolves its own default `redis` image, so it won't pick up this change.

Looks like a one liner via `embedded.redis.dockerImage` in the test config if you want to fold it in. One thing worth checking if you do: Playtika's wait strategy shells out to the client binary, so worth confirming `valkey/valkey:8` exposes a `redis-cli` before assuming it's a drop-in.

`kayenta-graphite.gradle:7`:
https://github.com/spinnaker/spinnaker/blob/2f9a8c864c73a4bd0be972f60024ab01b355f32e/kayenta/kayenta-graphite/kayenta-graphite.gradle#L7

`kayenta-signalfx.gradle:7`:
https://github.com/spinnaker/spinnaker/blob/2f9a8c864c73a4bd0be972f60024ab01b355f32e/kayenta/kayenta-signalfx/kayenta-signalfx.gradle#L7

Both `it.ozimov:embedded-redis:0.7.2` on the buildscript classpath driving the start/stop tasks. That's an in-process Redis binary rather than a container, so it's a different problem and fair to leave out of scope here. Maybe worth a line in the description as a follow-up so the trail doesn't get lost.

Happy to pick up either of those as a separate PR if you'd rather keep this one tight.

cristhian-castaneda: Last one, more a question than a request: after this, CI runs against Valkey 8 (Redis 7.2 command surface) while the reference deployment in

`spinnaker-kustomize/components/redis/base/statefulset.yml:12`:

https://github.com/spinnaker/spinnaker/blob/2f9a8c864c73a4bd0be972f60024ab01b355f32e/spinnaker-kustomize/components/redis/base/statefulset.yml#L12

still ships `redis:5`, and 

`kayenta/docker-compose.yaml:4`:

https://github.com/spinnaker/spinnaker/blob/2f9a8c864c73a4bd0be972f60024ab01b355f32e/kayenta/docker-compose.yaml#L4

ships `redis:latest`. 

Before this PR those agreed on Redis 5, so the test env was effectively the floor. After it, code using a 6+ command (`COPY`, `GETEX`, `ZRANGESTORE`, `LMPOP`, RESP3 paths) goes green in CI and breaks on a stock deploy.

Not a reason to hold this up, those pins were stale either way. But is the intent that Valkey 8 becomes the documented minimum with kustomize following? If so I'm happy to send that follow-up so the two don't sit out of sync.

cristhian-castaneda (changes requested):
Code changes LGTM. All the `library/redis:5-alpine` `GenericContainer` pins are covered, `echo-integration` was already on `valkey/valkey:8` so this is consistent with what's there, and the `*-integration` suites are green, so the unchecked box in the test plan can be ticked.

Using **"request changes"** mainly so the plan doc question doesn't get lost, not because anything in the code needs fixing. The rest is comments and one question, none of it blocking. Details inline.

> Minor context on the rationale, since it may come up later: `redis:5-alpine` is BSD-3, the relicensing only starts at 7.4. So the argument for moving now is really the maintenance one (no 5.0 release since 2021), with the license question applying to future bumps rather than this one.

kayenta/kayenta-integration/src/test/java/com/netflix/spinnaker/kayenta/StandaloneContainerTest.java
jasonmcintosh: Yeah let me get the statefulset updated - that REALLY should be on a current version and thought I'd set in release notes to upgrade ;) This is an OLD headache on these.

plans/2026/pub-sub-support.md
jasonmcintosh: Yeah let me move this ;) and I'd considered the governance.  SINCE there was already some... start to pub/sub... wasn't sure if this was an "implementation plan" or a "RFC" level conversation.  I've been wondering if these kinds of things would be better here for more visibility - governance is ALSO not getting much attention.  EITHER way i"ll do a separate PR on this with the plan moved there.

kayenta/kayenta-integration/src/test/java/com/netflix/spinnaker/kayenta/StandaloneContainerTest.java
jasonmcintosh: Let me also fix a few more of those places... I'd been focused on some of the test containers, but... let me look at the others you found :)

kayenta/kayenta-integration-tests/src/test/resources/application-base.yml
```diff
@@ -1,5 +1,6 @@
 
 redis.connection: redis://${embedded.redis.user}:${embedded.redis.password}@localhost:${embedded.redis.port}
+embedded.redis.dockerImage: valkey/valkey:8
```
jasonmcintosh, line 3: I'll note NOT entirely sure this is needed with other changes

cristhian-castaneda (approved):
LGTM, thank you so much for the updates!

PR #8034, opened 2026-09-21
fix(rosco): close remaining helmfile gaps on some operations

jasonmcintosh:
The prior hooks/postRenderers guard silently skipped validation on unparseable YAML (e.g. Go template control-flow blocks, or content nested past SnakeYAML's 50-level limit), never recursed into helmfile's `helmfiles:` key, and only warned (rather than rejected) on remote/templated `bases:`/`helmfiles:` references it couldn't fetch and inspect. Any of these let a hostile helmfile.yaml smuggle `hooks:`/`postRenderers:` past the guard undetected.

Fail closed instead: refuse to bake content the guard cannot fully parse, recurse into `helmfiles:` alongside `bases:`, and reject unresolvable remote/templated references by default.

Static YAML inspection can't see everything, though: helmfile's own exec/envExec/readFile/readDir/readDirEntries template functions, and its fetching of remote bases/helmfiles/values/chart repos, aren't expressible as a YAML key at all. Set HELMFILE_DISABLE_HOOKS=true and HELMFILE_DISABLE_INSECURE_FEATURES=true on the helmfile subprocess by default (unless allowHooksAndPostRenderers opts a trusted source out), threading an env map through BakeRecipe -> JobRequest -> JobExecutorLocal for the first time.


christosarvanitis (approved):

jasonmcintosh:
@mergifyio backport release-2026.3.x release-2026.2.x release-2026.1.x

PR #8039, opened 2026-09-21
deps: update several libraries as easy-wins

jgrumboe:
## Summary

Bumps eight third-party dependencies that are pinned directly in
`spinnaker-dependencies.gradle` with no documented constraint rationale.
All are isolated utility or test-only libraries, so the blast radius is
limited to the artifact itself rather than the managed platform.

No changes to the Spring Boot / Spring Cloud baseline.

## Changes

| Artifact | From | To |
| --- | --- | --- |
| `org.objenesis:objenesis` | 2.5.1 | 3.6 |
| `org.jetbrains:annotations` | 19.0.0 | 26.1.0 |
| `org.threeten:threeten-extra` | 1.0 | 1.10.0 |
| `org.bitbucket.b_c:jose4j` | 0.9.6 | 0.9.7 |
| `com.nimbusds:nimbus-jose-jwt` | 10.9 | 10.10 |
| `com.google.guava:guava` | 33.0.0-jre | 33.7.1-jre |
| `io.mockk:mockk` | 1.13.17 | 1.14.11 |
| `org.pf4j:pf4j` | 3.14.1 | 3.15.1 |

## Why these, and why now

Each of these sits in `constraints {}` with a bare version literal and no
surrounding comment explaining the pin, which distinguishes them from the
deliberate pins in the same file (Groovy, Bouncy Castle, the Kubernetes
client, netty). Several had drifted a long way from the current release:
`objenesis` was still on a January 2017 build and `threeten-extra` on 1.0.

## Follow-up

Since Spring Framework 7 standardizes on JSpecify nullness annotations, we should migrate from JetBrains annotations to org.jspecify:jspecify separately. This PR preserves existing semantics and scope.


jasonmcintosh:
I'm really hating that flakey front50 test... and need to get it fixed...

jasonmcintosh (approved):

PR #8044, opened 2026-09-21
feat(deck): distinguish same-kind K8s manifests in Deploy (Manifest) status

apoorvmahajandd:
Include the manifest name alongside its kind in the Deploy (Manifest) stage status so users can tell apart multiple manifests of the same kind (e.g. several ConfigMaps or Secrets) at a glance.

# Before
Screenshot 2026-09-21 at 4 41 04 PM

# After
Screenshot 2026-09-21 at 4 41 12 PM


jasonmcintosh (approved):
Ohhh I like this!! :)

PR #8045, opened 2026-09-22
perf(clouddriver/orca): use label selectors for kubernetes artifact lookups

apoorvmahajandd:
## Motivation

1. When deploying a manifest, there are 3 instances where a `kubectl get` call is made to fetch resources of a certain K8s Kind from a namespace. This causes huge memory spikes when performed against a namespace that has a lot of resources.
Kubectl can effectively perform a `kubectl get` if we pass label selectors to it. So this PR makes it so that you can optionally provide label selectors and these 3 calls will make use of label selectors, otherwise it'll default to the original behavior.

2. By passing label selectors down to clouddriver, we can optimize some kubectl gets calls as mentioned above. But in CleanupArtifactsTask, this value wasn't being passed down to clouddriver from orca. So this PR also fixes that gap.

## Testing
- Unit tests

- Verified in clouddriver that it receives this value:
```
DEBUG 1 --- [0.0-7002-exec-3] o.a.coyote.http11.Http11InputBuffer      : Received [:{"podAntiAffinity":{"preferredDuringSchedulingIgnoredDuringExecution":[{"podAffinityTerm":{"labelSelector":{"matchExpressions":[{"values":["test-service"],"key":"bar","operator":"In"},{"values":["test"],"key":"foo","operator":"In"}]}},{"podAffinityTerm":{"labelSelector":{"matchExpressions":[{"values":["test-service"],"key":"bar","operator":"In"},{"values":["test"],"key":"foo","operator":"In"}]}},"weight":100.0}]}}}},"replicas":2.0,"revisionHistoryLimit":10.0,"selector":{"matchLabels":{"foo":"test","bar":"test-service"}},"strategy":{"type":"RollingUpdate","rollingUpdate":{"maxSurge":"25%","maxUnavailable":0.0}},"progressDeadlineSeconds":600.0},"status":{"replicas":2.0,"readyReplicas":2.0,"updatedReplicas":2.0,"availableReplicas":2.0,"observedGeneration":355.0}}],"labelSelectors":{"selectors":[{"kind":"EQUALS","values":["test"],"key":"foo"},{"kind":"EQUALS","values":["test-service"],"key":"bar"}]}}}]]
```

and
```
INFO 1 --- [tionProcessor-2] c.n.s.c.k.op.job.KubectlJobExecutor      : Getting list of kinds [deployment] in namespace test-service
INFO 1 --- [tionProcessor-2] c.n.s.c.k.op.job.KubectlJobExecutor      : with selectors: foo = test,bar = test-service
INFO 1 --- [tionProcessor-2] c.n.s.c.jobs.local.JobExecutorLocal      : Starting job: 'kubectl --kubeconfig=/tmp/dev --namespace=test-service -o json get deployment -l=foo = test,bar = test-service'...
```


clouddriver/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/jobs/local/JobExecutorLocal.java
```diff
@@ -60,8 +59,7 @@ public <T> JobResult<T> runJob(final JobRequest jobRequest, ReaderConsumer<T> re
 
   private <T> JobResult<T> executeWrapper(
       final JobRequest jobRequest, RequestExecutor<T> requestExecutor) {
-    log.debug(String.format("Starting job: '%s'...", jobRequest.toString()));
-    final String jobId = UUID.randomUUID().toString();
+    log.info(String.format("Starting job: '%s'...", jobRequest.toString()));
```
jasonmcintosh, line 62: NIT:  This is a LOT of logs that are potentially going ot happen, given this happens on every k8s operation.  I'd vote for debug

clouddriver/clouddriver-core/src/main/java/com/netflix/spinnaker/clouddriver/jobs/local/JobExecutorLocal.java
```diff
@@ -72,7 +70,9 @@ private <T> JobResult<T> executeWrapper(
     }
 
     if (jobResult.isKilled()) {
-      log.warn(String.format("Job %s timed out (after %d minutes)", jobId, timeoutMinutes));
+      log.warn(
+          String.format(
+              "Job %s timed out (after %d minutes)", jobRequest.toString(), timeoutMinutes));
```
jasonmcintosh, line 75: MIGHT not be bad to wrap the %s in backticks or similar to make it clearer as it's the command line to be executed.  This isn't bad nor a "don't merge" just a thought. Still useful to have this on failure :)

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/caching/view/provider/ArtifactProvider.java
```diff
@@ -34,9 +34,42 @@
...
+   */
   public ImmutableList<Artifact> getArtifacts(
-      KubernetesKind kind, String name, String location, KubernetesCredentials credentials) {
-    return credentials.list(kind, location).stream()
+      KubernetesManifest manifest,
+      String name,
+      KubernetesCredentials credentials,
+      KubernetesSelectorList labelSelectors) {
+    ImmutableList<KubernetesManifest> candidateManifests;
+    if (labelSelectors.isEmpty()) {
```
jasonmcintosh, line 65: NIT:  Though don't think it can happen, NonNull annotation OR a null check/objectutils that checks for these.

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/description/KubernetesAtomicOperationDescription.java
```diff
@@ -40,6 +41,7 @@ public String getAccount() {
   }
 
   KubernetesNamedAccountCredentials credentials;
+  private KubernetesSelectorList labelSelectors = new KubernetesSelectorList();
```
jasonmcintosh, line 44: Given this is set here to a new object and always set... the above comment is PROBABLY not needed... but DOES catch my eye on sync operations where I've seen goofy things happen ;)

clouddriver/clouddriver-kubernetes/src/main/java/com/netflix/spinnaker/clouddriver/kubernetes/op/job/KubectlJobExecutor.java
```diff
@@ -420,7 +420,7 @@ public Void rollingRestart(
   @Nullable
   public KubernetesManifest get(
       KubernetesCredentials credentials, KubernetesKind kind, String namespace, String name) {
-    log.debug(
+    log.info(
```
jasonmcintosh, line 423: .info logs again.. really like to avoid extra logs for this kinda stuff... debug please!

jasonmcintosh: Same on the other log.info changes!

orca/orca-clouddriver/src/main/java/com/netflix/spinnaker/orca/clouddriver/tasks/artifacts/CleanupArtifactsTask.java
```diff
@@ -27,18 +26,20 @@
...
 import org.springframework.stereotype.Component;
 
 @Component
 public class CleanupArtifactsTask implements CloudProviderAware, Task {
   public static final String TASK_NAME = "cleanupArtifacts";
 
-  @Autowired KatoService kato;
+  KatoService katoService;
 
-  @Autowired ObjectMapper objectMapper;
```
jasonmcintosh, line 41: Rather surprised, though guess I shouldn't be the objectMapper wasn't used :)

jasonmcintosh:
No MAJOR concerns, but prefer to lower the log levels to reduce "noise" on logging systems

